PR #3, opened 2016-09-19
Add a status command

nono:


cmd/status.go
```diff
@@ -0,0 +1,37 @@
...
+
+// statusCmd represents the status command
+var statusCmd = &cobra.Command{
+	Use:   "status",
+	Short: "Check if the HTTP server is running",
+	Long:  `Check if the HTTP server has been started and answer 200 for /status.`,
+	Run: func(cmd *cobra.Command, args []string) {
+		port := os.Getenv("PORT")
+		if len(port) == 0 {
+			port = "8080"
```
aenario: We should let viper handle the whole config or ENV or default. Can be the object of a separate PR.

cmd/status.go
```diff
@@ -0,0 +1,37 @@
...
+		}
+		resp, err := http.Get("http://localhost:" + port + "/status")
+		if err != nil {
+			fmt.Println("Error the HTTP server is not running:", err)
+			os.Exit(1)
```
aenario: Should we have significant error code ? In which case we should also named constants.

cmd/status.go
nono: Yes, it should be better. For the moment, it's just the default way of how gin works. But later, we should definetively move that in viper.

cmd/status.go
nono: I think we should just have 0 for OK and something else for an error. Error messages are more convenient to know what has failed.

PR #4, opened 2016-09-19
Documentation about the files service

nono:
@aenario it's work in progress but don't hesitate to tell me what you think of it


nono:
@aenario I think it's complete. If it's OK for you, it can be merged

aenario:
One thing I forgot : In javascript uploading a file as the body of the request is way harder then uploading  a multipart form. May be we should allow upload using a multipart form also eventually on a different endpoint.

nono:
OK, it's noted. Let's start without this endpoint and if it's a pain later to upload files from javascript, we will add this new endpoint (probably `/files/upload`).

PR #6, opened 2016-09-23
Fix src path of cozy-stack

tomsquest:
Go source code is downloaded to `$GOPATH/src`.
So, `cd $GOPATH/github.com/cozy/cozy-stack` should be `cd $GOPATH/src/github.com/cozy/cozy-stack`.


nono:
You are right, thanks

PR #7, opened 2016-09-23
Add documentation for next sprint CRUD API

aenario:
RFC @nono

First shot at documenting the API for data manipulation.

Basically just a JSONAPI wrapper to Couchdb

Decision to be discussed : 
- Do we really want JSON-API for these route too or should we consider them lower level.
- The former DS hided `_rev` from the client, I believe this was unwise as we ended up reimplementing it in places for synchronisation with other sources (*dav and imap)
- Should we really allow the creation of a doc with a given `_id` ? The only proper way to use this API is to wrap it in a trycatch and do a `PUT` instead if the doc already exists. It seems we should push devs to always do a `PUT`, or eventually a `PATCH` if only changing some fields.
- keeping the type (formerly `docType`) in the document seems convenient, however, it should be marked as a special field with `_`. 
- I added the `type` to the URL, it is not a big effort on developper part but will allow some optimisations later.
- Is the error format lacking something ? One of the issue we had in previous stack was the lack of uniformity and lost of information in error reporting.

### DIFF
--- a/docs/crud.md
+++ b/docs/crud.md
@@ -2,9 +2,15 @@
 
 ### Typing
 
-The notion of document type does not exist in couchdb.
-Cozy-stack introduce this notion with a special `_type` field. This type cannot contain "/", it should be unique among all developers, it is recommend to use the Java naming convention with a domain you own.
-All CozyCloud types will be prefixed by io.cozy
+The notion of document type does not exist in Couchdb.
+
+Cozy-stack introduce this notion by prefixing each document's `_id` with a fully qualified type name followed by `/`.
+
+This type name cannot contain `/`, and it should be unique among all developers, it is recommended to use the Java naming convention with a domain you own.
+
+All CozyCloud types will be prefixed by io.cozy and be pluralized.
+Example ID : `io.cozy.events/6494e0ac-dfcb-11e5-88c1-472e84a9cbee`
+Where, `io.cozy.` is the developer specific prefix, `events` the actual type, and `6494e0ac-dfcb-11e5-88c1-472e84a9cbee` the document's unique (within a type) id .
 
 ------------------------------------------------------------------------------
 
@@ -30,7 +36,6 @@ Etag: "3-6494e0ac6494e0ac"
 {
     "_id": "io.cozy.events/6494e0ac-dfcb-11e5-88c1-472e84a9cbee",
     "_rev": "3-6494e0ac6494e0ac",
-    "_type": "io.cozy.events",
     "startdate": "20160823T150000Z",
     "enddate": "20160923T160000Z",
     "summary": "A long month",
@@ -79,7 +84,6 @@ Accept: application/json
 ```
 ```json
 {
-    "_type": "io.cozy.events",
     "startdate": "20160712T150000",
     "enddate": "20160712T150000",
 }
@@ -99,7 +103,6 @@ Content-Type: application/json
   "data": {
     "_id": "io.cozy.events/6494e0ac-dfcb-11e5-88c1-472e84a9cbee",
     "_rev": "1-6494e0ac6494e0ac",
-    "_type": "io.cozy.events",
     "startdate": "20160712T150000",
     "enddate": "20160712T150000"
   }
@@ -115,7 +118,6 @@ Content-Type: application/json
 ### Details
 
 - A doc cannot contain an `_id` field, if so an error 400 is returned
-- A doc cannot contain a `_type` different from the URL one, if so an error 400 is returned
 - A doc cannot contain any field starting with `_`, those are reserved for future cozy & couchdb api evolution
 
 
--- a/docs/crud.md
+++ b/docs/crud.md
@@ -106,3 +106,62 @@ Content-Type: application/json
 - A doc cannot contain an `_id` field, if so an error 400 is returned
 - A doc cannot contain a `_type` different from the URL one, if so an error 400 is returned
 - A doc cannot contain any field starting with `_`, those are reserved for future cozy & couchdb api evolution
+
+
+--------------------------------------------------------------------------------
+
+# Delete a document
+
+### Request
+```http
+DELETE /data/:type/:id
+```
+```http
+DELETE /data/types.cozy.io/events/6494e0ac-dfcb-11e5-88c1-472e84a9cbee
+Accept: application/json
+
+```
+
+### Response OK
+```http
+200 OK
+Content-Length: ...
+Content-Type: application/json
+```
+```json
+{
+    "id": "types.cozy.io/events/6494e0ac-dfcb-11e5-88c1-472e84a9cbee",
+    "ok": true,
+    "rev": "2-056f5f44046ecafc08a2bc2b9c229e20",
+    "_deleted": true
+}
+```
+### Possible errors :
+- 400 bad request
+- 403 unauthenticated
+- 401 unauthorized
+- 404 not_found
+  - reason: missing
+  - reason: deleted
+- 500 unkown
+
+### Conflict prevention
+
+It is possible to use either a `rev` query string parameter or a HTTP `If-Match` header to prevent conflict on deletion:
+- If both are passed and are different, an error 400 is returned
+- If only one is passed or they are equals, the document will only be deleted if its `_rev` match the passed one. Otherwise, an error 412  is returned.
+- If none is passed, the document will be force-deleted.
+
+**Why shouldn't you force delete** (contrieved example) the user is syncing contacts from his mobile, a contact is created with name but no number, the user see it in the contact app. In parallel, the number is added by sync and the user click "delete" because a contact with no number is useless. The decision to delete is based on outdated data state and should therefore be aborted.
+Couchdb will prevent this, the stack API allow it for fast prototyping but it should be avoided for serious applications.
+
+### Binary attachments
+
+When a document is deleted and it was the last reference to a binary, said binary is deleted as well.
+
+If you are moving binary from one document to another, you will need to create the new document with binary link first, and only afterward delete the previous document.
+
+### Details
+
+- If no id is provided in URL, an error 400 is returned
+-
--- a/docs/crud.md
+++ b/docs/crud.md
@@ -56,12 +56,12 @@ Content-Type: application/json
 ```
 
 ### possible errors :
-- 403 unauthenticated
-- 401 unauthorized
+- 401 unauthorized (no authentication has been provided)
+- 403 forbidden (the authentication does not provide permissions for this action)
 - 404 not_found
   - reason: missing
   - reason: deleted
-- 500 unkown
+- 500 internal server error
 
 --------------------------------------------------------------------------------
 
@@ -107,9 +107,10 @@ Content-Type: application/json
 ```
 
 ### possible errors :
-- 403 unauthenticated
-- 401 unauthorized
-- 500 unkown
+- 400 bad request
+- 401 unauthorized (no authentication has been provided)
+- 403 forbidden (the authentication does not provide permissions for this action)
+- 500 internal server error
 
 ### Details
 
@@ -148,12 +149,13 @@ Content-Type: application/json
 ```
 ### Possible errors :
 - 400 bad request
-- 403 unauthenticated
-- 401 unauthorized
+- 401 unauthorized (no authentication has been provided)
+- 403 forbidden (the authentication does not provide permissions for this action)
 - 404 not_found
   - reason: missing
   - reason: deleted
-- 500 unkown
+- 412 Conflict (see Conflict prevention section below)
+- 500 internal server error
 
 ### Conflict prevention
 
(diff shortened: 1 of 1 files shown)


nono:
FYI, I've renamed the `doc/` directory in `docs/`, as it is the convention on github. It should allow us to publish the documentation on https://cozy.github.io/cozy-stack/ if we find how to tell github to transform markdown to html (it probably needs to install jekyll and have a config file for it).

nono:
> Do we really want JSON-API for these route too or should we consider them lower level.

If possible, I'd really like the possibility to have the same API that CouchDB. It will offer some nice possibilities later, like using pouchdb as a client of our API. The one database per doctype would probably be reflected on the API, like saying to pouchdb that the remote URL is `https://cozy.example.com/data/events`, but I think it's OK.

> The former DS hided _rev from the client, I believe this was unwise as we ended up reimplementing it in places for synchronisation with other sources (*dav and imap)

:+1: for not hiding `_rev`

> Should we really allow the creation of a doc with a given _id ?

I think the proper way to create a document with a given `_id` is a `PUT` on an URL finishing by `/` and the given `_id`. IIRC, it's how CouchDB works. What do you think of it?

> keeping the type (formerly docType) in the document seems convenient, however, it should be marked as a special field with _.

OK. Is this special field `_type` saved in the database, or manipulated by the stack on each request?

> I added the type to the URL, it is not a big effort on developper part but will allow some optimisations later.

I also think that the type should appear in the URL, but it will contain some slashes as it should be namespaced. Example `github.com/cozy/cozy-types/events`.

doc/crud.md
```diff
@@ -0,0 +1,106 @@
+# Access a document
+
+#### Request
+```http
+GET /data/:type/:id
```
nono: Don't forget that type is expected to be namespaces, like `github.com/cozy/cozy-types/events`

doc/crud.md
```diff
@@ -0,0 +1,106 @@
...
+- 404 missing
+- 500 unkown
+
+--------------------------------------------------------------------------------
+
+# Create a document
+
+### Request
+```http
+POST /data/:type/(:id ?)
```
nono: I think we should have 2 distinct routes: POST /data/:type/ and PUT /data/:type/:id

docs/crud.md
```diff
@@ -0,0 +1,103 @@
...
+{
+  "errors": [ {
+    "code": "deleted",
+    "status": 404,
+    "title": "Event deleted",
+    "details": "Event 6494e0ac-dfcb-11e5-88c1-472e84a9cbee was deleted",
+    "links": {"about": "http://github.com/cozy/cozy-stack/docs/errors/deleted"}
+  }]
+}
+```
```
nono: @aenario for something that has existed but is now deleted, it's better to use the code `410 Gone`

aenario:
I made some changes : 

- moved from /doc to /docs
- Removed the possibility to create a doc with a set `_id`, I agree this should be moved to PUT.

- Regarding using pouchdb as a client: :+1: , we dont event need to have the db be `my.cozycloud.cc/data/event` as the _id contains `event/`

Some problematic points : 

- Type Namespacing : Regarding types namespacing, we should agree on a a format, either java like `io.cozy.types.events`, or actual URL `https://cozy.io/types/event` (but we can put those in the URL) or go like `cozy.io/types/events`. 
My vote is for golike.

- Errors : either we go with the JSON-API format, or we keep the couchdb one. We can also, as done for creation OK response extends the couchdb one with more informations. The current jsonapi format will not be properly understood by pouchdb.

- Creation OK response : couchdb doesnt include the created doc in response, I think this is useful, especially when we will start doing some magic, like doctype validation, meta-data extraction, tag matching, ect .... I have put a format which should be compatible with couch client and also include the complete created doc in a JSONAPI like manner.

- The `_type` field : It cannot be saved in couchdb doc (starts with `_`), so we can either
A. Rename it in the public API `cozytype`, `$type`, `çtype`, ...
B. Rename it in couchdb with a complex name to not conflict `x-cozy-type`
C. Dont save it in couchdb doc, but instead let it be the database name.

I prefer keeping some king of special character in the public API for all special fields (_id, _rev, _type, _tags, _attachments, ...). This goes further than type, so we should eliminate Option C. I think we should have the same special character for cozy-special (tags, type) as couchdb-special (rev, id).

nono:
> Regarding using pouchdb as a client: :+1: , we dont event need to have the db be my.cozycloud.cc/data/event as the _id contains event/

Well, I think it's safer to have the type in the URL. The _id is not always available when executing a request. For example, if you want to replicate the events from CouchDB to PouchDB, I think it's mandatory to have the type in the URL (at least, I don't see how to do that otherwise).

> Type Namespacing: My vote is for golike.

:+1:

> Errors : either we go with the JSON-API format, or we keep the couchdb one. We can also, as done for creation OK response extends the couchdb one with more informations. The current jsonapi format will not be properly understood by pouchdb.

Let's do it the CouchDB way.

nono:
> Creation OK response : couchdb doesnt include the created doc in response, I think this is useful, especially when we will start doing some magic, like doctype validation, meta-data extraction, tag matching, ect .... I have put a format which should be compatible with couch client and also include the complete created doc in a JSONAPI like manner.

I'm OK with the format. By looking at the [CouchDB API documentation](http://docs.couchdb.org/en/2.0.0/api/database/common.html#post--db), I've seen that the `POST /:db` accepts a `batch` parameter to send several documents at once. I think we should support it too.

aenario:
`POST /:db/?batch=ok` batch mode does not allow to create several documents at once, it only allows more throughput by not waiting for couchdb to store docs on disk. 

The api to create / update several doc at once is [bulk_docs](http://docs.couchdb.org/en/2.0.0/api/database/bulk-api.html#db-bulk-docs)

nono:
My bad, you are right, batch is not for creating several documents at once.

docs/crud.md
```diff
@@ -0,0 +1,103 @@
...
+```http
+HTTP/1.1 200 OK
+Date: Mon, 27 Sept 2016 12:28:53 GMT
+Content-Length: ...
+Content-Type: application/json
+Etag: "3-6494e0ac6494e0ac"
+```
+```json
+{
+    "_id": "types.cozy.io/events/6494e0ac-dfcb-11e5-88c1-472e84a9cbee",
```
tomsquest: Shouldn't it be `"_id": "events/6494e0ac-dfcb-11e5-88c1-472e84a9cbee",` as below in the Create Response ?

docs/crud.md
```diff
@@ -0,0 +1,103 @@
...
+  - reason: deleted
+- 500 unkown
+
+--------------------------------------------------------------------------------
+
+# Create a document
+
+### Request
+```http
+POST /data/:type/
```
tomsquest: Why is `:type` required ? Is it this already part of the payload (and mandatory there), it is a duplicate information and less obvious.

docs/crud.md
```diff
@@ -0,0 +1,103 @@
...
+```http
+HTTP/1.1 404 Not Found
+Content-Length: ...
+Content-Type: application/json
+```
+```json
+{
+  "status": 404,
+  "error": "not_found",
+  "reason": "deleted",
```
tomsquest: How do you think the deleted status would be keep ?
I mean: doesn't all docs won't be really deleted by **marked** as deleted.
Won't it occur a performance penality (disk space, memory, sharding) ?
Won't it imply a purging functionallity ? (removing deleted docs older than X days)
Should all query skip deleted docs ? (eg. while indexing, select only docs where `_status != 'deleted')`

docs/crud.md
```diff
@@ -0,0 +1,103 @@
+# Access a document
+
+#### Request
+```http
+GET /data/:type/:id
+```
+```http
+GET /data/types.cozy.io/events/6494e0ac-dfcb-11e5-88c1-472e84a9cbee
```
tomsquest: Why setting `cozy.io` in the type name ? eg. `/data/types.cozy.io/events/123`
Shouldn't it be simpler just named `types` ? eg. `/data/types/events/123`
One reason I see is having many kind of types and not all managed by Cozy, which is also wrong because the value can be rewritten by the couchDb access layer.

tomsquest: Why the string `types.cozy.io` is required at all ? 
This `/data/events/123` could be enough.

aenario: The all types should be fully qualified to distinguish between types.cozy.io/messages (email) and my-great-irc-app.xyz/messages for instance.

docs/crud.md
aenario: This is linked to couchdb : deleting a doc is equivalent to create a new tombstone revision ( `{"_id": "types.cozy.io/events/8239239283", "_rev":"5-72372382873", "_deleted": true}` ). This is necessary to ensure proper replication.

Former revisions of the document are also kept by couchdb, but will be removed by a **compact** operation when disk space starts to become a problem. However, the tombstone will remain.
Removing tombstone is done with the **purge** operation, this is rarely necessary as the tombstone weight only a few 10s bytes.

docs/crud.md
aenario: I must admit I am not sure about this part as it is indeed a duplicate information.

Having it in the URL allow us to access it without parsing the whole body, eventually ignoring the body if we know from the type this request wont be accepted (permission/protection against memory bomb) or starting some stream-parsing-analysis without keeping the whole body in memory if we know we will need it (images).
 
On the other hand, having it in the JSON body itself might prove helpful for consumption from client-side frameworks, this is more of a gut feeling, so I am not sure we wil actually need it in the body.

docs/crud.md
aenario: It's actually the other one that is wrong, see my comment [below](https://github.com/cozy/cozy-stack/pull/7#discussion_r80943692)

aenario:
Just pushed some changes 

1. Added the delete API, as it fall within this sprint exploration scope
2. Fix the doctype was not fully qualified in creation's OK response.


While implementing the GET (PR incoming), I am bothered by the fully qualified type identifier containing "/" (where do the type stop, and the id start ? I cant just use gin's router, i need to do some parsing by hand). While it's still an option, I think we might want to switch to a java-like format `io.cozy.types.events`

docs/crud.md
```diff
@@ -0,0 +1,177 @@
...
+    "_deleted": true
+}
+```
+### Possible errors :
+- 400 bad request
+- 403 unauthenticated
+- 401 unauthorized
+- 404 not_found
+  - reason: missing
+  - reason: deleted
```
nono: Don't forget 412, as explained below in conflict prevention

docs/crud.md
```diff
@@ -0,0 +1,177 @@
...
+- 500 unkown
```
nono: 500 is internal server error, not really "unkown"

docs/crud.md
```diff
@@ -0,0 +1,177 @@
...
+    "id": "io.cozy.events/6494e0ac-dfcb-11e5-88c1-472e84a9cbee",
+    "ok": true,
+    "rev": "2-056f5f44046ecafc08a2bc2b9c229e20",
+    "_deleted": true
+}
+```
+### Possible errors :
+- 400 bad request
+- 403 unauthenticated
+- 401 unauthorized
```
nono: Explained like that, it's not really clear what 401 and 403 means.

401 unauthorized is when the user is not authenticated and should try again after being authed
403 forbidden is when the client should not try again, the permissions are not given to do that operation

docs/crud.md
```diff
@@ -0,0 +1,177 @@
...
+- If none is passed, the document will be force-deleted.
+
+**Why shouldn't you force delete** (contrieved example) the user is syncing contacts from his mobile, a contact is created with name but no number, the user see it in the contact app. In parallel, the number is added by sync and the user click "delete" because a contact with no number is useless. The decision to delete is based on outdated data state and should therefore be aborted.
+Couchdb will prevent this, the stack API allow it for fast prototyping but it should be avoided for serious applications.
+
+### Binary attachments
+
+When a document is deleted and it was the last reference to a binary, said binary is deleted as well.
+
+If you are moving binary from one document to another, you will need to create the new document with binary link first, and only afterward delete the previous document.
```
nono: For moving files, I think we should just advise to use the files API (it has a way to do a move in just one HTTP request).

aenario: You're right, I removed this paragraph until we figure out the API/mechanism to attach file to another doc than files

docs/crud.md
aenario: fixed

docs/crud.md
```diff
@@ -0,0 +1,172 @@
...
+```http
+HTTP/1.1 200 OK
+Date: Mon, 27 Sept 2016 12:28:53 GMT
+Content-Length: ...
+Content-Type: application/json
+Etag: "3-6494e0ac6494e0ac"
+```
+```json
+{
+    "_id": "io.cozy.events/6494e0ac-dfcb-11e5-88c1-472e84a9cbee",
```
nono: Why do you put the type in the `_id` field? If we keep it, I don't think it's useful to put also the type in the `type` field in the payload

aenario: I thought there might be a chance we want to keep documents from several doctypes into a single database later on (data-proxy / pouchdb). If the _id is not prefixed, and allowing to create a doc with a specific ID, this might lead to conflicts (`io.cozy.folders/root` vs `password.sync.app/root`).

I agree this removes the need for the _type in document. fixing it now.

nono: Let's make it a good practice: if a developer want to force an id, it should prefix it with the type. But for uuid, no need to keep the _type.

PR #8, opened 2016-09-26
Handle error when running the root command

tomsquest:
Initialize Cobra as per the `README.md`, which handles a potential error in the `RootCmd`.  
Src: https://github.com/spf13/cobra#getting-started


nono:
Thanks!

PR #22, opened 2016-10-03
First shot at read api.

aenario:
So this is the first shot at GetDoc API

It is based on 3 commits by @nono which introduce the instance concept.

Some design decisions : 
- a couchdb packages at cozy-stack/couchdb
- Introducing the notion of database prefix in instance, which joined with doctype give use the database name
- using `gin.ErrorLogger` to output json errors
- Changed status test for healthy and added couchdb to travis

Left TODO :
- the database naming should be better analysed, for now it will fail if there is a character outside of `_$()+-/a-zA-Z0-9` in host or doctype, or if the host doesnt start with a `a-z`.
- Need more code comment, will be working on it in parallel with review

@nono, there is a merge conflict with master. I could fix-it by merging master into this branch, or rebasing this branch on master, but it gets messy with your commits included here.

### DIFF
--- a/couchdb/couchdb.go
+++ b/couchdb/couchdb.go
@@ -1,0 +1,64 @@
+package couchdb
+
+import (
+  "net/url"
+  "io/ioutil"
+  "fmt"
+  "strings"
+  "net/http"
+  "encoding/json"
+)
+// CouchDBURL is the URL where to check if CouchDB is up
+func CouchURL() string {
+  return "http://localhost:5984/"
+}
+var CouchdbClient = &http.Client{}
+
+func makeDBName(dbprefix, doctype string) string{
+  // @TODO This should be better analysed
+  dbname := dbprefix + doctype
+  dbname = strings.Replace(dbname, ".", "-", -1)
+  dbname = strings.ToLower(dbname)
+	return url.QueryEscape(dbname)
+}
+
+func makeDBURL(dbprefix, doctype string) string{
+  return CouchURL() + makeDBName(dbprefix, doctype)
+}
+
+func makeDocID(doctype string, id string) string{
+  return url.QueryEscape(doctype + "/" + id)
+}
+
+func DocURL(dbprefix, doctype, id string) string{
+  return makeDBURL(dbprefix, doctype) + "/" + makeDocID(doctype, id)
+}
+
+func GetDoc(dbprefix, doctype, id string, out interface{}) error {
+  url := DocURL(dbprefix, doctype, id)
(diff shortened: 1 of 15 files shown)


couchdb/couchdb.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		return &CouchdbError{http.StatusInternalServerError,
+      []byte("{\"error\":\"Wrong configuration for couchdbserver\"}")};
+	}
+	resp, err := CouchdbClient.Do(req)
+	if err != nil {
+		return &CouchdbError{http.StatusServiceUnavailable,
+      []byte("{\"error\":\"No couch to seat on.\"}")}
+	}
+  body, err := ioutil.ReadAll(resp.Body)
```
nono: I think you need to close the body:

```go
defer resp.Body.Close()
```

nono:
I think you should merge master in this branch to resolve the conflict

couchdb/couchdb.go
```diff
@@ -0,0 +1,64 @@
...
+
+func DocURL(dbprefix, doctype, id string) string{
+  return makeDBURL(dbprefix, doctype) + "/" + makeDocID(doctype, id)
+}
+
+func GetDoc(dbprefix, doctype, id string, out interface{}) error {
+  url := DocURL(dbprefix, doctype, id)
+  fmt.Printf("[couchdb request] %v\n", url)
+	req, err := http.NewRequest("GET", url, nil)
+  req.Header.Add("Accept", "application/json")
```
jinroh: this `req.Header.Add` should probably be after the `if err != nil`.

couchdb/couchdb.go
aenario: done in https://github.com/cozy/cozy-stack/pull/22/commits/91542eb03e8eda871b47723dda22bd71b55880b7

couchdb/couchdb.go
aenario: done in https://github.com/cozy/cozy-stack/pull/22/commits/91542eb03e8eda871b47723dda22bd71b55880b7

aenario:
OK, I figured out what was wrong with my setup (wrong GOPATH in profile made the linters atom plugins fail silently). So now, gofmt get applied, gometalinter too and travis is happy, which makes me happy.

Both your reviews comment have been fixed in https://github.com/cozy/cozy-stack/pull/22/commits/f214b15ddbd8bc0bb24a11c579483c30032820ee

Issue #21, opened 2016-10-03
Config enhancement


tomsquest:
## To be done

_Work started in:_ https://github.com/tomsquest/cozy-stack

* [x] Look in current directory for a config file
* [x] Handle error when reading config file
* [x] Create a `Config` struct containing the app configuration (see below)
* [x] Add a `config` command to display the parsed configuration (eg. `./cozy-stack config`)
* [x] Inject `Config` where needed, eg. `cmd/status.go` needs the database url

## Already done in `master`

* [x] Parse command line argument with Cobra
* [x] Read config file with Viper
* [x] Integrate Cobra with Viper

## The `Config` struct

This structure contains the final configured key-value pairs.

Goals:
* Be injected into code which depends on configuration
* Not having reference/depends the global `spf13/viper` in application code
* Level of indirection between the config flag and the real values (eg. renaming `prod` to `production` internally)

Pseudo-code for Config:

```go
type Config struct {
	Mode mode 
	Database struct {
	        Url string
        }
}

config := Config{
	Mode: Production(),
	Database: {
		Url: "http://db:5984",
	},
}
```

@nono @aeris preliminarily feedback appreciated

nono:
ping @aenario

nono:
@tomsquest looks good to me. I'd also be interested by having the possibility to override the config file with env variables (e.g. `COZY_MODE` for the mode).

tomsquest:
@nono Yes, this would be the defaults, and they are enforced by Viper.
AFAIK flag > env > config

From `viper.go`:
```go
// overrides
// flag
// env
// config
// key/value store
// default
```

nono:
Nice!

tomsquest:
PR in #24

# Info

* Configured params:

```
$ go run main.go --help
Cozy is a platform that brings all your web services in the same private space.
With it, your web apps and your devices can share data easily, providing you
with a new experience. You can install Cozy on your own hardware where no one
profiles you.

Usage:
  cozy [flags]
  cozy [command]

Available Commands:
  config      Display the configuration
  serve       Start the stack and listens for HTTP calls
  status      Check if the HTTP server is running

Flags:
  -a, --address string   Address on which the server will listen (default "localhost")
  -c, --config string    config file (default is $HOME/.cozy.yaml)
  -m, --mode string      server mode: development or production (default "development")
  -p, --port int         Port on which the server will listen (default 8080)

Use "cozy [command] --help" for more information about a command.
```

[...]

PR #24, opened 2016-10-04
Configuration enhancement

tomsquest:
Solves #21  "Configuration enhancement", see there for details

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -19,8 +19,12 @@ the given parameters to display the configuration.`,
 		}
 
 		cfg, err := json.MarshalIndent(config.GetConfig(), "", "  ")
+		if err != nil {
+			return err
+		}
+
 		fmt.Println(string(cfg))
-		return err
+		return nil
 	},
 }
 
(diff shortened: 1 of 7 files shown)


nono:
Thanks for the PR.

@aenario @jinroh can you look at this PR, and merge it if good for you please?

jinroh:
Great work, thanks @tomsquest !

I know this is not from your modifications, but I would also remove the error printing from the `Execute` error handler in `main.go`. Errors are [printed twice](https://github.com/spf13/cobra/blob/d671d98363b02cd8127087b4675ae77396db5843/command.go#L700-L702):

```go
func main() {
	if err := cmd.RootCmd.Execute(); err != nil {
		fmt.Println(err) // <- redundant (and print in stdout...)
		os.Exit(-1)
	}
}
```

Also the exit code should probably be 1 instead of -1 (255)...

If you're OK with this little change, I'll happily merge your PR.

tomsquest:
@jinroh done + some more (see logs)

jinroh:
Merged. Thanks @tomsquest.

PR #26, opened 2016-10-05
[WIP] Create directory

jinroh:
**Still work in progress**

Lots of stuff are missing and would require connections with couchdb. However since the directory creation share some codes with the uploading part, I prefer to make this PR ASAP so that we can work in parallel 🎿 .

This PR introduces:
  - better error handling
  - some input sanitization
  - draft of the directory creation on VFS

### DIFF
--- a/web/files/files.go
+++ b/web/files/files.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"regexp"
 	"strings"
 
 	"github.com/cozy/cozy-stack/web/jsonapi"
@@ -34,7 +35,10 @@ const (
 var (
 	errDocAlreadyExists = errors.New("Directory already exists")
 	errDocTypeInvalid   = errors.New("Invalid document type")
+	errIllegalFilename  = errors.New("Invalid filename: empty or contains one of these illegal characters: / \\ : ? * \" |")
 )
+
+var regFileName = regexp.MustCompile("[\\/\\\\:\\?\\*\"|]+")
 
 // DocMetadata encapsulates the few metadata linked to a document
 // creation request.
@@ -47,7 +51,38 @@ type DocMetadata struct {
 }
 
 func (metadata *DocMetadata) path() string {
-	return escapeSlash(metadata.FolderID) + "/" + escapeSlash(metadata.Name)
+	return metadata.FolderID + "/" + metadata.Name
+}
+
+// NewDocMetadata is the DocMetadata constructor. All inputs are
+// validated and if wrong, an error is returned.
+func NewDocMetadata(docTypeStr, name, folderID, tagsStr string, executable bool) (*DocMetadata, error) {
+	docType, err := parseDocType(docTypeStr)
+	if err != nil {
+		return nil, err
+	}
+
+	if err = checkFileName(name); err != nil {
+		return nil, err
+	}
+
+	// FolderID is not mandatory. If empty, the document is at the root
+	// of the FS
+	if folderID != "" {
+		if err = checkFileName(folderID); err != nil {
+			return nil, err
+		}
+	}
+
+	tags := parseTags(tagsStr)
+
+	return &DocMetadata{
+		Type:       docType,
+		Name:       name,
+		FolderID:   folderID,
+		Tags:       tags,
+		Executable: executable,
+	}, nil
 }
 
 // Upload is the method for uploading a file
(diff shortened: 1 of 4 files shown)


web/files/files_test.go
```diff
@@ -0,0 +1,125 @@
...
+	assert.Equal(t, 409, res.StatusCode)
+	res.Body.Close()
+}
+
+func TestCreateDirSuccess(t *testing.T) {
+	res := createDir(t, "/files/123?Name=coucou&Type=io.cozy.folders")
+	assert.Equal(t, 201, res.StatusCode)
+	res.Body.Close()
+
+	storage, _ := instance.GetStorageProvider()
```
aenario: I guess gometalinter wants you to factorize this part of the code with TestCreateDirWithSlashInName into a `testDirExist(t, instance, expectedPath)` function

web/files/files_test.go
```diff
@@ -0,0 +1,125 @@
...
+	exists, err := afero.DirExists(storage, "123/coucou")
+	assert.NoError(t, err)
+	assert.True(t, exists)
+}
+
+func TestCreateDirWithSlashInName(t *testing.T) {
+	res := createDir(t, "/files/123?Name=coucou/les/copains!&Type=io.cozy.folders")
```
aenario: I am not sure we want to allow creation of a folder with "/" in names as it will not work when mapping it to an actual FS (like cozy-desktop does). We should probalby just throw a 400 "no slash in name".

web/middlewares/instance.go
```diff
@@ -70,3 +70,18 @@ func SetInstance() gin.HandlerFunc {
 		c.Set("instance", instance)
 	}
 }
+
+// GetInstance will return the instance linked to the given gin context
+func GetInstance(c *gin.Context) (*Instance, error) {
```
aenario: In my own PR, I use c.MustGet("instance"), which panic if there is no instance. This avoids having to test if exists and returning an error in every handler.
  
If the instance middleware aborts when it cant make or get the instance for a request, it should be sufficiently robust. To be discussed

aenario:
I agree that the cyclomatic complexity isnt really helpful for CreationHandler as it is only caused by all the error handling. 

If we decide to keep it, we could divide this function into a `makeMetaDataFromRequest(context)` and an custom error type with a `StatusCode` property, which will be handled in a middleware.

web/files/files_test.go
jinroh: Indeed. Made changes to reject illegal characters in 3ffba188bc2d3c5fd6ec7bff71d6655b1f9e28f7

web/middlewares/instance.go
jinroh: I would tend to agree. For commodity reasons and also because if we do not have an instance in the context, `panic`ing seems the right thing to do...

web/files/files.go
```diff
@@ -76,5 +184,65 @@ func Upload(c *gin.Context) {
...
+	default:
+		code = http.StatusInternalServerError
+	}
+	return
+}
+
+func parseTags(str string) []string {
+	var tags []string
+	for _, tag := range strings.Split(str, ",") {
+		// @TODO: more sanitization maybe ?
```
nono: I think you can remove this TODO. I don't see a reason to restrict what the tags can be

web/files/files.go
```diff
@@ -19,54 +22,159 @@ import (
...
+	// FileDocType is document type
+	FileDocType DocType = "io.cozy.files"
+	// FolderDocType is document type
+	FolderDocType = "io.cozy.folders"
+)
+
+var (
+	errDocAlreadyExists = errors.New("Directory already exists")
+	errDocTypeInvalid   = errors.New("Invalid document type")
+	errIllegalFilename  = errors.New("Invalid filename: empty or contains one of these illegal characters: / \\ : ? * \" |")
```
nono: How did you choose what chars are legal / illegal?

web/files/files.go
```diff
@@ -19,54 +22,159 @@ import (
...
+)
+
+var regFileName = regexp.MustCompile("[\\/\\\\:\\?\\*\"|]+")
```
nono: I don't think it's necessary to escape `/`, `?` and `*` in this regexp.

nono: Oh, and it may be simpler to use [strings.IndexAny](https://golang.org/pkg/strings/#IndexAny) to check for forbidden chars

web/files/files.go
jinroh: By looking at Windows and its naming conventions... https://msdn.microsoft.com/en-us/library/aa365247(v=vs.85).aspx (I'm missing `>` and `<` btw)...

Don't you think it may cause us problem if one day we want to sync on a win fs ?

web/files/files.go
jinroh: Thanks, [strings.ContainsAny](https://golang.org/pkg/strings/#ContainsAny) is even more appropriate.

web/files/files.go
jinroh: If we don't want to enforce these rules, I don't think its necessary to forbid these special characters either. I'll just leave `/` as an illegal one for now.

New PR: #30

PR #27, opened 2016-10-05
Add API for document creation

aenario:
Test incoming, but the function itself is here.

What was done : 
- extract from GetDoc the code that will be common into a makeRequest function
- create a CreateDoc function
- create the associated controller using `gin.Bind` for JSON parsing as it seems the best practice
- some cleanup on error handling

What is left TODO
- Test for creation
- Test cleanup : not sure what's the best practice, my guess  would be
   - couchdb_test where we mock couchdb HTTP interface and test the behaviour from these functions (poper JSON, proper _id attribution, ect ...)
   - data_test where we mock couchdb (proper error codes)
   - some e2e test with an actual couchdb (the stuff actually work)
- Error handling cleanup : what's the best practice to interface `HTTP response` -> `couchdb.Error` -> `gin.Error` -> `JSON output`


jinroh:
OK for me.
@nono ?

couchdb/couchdb.go
```diff
@@ -24,46 +31,96 @@ func makeDBName(dbprefix, doctype string) string {
...
+	createErr := CreateDB(dbprefix, doctype)
+	if createErr != nil {
+		return createErr
+	}
+	return CreateDoc(dbprefix, doctype, doc)
+}
+
+// CreateDoc creates a document
+// created is populated with keys from
+func CreateDoc(dbprefix, doctype string, doc Doc) error {
```
nono: I think you have to pass doc as a pointer (`doc *Doc`) to make `_id` and `rev` available from the caller

couchdb/couchdb.go
```diff
@@ -24,46 +31,96 @@ func makeDBName(dbprefix, doctype string) string {
...
+	err := makeRequest("POST", makeDBName(dbprefix, doctype), &doc, &response)
+	if isNoDatabaseError(err) {
+		return attemptCreateDBAndDoc(dbprefix, doctype, doc)
+	} else if err != nil {
+		return err
+	} else if !response["ok"].(bool) {
+		return fmt.Errorf("couchdb replied with 200 ok=false")
+	}
+	// assign extracted values to the given doc
+	// doubt : should we instead try to be more immutable and make a new map ?
```
nono: It's OK to modify the doc here. You can say in the comment of CreateDoc that doc will be modify in place

nono:
Looks good to me so far

couchdb/couchdb.go
```diff
@@ -24,46 +31,96 @@ func makeDBName(dbprefix, doctype string) string {
...
+	}
+	return CreateDoc(dbprefix, doctype, doc)
+}
+
+// CreateDoc creates a document
+// created is populated with keys from
+func CreateDoc(dbprefix, doctype string, doc Doc) error {
+	var response map[string]interface{}
+
+	doc["_id"] = doctype + "/" + uuid.NewV4().String()
```
aenario: Fixed in https://github.com/cozy/cozy-stack/pull/27/commits/0262a2b977cb88f88911c328c1ef5ad9f28429e2 and https://github.com/cozy/cozy-stack/pull/27/commits/28e2bfeed9e7af981349635a254550a2a20c9819

couchdb/couchdb.go
aenario: fixed in 16a7b60

web/data/data.go
```diff
@@ -2,83 +2,58 @@
...
-func GetDoc(c *gin.Context) {
-	// @TODO this should be extracted to a middleware
-	instance, exists := c.Get("instance")
-	if !exists {
-		abortNoInstance(c)
-		return
-	}
+func getDoc(c *gin.Context) {
+	instance := c.MustGet("instance").(*middlewares.Instance)
+	doctype := c.MustGet("doctype").(string)
```
nono: @aenario why do you use `c.MustGet("doctype").(string)` instead of `c.Param("doctype")` here?

PR #28, opened 2016-10-05
Document the /apps service, and related things

nono:


docs/apps.md
```diff
@@ -0,0 +1,505 @@
...
+The `file/app` permission is powerful, it gives the app the possibility to
+modify itself. It can be dangerous, but it allows to create some static files
+for when JS is not an option. For example, a blog application can generate an
+RSS feed and upload it to this folder.
+
+For jobs, the permission key is composed of `jobs/` and the worker name. Some
+workers can use the `access` to restrict the permission (e.g. `konnectors` use
+the `access` to say which konnector can be used).
+
+For settings, the permission key is composed of `settings/` and a type. The
```
aenario: Why separate settings instead of having `data/io.cozy.settings/locale` for instance ?

docs/apps.md
```diff
@@ -0,0 +1,505 @@
...
+---------------------------
+
+### GET /apps
+
+An application can be in one of these states:
+
+- `ready`, the user can use it
+- `installing`, the installation is running and the app will soon be usable
+- `upgrading`, a new version is being installed
+- `uninstalling`, the app will be removed, and will return to the `available` state.
```
aenario: There is no `available` state  in this list. I guess it just returns to `/apps/manifests`

docs/apps.md
```diff
@@ -0,0 +1,505 @@
...
+Manage the marketplace
+----------------------
+
+### GET /apps/manifests
+
+List applications in the marketplace.
+
+### POST /apps/manifests
+
+Add an application to the marketplace. The payload is a subset of the
```
aenario: Do you mean the application is added to this instance copy of the marketplace or to the common marketplace shared by all cozies ?

docs/apps.md
```diff
@@ -0,0 +1,505 @@
...
+Policy](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy)
+for that. If the token is put in the html of the index.html page of an app,
+another app can use the fetch API to get the html page, parse its content and
+extract the token. We can think of other ways to inject the token (indexeddb,
+URL, cookies) but none will provide a better isolation. Maybe in a couple
+years, when [the origin spec](https://w3c.github.io/webappsec-suborigins/)
+will be more advanced.
+
+So, if having the apps on the same domain is not possible, we have to put them
+on several domains. One interesting way to do that is using sandboxed iframes.
```
aenario: I took a few minutes to understand this paragraph, while knowing the problem. I thin just replacing `domain` by `origin` in this line will make things clearer : 
- we need different origin
- we could do it with iframe, but it's very very complex
- only choice : several domains

docs/instance.md
```diff
@@ -0,0 +1,37 @@
...
+- `io.cozy.apps`
+- `io.cozy.manifests`
+- `io.cozy.files`
+- `io.cozy.notifications`
+- `io.cozy.settings`
+
+Then, it creates some folders:
+
+- `/`, with the id `io.cozy.folders-root`
+- `/Apps`, with the id `io.cozy.folders-apps`
```
aenario: Should we explicit these folders will be created with proper locales

docs/apps.md
nono: I don't want to have permissions per field for data. It's too complicated. So, the permissions for settings will be for the `/settings` endpoint, and we don't have to enter in this complicated area for the `/data` endpoint.

docs/apps.md
nono: Yeah, I'm not really sure about what I've done here. I hesitated a lot. My final decision was to have the installed apps on `/apps` and the marketplace and `/apps/manifests`.

docs/apps.md
nono: I think this method will be used by the worker that fetches the registry every night to make new applications visible in the marketplace (or App Center as it is now called).

docs/instance.md
nono: It's a good question and I don't have an answer for the moment.

I'm not sure we will have the locale when the instance is created. Maybe, we will have it only later, when the user fills the registration/onboarding.

docs/apps.md
nono: Done. It's difficult for me to explain clearly the issues here. The advices are welcomed :smile:

docs/apps.md
aenario: OK, my idea was more to split settings in several documents, but it is still permissions per document. 
:+1: for special route

docs/apps.md
aenario: Oh, I thought this will be done internally and didn't see the need for an HTTP route, that does open an interesting topic : will the workers communicate with the stack through http ? Shouldn't we explictly move this kind of route to a separate endpoint `/workerapi` ?

docs/apps.md
aenario: Mmmh, one of the complex thing to wrap the mind around in previous stack was the fact that the same document App was used liberally for market & apps (same model Backbone in home, controller received one or other, ect ....). 

I think it's important in this new stack to really differentiare between installed `io.cozy.apps` at `/apps` and `io.cozy.manifests` at `/appcenter` : different jsonapitype / doctype. Apps are always identified by slug (we can make fixed ID with slug), and manifests by Source. Which is what you did :+1:

docs/apps.md
```diff
@@ -0,0 +1,505 @@
...
+the `access` to say which konnector can be used).
+
+For settings, the permission key is composed of `settings/` and a type. The
+access can be `read`, `write` and `readwrite`. The type can be:
+
+`locale`     | the default locale for the cozy instance
+`background` | the background for the home
+`theme`      | the CSS theme
+`owner`      | the name of the owner of this cozy instance
+`all`        | all the things list above
```
tomsquest: _the markup is broken here__

docs/instance.md
tomsquest: Question: why `io.cozy.folders-root` instead of `io.cozy.folders.root` (dot separator) ?

docs/apps.md
nono: Good catch, I fix it

docs/instance.md
nono: I don't know. It was the first thing that came to my mind...

aenario: If we keep the current API, the _id will end up being "io.cozy.folders/apps" ie a specific id "apps" of type "io.cozy.folder".

docs/apps.md
aenario: Ok, :+1:

PR #29, opened 2016-10-06
Upload with integrity check

jinroh:
Adds MD5 checksum on uploaded content.

In 6fbe3ce40d2e57268aeb8463166c1aaa87628ed2 I tested a concurrent approach which is a little faster but with added complexities. I came to a simpler (and a little slower) solution finally...

### DIFF
--- a/web/files/upload.go
+++ b/web/files/upload.go
@@ -1,0 +1,94 @@
+// Package files is for storing files on the cozy, including binary ones like
+// photos and movies. The range of possible operations with this endpoint goes
+// from simple ones, like uploading a file, to more complex ones, like renaming
+// a folder. It also ensure that an instance is not exceeding its quota, and
+// keeps a trash to recover files recently deleted.
+package files
+
+import (
+	"bytes"
+	"io"
+
+	"github.com/spf13/afero"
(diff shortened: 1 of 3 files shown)


web/files/upload.go
```diff
@@ -0,0 +1,57 @@
+// Package files is for storing files on the cozy, including binary ones like
+// photos and movies. The range of possible operations with this endpoint goes
+// from simple ones, like uploading a file, to more complex ones, like renaming
+// a folder. It also ensure that an instance is not exceeding its quota, and
+// keeps a trash to recover files recently deleted.
+package files
+
+import (
+	"bytes"
+	"crypto/md5"
```
nono: It looks like travis build is red because gometalinter uses a checker called "gas" that verify some security rules, including that no DES, RC4 or MD5 is used. It's possible to disable those checks with a `// #nosec` comment, as explained here: https://github.com/HewlettPackard/gas#annotating-code

nono:
Looks good to me. I prefer the slower but simpler solution.

@jinroh can you fix the conflict?

jinroh:
rebased

PR #32, opened 2016-10-07
Serve help message

jinroh:
Little *morning nitpicking* PR on naming and help messages:

  - rephrased the outdated `help serve` message
  - renamed the `Address` config param to `Host`, which I think is more correct..
  - more consistent help messages for flags

### DIFF
--- a/cmd/status.go
+++ b/cmd/status.go
@@ -20,8 +20,8 @@ var statusCmd = &cobra.Command{
 			return err
 		}
 
-		address := "http://" + config.GetConfig().Address + ":" + strconv.Itoa(config.GetConfig().Port) + "/status"
-		resp, err := http.Get(address)
+		url := "http://" + config.GetConfig().Host + ":" + strconv.Itoa(config.GetConfig().Port) + "/status"
+		resp, err := http.Get(url)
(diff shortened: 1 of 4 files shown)


cmd/status.go
```diff
@@ -20,8 +21,12 @@ var statusCmd = &cobra.Command{
 			return err
 		}
 
-		address := "http://" + config.GetConfig().Address + ":" + strconv.Itoa(config.GetConfig().Port) + "/status"
-		resp, err := http.Get(address)
+		url := &url.URL{
+			Scheme: "http",
+			Host:   config.GetConfig().Host + ":" + strconv.Itoa(config.GetConfig().Port),
+			Path:   "/status",
```
tomsquest: `/` can be removed as URL adds it automatically

tomsquest:
:smile: I used `host` as my first version, then switched to `address` because gin uses it as the param name. But reading the code, `host` is far more understandable !

cmd/status.go
jinroh: 8ae0c20151a2d74b917040d855ab07d7ba1cfe36

PR #34, opened 2016-10-07
Fix status conversion in Couchdb Error

tomsquest:
* Converting the status code from `int` to `string` was done using `string()` instead of `strconv.ItoA()`.
* Test case created.
* Minor fixes/cleanup

### DIFF
--- a/couchdb/errors.go
+++ b/couchdb/errors.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"strconv"
 )
 
 // This file contains error handling code for couchdb request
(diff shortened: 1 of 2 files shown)


nono:
Thanks for the PR. I was working on the same files, you should look at my PR: #33

tomsquest:
If there are conflicts between our two branches, could you cherry-pick/apply these/changes and closes this PR (I am not emotionally attached to it :wink: ).

tomsquest:
@nono Just thinking about it: the `status` field in ther Error json is a `string`. Shouldn'it it be kept as an `int`, or it could be renamed as `statusCode` (principle of least surprise). WDYT ?

nono:
@tomsquest we want to stay close to the couchdb API, and couchdb uses `status` with a string, so we will keep that

jinroh:
@tomsquest I've merged #33 and opened another PR #35 with your rebased changes.

PR #35, opened 2016-10-07
Fix status conversion in Couchdb Error (bis)

jinroh:
See #34.


nono:
LGTM. When travis build is done, we can merge it (if green).

Issue #36, opened 2016-10-10
CouchDB Doc interface


jinroh:
Working on the interaction between the VFS and Couch, I have some suggestions to change the current API of our data layer, mainly on the `couch.Db` type. I'll try to list the different options that I've seen and the and the one I think is best.

I hope this is not something part of a previous discussion. I have not seen it.

## `Doc` as a `map[string]interface{}`

The current one. While it's maybe the most direct and basic solution, and perfect for the lose `/data` api, it does not give us many possibilities for inner usages.

Working with map is not very idiomatic when we would like to enforce more type safety on the documents we store in database. And converting maps to structs is neither efficient nor desirable all over our code base.

Finally, map being implemented as a hashmap, it is less efficient than a packed `struct`.

## `Doc` as a `interface{}`

By using the empty iface type, we have one level more of indirection. It allows us to use type structs or maps, which one is better to represent our data.

Although we have to rely on reflection and convention (an `_id` field on the struct for instance) to solve the "id problem". We do not gain much in type safety... while Go can provide us better possibilities by representing this behaviour through methods.

## `Doc` as a `interface { SetID, GetID }`

Finally, we can enforce the existence of setter/getter on the type's methods to solve this issue. We of course still have one level on indirection more, but I think it is a useful abstraction in this case with better type safety and more idiomatic.

Regarding the `rev`, we can either enforce `SetRev/GetRev` methods on the doc types through the interface, or returning the `rev` at the return site of the `CreateDoc` func:

```go
type Doc interface {
  GetID() string
  SetID(string id)
}

func CreateDoc(..., doc Doc) (rev string, error)
```

I think this last solution is better. See an implementation [here](https://github.com/jinroh/cozy-stack/commit/c3bb5673afa9d12fefd1f277f036db5bb2b03d14).

What do you think ?

nono:
:+1: for GetRev/SetRev

@aenario you have also asked to add `Doctype` to the interface. Am I right?

For a map, it can return one the field. For a struct, it can be hardcorded:

```go
type File struct {
  // ...
}

func (f *File) Doctype() string {
  return "io.cozy.files"
}
```

jinroh:
Ok so we all agree an the third solution. Let's agree on the interface per se.

Can we agree on this ? :

```go
type Doc interface {
  ID() string
  Rev() string
  DocType() string // <- or a DocType type string maybe... we'll see later

  SetID(id string)
  SetRev(rev string)
}
```

[...]

PR #37, opened 2016-10-10
couch.Doc interface

jinroh:
See #36.

The `doctype` on the JSONDoc felt weird and we may fallback on this one... This change will affect the crud api since it adds a explicit `doctype` field to documents.. while the `_id` field also contains it. I'm not really sure about that.

Other than that, I don't see many issue.

### DIFF
--- a/couchdb/couchdb.go
+++ b/couchdb/couchdb.go
@@ -162,9 +162,7 @@ func ResetDB(dbprefix, doctype string) error {
 	return CreateDB(dbprefix, doctype)
 }
 
-func createDocOrDb(dbprefix string, doc Doc, response interface{}) (err error) {
-	doctype := doc.DocType()
-
+func createDocOrDb(dbprefix, doctype string, doc Doc, response interface{}) (err error) {
 	db := makeDBName(dbprefix, doctype)
 	err = makeRequest("POST", db, doc, response)
 	if err == nil || !isNoDatabaseError(err) {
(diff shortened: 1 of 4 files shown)


web/data/data.go
```diff
@@ -39,28 +39,28 @@ func getDoc(c *gin.Context) {
...
 
-	var doc couchdb.Doc
+	var doc couchdb.JSONDoc
 	if err := c.BindJSON(&doc); err != nil {
 		c.AbortWithError(http.StatusBadRequest, err)
 		return
 	}
 
-	err := couchdb.CreateDoc(prefix, doctype, doc)
+	doc["doctype"] = doctype
```
jinroh: 😕

nono:
What can we do to avoid the doctype in the JSON?

One possibility is to add a `ToJSON()` method to our interface and modify JSONDoc to:

```go
type JSONDoc struct {
  DocType string
  Data map[string]interface{}
}
```

aenario:
How about
```go
type JSON map[string]interface{}
func (d *JSONDoc) Doctype {
id, ok = d["_id"].(string)
if !ok || id=="" { 
    return ""
}
return id[0:strings.Index(id, "/")]
}
```

?

nono:
@aenario and how do you create a new document? You create the id in go, and not on relying on couchdb?

aenario:
Yep, i did not think this in depth...

Alternatively, we can keep the doctype outside of the structure itself, then we can imagine having two APIs : 

1. The full-of-string one for /data API (with `Create(prefix, doctype, doc)`)
2. The go clean one : 
```go
type File 
//... 
files = couchdb.Doctype(File.Doctype())
//...
func upload()...{
    files.Create(myFile)
}


```

couchdb/couchdb.go
```diff
@@ -19,13 +19,48 @@ type updateResponse struct {
...
+	DocType() string
+
+	SetID(id string)
+	SetRev(rev string)
+}
+
+// JSONDoc is a map representing a simple json object that implements
+// the Doc interface.
+type JSONDoc struct {
+	Type string
```
tomsquest: Why `Type` is exported since there is `DocType()` ?

nono: We want to specify the type when creating a document that we will save.

jinroh:
I want to keep this PR simple for now. We can still discuss a change of API or adding the notion of serialization/deserialization later on.

Tell me if its good to merge for you (I'll remove the revert commits).

PR #38, opened 2016-10-10
Add couchdb & HTTP API for update & delete

aenario:
Tests & doc incoming.

### DIFF
--- a/couchdb/couchdb.go
+++ b/couchdb/couchdb.go
@@ -47,7 +47,8 @@ func (j JSONDoc) Rev() string {
 
 // DocType returns the document type
 func (j JSONDoc) DocType() string {
-	return j["doctype"].(string)
+	qid := j["_id"].(string)
+	return qid[0:strings.Index(qid, "/")]
 }
 
 // SetID is used to set the identifier of the document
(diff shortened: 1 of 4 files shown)


couchdb/couchdb.go
```diff
@@ -37,17 +37,29 @@ type JSONDoc map[string]interface{}
...
+	if ok {
+		return id
+	}
+	return ""
 }
 
 // Rev returns the revision field of the document
 func (j JSONDoc) Rev() string {
-	return j["_rev"].(string)
+	id, ok := j["_rev"].(string)
```
nono: s/id/rev/

couchdb/couchdb.go
```diff
@@ -37,17 +37,29 @@ type JSONDoc map[string]interface{}
 
 // ID returns the identifier field of the document
```
nono: You should probably say that it is the quantified id, or give an example

couchdb/couchdb.go
```diff
@@ -37,17 +37,29 @@ type JSONDoc map[string]interface{}
...
 func (j JSONDoc) ID() string {
-	return j["_id"].(string)
+	id, ok := j["_id"].(string)
```
nono: s/id/qid/

couchdb/couchdb.go
```diff
@@ -154,12 +166,57 @@ func DeleteDB(dbprefix, doctype string) error {
...
-func ResetDB(dbprefix, doctype string) error {
-	err := DeleteDB(dbprefix, doctype)
-	if err != nil {
-		return err
+func ResetDB(dbprefix, doctype string) (err error) {
+	err = DeleteDB(dbprefix, doctype)
+	if err == nil {
+		return CreateDB(dbprefix, doctype)
+	}
+	return
```
nono: I'd prefer:

```go
func ResetDB(dbprefix, doctype string) error {
  err := DeleteDB(dbprefix, doctype)
  if err != nil {
    return err
  }
  return CreateDB(dbprefix, doctype)
}
```

nono: The goal is to make the code easier to change if we had to add a third step, like this:

```go
func ResetDB(dbprefix, doctype string) error {
  err := DeleteDB(dbprefix, doctype)
  if err != nil {
    return err
  }
  err = CreateDB(dbprefix, doctype)
  if err != nil {
    return err
  }
  return ThirdStep()
}
```

nono: And I think it's a golang convention to have the good case path in the final return, and errors in the if cases.

web/data/data_test.go
```diff
@@ -136,3 +143,100 @@ func TestWrongID(t *testing.T) {
 		assert.Equal(t, "missing", out["reason"], "should give a reason")
 	}
 }
+
+// @TODO uncomment me when we stop falling back to Host = dev
```
nono: Maybe we can use [Skip](https://golang.org/pkg/testing/#T.Skip) for this test

couchdb/couchdb.go
aenario: :+1: fixed in 9e1089c6643ac2cfee0405f0ea310abe444e1595

couchdb/couchdb.go
aenario: fixed in 9e1089c6643ac2cfee0405f0ea310abe444e1595

couchdb/couchdb.go
aenario: Yep, I just discovered this cool named return and may have been carried away :smile: 
Fixed in 9e1089c6643ac2cfee0405f0ea310abe444e1595

web/data/data_test.go
aenario: Done in 9e1089c6643ac2cfee0405f0ea310abe444e1595

PR #39, opened 2016-10-11
Vfs with couchdb

jinroh:
This PR implement an almost complete API to upload files and create directories.

Still missing (indicated by `@TODO`s in the code):
  - a proper implementation of our jsonapi: this PR introduces a temporary `JSONApier` interface to do the job
  - ~~proper error handling to make the creation of a file/directory "atomic" to make sure out couchdb and VFS are coherent: this requires a `couchdb.Delete/Update`~~
  - ~~`Mime` and `Class` fields in file attributes (not sure how we want to handle that)~~


web/files/file.go
```diff
@@ -0,0 +1,151 @@
...
+	md5H := md5.New() // #nosec
+	_, err = io.Copy(f, io.TeeReader(r, md5H))
+	if err != nil {
+		return err
+	}
+
+	calcMD5 := md5H.Sum(nil)
+	if !bytes.Equal(m.GivenMD5, calcMD5) {
+		err = errInvalidHash
+		return
```
nono: I think you can replace these 2 lines by this one:

```go
return errInvalidHash
```

web/files/file.go
```diff
@@ -0,0 +1,151 @@
...
+		if err == nil {
+			return
+		}
+		_, isCouchErr := err.(*couchdb.Error)
+		if isCouchErr {
+			couchdb.DeleteDoc(dbPrefix, doc)
+		} else {
+			fs.Remove(pth)
+		}
+	}()
```
nono: This error handler looks weird to me. I'd have simply use that:

```go
defer func() {
  if err != nil {
    fs.Remove(pth)
  }
}
```

What am I missing?

web/files/directory.go
```diff
@@ -0,0 +1,111 @@
...
+		"attributes": d.Attrs,
+	}
+	m := map[string]interface{}{
+		"data": dat,
+	}
+	return json.Marshal(m)
+}
+
+// CreateDirectory is the method for creating a new directory
+func CreateDirectory(m *DocMetadata, fs afero.Fs, dbPrefix string) (jsonapier jsonapi.JSONApier, err error) {
```
jinroh: This is what i did initially, and gometalinter hinted that I was returning a private struct on a public func.

I didn't really think it is necessary to make dirDoc a public struct. I could also make createDirectory a private func for now.

web/files/files_test.go
```diff
@@ -9,10 +12,13 @@ import (
 
 	"github.com/cozy/cozy-stack/web/middlewares"
 	"github.com/gin-gonic/gin"
+	"github.com/sourcegraph/checkup"
 	"github.com/spf13/afero"
 	"github.com/stretchr/testify/assert"
 )
 
+const COUCHURL = "http://localhost:5984/"
```
jinroh: copypasta. done d72690e147da9743ead284cd008ed97b72dd69c9

web/files/file.go
jinroh: done fbb9fb1c850b6b3b4b984e3b9e960d3d5de3270c

web/files/directory.go
nono: Well, I think that making `dirDoc` public may be the good choice. We want other packages to use `CreateDirectory` and it make senses that they know what is a `dirDoc` to use the returned value.

jinroh: Let's go for it. I'll do the same for `fileDoc`.

jinroh: done 83f36d4043d3247b2ac31b354bd64af89a434ebb

nono:
> `Mime` and `Class` fields in file attributes (not sure how we want to handle that)

`Mime` is the `Content-Type` HTTP header in the request (you can use `application/octet-stream` for the default value).

`Class` is computed from `Mime`. In the actual cozy, each application has its own rules for setting the `class`. https://github.com/cozy/cozy-files/blob/ea9f31c772c7002ef5853361ba8ffe6a6942b5b3/server/helpers/file.coffee#L18-L25 for example. For the moment, let's say we use the first part of the `Mime` (`image`, `audio`, `video`, `text`, etc.) and we will improve this for specific mime-type later.

web/files/file.go
```diff
@@ -163,3 +167,14 @@ func copyOnFsAndCheckIntegrity(m *DocMetadata, fs afero.Fs, pth string, r io.Rea
...
 }
+
+func extractMimeAndClass(contentType string) (mime, class string) {
+	if contentType == "" {
+		contentType = DefaultContentType
+	}
+
+	// @TODO improve for specific mime types
+	mime = contentType
+	class = contentType[:strings.Index(contentType, "/")]
```
nono:
```
class = strings.Split(contentType, "/")[1]
```

is more explicit, no?

jinroh:
I've just noted that I missed the file size. We should probably handle the `Content-Length` and return an error on:

  - length mismatch (contentlength != copied)
  - content-length not specified

nono:
> length mismatch (contentlength != copied)

Yes!

> content-length not specified

We can let `content-length` be optional.

web/files/file.go
nono: it's OK for me

jinroh:
Added content length mismatch support (22c2cd5b61605c5552b180400078b914df51da1a) and removed charset from mime (3022445f0a9b135138cdba8f031180412a394ab0).

PR #41, opened 2016-10-12
[VFS] Download files

jinroh:
This PR add the API to download a file by id and path:

  - `GET /files/:file-id`
  - `GET /files/download?path=`

I started by implementing a simple `io.Copy` streaming (8d234b0ed1c8772dc4fc2a7f5d47d8d9662d372d) and finally decided to use `http.ServeContent`. It allowed me to add support `Range`, `If-Modified-Since` and `If-None-Match` headers.

I'm just checking you're all OK with that before updating the documentation to reflect these possibilities.

### DIFF
--- a/web/files/file.go
+++ b/web/files/file.go
@@ -5,7 +5,8 @@ import (
 	"crypto/md5" // #nosec
 	"encoding/json"
 	"io"
-	"os"
+	"net/http"
+	"path"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


web/files/file.go
```diff
@@ -87,6 +89,70 @@ func (f *FileDoc) ToJSONApi() ([]byte, error) {
...
+// offering support to Range, If-Modified-Since and If-None-Match
+// requests. It uses the revision of the file as the Etag value for
+// non-ranged requests
+//
+// The content disposition is inlined.
+func ServeFileContent(fileDoc *FileDoc, req *http.Request, w http.ResponseWriter, fs afero.Fs) (err error) {
+	attrs := fileDoc.Attrs
+	header := w.Header()
+	header.Set("Content-Type", attrs.Mime)
+	header.Set("Content-Disposition", "inline; filename="+attrs.Name+"")
```
nono: Why `+""`?

web/files/file.go
```diff
@@ -87,6 +89,70 @@ func (f *FileDoc) ToJSONApi() ([]byte, error) {
...
+	return
+}
+
+// ServeFileContentByPath replies to a http request using the content
+// of a file identified by its full path on the VFS. Unlike
+// ServeFileContent, this method does not require the full file
+// document but only its path.
+//
+// It also uses internally http.ServeContent but does not provide an
+// Etag.
```
nono: It's something that we will probably want to change later. Having Etag on this route would be very nice, because this route will be used to serve JS and CSS files for client-side apps and we want the browsers to avoid to download them each time the app loads.

web/files/files.go
```diff
@@ -156,8 +158,51 @@ func CreationHandler(c *gin.Context) {
...
+	} else {
+		var doc *FileDoc
+		doc, err = GetFileDoc(string(FileDocType)+"/"+fileID, dbPrefix)
+		if err == nil {
+			err = ServeFileContent(doc, c.Request, c.Writer, storage)
+		}
+	}
+
+	if err != nil {
+		c.AbortWithError(http.StatusInternalServerError, err)
```
nono: Don't forget to send a 404 error when the file is missing

jinroh: just in time 😄 3e173ed56bbc1015b26a7912e2b7d1094f3b3a76

web/files/file.go
jinroh: Shall we do a CRC or something like that for some contents (or all in this route...) ?

jinroh: OK, I'm not really decided on how to do this lookup for now..

BTW, I used the doc revision instead of the MD5 as an Etag, I'll change that (the MD5 is more appropriate since the rev can change without the file content changing)

PR #42, opened 2016-10-12
Remove qid vs id, add _type field

aenario:


couchdb/couchdb.go
```diff
@@ -58,21 +58,34 @@ func (j JSONDoc) Rev() string {
...
 func (j JSONDoc) DocType() string {
-	qid, ok := j["_id"].(string)
-	if !ok {
-		return ""
-	}
-	return qid[0:strings.Index(qid, "/")]
+	return j.Type
 }
 
 // SetID is used to set the qualified identifier of the document
```
jinroh: `s/qualified indentifier/identifier/` (there may be other comments using this term)

PR #46, opened 2016-10-13
Format errors as JSON-API

nono:
~~@aenario @jinroh I'm working on JSON-API formatting. This PR is not yet ready to be merged, but you may want to look at it to see where I'm going~~

It's now ready!


jinroh:
@nono no `api2go` finally, or it's coming in ?

web/jsonapi/errors.go
```diff
@@ -0,0 +1,91 @@
...
+	case vfs.ErrDocDoesNotExist:
+		return NotFound(err)
+	case vfs.ErrParentDoesNotExist:
+		return NotFound(err)
+	case vfs.ErrDocTypeInvalid:
+		return InvalidAttribute("type", err)
+	case vfs.ErrIllegalFilename:
+		return InvalidParameter("folder-id", err)
+	case vfs.ErrInvalidHash:
+		return InvalidParameter("Content-MD5", err)
```
jinroh: Same remark for `ErrContentLengthMismatch` which can fall into the same category.

nono: You are right, I'm going to fix that!

nono:
@jinroh errors in `api2go` was identified as a limitation. It's tied to using the routing stuff from `api2go`. I don't want to use the routing stuff of `api2go`, only the serialization/deserialization of objects. So, I have to code it myself ;-)

aenario:
I am OK to merge this PR as is, but in the future think we should have a similar structure for error handling in both CRUD and VFS api.

In CRUD, I opted to use the gin error handling mechanism, with the combo `HTTPStatus()`, `c.AbortWithError` and then an error handling middleware which renders the couchdb.Error to JSON if it is one. This is not so clean : 

- HTTPStatus is only useful because, we need the status code to call AbortWithError. If there was another (*Context) function equivalent to c.Error(err) && c.Abort(), it could all be handled in the error handling middleware (like you did in jsonapi.AbortWithError())
- gin wraps the errors we pass into gin.Error and maintains a slice of them, which is pretty useless, but it can be configured to log some errors, which might prove useful.

In VFS, you went with having a `jsonapi.AbortWithError()` which immediately do the render, without passing the error to gin, which is way cleaner and doesnt involve 

---------------------------------------------------

In the future we could

A. switch to echo v3, where we simply return error and then have an error handling function wich does all the triaging/formating.
B. If we stay with gin, we might do something like :
- a common interface, `type StackError interface  Error(), Title(), Reason(), JSON(), JSONAPI() StatusCode()` implemented by both VFS errors, couchdb errors, other errors we will need in the future and a oneline constructor.
- a function Abort(context, StackError), which call c.Error(), c.JSON(), c.Abort() and renders the errors with proper formating depending on c.NegotiateFormat.


my 2c

jinroh:
Rebased and merged 226e75a0f8989cfec1f5deded17e0560689113d8

PR #48, opened 2016-10-14
[VFS] Update file content

jinroh:
This PR introduce the new route `PUT /files/:files-id` along with a new method `vfs.ModifyFileContent`.

==

*Some notes*:

In order to add this feature, I also removed the intermediary (and awkward) struct `vfs.DirAttributes` that was used to stack all the contents of a creation request for files *and* directories. Now, `web/file` can use `vfs.NewFileDoc` or `vfs.NewDirDoc` directly and appropriately.

I also increased the gometalinter dupl-threshold, because it annoyed me on some trivial code (in test)...

### DIFF
--- a/vfs/directory.go
+++ b/vfs/directory.go
@@ -8,14 +8,6 @@ import (
 	"github.com/spf13/afero"
 )
 
-// DirAttributes is a struct with the attributes of a directory
-type DirAttributes struct {
-	Name      string    `json:"name"`
-	CreatedAt time.Time `json:"created_at"`
-	UpdatedAt time.Time `json:"updated_at"`
-	Tags      []string  `json:"tags"`
-}
-
 // DirDoc is a struct containing all the informations about a
 // directory. It implements the couchdb.Doc and jsonapi.JSONApier
 // interfaces.
@@ -24,12 +16,16 @@ type DirDoc struct {
 	DID string `json:"_id,omitempty"`
 	// Directory revision
 	DRev string `json:"_rev,omitempty"`
-	// Directory attributes
-	Attrs *DirAttributes `json:"attributes"`
+	// Directory name
+	Name string `json:"name"`
 	// Parent folder identifier
 	FolderID string `json:"folderID"`
 	// Directory path on VFS
 	Path string `json:"path"`
+
+	CreatedAt time.Time `json:"created_at"`
+	UpdatedAt time.Time `json:"updated_at"`
+	Tags      []string  `json:"tags"`
 }
 
 // ID returns the directory qualified identifier (part of couchdb.Doc interface)
@@ -63,11 +59,17 @@ func (d *DirDoc) SetRev(rev string) {
 // ToJSONApi implements temporary interface JSONApier to serialize
 // the directory document
 func (d *DirDoc) ToJSONApi() ([]byte, error) {
+	attrs := map[string]interface{}{
+		"name":       d.Name,
+		"created_at": d.CreatedAt,
+		"updated_at": d.UpdatedAt,
+		"tags":       d.Tags,
+	}
 	data := map[string]interface{}{
 		"type":       d.DocType(),
 		"id":         d.ID(),
 		"rev":        d.Rev(),
-		"attributes": d.Attrs,
+		"attributes": attrs,
 	}
 	m := map[string]interface{}{
 		"data": data,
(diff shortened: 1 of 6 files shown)


vfs/directory.go
```diff
@@ -63,42 +59,50 @@ func (d *DirDoc) SetRev(rev string) {
...
-	if m.docType != FolderDocType {
-		err = ErrDocTypeInvalid
-		return
-	}
-
-	pth, _, err := createNewFilePath(m, fs, dbPrefix)
-	if err != nil {
+// NewDirDoc is the DirDoc constructor. The given name is validated.
+func NewDirDoc(name, folderID string, tags []string) (doc *DirDoc, err error) {
+	if err = checkFileName(name); err != nil {
```
aenario: Coding style decision to be made here : 
@nono reverted this kind of set-output-value-and-return-empty in one of its PR (although it was only 1 value)

We need to pick a style and stick to it. Personnaly, I am not sure : this is cleaner than `return nil, err` but the intent is less readable. When in doubt, it is better to pick more readability rather than a few less chars.

vfs/file.go
```diff
@@ -192,38 +202,96 @@ func CreateFileAndUpload(m *DocAttributes, fs afero.Fs, dbPrefix string, body io
...
+	newDoc.CreatedAt = oldDoc.CreatedAt
+	newDoc.UpdatedAt = updateDate
+
+	return couchdb.UpdateDoc(dbPrefix, newDoc)
 }
 
-func copyOnFsAndCheckIntegrity(m *DocAttributes, fs afero.Fs, pth string, r io.ReadCloser) (written int64, md5Sum []byte, err error) {
-	f, err := fs.Create(pth)
+func copyOnFsAndCheckIntegrity(pth string, givenMD5 []byte, executable bool, fs afero.Fs, r io.Reader) (written int64, md5Sum []byte, err error) {
+	// We want to write only (O_WRONLY), create the file if it does not
```
aenario: This comment is at the wrong place (should be before L279). Another comment about "magic numbers" 0755 & 644 would be cool.

vfs/file.go
```diff
@@ -192,38 +202,96 @@ func CreateFileAndUpload(m *DocAttributes, fs afero.Fs, dbPrefix string, body io
...
 	}
 
-	defer f.Close()
-	defer r.Close()
+	err = fs.Chmod(pth, mode)
+	if err != nil {
+		return
+	}
+
+	defer func() {
```
aenario: The defer call should be before fs.Chmod, no?

web/files/files.go
```diff
@@ -134,11 +192,58 @@ func ReadHandler(c *gin.Context) {
...
 	router.POST("/", CreationHandler)
 	router.POST("/:folder-id", CreationHandler)
+
+	router.PUT("/:file-id", OverwriteFileContentHandler)
+}
+
+func getFsAndDBPrefix(c *gin.Context) (fs afero.Fs, dbPrefix string, err error) {
+	instance := middlewares.GetInstance(c)
+	dbPrefix = instance.GetDatabasePrefix()
+	fs, err = instance.GetStorageProvider()
```
aenario: [Future] May be all this instanciating of request specific structs (instance, dbprefix, fs) could be moved to middlewares and c.MustGet.

web/files/files_test.go
```diff
@@ -332,7 +443,9 @@ func TestMain(m *testing.M) {
 	router.Use(injectInstance(instance))
 	router.POST("/files/", CreationHandler)
 	router.POST("/files/:folder-id", CreationHandler)
-	router.GET("/files/:file-id", ReadHandler)
+	router.PUT("/files/:file-id", OverwriteFileContentHandler)
+	router.HEAD("/files/:file-id", ReadFileHandler)
```
nono: I'm curious. Do you know what we have to do to support HEAD with Gin?

vfs/directory.go
jinroh: Sure.. I find named returns useful when I change the signature of the function (to have smaller diffs to..). But in the end I don't really care that much. @nono an opinion ?

vfs/file.go
jinroh: You're right, done https://github.com/cozy/cozy-stack/pull/48/commits/108945bd0af35731fb716735feefc81ff8317788

vfs/directory.go
jinroh: Here I would have to do `return "", nil, err` which I think is a bit annoying..
Although for functions returning only 0 or 1 value + error, `return nil, err` may be a bit cleaner...

nono: I like to give a name to return values to explain what they are. But I don't like the naked returns. It's really too easy to shadow `err` by mistake. This is an example of what I means: https://play.golang.org/p/0ZZ7bXElXq.

web/files/files_test.go
nono: Good to know!

PR #49, opened 2016-10-17
[VFS] Concurrent access

jinroh:
Fix concurrent access to file and directory creation and add tests on the matter.

  - rely on one call to the FS using `O_CREATE|O_EXCL`
  - rely on `MkDir` atomicity *before* adding the document in Couch
  - rely on `Rename` atomicity to modify file content using a temporary file

--

This should be merged after #48

### DIFF
--- a/vfs/directory.go
+++ b/vfs/directory.go
@@ -97,10 +97,13 @@ func NewDirDoc(name, folderID string, tags []string) (doc *DirDoc, err error) {
 }
 
 // CreateDirectory is the method for creating a new directory
-func CreateDirectory(doc *DirDoc, fs afero.Fs, dbPrefix string) error {
-	var err error
-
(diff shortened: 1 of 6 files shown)


jinroh:
Note that I now use a `file://` instead of `mem://` storage since `afero/memmap` lacks some functionalities (mainly O_EXCL).

vfs/directory.go
```diff
@@ -113,13 +116,5 @@ func CreateDirectory(doc *DirDoc, fs afero.Fs, dbPrefix string) error {
...
-	if err = couchdb.CreateDoc(dbPrefix, doc); err != nil {
-		return err
-	}
-
-	if err = fs.Mkdir(pth, 0755); err != nil {
-		return err
-	}
-
-	return nil
+	return couchdb.CreateDoc(dbPrefix, doc)
```
aenario: Should we not remove the folder in case of couchdb create error to ensure they stay in sync?

PR #51, opened 2016-10-17
[VFS] Patch file meta

jinroh:
Adds the `PATCH /files/:file-id` route to modify the metadata of a file, move it and rename it.

I've also added a `vfs.Context` struct to stop passing around fs / dbprefix values...

### DIFF
--- a/vfs/file.go
+++ b/vfs/file.go
@@ -123,9 +123,9 @@ func NewFileDoc(name, folderID string, size int64, md5Sum []byte, mime, class st
 
 // GetFileDoc is used to fetch file document information form our
 // database.
-func GetFileDoc(fileID, dbPrefix string) (doc *FileDoc, err error) {
+func GetFileDoc(c *Context, fileID string) (doc *FileDoc, err error) {
 	doc = &FileDoc{}
-	err = couchdb.GetDoc(dbPrefix, string(FileDocType), fileID, doc)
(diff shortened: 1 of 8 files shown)


vfs/file.go
```diff
@@ -218,85 +218,154 @@ func CreateFileAndUpload(doc *FileDoc, fs afero.Fs, dbPrefix string, body io.Rea
...
+		}
+		newpath = path.Join(parentDoc.Path, olddoc.Name)
+		newfolderID = data.FolderID
+	}
+
+	if data.Tags != nil {
+		newtags = data.Tags
+	}
+
+	newdoc, err = NewFileDoc(
```
aenario: Why not keeping oldpath and modifying the old doc in place ?

vfs/vfs.go
```diff
@@ -26,6 +25,26 @@ const (
...
+// file and directory documents.
+type DocMetaAttributes struct {
+	Name     string   `json:"name,omitempty"`
+	FolderID string   `json:"folderID,omitempty"`
+	Tags     []string `json:"tags,omitempty"`
+}
+
+// Context is used to convey the afero.Fs object along with the
+// CouchDb database prefix.
+type Context struct {
```
aenario: [Future] This could get complicated fast between gin.Context, couchdb.Prefix (+ some auth later), vfs.Context, middlewares.Instance. We should at least have some out-of-code doc on it, but better a proper discussion and normalisation.

PR #52, opened 2016-10-17
Add more information to instances management

aenario:


docs/instance.md
```diff
@@ -1,7 +1,17 @@
 Instances
 =========
 
-**TODO** it's just a first draft that needs to be completed and more explicit
+**TODO** it's still a work in progress that needs to be completed.
+
+A single cozy-stack can manage several instances. Requests to different instances are identified through the `Host` HTTP Header, any reverse proxy placed in front of the cozy-stack should forward this header. 
+
+To simplify development, a `dev` instance name is used when no Host Header is provided. This behaviour will be kept when the stack is started in dev mode but will be blocked
```
nono: It looks like the sentence was not finished

docs/instance.md
```diff
@@ -11,16 +21,42 @@ An instance is created on the command line:
...
+With some possible additional options
 
-It registers the instance in a global database (**TODO** which one), and
-creates some databases for these doctypes:
+- `--locale <lang>` The
+- `--email <email>`
+- `--environment <dev/test/production>`
+- `--apps <app1,app2,app3>`
+- `--home htts://github.com/cozy/cozy-home`
+- `--onboarding htts://github.com/cozy/cozy-home`
```
nono: s/htts/https/g

docs/instance.md
```diff
@@ -11,16 +21,42 @@ An instance is created on the command line:
 ```sh
 $ cozy-stack instances add <domain>
 ```
+With some possible additional options
 
-It registers the instance in a global database (**TODO** which one), and
-creates some databases for these doctypes:
+- `--locale <lang>` The
```
nono: What the `The` means?

docs/instance.md
```diff
@@ -31,9 +67,55 @@ Then, it creates some folders:
 - `/Documents/Music`, with the id `io.cozy.folders-music`
 - `/Documents/Videos`, with the id `io.cozy.folders-videos`
 
-The ids are forced to known values. So, even if these folders are moved or
-renamed, they can still be found for the permissions.
+**The ids are forced to known values :**  even if these folders are moved or
```
nono: In english, there is no space before `:`

docs/instance.md
```diff
@@ -31,9 +67,55 @@ Then, it creates some folders:
...
+renamed, they can still be found for the permissions. 
+
+**The names are localized :** If a locale is provided through the CLI, the folders will be created with 
+names in this locale. Otherwise, theses folders will be created in english and renamed to localized name the first time the locale is set (during onboarding).
+
+Then it creates the basic settings 
+
+- `owner_email` if an email was provided through the CLI
+- `locale` if a locale was provided through the CLI
+- `home_app` URL of the app to use as home, github.com/cozy/cozy-home by default
```
nono: I think in the settings we will have the slug of the home. The home app will be installed like any other app. The only special thing is that the cozy-stack will redirect the user to the home after a successful authentication.

docs/instance.md
```diff
@@ -31,9 +67,55 @@ Then, it creates some folders:
...
+- `onboarding_app` URL of the app to use as onboarding, github.com/cozy/cozy-onboarding by default
+- `market_url` URL of the market place to fetch manifests of installable applications.
```
nono: s/market/registry/ maybe? In my mind, the market/store is the app inside the cozy, and the registry is a list of available apps on a server.

docs/instance.md
```diff
@@ -31,9 +67,55 @@ Then, it creates some folders:
...
+
+An instance is renamed through the command line.
+
+```sh
+$ cozy-stack rename <olddomain> <newdomain>
+```
+
+Renaming an instance only change the HostName in global/instances base.
+
+TODO : should the map Host -> instance be partially cached in cozy-stack RAM ?
```
nono: Globally, we will have cache for the cozy-stack. For single host, the cache will be in the cozy-stack RAM. For multiple servers, the cache will be in redis, so it can be shared between the servers. I don't think we should mention cache in this document.

aenario:
I fixed all your comments except for the home URL vs slug one as it raise more questions.

What is the relation between URL and slug ?
My understanding of this feature was to easily install / update different versions of the core app
(https://github.com/cozy/cozy-home vs https://github.com/cozy-partner1/cozy-home).

Should both these apps have `home` slug ?  Is slug provided in the app manifest / the registry ? Or computed from the URL ?

nono:
A default value for the slug is given in the manifest, but it should be possible to override it. In your example, the 2 apps can have the same default value for the slug, but if the user installs both of them, the cozy-stack will compute a new slug for the second app.

aenario:
Changed to slug like.

docs/instance.md
```diff
@@ -11,16 +21,42 @@ An instance is created on the command line:
...
+- `--home <cozy-home>`
+- `--onboarding <cozy-onboarding>`
+- `--registry https://registry.cozycloud.cc`
+
+It registers the instance in a global couchdb database `global/instances`
+```json
+{
+    "hostname": "example.cozycloud.cc",
+    "dbprefix": "example-clozycloud-cc/",
+    "fsroot": "/usr/lib/cozy/example.cozycloud.cc/fs/"
```
nono: `/var/lib` would be better than `/usr/lib`. It can be sometimes a separate partition, with more disk space.

docs/instance.md
```diff
@@ -31,9 +67,53 @@ Then, it creates some folders:
...
 
-Finally, the default application for the current environment are installed.
+**The names are localized:** If a locale is provided through the CLI, the folders will be created with
+names in this locale. Otherwise, theses folders will be created in english and renamed to localized name the first time the locale is set (during onboarding).
+
+Then it creates the basic settings
+
+- `owner_email` if an email was provided through the CLI
+- `locale` if a locale was provided through the CLI
+- `home_app` URL of the app to use as home, github.com/cozy/cozy-home by default
```
nono: Slug, not URL, we have said :wink:

docs/instance.md
```diff
@@ -11,16 +21,42 @@ An instance is created on the command line:
...
+```
 
-It registers the instance in a global database (**TODO** which one), and
-creates some databases for these doctypes:
+and creates the proper databases ($PREFIX/$DOCTYPE) for these doctypes:
 
 - `io.cozy.apps`
 - `io.cozy.manifests`
 - `io.cozy.files`
+- `io.cozy.folders`
```
nono: [future] I think that we will merge the files and folders inside the same database, so that we can send several documents to update in the same request (`_bulk_docs`)

PR #54, opened 2016-10-18
[VFS] Get file metadata

jinroh:
Use CouchDB indexes to fetch files by path and export the content with `GET /file/metadata`.

Note that the functionality is not working since we are still laking the creation of indexes at startup.

### DIFF
--- a/web/files/files.go
+++ b/web/files/files.go
@@ -231,15 +231,58 @@ func modFileHandler(vfsC *vfs.Context, req *http.Request, patchData *jsonData) (
 	return doc, nil
 }
 
-// ReadFileHandler handles all GET requests on /files/:file-id aiming
-// at downloading a file. It serves two main purposes in this regard:
+// ReadFileHandler handles all GET requests on /files/:file-id,
+// /files/download and /files/metadata and dispatches to the right
+// handler. See ReadMetadataHandler and ReadFileContentHandler.
+func ReadFileHandler(c *gin.Context) {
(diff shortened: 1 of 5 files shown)


web/files/files.go
```diff
@@ -251,21 +294,27 @@ func ReadFileHandler(c *gin.Context) {
 
 	// Path /files/download is handled specifically to download file
 	// form their path
+	var doc *vfs.FileDoc
+	var disposition string
 	if fileID == "download" {
-		pth := c.Query("path")
-		err = vfs.ServeFileContentByPath(vfsC, pth, c.Request, c.Writer)
+		disposition = "attachment"
```
nono: I'm not sure it's a good idea to use an attachment disposition here. `/files/download` will be used by the clientside apps to load their JS, CSS and images, and I don't know how the browsers handle the attachment disposition for these cases. We can start with that and see later if it causes some issues.

nono:
LGTM

aenario:
I thought we did not want to have a "path" field in couchdb ? 

There is a tradeof here : 

- having this field makes GET simple & fast, but Renaming & Moving very hard (what happens if we cant change one file's path, ect ...).
- not having this field makes Renaming & Moving trivial (change Name / ParentID), but makes GET by path more complex : O(file's depth). Unless we use the VFS, to find doc's couchdb ID from its path.

jinroh:
Yes, the path field is temporary. We'll remove it when implementing renaming directories.

PR #55, opened 2016-10-19
JSON-API serialization

nono:
It should be enough for the files API (serialization only, unmarshalling in a future PR).

### DIFF
--- a/web/jsonapi/data.go
+++ b/web/jsonapi/data.go
@@ -65,21 +65,25 @@ type ObjectMarshalling struct {
 // MarshalObject serializes an Object to JSON.
 // It returns a json.RawMessage that can be used a in Document.
 func MarshalObject(o Object) (json.RawMessage, error) {
-	self := o.SelfLink()
 	id := o.ID()
-	o.SetID("")
 	rev := o.Rev()
+	self := o.SelfLink()
+	rels := o.Relationships()
+
+	o.SetID("")
 	o.SetRev("")
 	defer func() {
 		o.SetID(id)
 		o.SetRev(rev)
 	}()
+
 	data := ObjectMarshalling{
-		Type:       o.DocType(),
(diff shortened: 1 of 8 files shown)


web/jsonapi/data.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/cozy/cozy-stack/couchdb"
+)
+
+// Object is an interface to serialize something to a JSON-API Object
+type Object interface {
+	couchdb.Doc
+	SelfLink() string
+	Relationships() RelationshipMap
+	Included() []Object
+}
```
jinroh: Wouldn't it be useful to inherit the `json.Marshaler` interface here ?

jinroh:
LGTM 👍

web/jsonapi/data.go
nono: It's not in couchdb.Doc?

jinroh: Nope. I think it makes sense to put it in `jsonapi.Object` to make it clear that we are talking about the serialisation for the public API side (and not couch for instance)...

web/jsonapi/jsonapi.go
```diff
@@ -12,18 +12,43 @@ import (
...
+	if err != nil {
+		AbortWithError(c, InternalServerError(err))
+		return
+	}
+	doc := Document{
+		Data:     &data,
+		Links:    links,
+		Included: included,
+	}
+	body, err := json.Marshal(doc)
```
jinroh: Also, the `included` elements are not serialized adequatly here.

web/jsonapi/data.go
nono: Well, in fact, it looks like `struct`s don't implement the `json.Marshaler` interface, even if we can serialize them to JSON.

PR #56, opened 2016-10-19
[VFS] Change directory metadata, move and rename them

jinroh:
I opted for an intermediary solution for now to handle path and move actions: remove the `Path` field from `FileDoc` and keep it in `DirDoc`.

Files and Directories are still in different databases, but this will be removed soon.

### DIFF
--- a/vfs/directory.go
+++ b/vfs/directory.go
@@ -80,9 +80,9 @@ func (d *DirDoc) Included() []jsonapi.Object {
 	return []jsonapi.Object{}
 }
 
-func fetchChildren(c *Context, parent *DirDoc, doctype DocType, docs interface{}) error {
+func fetchChildrenDeep(c *Context, parent *DirDoc, doctype DocType, docs interface{}) error {
 	req := &couchdb.FindRequest{
-		Selector: mango.Equal("folder_id", parent.ID()),
+		Selector: mango.StartWith("path", parent.Path+"/"),
 	}
 	return couchdb.FindDocs(c.db, string(doctype), req, docs)
 }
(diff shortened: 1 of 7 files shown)


jinroh:
NOTE: They are still many things that we can improve in term of number of roundtrip to couchdb per requests. This is only a first implementation.

jinroh:
Rebased on #55 (to merge after)

web/files/files.go
```diff
@@ -208,46 +218,61 @@ func ModificationHandler(c *gin.Context) {
 	}
 
 	patchData := container.Data
-	docType, err := vfs.ParseDocType(patchData.Type)
+
+	fileID := c.Param("file-id")
+
+	var docType vfs.DocType
+	if fileID == MetadataPath {
```
aenario: It would be clearer to have two handler, 
```
router.GET("/metadata", ModificationMetadataHandler)
router.GET("/:file-id", ModificationContentHandler)
```

web/files/files.go
```diff
@@ -208,46 +218,61 @@ func ModificationHandler(c *gin.Context) {
...
-		// @TODO
-		err = fmt.Errorf("Not implemented")
+	var doc couchdb.Doc
+	if fileID == MetadataPath {
+		switch docType {
+		case vfs.FileDocType:
+			doc, err = vfs.GetFileDocFromPath(vfsC, c.Query("path"))
+		case vfs.FolderDocType:
+			doc, err = vfs.GetDirectoryDocFromPath(vfsC, c.Query("path"))
+		}
```
aenario: What happens if the doctype is neither ? We should return a readable error.

vfs/directory.go
```diff
@@ -126,3 +161,111 @@ func CreateDirectory(c *Context, doc *DirDoc) (err error) {
...
+	}
+
+	err = couchdb.UpdateDoc(c.db, newdoc)
+	return
+}
+
+// @TODO remove this method and use couchdb bulk updates instead
+func bulkUpdateDocsPath(c *Context, olddoc *DirDoc, newpath string) error {
+	var children []*DirDoc
+	err := fetchChildren(c, olddoc, FolderDocType, &children)
```
aenario: I think you are only changing the path of immediate children. It should rename all descendants folder.

aenario: You can use mango.StartWith to find them.

web/files/files.go
jinroh: This error is handled by the `ParseDocType` function. But I can be more explicit by having another check.

vfs/directory.go
```diff
@@ -126,3 +161,111 @@ func CreateDirectory(c *Context, doc *DirDoc) (err error) {
...
+	errc := make(chan error)
+
+	for _, child := range children {
+		go func(child *DirDoc) {
+			child.Path = path.Join(newpath, child.Name)
+			errc <- couchdb.UpdateDoc(c.db, child)
+		}(child)
+	}
+
+	for i := 0; i < len(children); i++ {
```
nono: I'd have written:

```go
for range children {
```

to make it explicit we are looping on the same thing.

web/files/files.go
jinroh: The reason I did not do that is that it appears that Gin does not allow it... I get a panic `wildcard route ':file-id' conflicts with existing children in path '/files/:file-id`

vfs/directory.go
jinroh: My bad...

web/files/files.go
aenario: Wow, this appears to be an heated discussion on both gin and the underlying httprouter. It should be solved by httprouter v2 (no timeline) or when/if we move to echo. 

Meanwhile however, it might be a good idea to be futurerproof and have the code be closer to the intent, may be with a anonymous function to do the split. something like
```go
router.GET("/:file-id", func (c *gin.Context) {
// dropping the constant as we only need it once and it's part of routing
 if c.Param("file-id") == "metadata" {      ModificationMetaHandler(c) 
 }else{ ModificationContentHandler(c) }
});
```

web/files/files.go
aenario: If it's handled elsewhere, nevermind :-)

PR #59, opened 2016-10-26
[VFS] Get dir meta

jinroh:
This PR is quite simple but is has some implications and changes regarding the current doc.

By having files and directories sharing the same doctype `io.cozy.files`, I made these changes:

  - `DirDoc` and `FileDoc` now have a `Type string` field (either `"file"` or `"directory"`)
  - The ID of the root directory is `io.cozy.files.rootdir`

I did not change **input** side of the API, which still has the notion of `io.cozy.folders`. So clearly, this PR is a work in progress and I would understand if we don't want to merge it as-is.

--

Note: the notion of RootDirectory is handled "manually" in the code. I did not introduce a special object in the database.

### DIFF
--- a/vfs/directory.go
+++ b/vfs/directory.go
@@ -1,7 +1,6 @@
 package vfs
 
 import (
-	"encoding/json"
 	"fmt"
 	"os"
 	"path"
@@ -22,9 +21,9 @@ type DirDoc struct {
 	// from couch.
 	Type string `json:"type"`
 	// Qualified file identifier
-	DID string `json:"_id,omitempty"`
+	ObjID string `json:"_id,omitempty"`
 	// Directory revision
-	DRev string `json:"_rev,omitempty"`
+	ObjRev string `json:"_rev,omitempty"`
 	// Directory name
 	Name string `json:"name"`
 	// Parent folder identifier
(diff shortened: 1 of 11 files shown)


vfs/directory.go
```diff
@@ -257,6 +335,45 @@ func bulkUpdateDocsPath(c *Context, olddoc *DirDoc, newpath string) error {
...
+	err = couchdb.FindDocs(c.db, FsDocType, req, &docs)
+	if err != nil {
+		return
+	}
+
+	var t struct {
+		Type string `json:"type"`
+	}
+
+	// @NOTE: this unmarshaling technique is very slow...
```
aenario: How about having a common struct for unmarshalling, that can then be converted to File or Directory ?

jinroh: The only way I found to do that was to have a wrapper struct `{ Type string, Data json.RawMessage }`, and I wanted to avoid nested struct for now...

Also the slowness is relative.. I measured ~18ms for 1000+ files, which is still quite low compared to CouchDB response which took about 300ms.

jinroh:
I'll update the docs.

vfs/directory.go
nono: I think what @aenario has in mind is something like that:

```go
type FileOrFolder struct {
  Type string
  ID string
  Name string `json:"name"`
  MD5Sum     []byte    `json:"md5sum"`  // Only used by file
  // ...
}

func Foo() {
  var f FileOrFolder
  err = json.NewDecoder(resp.Body).Decode(&f)
  if f.Type == "io.cozy.folders" {
    folder := Folder{
      ID: f.ID,
      Name: f.Name,
      // ...
    }
    doSomethingWithAFolder(folder)
  } else {
    file := File{
      ID: f.ID,
      Name: f.Name,
      MD5Sum: f.MD5Sum,
      // ...
    }
    doSomethingWithAFile(file)
  }
}
```

jinroh:
I updated the docs to reflect the current implementation. The docs are an intermediary state now. As you can see, they is still the notion of `io.cozy.folders`.

aenario:
I think this is good to merge, let's put a card on Technical Debt and fine-tune it when we next work here.

PR #60, opened 2016-10-27
[VFS] Remove the requirement to pass Type query param for modify/read

jinroh:
This PR removes the need to pass `?Type=` query param for `GET / PATCH` requests.


nono:
LGTM

PR #62, opened 2016-10-28
[VFS] Tags replace on patch

jinroh:
**To merge after #61**

In current version of patching, the tags field was in append mode only. This had two bad side-effect, IMHO:
- stateful API
- impossible to remove or replace

I don't think we want to go to the [jsonpatch](http://jsonpatch.com/) path, so I just rely on a simple replace.

---

This PR also mutualise some code between dirs and files and removes some legacy code used to unmarshal jsonapi document. A `jsonapi.Bind` function has been introduced to do that.

### DIFF
--- a/vfs/directory.go
+++ b/vfs/directory.go
@@ -263,43 +263,40 @@ func CreateDirectory(c *Context, doc *DirDoc) (err error) {
 // ModifyDirectoryMetadata modify the metadata associated to a
 // directory. It can be used to rename or move the directory in the
 // VFS.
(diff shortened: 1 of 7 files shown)


nono:
Travis build is red. Just a missing comment, I think.

> I don't think we want to go to the jsonpatch path, so I just rely on a simple replace.

In my mind, it was a simple replace, so :+:1

nono:
LGTM

jinroh:
rebased on master and added comment.

PR #63, opened 2016-10-28
Add actual instances management

aenario:
- Instances are created through the CLI
- Instances are stored in a Couchdb DB named global/instances
- When an instances is created, we also create vfs root directory and vfs indexes.
- If there is no instance for the request's Host header, an error is returned (exception for localhost -> dev)

@jinroh I did not change the several places where you commented about the root folder. Let's discuss next week. 

**FUTURE**
- Handle subdomains (someapp.bob.cozycloud.cc)
- Move creation of `global/instances byDomain` request into stack initialization (Not a priority, win a few ms on instance creation ...)

### DIFF
--- a/debug.log
+++ b/debug.log
@@ -1,28 +1,0 @@
-[couchdb request] DELETE dev%2Fio-cozy-files 
-[couchdb request] PUT dev%2Fio-cozy-files 
-[couchdb request] POST dev%2Fio-cozy-files/_index {"index":{"fields":["folder_id","name"]}}
-[couchdb request] PUT dev%2Fio-cozy-files/io.cozy.files.rootdir {"type":"","_id":"io.cozy.files.rootdir","name":"","folder_id":"","created_at":"0001-01-01T00:00:00Z","updated_at":"0001-01-01T00:00:00Z","path":"/","tags":null}
-=== RUN   TestGetFileDocFromPathAtRoot
-[couchdb request] POST dev%2Fio-cozy-files {"type":"file","name":"toto","folder_id":"io.cozy.files.rootdir","created_at":"2016-11-02T15:39:41.073890175+01:00","updated_at":"2016-11-02T15:39:41.073890175+01:00","size":"7","md5sum":"MHm6zuV/0aAEzvWpObXh4A==","mime":"foo/bar","class":"foo","executable":false,"tags":[]}
-[couchdb request] POST dev%2Fio-cozy-files/_find {"selector":{"path":"/"},"limit":1}
-[couchdb request] POST dev%2Fio-cozy-files/_find {"selector":{"$and":[{"folder_id":"io.cozy.files.rootdir"},{"name":"toto"},{"type":"file"}]},"limit":1}
-[couchdb request] POST dev%2Fio-cozy-files/_find {"selector":{"path":"/"},"limit":1}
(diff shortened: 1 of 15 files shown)


nono:
The build on travis is red, with this message:

> instance/instance.go:79:2:warning: 'if err != nil { return err }; return nil' can be simplified to 'return err' (gosimple)

instance/instance.go
```diff
@@ -0,0 +1,167 @@
...
+func (i *Instance) SetID(v string) { i.DocID = v }
+
+// Rev implements couchdb.Doc
+func (i *Instance) Rev() string { return i.DocRev }
+
+// SetRev implements couchdb.Doc
+func (i *Instance) SetRev(v string) { i.DocRev = v }
+
+// ensure Instance implements couchdb.Doc
+var _ couchdb.Doc = (*Instance)(nil)
```
nono: I didn't know of this trick. Looks cool!

instance/instance.go
```diff
@@ -0,0 +1,167 @@
...
+// CreateRootFolder creates the root folder for this instance
+func (i *Instance) createRootFolder() error {
+	root := vfs.MakeRoot()
+	prefix := i.GetDatabasePrefix()
+	err := couchdb.CreateDB(prefix, root.DocType())
+	if err != nil {
+		return err
+	}
+	return couchdb.CreateNamedDoc(prefix, root)
+	// TODO (vfs) should we also create the root folder on FS ?
```
nono: Yes, we should

vfs/vfs.go
```diff
@@ -142,6 +142,14 @@ func getRootDirDoc() *DirDoc {
 	}
 }
 
+// MakeRoot returns a new instance of a DirDoc usable as root
+func MakeRoot() *DirDoc {
+	return &DirDoc{
+		ObjID: RootFolderID,
+		Path:  "/",
```
nono: #61 was merged in master, and `Path` was renamed to `Fullpath` in it

debug.log
```diff
@@ -0,0 +1,28 @@
...
+	assertions.go:241: 
                        
	Error Trace:	vfs_test.go:67
+		
	Error:		Received unexpected error:
+			file does not exist
+		
+FAIL
+exit status 1
+FAIL	github.com/cozy/cozy-stack/vfs	0.165s
```
jinroh: oups.. a .log leak 😄

instance/instance.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// GetVFSContext returns a vfs.Context for this Instance
+func (i *Instance) GetVFSContext() (c *vfs.Context, err error) {
+	dbprefix := i.GetDatabasePrefix()
+	fs, err := i.GetStorageProvider()
+	if err != nil {
+		return nil, err
+	}
+	return vfs.NewContext(fs, dbprefix), nil
```
jinroh: 👍

jinroh:
LGTM (except for the .log file)

aenario:
Ok, this costed me a lot more time than expected, but I added creation of the root folder to it as well.

We have a bit of a blur on the way the various tests are instantiated.
I think we should mix instance.Instance & vfs.Context at some point.

jinroh:
@aenario yes, a "global" context would be better i think.

nono:
LGTM.

BTW, we should take care to not have too many TODOs in our code (I have seen some ones in this PR).

PR #64, opened 2016-10-31
[APPS] Clone git application

jinroh:
~~This is still work in progress.~~

For now, this allow to clone an application from a git source. It is still missing tests and suffering from really bad performance.


nono:
> It is still missing tests and suffering from really bad performance.

As Kent Beck said, make it work, make it right, make it fast. Source: http://wiki.c2.com/?MakeItWorkMakeItRightMakeItFast

nono:
TODO for later: move most of the code inside `web/apps/` to `apps/`, and add a `cozy app` CLI.

web/apps/git.go
```diff
@@ -0,0 +1,290 @@
...
+func (fs *gfs) Base() string {
+	return fs.base
+}
+
+var (
+	_ Installer        = &GitInstaller{}
+	_ gitFS.Filesystem = &gfs{}
+	_ gitFS.File       = &gfileWrite{}
+	_ gitFS.File       = &gfileRead{}
+)
```
nono: Some comments will be welcomed in this file ;-)

nono:
For performances, the first thing that comes to my mind is bulk operations to couchdb. Do you see anything else?

jinroh:
I'm working on a "data layer" laying between the VFS and CouchDB, with a go interface.

I have an implementation for a single node (not distributed) relying on a simple thread-safe cache LRU. I'm already seeing 300% improvement even if its still missing some obvious optimisations like batching with couchdb bulk operations.

This abstraction layer feels quite good because it clearly looks like the good place to implement our future code for synchronisation and optimisation for a distributed node — talking to our groupcache/redis layer.

It also removes some bad states that were laying around in the VFS code, like files/dirs parent memoization...

nono:
Nice!

jinroh:
OK, I think I have something mergeable. It is still suffering form bad performance but it is a good starting point for a review.

I created the `package apps` in `/apps`

jinroh:
Yes #66 is welcome 👍 

What makes performance poor is that the performance of the VFS in general are really bad for now. Anything that create quite a lot a files (here the .git directory and any file in the working tree) will have tons of useless RTTs with CouchDB (for instance, for EVERY file creation, it calls CouchDB twice) which make the system very cumbersome.

PR #67, opened 2016-11-03
[APPS] Add listing of applications

jinroh:
Introduces the `GET /apps/` endpoint.

Added `jsonapi.DataList` to marshal list of `jsonapi.Object`.

### DIFF
--- a/web/files/files_test.go
+++ b/web/files/files_test.go
@@ -274,7 +274,7 @@ func TestCreateDirConcurrently(t *testing.T) {
 	for i := 0; i < n; i++ {
 		select {
 		case res := <-errs:
-			assert.Equal(t, 409, res.StatusCode)
+			assert.True(t, res.StatusCode == 409 || res.StatusCode == 503)
 		case <-done:
 			c = c + 1
 		}
--- a/web/jsonapi/jsonapi.go
+++ b/web/jsonapi/jsonapi.go
@@ -54,7 +54,7 @@ func Data(c *gin.Context, statusCode int, o Object, links *LinksList) {
 
 // DataList can be called to send an multiple-value answer with a
 // JSON-API document contains multiple objects.
-func DataList(c *gin.Context, statusCode int, objs []Object) {
+func DataList(c *gin.Context, statusCode int, objs []Object, links *LinksList) {
 	var err error
 	objsMarshaled := make([]json.RawMessage, len(objs))
 	for i, o := range objs {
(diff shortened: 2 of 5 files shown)


web/files/files_test.go
```diff
@@ -274,7 +274,7 @@ func TestCreateDirConcurrently(t *testing.T) {
 	for i := 0; i < n; i++ {
 		select {
 		case res := <-errs:
-			assert.Equal(t, 409, res.StatusCode)
+			assert.True(t, res.StatusCode == 409 || res.StatusCode == 503)
```
nono: Why this change?

web/jsonapi/jsonapi.go
```diff
@@ -52,6 +52,39 @@ func Data(c *gin.Context, statusCode int, o Object, links *LinksList) {
 	c.Data(statusCode, ContentType, body)
 }
 
+// DataList can be called to send an multiple-value answer with a
+// JSON-API document contains multiple objects.
+func DataList(c *gin.Context, statusCode int, objs []Object) {
```
nono: Like `Data`, we should allow to pass `links *LinksList` in parameters. It will be used for pagination, for example.

web/files/files_test.go
jinroh: meh.. sometimes Couch does not handle very well that much concurrent call (100 here) on my laptop and returns 503. I wanted to consider these as "false negative"...

web/jsonapi/jsonapi.go
```diff
@@ -52,6 +52,39 @@ func Data(c *gin.Context, statusCode int, o Object, links *LinksList) {
...
+	var err error
+	objsMarshaled := make([]json.RawMessage, len(objs))
+	for i, o := range objs {
+		var j json.RawMessage
+		j, err = MarshalObject(o)
```
nono: Why not remove the `var err error` at top of the func and `var j json.RawMessage`? And initialize them with:

```go
j, err := MarshalObject(o)
```

jinroh: I thought the linter would cry about that, but I was wrong ...

web/jsonapi/jsonapi.go
jinroh: done 8873ad40bf86ed49b6e33851407474ab5301ac05

web/jsonapi/jsonapi.go
jinroh: Premature linting is the root of all evil 😄

PR #68, opened 2016-11-03
Add simple logging with logrus

jinroh:
This PR adds [`github.com/Sirupsen/logrus`](http://github.com/Sirupsen/logrus).

A `--log-level` option can be used to change the logger level. No more option for now.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -57,8 +57,11 @@ func init() {
 	flags.IntP("port", "p", 8080, "server port")
 	viper.BindPFlag("port", flags.Lookup("port"))
 
-	flags.StringP("database-url", "d", "http://localhost:5984", "couchdb database address")
-	viper.BindPFlag("database.url", flags.Lookup("database-url"))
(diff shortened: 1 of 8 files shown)


cmd/root.go
```diff
@@ -2,12 +2,27 @@ package cmd
...
+const ConfigFilename = ".cozy"
+
+// ConfigPaths is the list of directories used to search for a
+// configuration file
+var ConfigPaths = []string{
+	".cozy",
+	"/etc/cozy",
+	"$HOME",
+	".",
+}
```
nono: It looks like you have added a config file. It's a good idea, but we should document it somewhere. Maybe in a `cli.md`, with other informations about the command-line (like the server mode)

jinroh: The config file already existed, it is just more explicit in this PR. I can add it in the doc.

jinroh:
🆙  should be good to go. I added an example file `cozy.dist.yaml`. I also renamed `database` to `couchdb` to be more explicit.

Issue #66, opened 2016-11-02
Unifying Contexts across the stack


aenario:
# State of the union

We currently have 4 context-like objects
- `gin.Context` is the request context as provided by gin framework and is passed to the various middlewares, it host informations about the http.Request and http.Response to the stack as well as errors bound to the request.
- `vfs.Context` is an object we made to pass both the dbPrefix and the afero.fs that are used by the vfs package to perform its operations
- `instance.Instance` is an object we made that represent a cozy instance, it contains the dbprefix and a "storageURL" which is passed to afero to create. It has a `.getVFSContext` method which is used to create a VFS context.
- `couchdb dbPrefix` is a string (but we might need to add host/auth later) which is used by the couchdb package

We expect to at least add
- `cache.Location` which will indicate for each instance where to store the  memory cache (redis...)

# Proposal

To avoid too much coupling but still keeping this simple, `vfs.Context` and`couchdb.DBprefix` should be made into interfaces. 

Instance can then implement these interface: 
- we only pass a &Instance around in the normal stack flow
- we create TestVFSContext and TestPrefix to keep vfs and couchdb test package-scoped

# Going further 

I have a feeling the way we test gin.Handlers by recreating the router in every web/xxx_test.go files is weird, but I cant think of a better solution. Let's not dwell on it if we end up changing the routing framework.

PR #69, opened 2016-11-03
[WIP] [VFS] Add cache layer between vfs and couchdb

jinroh:
**~~Still work in progress~~, open for comments.**

In this PR I add a cache layer between the VFS and CouchDB.

This layer for now only does caching, it does not do any bulk operation. I measured from 3x to 5x amelioration when cloning a repository with 1000 files.

Some parts of this cache layer could be generalised to any CouchDB access in our code, but for now it focuses on the VFS only.

As mentioned in #66, this cache would benefit from a rationalisation of our context system.


jinroh:
🆙  I think this is ready for a proper review to be merged.

I added a `FSCache() Cache` method to our `vfs.Context` interface.

vfs/cache_local.go
```diff
@@ -0,0 +1,407 @@
...
+// LocalCache implements the VFS Cache interface and should be used
+// for mono-stack usage, where only one cozy-stack is accessing to the
+// VFS at a time.
+//
+// Internally it provides some optimisations to cache file attributes
+// and avoid having multiple useless RTTs with CouchDB.
+type LocalCache struct {
+	mud  sync.RWMutex       // mutex for directories data-structures
+	lrud *lru.Cache         // lru cache for directories
+	pthd map[string]*string // path directory to id map
```
nono: Why `map[string]*string` and not `map[string]string`?

vfs/cache_local.go
```diff
@@ -0,0 +1,407 @@
...
+
+	muf  sync.RWMutex       // mutex for files data-structures
+	lruf *lru.Cache         // lru cache for files
+	pthf map[string]*string // (folderID, name) file pair to id map
+}
+
+// NewLocalCache creates an new LocalCache. The maxEntries parameter
```
nono: s/an/a/

instance/instance.go
```diff
@@ -25,6 +25,13 @@ var (
 	ErrIllegalDomain = errors.New("Domain name contains illegal characters")
 )
 
+var (
+	// fsLocalCache is the FS Cache used for a classical (non-
+	// distributed) filesystem. It is base on a simple in-memory cache.
+	// There is no fundamental theory behind the size of the cache.
+	fsLocalCache = vfs.NewLocalCache(50)
```
nono: I think you can be a lot more aggressive on caching for local cache. Maybe, 256 instead of 50

vfs/cache_local.go
```diff
@@ -0,0 +1,407 @@
...
+	newpath, err := newdoc.Path(c)
+	if err != nil {
+		return err
+	}
+
+	if oldpath == newpath {
+		return nil
+	}
+
+	var children []*DirDoc
```
nono: Is `children` a slice of only directories? Or directories+files?

If it is a slice of directories+files, how do you track in `updateDirDoc` which lru cache to update?

vfs/cache_local.go
```diff
@@ -0,0 +1,407 @@
...
+// Len returns the total number of elements currently cached
+func (lc *LocalCache) Len() int {
+	lc.mud.RLock()
+	lc.muf.RLock()
+	defer lc.mud.RUnlock()
+	defer lc.muf.RUnlock()
+	return lc.lrud.Len() + lc.lruf.Len()
+}
+
+func (lc *LocalCache) touchDir(doc *DirDoc) {
```
nono: I think we should rename this function. For me, `touch` is a write operation that changes the timestamp (like `touch` in the shell).

vfs/cache_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+	assert.Equal(t, dir1, dir2, "should have been retrived from cache")
+
+	dir3, err := cache.DirByID(vfsC, dir1.ID())
+	if !assert.NoError(t, err) {
+		return
+	}
+
+	assert.Equal(t, dir1, dir3, "should have been retrived from cache")
+}
```
nono: I really think we should have more tests here.

For example, I have some questions about renaming a folder. We should have a test that renames a folder (`foo/` -> `qux/`) with a file (`bar`) in it, and check later that accessing this file (`foo/bar`) doesn't give the cached file.

jinroh: Indeed, I'll add more tests

vfs/cache_local.go
jinroh: What about `memDir/memFile` (for memorize) or `tapDir/tapFile` ?

nono: They are both good. I have a small preference for `tapDir/tapFile`

vfs/cache_local.go
jinroh: It is a slice of only of directories. This part of the code is actually responsible for updating of the `path` attributes of the directories and subdirectories when the parent directory's path is changed. I'll add a comment to make it clearer.

nono: So, how do you clear the cache for files inside the renamed directory?

jinroh: Indeed I'll have to fix that.

I'll probably use a radix tree instead of a map to store directories by path. I made [this lib](https://github.com/jinroh/radix) few weeks ago, time to use it.

nono: Awesome go gives me these packages too:

- https://github.com/plar/go-adaptive-radix-tree
- https://github.com/Workiva/go-datastructures
- https://github.com/emirpasic/gods
- https://github.com/derekparker/trie

PR #70, opened 2016-11-03
Authentication and access delegation

nono:


docs/auth.md
```diff
@@ -0,0 +1,444 @@
...
+- Resource owner credentials grant type.
+
+On cozy, only the most typical one is used: authorization code. To start this
+flow, the client must have a `client_id` and `client_secret`. The Cozy stack
+implements the OAuth2 Dynamic Client Registration Protocol (an extension to
+OAuth2) to allow the clients to obtain them.
+
+OAuth2 has also 3 ways to use a token:
+
+- in the query-string (even if the spec it's not recommended)
```
jinroh: typo: even if the spec *does not* recommend it

jinroh:
Don't know if it is the right place, but we could be explicit in this document about how we want to store the passwords, describing which key derivation function we want to use (I would suggest scrypt).

Also, the [OWASP Authentication Cheat Sheet](https://www.owasp.org/index.php/Authentication_Cheat_Sheet) could be useful to link to.

docs/auth.md
```diff
@@ -0,0 +1,462 @@
...
+implements the OAuth2 Dynamic Client Registration Protocol (an extension to
+OAuth2) to allow the clients to obtain them.
+
+OAuth2 has also 3 ways to use a token:
+
+- in the query-string (even if the spec does not recommended it)
+- in the POST body
+- in the HTTP Authorization header.
+
+On cozy, only the HTTP header is supported.
```
aenario: We will need to support query-string token at least for vfs download route to allow using vfs files in html `<audio>` tags. (see https://github.com/cozy/cozy-data-system/commit/f1f950eedba31699540de4ee99fc0fb65951a76d).

If we really dont want to, we will need a "create temporary unauthenticated route for this file" mechanism.

docs/auth.md
```diff
@@ -0,0 +1,462 @@
...
+
+When an OAuth2 client wants to get access to the data of the cozy owner, it
+starts the OAuth2 dance with this step. The user is shown what the client asks
+and has an accept button if she is OK with that.
+
+The parameters are:
+
+- `client_id`, that identify the client
+- `redirect_uri`, it has be exactly the same of the one used in registration
+- `state`, it's a protection against CSRF on the client (a random string
```
aenario: If the fragment is overwritten and the user use a link & exit the apps page, the state is not only a random string but also contains the current state of the client if it is a SPA  / a sessionid if it's not.

docs/auth.md
```diff
@@ -0,0 +1,462 @@
...
+attacks like session fixation).
+
+```http
+DELETE /auth/login?CtxToken=token-for-a-private-context HTTP/1.1
+Host: cozy.example.org
+```
+
+### POST /auth/register
+
+This route is used by OAuth2 clients to dynamically register them-selves.
```
aenario: I am not sure I understand the added security from this step : 

- Either we start an actual cozy partnership program and dynamic registration requires something cryptosigned by the cozy cloud company or for the cozy to be in "dev" mode. In which case, it might be important to specify it here.
- If there is no such verification, this is (IMHO) **less** secure than simply accepting any redirect URL and warning the user : fishing is way easier with added description, ie difference between
`[LOGO Partner] The partner company SA  (https://almosttheprtnerurl.com)` vs `You are about to share data with https://almosttheprtnerurl.com, BE CAREFUL, make sure its the correct url, cozy can make no guarantee other than the url`

docs/auth.md
```diff
@@ -0,0 +1,462 @@
...
+```http
+HTTP/1.1 302 Moved Temporarily
+Location: https://client.org/?state=Eh6ahshepei5Oojo&access_code=Aih7ohth
+```
+
+### POST /auth/access_token
+
+Now, the client can check that the state is correct, and if it is the case,
+ask for an `access_token`. It can use this route with the `code` (ie
+access_code) given below.
```
aenario: s/below/above/

docs/auth.md
```diff
@@ -0,0 +1,462 @@
...
+### How to register the application?
+
+The application is registered at install. See [app management](apps.md) for
+details.
+
+### How to get a token?
+
+When a user access an application, she first loads the HTML page. Inside this
+page, the `<body>` tag has a `data-cozy-token` attribute with a token. This
+token is specific to a context, that can be either public or private.
```
aenario: If we have a domain/app, why not cookies? There are easier to set from the server without parsing the HTML / requiring the app dev to put a placehoder and as easy to read from the javascript side. They also will apply to any subrequest without the need for the app developer to handle them manually.

docs/auth.md
```diff
@@ -0,0 +1,462 @@
...
+--------------------
+
+### How to register the application?
+
+If a third-party websites would like to access a cozy, it had to register
+first. For example, a big company can have data about a user and may want
+to offer her a way to get her data back in her cozy. When the user is
+connected on the website of this company, she can give her cozy address. The
+website will then register on this cozy, using the OAuth2 Dynamic Client
+Registration Protocol, as explained [below](#post-authregister).
```
aenario: s/below/above/

docs/auth.md
```diff
@@ -0,0 +1,462 @@
...
+is given to the client by redirecting the user to an URL controlled by the
+client. But devices and browser extensions don't have an obvious URL for that.
+
+The IETF has published an RFC called [OAuth 2.0 for Native
+Apps](https://tools.ietf.org/html/draft-ietf-oauth-native-apps-05).
+
+### Native apps on desktop
+
+A desktop native application can start an embedded webserver on localhost. The
+redirect_uri will be something like `http://127.0.0.1:19856/callback`.
```
aenario: Without additional crypto, it is important to note (and show to the user) "an application which might or might not be the cozy-desktop client".

docs/auth.md
```diff
@@ -0,0 +1,462 @@
...
+For example, an application that has been dynamically registered can't ask the
+cozy owner to give it the right to install other applications. This limitation
+should improve security, as avoiding too powerful scopes to be used with
+unknown applications.
+
+The cozy stack will apply rate limiting to avoid brute-force attacks.
+
+The cozy stack offers
+[CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS)
+for most of its services. But it's disabled for `/auth` (it doesn't make sense
```
aenario: Also important to not CORS the other applications `index.html` as they contains tokens.

docs/auth.md
nono: This part is about OAuth2 and client-side apps don't use this. `<audio>` tags will be used on client-side apps, where the token is passed in the query-string.

docs/auth.md
nono: We won't start a partnership program. We will warn the user when we ask him/her to give permissions for the app. But forcing the app to register its-self with a `redirect_uri` is still a good security measure, to avoid abusing the TOFU principle.

docs/auth.md
nono: Well, the OAuth2 spec just says that `state` should have at least 2^128 entropy. You can use a session_id for that. SPA are a special case, and not a easy target for OAuth2. It's possible to generate a random string, save it in the local storage, doing th OAuth2 and check the state at the end.

docs/auth.md
nono: This token is meant to authentify the app when talking with the stack. A cookie is sent to the stack only if it is set on the domain of the stack, not on the sub-domain of the app. And then, it is no longer specific to the app, but shared by all the apps, which is exactly what we don't want to do.

The cookies are also prone to race conditions when several contexts are involved. For example, if a user shares 2 album with photos, and send the 2 links to one of his friend. If the friend opens the 2 links just one after the other, there is a risk than the cookie of the second link overwrites the cookies of the first link, and when the JS in the first page read the cookies, it will obtain a context token for the second page and that will let it show the photos of the first album.

docs/auth.md
nono: Yes, the best we can say is that the app declared its-self that it is the cozy-desktop (but we have no garantee of that).

docs/auth.md
nono: Yes, I should make it explicit that CORS is only for the stack domain, and not allowed for the client-side apps on the subdomain.

PR #75, opened 2016-11-07
Build.sh

jinroh:
This PR adds a script to automate binary releases. It only does build task for now. It may feel a bit premature since we don't have yet a release yet, but at least we're buckled up 😄 .

It also adds a `BuildMode` and `BuildTime` constants in the binary. `GET /version` outputs all these constants in addition to the binary version and the GO runtime version.

The `--mode` flag passed at runtime is checked against the `BuildMode` to forbid a `development` mode in a production release.

-------

Copy pasta from the comment describing how versions are generated:

```
The version string is deterministic and reflects entirely the state
of the working-directory from which the release is built from. It is
generated using the following format:

    <TAG>[-<NUMBER OF COMMITS AFTER TAG>][-dirty][-dev]

Where:
 - <TAG>: closest annotated tag of the current working directory. If
   no tag is present, is uses the string "v0". This is not allowed in
   a production release.
 - <NUMBER OF COMMITS AFTER TAG>: number of commits after the
   closest tag if the current working directory does not point
   exactly to a tag
 - -dirty: added if the working if the working-directory is not
   clean (contains un-commited modifications). This is not allowed
   in production release.
 - -dev: added for a development mode relase

The outputed binary is named "cozy-stack-${VERSION_STRING}". A
SHA256 checksum of the binary is also generated in a file named
"cozy- stack-${VERSION_STRING}.sha256".
```

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -4,7 +4,9 @@ COZY_ENV_DFL=production
 
 [ -z ${COZY_ENV} ] && COZY_ENV=${COZY_ENV_DFL}
 
-WORK_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+pushd `dirname $0` > /dev/null
+WORK_DIR=`pwd`
+popd > /dev/null
 
 usage() {
 	echo -e "Usage: ${1} [release]"
@@ -35,7 +37,7 @@ usage() {
 #
 # The outputed binary is named "cozy-stack-${VERSION_STRING}". A
 # SHA256 checksum of the binary is also generated in a file named
-# "cozy- stack-${VERSION_STRING}.sha256".
+# "cozy-stack-${VERSION_STRING}.sha256".
 do_release() {
 	set -e
 
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,104 @@
+#!/bin/bash
+
+COZY_ENV_DFL=production
+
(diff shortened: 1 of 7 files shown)


build.sh
```diff
@@ -0,0 +1,104 @@
+#!/bin/bash
+
+COZY_ENV_DFL=production
+
+[ -z ${COZY_ENV} ] && COZY_ENV=${COZY_ENV_DFL}
+
+WORK_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
nono: I'm using zsh, I don't think `BASH_SOURCE` will work with it. And I think it's dash on travis (if we want to use this script on travis later).

build.sh
```diff
@@ -0,0 +1,104 @@
...
+#    closest tag if the current working directory does not point
+#    exactly to a tag
+#  - -dirty: added if the working if the working-directory is not
+#    clean (contains un-commited modifications). This is not allowed
+#    in production release.
+#  - -dev: added for a development mode relase
+#
+# The outputed binary is named "cozy-stack-${VERSION_STRING}". A
+# SHA256 checksum of the binary is also generated in a file named
+# "cozy- stack-${VERSION_STRING}.sha256".
```
nono: s/cozy- stack/cozy-stack/ (no space)

PR #76, opened 2016-11-07
Update git filesystem to support go-git-v4.rc2

jinroh:
These changes add support to go-git-v4.rc2 that added the `OpenFile(name, flag, perm)` to its `FileSystem` interface.

I took the opportunity to have add `*File` handle that implement `io.Reader`, `io.Seeker`, `io.Closer`, `io.Writer`.

I had to change the way file modification is handled, since go-git needs has a handle to both read and write a same file:

  - before: `create tempfile -> write -> safe rename on close or if error remove the tempfile`
  - now: `safe rename in backup file -> safe file create -> write -> remove backup on close or if error reuse backup`

This scheme looks better for when we will want to keep multiple file revisions in the vfs.

### DIFF
--- a/apps/git.go
+++ b/apps/git.go
@@ -13,7 +13,6 @@ import (
 	"time"
 
 	"github.com/cozy/cozy-stack/vfs"
-	"github.com/spf13/afero"
 	git "gopkg.in/src-d/go-git.v4"
 	gitSt "gopkg.in/src-d/go-git.v4/storage/filesystem"
 	gitFS "gopkg.in/src-d/go-git.v4/utils/fs"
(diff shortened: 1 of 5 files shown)


apps/git.go
```diff
@@ -307,9 +275,15 @@ func (fs *gfs) ReadDir(dirname string) ([]gitFS.FileInfo, error) {
 
 func (fs *gfs) TempFile(dirname, prefix string) (gitFS.File, error) {
 	// TODO: not really robust tempfile...
```
nono: For the record, this is how `TempFile` is implemented in the stdlib: https://golang.org/src/io/ioutil/tempfile.go?s=1285:1342#L37

vfs/vfs.go
```diff
@@ -269,6 +305,14 @@ func (c *Context) Rename(oldpath, newpath string) error {
 	return err
 }
 
+// Remove removes the specified named file or directory.
+func (c *Context) Remove(name string) error {
+	// TODO: fix this remove method implemented for now only to support
+	// go-git. This method should also remove the document from
+	// database.
```
nono: Don't forget this TODO :wink:

jinroh: Indeed ! Hope It'll remain obvious when we'll do the remove 😉 .

PR #77, opened 2016-11-07
Move mango indexes definition inside VFS package

jinroh:
This PR main goal is to have the definition of all VFS indexes directly inside the vfs package. This helps for tests and, IMHO, is clearer.

@aenario I changed a few of the names in `mango/index`, if you ok with that ? :
  - `s/IndexDefinition/IndexFields/`
  - `s/IndexDefinitionRequest/Index/`

### DIFF
--- a/vfs/vfs.go
+++ b/vfs/vfs.go
@@ -19,21 +19,12 @@ import (
 
 // Indexes is the list of required indexes by the VFS inside CouchDB.
 var Indexes = []mango.Index{
-	mango.Index{
-		Name:  "FileFromParent",
-		DDoc:  "used to lookup a file given its parent",
-		Index: mango.NewIndexFields("folder_id", "name", "type"),
-	},
-	mango.Index{
-		Name:  "DirectoryPath",
-		DDoc:  "used to lookup a directory given its path",
-		Index: mango.NewIndexFields("path"),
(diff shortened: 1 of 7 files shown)


vfs/vfs.go
```diff
@@ -13,9 +13,29 @@ import (
...
 
 	"github.com/cozy/cozy-stack/couchdb"
+	"github.com/cozy/cozy-stack/couchdb/mango"
 	"github.com/spf13/afero"
 )
 
+// Indexes is the list of required indexes by the VFS inside CouchDB.
+var Indexes = []mango.Index{
+	mango.Index{
+		Name:  "FileFromParent",
```
aenario: Name And DDoc are couchdb internals : "_design/{Ddoc}" is the name of the design document where the index  is defined. "name" is the name of this index in the design doc. 

If we dont let couchdb pick those automatically, there will be a wasteful double indexing if two apps define indexes on same field but with different name.

jinroh: @aenario Oh i see. Thanks for the info. I'll just leave a comment in our Go code then.

aenario: Thanks !

nono:
@jinroh travis is not happy

PR #78, opened 2016-11-07
Unify Contexts

aenario:
Following #66 here is a first shoot at unifying all the contexts, Basically : 

- replace `dbprefix string` by a `couchdb.Database interface{GetPrefix() string}`
- replace `vfs.Context struct` by an interface extending `couchdb.Database` with `GetFS() afero.Fs`
- make Instance implements both Interfaces
- only pass reference to the instance around
- create appropriate test for each package (in couchdb a custom couchdb.Database, in vfs a custom vfs.Context, in data Instance)

### DIFF
--- a/instance/instance.go
+++ b/instance/instance.go
@@ -13,7 +13,12 @@ import (
 	"github.com/spf13/afero"
 )
 
-const globalDBPrefix = "global/"
+type dbPrefix struct{ prefix string }
+
+func (p *dbPrefix) Prefix() string { return p.prefix + "/" }
+
+var globalDBPrefix = &dbPrefix{"global"}
+
 const instanceType = "instances"
 
 var (
@@ -51,8 +56,9 @@ func (i *Instance) Rev() string { return i.DocRev }
 // SetRev implements couchdb.Doc
 func (i *Instance) SetRev(v string) { i.DocRev = v }
 
-// ensure Instance implements couchdb.Doc
+// ensure Instance implements couchdb.Doc & vfs.Context
 var _ couchdb.Doc = (*Instance)(nil)
+var _ vfs.Context = (*Instance)(nil)
 
 // CreateInCouchdb create the instance doc in the global database
 func (i *Instance) createInCouchdb() (err error) {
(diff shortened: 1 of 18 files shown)


jinroh:
👍  I like it.

Nitpick: what do you think about replacing `GetFS()/GetPrefix()` by `FS()/Prefix()` which is a more go idiomatic ?

nono:
I'm asking myself if we are not missing something as not using [context from stdlib](https://blog.golang.org/context). It could be a way to pass things (what we are calling context now), but also to have a way to cancel an operation (or have a timeout on it).

aenario:
Regarding `context.Context`, it appears from what I have read it is not recommended to be used to pass things necessary for the function to run (WithValue is only suposed to pass around logging details).

We can however create our own context implementing struct, which also provides the necessary fields (db,vfs root ....).
```
type InstanceContext struct{
   context.Context
   Instance instance.Instance
}
```


We could however add our own

instance/instance.go
```diff
@@ -75,8 +71,8 @@ func (i *Instance) createRootFolder() error {
 		return err
 	}
 
-	if err = vfs.CreateRootDirDoc(vfsC); err != nil {
-		rootFs.Remove(domainURL.Path)
+	if err = vfs.CreateRootDirDoc(i); err != nil {
+		rootFs.Remove(i.Domain)
```
jinroh: this should remain the entire path `domainURL.Path`

instance/instance.go
```diff
@@ -107,32 +102,51 @@ func Create(domain string, locale string, apps []string) (*Instance, error) {
 		StorageURL: domainURL.String(),
 	}
 
-	err := i.Create()
+	var err error
+
+	if err != nil {
+		return nil, err
+	}
```
jinroh: this err checking necessary ?

instance/instance.go
```diff
@@ -107,32 +102,51 @@ func Create(domain string, locale string, apps []string) (*Instance, error) {
...
+		return err
+	}
+	switch u.Scheme {
+	case "file":
+		i.storage = afero.NewBasePathFs(afero.NewOsFs(), u.Path)
+	case "mem":
+		i.storage = afero.NewMemMapFs()
+	default:
+		return fmt.Errorf("Unknown storage provider: %v", u.Scheme)
+	}
```
jinroh: this could use the `createFs` helper (end of the file)

instance/instance.go
aenario: I find it cleaner for the function to always return result XOR error, but never both.

jinroh: but the var `err` here should always be nil since its defined just above.

aenario:
:tada: Merci :tada:

PR #79, opened 2016-11-08
Fix root directory creation

jinroh:
This is a fix for the root directory creation code and `instances add`. In the current code, the directory creation on the actual file system could not work since we were using a namespaced fs to build its own root directory.

~~I left the `globalFsURL` pointing at `file://localhost/tmp/cozy2` for now. I'm working a way to pass that from the configuration (and have a way to inject configuration from tests).~~

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -45,6 +45,11 @@ profiles you.`,
 var cfgFile string
 
 func init() {
+	pwd, err := os.Getwd()
+	if err != nil {
+		panic(err)
+	}
+
 	flags := RootCmd.PersistentFlags()
 	flags.StringVarP(&cfgFile, "config", "c", "", "configuration file (default \"$HOME/.cozy.yaml\")")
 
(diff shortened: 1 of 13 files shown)


jinroh:
Sorry for extending the scope of this PR, but I had a hard time to fix tests without doing some weird things that would have been temporary.

I opted for the "clean" solution and finally added the possibility to pass the "global url" of our VFS via configuration, and added the possibility to inject configurations in tests.

cmd/root.go
```diff
@@ -57,6 +57,9 @@ func init() {
 	flags.IntP("port", "p", 8080, "server port")
 	viper.BindPFlag("port", flags.Lookup("port"))
 
+	flags.String("fs-url", "files://localhost/tmp/cozy2", "filesystem url")
```
nono: Maybe we can use a path relative to the current directory (`./storage`) instead of using `/tmp/cozy2`

PR #81, opened 2016-11-08
Create all directories required on instance creation

jinroh:
Re-implement what was the good behavior: cozy-stack should create *all* the necessary directories to run on `instances add`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 /storage
 swagger-ui
 /.cozy
+/storage/*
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -2,3 +2,4 @@
 /storage
 swagger-ui
 /.cozy
+/storage/*
```
nono: `/storage` is ignored at line 2. I think it's enough

PR #82, opened 2016-11-08
Access to an app

nono:
It's work in progress. There are some known bugs and I should add some tests. But you can look at it

### DIFF
--- a/web/apps/apps_test.go
+++ b/web/apps/apps_test.go
@@ -75,11 +75,7 @@ func installMiniApp() error {
 		return err
 	}
 	err = createFile(pubdir, "index.html", "this is a file in public/")
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return err
 }
 
 func assertGet(t *testing.T, path, contentType, content string) {
--- a/web/apps/apps_test.go
+++ b/web/apps/apps_test.go
@@ -1,0 +1,147 @@
+package apps
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"path"
+	"testing"
+
+	"github.com/cozy/cozy-stack/apps"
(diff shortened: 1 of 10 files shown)


web/middlewares/instance.go
```diff
@@ -3,21 +3,68 @@
...
+			if couchdb.IsNotFoundError(err) {
+				jsonapi.AbortWithError(c, jsonapi.NotFound(err))
+				return
+			}
+			jsonapi.AbortWithError(c, jsonapi.InternalServerError(err))
+		}
+
+		instInterface, ok := c.Get("instance")
+		if !ok {
+			jsonapi.AbortWithError(c, jsonapi.InternalServerError(errors.New("no instance")))
```
jinroh: missing `return`

web/middlewares/instance.go
```diff
@@ -3,21 +3,68 @@
...
+func NeedInstance() gin.HandlerFunc {
+	return func(c *gin.Context) {
+		errInterface, ok := c.Get("instance_error")
+		if ok {
+			err := errInterface.(error)
+			if couchdb.IsNotFoundError(err) {
+				jsonapi.AbortWithError(c, jsonapi.NotFound(err))
+				return
+			}
+			jsonapi.AbortWithError(c, jsonapi.InternalServerError(err))
```
jinroh: missing `return`

web/apps/apps_test.go
```diff
@@ -0,0 +1,147 @@
...
+	err = createFile(appdir, "index.html", "this is index.html")
+	if err != nil {
+		return err
+	}
+	err = createFile(appdir, "hello.html", "world")
+	if err != nil {
+		return err
+	}
+	err = createFile(pubdir, "index.html", "this is a file in public/")
+	if err != nil {
```
aenario: gometalinter & travis dislike this condition (if err != nil { return err }; return nil)

PR #83, opened 2016-11-08
README: Add instructions to create the dev instance

sebn:
Moved the CouchDB section first, so newcomers have a database
instance before running the `instances add` command.

Also added instructions to create the `storage/` directory.

Merge `--ff`?


sebn:
Looks like creating the `storage/` directory will be useless once #81 is merged.

PR #87, opened 2016-11-09
Add instances remove

jinroh:
This PR add a command to remove an instance and all its content.

It also fixes a issue that allowed to create multiple instances with the same domain.

### DIFF
--- a/cmd/instance.go
+++ b/cmd/instance.go
@@ -1,8 +1,6 @@
 package cmd
 
 import (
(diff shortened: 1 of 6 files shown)


aenario:
Actually I approved a bit too fast.

We have some docs for instance in docs/instance.md which define the instance removal function as instance destroy.

I dont really have a preference between rm and destroy, but the docs should match code or be removed.

PR #89, opened 2016-11-09
[VFS] Fix error returned on a conflictual file modification

jinroh:
This PR should fix a misleading error returned in certain conditions when a file is modified concurrently:

```
in case of a concurrent access to this method, it can happend
that the file has already been renamed. In this case the
safeRenameFile will return an os.ErrNotExist error. But this
error is misleading since it does not reflect the conflict.
```

@aenario waiting for your return on this one.


aenario:
:+1: 30 loops and still no bug, I think you got it

PR #90, opened 2016-11-09
Fix Table of Content -> Table of Contents typo

aenario:


jinroh:
looks like travis is seeing what #89 is fixing

PR #92, opened 2016-11-10
Fix /version/ and /status/ when no instance

jinroh:
This PR reorder the web router to allow `/version/` and `/status/` routes to respond even if no instance are created.


nono:
OK for `/version/`. But for `/status/`, it's not clear. In my mind, `/status/` was tied to an instance and gives the status for this instance.

PR #97, opened 2016-11-14
Onboarding doc

aenario:


docs/onboarding.md
```diff
@@ -0,0 +1,157 @@
...
+
+The CLI also allows to specify which source to use for `onboarding` and `home` applications. The defaults will be hosted on `github.com/cozy`.
+
+After creation, an instance has a `registerToken` generated randomly.
+
+All cozy-instance fields are considered private unless explictly prefixed by `public-`. Public instance field can be seen from the `instance` endpoint. Only public fields we can think of for now are `public-name`, `public-locale`.
+
+## Onboarding steps
+
+This document and the cozy-stack are only concerned with login and password registering step which are important for security.
```
CPatchane: Does it mean that other unauthenticated steps (welcome and CGU) won't need this `registerToken`? For now, we also need to save the onboarding progression and the `isCGUaccepted` boolean, so it would be better to save only using a token.

Also, I already think and work about a token usage for unauthenticated onboarding steps in [this closed PR](https://github.com/cozy/cozy-proxy/pull/309/files#diff-bafd0c87fdf0cc50bd63327c02fcecc1R146) that we didn't include in the current work. Is it the same working as here?

docs/onboarding.md
```diff
@@ -0,0 +1,157 @@
...
+
+All other steps are handled by the `onboarding` application.
+
+The `onboarding` application SHOULD therefore provide the following features
+- When started with a `registerToken`, allow the user to create a password
+- When started with a `contextToken` ([see auth doc](./auth.md#how-to-get-a-token)) use it to retrieve instance document.
```
CPatchane: The link here seems to be broken or maybe the document is not available yet. For my culture, what is the working of this `contextToken` here? Is it related to the authentication cookie?

docs/onboarding.md
```diff
@@ -0,0 +1,157 @@
...
+  - If the instance document is complete **according to the `onboarding` app**, redirect to `home` application.
+  - Otherwise, performs whatever steps it deems necessary to fill out the instance (ask for user email, help set up `myaccounts` accounts, say thank you...)
+
+This makes cozy-stack simple and safer while allowing behaviour modification for several install types by picking the correct `onboarding` application / branch.
+
+This makes it easier to add more onboarding steps and have them run on already-installed cozy : On next login after onboarding application update, it will ask the user.
```
CPatchane: Actually, it's a bit more difficult than that. For now, we decided to not run the new onboarding on already-installed cozy for functional reasons (as we don't have token usage, we reject new password saving in the unauthenticated way if the password is already set in the instance document). But I agree about adding more steps for "already-onboarded" cozies if we have token for safer process (and for `isRegistered` detection so).

docs/onboarding.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Redirections
+
+When an user attempts to access the root of its instance (`https://example.cozycloud.cc`) or an application (`https://contacts.example.cozycloud.cc`), and she is not logged-in, she is redirected :
```
CPatchane: I completely agree with this idea. That already worked like that with the new onboarding mecanism, except for `/apps`. Indeed, I think we need some bypasses for the `/apps*` routes because, for example, we need to access the MyAccounts app inside the onboarding accounts step (via iframe) while not allowing the user to jump to his cozy home page. If we do this way we will have an infinite loop of iframes in the step :/

docs/onboarding.md
```diff
@@ -0,0 +1,157 @@
...
+If the user is not logged in, display public instance information
+#### Request
+```http
+GET https://alice.example.com/instance
+```
+#### Response
+```json
+{
+  "public-locale":"fr",
+  "public-name":"Alice Martin"
```
CPatchane: Currenlty, we need a public boolean `hasValidInfos` to know if we have to an extra `infos` step (additional step to set email, username and timezone) or not. Indeed, the progression breadcrumb displays 5 steps if the user already have valid infos or 6 steps otherwise.

docs/onboarding.md
```diff
@@ -0,0 +1,157 @@
...
+  "domain": "alice.example.com",
+  "owner-email": "alice@example.com",
+  "public-locale":"fr",
+  "public-name":"Alice Martin"
+}
+```
+
+### POST /instance
+
+If the user is not logged in, allow to set the password
```
CPatchane: Maybe here, if I understand correctly, if the user is not logged in, that should allow to set the password if and only if the registerToken used by the user is correct, isn't it?

docs/onboarding.md
```diff
@@ -0,0 +1,157 @@
...
+```json
+{
+  "public-locale":"fr",
+  "public-name":"Alice Martin"
+}
+```
+
+If the user is logged in, display all instance information, except password
+#### Request
+```
```
CPatchane: Just a little detail.
Maybe you forget the `http`language specification here (compared to other requests code writing).

docs/onboarding.md
```diff
@@ -0,0 +1,157 @@
...
+The instance is created
+```json
+{
+  "domain": "alice.example.com",
+  "publicName": "Alice Martin",
+  "tz": "Europe/Paris",
+  "locale":"fr"
+}
+```
+- Eve knows Alice just had an instance created, she goes to `https://alice.cozycloud.cc`. There is no `registerToken`, so she only see a message (in french) along the lines of "This is the cozy for Alice Martin, this register link is incorrect, if you are Alice Martin please ask your sysadmin for a new link".
```
CPatchane: If I understand correctly, the registerToken will be required since the first onboarding steps here. Maybe we should be more explicit about that.

docs/onboarding.md
```diff
@@ -0,0 +1,157 @@
...
+
+Compatibility with the current developments on cozy onboarding is a goal : [The following documents has been consulted for this proposal]( https://github.com/cozy/cozy-proxy/blob/bf9af7f2342e3fc183a8b4e72bcedb909afa3eb8/docs/client/)
+
+## Instance creation
+
+Creating an instance is done through CLI or through the (future) partner farm manager system. Some **settings** can be pre-defined on instance creation. ([doc](./instance.md#creation)).
+
+The CLI also allows to specify which source to use for `onboarding` and `home` applications. The defaults will be hosted on `github.com/cozy`.
+
+After creation, an instance has a `registerToken` generated randomly.
```
m4dz: By curiosity, what solution do you plan to use to generate the registration token? Something based on `Base64TimeSigner` using HMAC?

docs/onboarding.md
aenario: My idea was to offload this decision to the onboarding app. 

In client-side state machine, you could have "If there is no registerToken, display a warning". 
The server-side doesn't know and doesn't care about the UX steps welcome & agreement.

I dont think we need to save the isCGUaccepted on the server side before password registration. 

If the lawyers want, we can add a `i_accept_cgu` field to the `POST /instance` route.
To be discussed with UX team, but the only use case it solves (vs a localStorage field) is :
"When I click [Accept CGU] on the mobile, then go to my cozy on my desktop, I dont want to click [Accept CGU] again", is it really that important ?

docs/onboarding.md
aenario: From the point where we did POST /instance, the user is logged in and can access any app. We probably dont want to prevent them from using them even before setting up his accounts. He will still get redirected to onboarding app everytime she logs in.

docs/onboarding.md
aenario: With the structure in this document, every thing that happens after POSTing /instance is the onboarding app in a loged-in state (login is handled by the platform). 
So no difference between first time and later (unlike cozy-proxy which also handles login)

docs/onboarding.md
aenario: Interesting point, let me sleep on it

docs/onboarding.md
aenario: :+1: thanks

docs/onboarding.md
aenario: :+1: will add this

docs/onboarding.md
aenario: :+1:

docs/onboarding.md
aenario: The registerToken only protect empty cozies from hijacking. 
It will probably be a from crypto/rand.Reader

docs/onboarding.md
CPatchane: That's true, but the first steps also need to store informations like the progression. Currently, we store the progression as an array of strings in the user document (`onboardedSteps`). How will you handle the progression saving in your case?

docs/onboarding.md
CPatchane: Ok, I see, that's a good point so 👍

docs/onboarding.md
CPatchane: Ok, that's good so 👍

docs/onboarding.md
CPatchane: Maybe we can omit to save the progression on the first steps, but I think it's about UX so (poke @poupotte).

docs/onboarding.md
m4dz: About `Base64TimeSigner` https://github.com/vmihailenco/signer

What've in mind is something similar to what [itsdangerous](https://pythonhosted.org/itsdangerous/#signatures-with-timestamps) is providing in python environment, with token that contains signature, expiration delay, etc

jinroh: So, if I understand correctly, it implements option 2) with a unix timestamp in the message so that it can expire without storing an additional date in DB. I think we would use a JWT for such things if your question is more on the formalism part.

I still don't think we don't want to introduce a global private key in this situation, a random token per user should hopefully be sufficient.

aenario:
Ok, let's sum it up : 

# Option A - Onboarding app does everything

This is the structure described by this document. 

- The only onboarding route exposed by the stack is POST /instance to create the password
- Pre-password onboarding steps can be stored in localstorage, but not synced across browser
- Post-password, onboarding is like any other cozy application and can store / read data from couchdb.
- All variations on onboarding process for different partners / cozy distributions are branch/fork on the onboarding app. We can eventually add more instance public / private field to link CLI <-> onboarding.
- The base onboarding app should be structured so that these branch/fork are easy to rebase.
- **[issue]** Onboarding app does NOT have access to instance state before password register. So it has to make a choice between not displaying a "step 1 of 5" message before password register, or have steps which are just confirmation of pre-filled data. Because the stack has no idea of what step does onboarding have, it cannot compute this information.

# Option B - Make onboarding a complex interaction between stack & client

This will need to be explored further, but the idea is to expose a similar API to cozy-proxy from the stack. So,

- Have the stack understand the notion of onboardingStep / onboardedStep (ie. step TODO and step DONE)
- The list of steps TODO needs to be computed from what was provided by the CLI instance creation and which information we have / if they are correct.
- There is a route to complete each steps with the proper information
- Adding a step implies development in both stack & onboarding
- Maintaining different version means branches of stack & onboarding


I vote :smiling_imp:

docs/onboarding.md
aenario: We should definitely use JWT for most tokens used by the stack. This one in particular, probably not as: 
- Making it expire will be a nightmare for support
- The only "private" piece of information it protects is user's email
- There is only one per instance, and we need to fetch the instance doc for domain recognition anyway.

docs/onboarding.md
aenario: It is the only **[issue]** I indicated with option A

gregorylegarec: Oops. My bad. Sorry for the inconvenience.

CPatchane:
Even if we seem to break some possibilities with the option A, it allows to keep the interactions between the stack and client much simpler, safer and easier to maintain. So I vote __A__ (even if I would want to keep the breadcrumb feature and the progression saving in first steps). We should be able to keep the onboardedSteps notion and its computation in the client side.
Also, the first steps between password saving should/will be processed only once in the cozy life compared to other steps (reset the 'tutorial', add new steps...).

gregorylegarec:
The option A is definitely simpler and we need to prefer simplest solutions. Maybe it will be time to change onboarding step order, and setting password step as the first one. I don't know how the agreement step has to be before the password one, in a legal perspective.

docs/onboarding.md
```diff
@@ -0,0 +1,158 @@
...
+  "tz": "Europe/Paris",
+  "locale":"fr"
+}
+```
+- Eve knows Alice just had an instance created, she goes to `https://alice.cozycloud.cc`. There is no `registerToken`, so she only see a message (in french) along the lines of "This is the cozy for Alice Martin, this register link is incorrect, if you are Alice Martin please ask your sysadmin for a new link".
+- Alice navigates to `https://alice.cozycloud.cc?registerToken=42...42`
+- She is redirected to the `onboarding` application
+- The `onboarding` application receive the registerToken. It is the default onboarding application and therefore display the cozy cloud agreement and then ask for a Password.
+- The `onboarding` application use its `registerToken` to register the password. Registering the password automatically log Alice in and redirect her back to the `onboarding` app but with its normal `data-cozy-token`.
+- The `onboarding` application can then do whatever it needs to do :
```
jinroh: We should not forget to have CSRF protection from here. The `registerToken` should protect us on the creation of the instance with the password, but for the next steps we should have another anti-csrf token to prevent an attacker from registering unwanted informations into the account (even if it does not known the password, it could forge requests to put its own email for instance).

Event though, this may be part of another "global action" against CSRF more than this specific document, it can be specified here..

aenario: At this point, the protection is the same than for all other cozy apps communication with the stack : to modify the instance, onboarding app use it's cozy-token provided as an attribute to its body.

jinroh: Ok I did not understand that the data-cozy-token was referring to the anti-csrf token. My bad !

PR #99, opened 2016-11-14
Force Files type manipulation to occur through the vfs

aenario:


web/data/errors.go
```diff
@@ -28,3 +28,7 @@ func HTTPStatus(err error) (code int) {
 func invalidDoctypeErr(doctype string) error {
 	return fmt.Errorf("Invalid doctype '%s'", doctype)
 }
+
+var errVfsNotData = fmt.Errorf(`
+	Type io.cozy.files belongs to the vfs and cannot be manipulated from the /data endpoint.
+`)
```
jinroh: Error string should preferably not contain \t or \n. Also `errors.New` is sufficient here I think.

aenario: :+1:

PR #100, opened 2016-11-15
[APPS] Improve error handling

jinroh:
This PR should improve the error handling of `/apps`:
  - distinguish error for manifest not reachable
  - add documentation of status codes
  - add documentation on the async nature of the `POST /apps/:slug` request
  - set persisted manifest in `errored` state when an error ocured while installing

### DIFF
--- a/couchdb/couchdb.go
+++ b/couchdb/couchdb.go
@@ -12,7 +12,6 @@ import (
 	log "github.com/Sirupsen/logrus"
 	"github.com/cozy/cozy-stack/config"
 	"github.com/cozy/cozy-stack/couchdb/mango"
-	"github.com/google/go-querystring/query"
 )
 
 // Doc is the interface that encapsulate a couchdb document, of any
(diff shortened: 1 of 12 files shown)


jinroh:
The second commit add proper initialisation of the apps DB and removes the use of `mango.Empty()` filter (used as a wildcard...) and adds a `couchdb.GetAllDocs` using the `_all_docs` endpoint.

couchdb/couchdb.go
```diff
@@ -430,6 +432,33 @@ func FindDocsRaw(db Database, doctype string, req interface{}, results interface
...
+// GetAllDocs returns all documents of a specified doctype. It filters
+// out the possible _design document.
+// TODO: pagination
+func GetAllDocs(db Database, doctype string, req *AllDocsRequest, results interface{}) error {
+	var response allDocsResponse
+	v, err := query.Values(req)
+	if err != nil {
+		return err
+	}
+	url := makeDBName(db, doctype) + "/_all_docs?included_docs=true&" + v.Encode()
```
aenario: I think it will be cleaner to have v.Add("include_docs", true). 
**Note** s/included/include

Alternatively, we could also just POST the AllDocRequest, which we will need to do anyway for large number of keys.

jinroh: Oh yes, I missed this from the documentation. Thanks !

jinroh:
I added tests on `apps` and fixes many issues I found while adding them

PR #104, opened 2016-11-16
Add basic register & login

aenario:


instance/instance.go
```diff
@@ -11,6 +11,8 @@ import (
 	"github.com/cozy/cozy-stack/couchdb"
 	"github.com/cozy/cozy-stack/couchdb/mango"
 	"github.com/cozy/cozy-stack/vfs"
+	"github.com/dchest/uniuri"
+	scrypt "github.com/elithrar/simple-scrypt"
```
jinroh: we should use [scrypt package](https://godoc.org/golang.org/x/crypto/scrypt) from more official golang/x/crypto

instance/instance.go
```diff
@@ -260,6 +272,32 @@ func (i *Instance) Prefix() string {
 	return i.Domain + "/"
 }
 
+// RegisterPassword replace the instance registerToken by a password
+func (i *Instance) RegisterPassword(pass string, tok string) error {
+	if i.RegisterToken == "" {
+		return errors.New("Missing registerToken")
+	}
+	if i.RegisterToken != tok {
```
jinroh: This should be constant time equality check (to avoid timing attacks): see [subtle package](https://golang.org/pkg/crypto/subtle/#ConstantTimeCompare)

instance/instance.go
```diff
@@ -118,8 +129,9 @@ func Create(domain string, locale string, apps []string) (*Instance, error) {
 
 	domainURL := config.BuildRelFsURL(domain)
 	i := &Instance{
-		Domain:     domain,
-		StorageURL: domainURL.String(),
+		Domain:        domain,
+		StorageURL:    domainURL.String(),
+		RegisterToken: uniuri.New(),
```
jinroh: i'm not sure we want to rely on external package (through github) to generate secure data

instance/instance.go
```diff
@@ -260,6 +272,32 @@ func (i *Instance) Prefix() string {
 	return i.Domain + "/"
 }
 
+// RegisterPassword replace the instance registerToken by a password
+func (i *Instance) RegisterPassword(pass string, tok string) error {
+	if i.RegisterToken == "" {
+		return errors.New("Missing registerToken")
+	}
+	if i.RegisterToken != tok {
+		return errors.New("Invalid registerToken")
```
jinroh: Maybe create a list of errors for `ErrInvalidRegisterToken` and `ErrMissingRegisterToken`

instance/instance.go
```diff
@@ -260,6 +272,32 @@ func (i *Instance) Prefix() string {
...
+	}
+
+	hash, err := scrypt.GenerateFromPassword([]byte(pass), scrypt.DefaultParams)
```
jinroh: We should have a random salt along with the password to generate the scrypt hash

web/auth/session.go
```diff
@@ -0,0 +1,137 @@
...
+	// expired session
+	if s.OlderThan(maxAgeDuration) {
+		return nil, ErrExpired
+	}
+
+	// if the session is older than half its maxAgeDuration,
+	// save the new LastSeen
+	if s.OlderThan(maxAgeDuration / 2) {
+		s.LastSeen = time.Now()
+		couchdb.UpdateDoc(i, &s)
```
jinroh: missing error handling (maybe just a log if you don't want it to bail)

web/auth/session.go
```diff
@@ -0,0 +1,137 @@
...
+
+	i := middlewares.GetInstance(c)
+	sid, err := c.Cookie(SessionCookieName)
+
+	// no cookie
+	if sid == "" {
+		return nil, ErrNoCookie
+	}
+
+	err = couchdb.GetDoc(i, SessionsType, sid, &s)
```
jinroh: you want to use the ID of the session document as token or is this just temporary ? if so maybe add a `TODO`.

web/auth/auth.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	redirectSuccessLogin(c)
+}
+
+func login(c *gin.Context) {
+	instance := middlewares.GetInstance(c)
+	var form loginForm
+	if err := binding.Form.Bind(c.Request, &form); err != nil {
+		c.AbortWithError(http.StatusBadRequest, err)
```
jinroh: missing `return`

web/auth/auth.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	if err := instance.CheckPassword(form.Password); err != nil {
+		c.AbortWithError(http.StatusBadRequest, err)
```
jinroh: missing important `return`

instance/instance.go
nono: And to avoid timing attacks, we should move the password generation to a goroutine too

instance/instance.go
jinroh: 16 byte random salt should be good

instance/instance.go
```diff
@@ -260,6 +272,32 @@ func (i *Instance) Prefix() string {
...
+	if err != nil {
+		return err
+	}
+
+	i.RegisterToken = ""
+	i.PasswordHash = string(hash)
```
nono: It would be nice to have a field that says that the hash algorithm for password is scrypt. It will ease the transition to another algorithm is scrypt is found to be weak in the future. It's a recommandation from @aeris

nono: And maybe also store the parameters from scrypt used to hash the password (`N` and `p`, I think)

nono: Oh, and we should start to call it a passphrase, not a password :wink:

instance/instance.go
```diff
@@ -30,11 +32,13 @@ var (
...
-	DocRev     string `json:"_rev,omitempty"` // couchdb _rev
-	Domain     string `json:"domain"`         // The main DNS domain, like example.cozycloud.cc
-	StorageURL string `json:"storage"`        // Where the binaries are persisted
-	storage    afero.Fs
+	DocID         string `json:"_id,omitempty"`  // couchdb _id
+	DocRev        string `json:"_rev,omitempty"` // couchdb _rev
+	Domain        string `json:"domain"`         // The main DNS domain, like example.cozycloud.cc
+	StorageURL    string `json:"storage"`        // Where the binaries are persisted
+	RegisterToken string `json:"registerToken,omitempty"`
+	PasswordHash  string `json:"passwordHash,omitempty"`
```
nono: I think the right type for `RegisterToken` and `PasswordHash` is `[]byte` and not `string` (the bytes sequence doesn't have to respect the UTF-8 encoding)

web/auth/auth.go
```diff
@@ -0,0 +1,72 @@
...
+func register(c *gin.Context) {
+	instance := middlewares.GetInstance(c)
+
+	var form registerForm
+	if err := binding.Form.Bind(c.Request, &form); err != nil {
+		c.AbortWithError(http.StatusBadRequest, err)
+	}
+
+	if err := instance.RegisterPassword(form.Password, form.Token); err != nil {
+		c.AbortWithError(http.StatusBadRequest, err)
```
nono: missing `return`

web/auth/session.go
```diff
@@ -0,0 +1,137 @@
...
+var (
+	// ErrNoCookie is returned by GetSession if there is no cookie
+	ErrNoCookie = errors.New("No session cookie")
+	// ErrInvalidID is returned by GetSession if the cookie contains wrong ID
+	ErrInvalidID = errors.New("Session cookie has wrong ID")
+	// ErrExpired is returned by GetSession if the cookie's session is old
+	ErrExpired = errors.New("Session cookie has wrong ID")
+)
+
+// A Session is an instance opened on a device
```
nono: s/device/browser/

web/auth/session.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+// ToCookie returns an http.Cookie for this Session
+func (s *Session) ToCookie() *http.Cookie {
+	return &http.Cookie{
+		Name:     SessionCookieName,
+		Value:    s.ID(),
+		MaxAge:   SessionMaxAge,
+		Path:     "/",
+		Domain:   s.Instance.Domain,
```
nono: I think that the domain should be prefixed by `.`, to make the browsers send the cookie when loading an application (`https://calendar.example.cozycloud.cc/`)

web/auth/session.go
```diff
@@ -0,0 +1,137 @@
...
+
+	return &s, nil
+
+}
+
+// ToCookie returns an http.Cookie for this Session
+func (s *Session) ToCookie() *http.Cookie {
+	return &http.Cookie{
+		Name:     SessionCookieName,
+		Value:    s.ID(),
```
nono: Are we sure that the generated IDs from couchdb can't be guessed? If an attacker has an account on the cozy cluster, he will an ID and can start trying IDs that are near this one

instance/instance.go
jinroh: I'm not really a fan of having dependencies like that to handle security in our code. Even if we copy/paste the library (or what's needed from it) it should *in-fine* be in our source tree and git repo.

We should be able to audit our code for security without having to go on github or any external source..

aenario: ok, I will copypaste some of simple-scrypt into cozy-stack. 
It handles all the complexity like the salt you mentionned below.

instance/instance.go
aenario: ok, adding an equivalent function.

instance/instance.go
aenario: will do, thanks

instance/instance.go
aenario: simple-scrypt include 16bit of salt with the hash. I will copy that behaviour

web/auth/session.go
aenario: It was my plan, at least until we are sure if session are going in couchdb / redis / other.

web/auth/session.go
aenario: will do

instance/instance.go
jinroh: 👍  it should be properly base64 en/decoded by [`encoding/json`](https://golang.org/pkg/encoding/json/#Marshal)

web/auth/session.go
jinroh: OK, np. Then just add a important security TODO, because it is not secure like that.

instance/instance.go
nono: No, I've just said something wrong. It doesn't matter

web/auth/auth.go
```diff
@@ -0,0 +1,84 @@
...
+	http.SetCookie(c.Writer, session.ToCookie())
+	c.Redirect(http.StatusSeeOther, instance.SubDomain(apps.OnboardingSlug))
+}
+
+func register(c *gin.Context) {
+	instance := middlewares.GetInstance(c)
+
+	var form registerForm
+	if err := binding.Form.Bind(c.Request, &form); err != nil {
+		fmt.Println(err, form)
```
jinroh: `Println` spotted

crypto/scrypt.go
```diff
@@ -0,0 +1,174 @@
...
+	"golang.org/x/crypto/scrypt"
+)
+
+// The code below is heavily inspired by https://github.com/elithrar/simple-scrypt
+
+// Scrypt params
+// TODO audit this parameters taken from simple-scrypt DefaultParams
+const n = 16384
+const r = 8
+const p = 1
```
jinroh: nitpick: we could rename those const to `defaultN`, `defaultR`, (or anything a bit more verbose...) to make sure they are not "shadowed" in a function and see more easily where they are actually used.

web/auth/auth.go
```diff
@@ -0,0 +1,82 @@
...
+
+	"github.com/cozy/cozy-stack/apps"
+	"github.com/cozy/cozy-stack/web/middlewares"
+	"github.com/gin-gonic/gin"
+	"github.com/gin-gonic/gin/binding"
+)
+
+type registerForm struct {
+	Passphrase string `form:"passphrase"`
+	Token      string `form:"registerToken"`
```
jinroh: Couldn't these fields be `[]byte` directly ?

aenario: Unlike `encoding/json`, `gin/binding/form` does not specialcase []byte, it will be considered a slice and will not be properly decoded [(source)](https://github.com/gin-gonic/gin/blob/bb159f9260a262fccbdbd6f37f8575d1e7e6aaa0/binding/form_mapping.go#L45)

web/auth/session.go
```diff
@@ -0,0 +1,143 @@
...
+	log "github.com/Sirupsen/logrus"
+	"github.com/cozy/cozy-stack/couchdb"
+	"github.com/cozy/cozy-stack/instance"
+	"github.com/cozy/cozy-stack/web/middlewares"
+	"github.com/gin-gonic/gin"
+)
+
+// SessionsType : The couchdb type for a session
+const SessionsType = "io.cozy.sessions"
+
```
jinroh: maybe all starting with `io.cozy` should be blacklisted ? i'm not sure.

web/data/data_test.go
```diff
@@ -62,6 +62,7 @@ func doRequest(req *http.Request, out interface{}) (jsonres map[string]interface
 	}
 	defer res.Body.Close()
 	body, err := ioutil.ReadAll(res.Body)
+	// fmt.Println("[result]", string(body))
```
nono: You can remove this line

web/auth/session.go
nono: @jinroh no, most `io.cozy` types will be used via `/data`, like contacts, calendars, settings, etc.

aenario:
Another issue we encountered during review with @poupotte : url encoded registerToken is not very pratical : for instance, httpie attemps some url-reencoding and we cant use httpie to test `/register`.

Will make a PR later to display the token as a hex string instead

PR #111, opened 2016-11-18
[VFS] Add a command server to execute ls / mv / ... commands through the CLI

jinroh:
This PR adds a command server to execute classical "commands" on the VFS. Commands are sent via ~~a TCP connection~~ the HTTP api to avoid spawning multiple cozy-stack. We could move this part of the code in a dedicated cli client at one point.

Even if we decide to change the way we implement this type of functionality, it has proven to be very useful for my developments, and I think it is not a luxury to have.

```
Available commands:

  mkdir [name]               Creates a directory with specified name
  touch [name]               Creates or change a file modifications time
  ls [-l] [-a] [-h] [name]   Prints the children of the specified directory
  tree [name]                Prints the tree structure of the specified directory
  attrs [name]               Prints the attributes of the specified file or directory
  cat [name]                 Echo the file content in stdout
  mv [from] [to]             Rename a file or directory
  rm [-f] [-r] [name]        Move the file to trash, or delete it permanently with -f flag
```


nono:
:+1: for the idea. But why don't you use the public Rest API?

jinroh:
@nono you're right, I took what I think was the simplest path to implement this but I agree we should use the HTTP api. Let me fix that :)

jinroh:
friendly 🆙 to merge this PR

aenario:
Coverall kindly reminds us that we will need more tests for this. 
@jinroh  , consider it if we have time on friday ;-)

PR #112, opened 2016-11-23
Update README: move docker specifics

jinroh:


jinroh:
I'll add some more thanks to @GoOz. Wait a little before merging.

PR #115, opened 2016-11-25
CORS middleware

jinroh:
Add a simple CORS middleware and fix the `Host` forwarding by the proxy.

### DIFF
--- a/web/router.go
+++ b/web/router.go
@@ -23,6 +23,11 @@
 package web
 
 import (
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+
 	"github.com/cozy/cozy-stack/web/apps"
 	"github.com/cozy/cozy-stack/web/auth"
 	"github.com/cozy/cozy-stack/web/data"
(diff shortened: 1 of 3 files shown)


web/router.go
```diff
@@ -38,10 +43,64 @@ func SetupRoutes(router *gin.Engine) {
 	router.Use(middlewares.ParseHost())
 	router.Use(middlewares.ServeApp(apps.Serve))
 	router.Use(middlewares.ErrorHandler())
+	router.Use(corsMiddleware("/apps", "/data", "/files"))
```
nono: Why did you insert the middleware here, and not like `middlewares.NeedInstance` in each `Group`?

web/router.go
```diff
@@ -38,10 +43,64 @@ func SetupRoutes(router *gin.Engine) {
...
+	preflightHeaders := make(http.Header)
+	preflightHeaders.Set("Access-Control-Allow-Credentials", "true")
+	preflightHeaders.Set("Access-Control-Allow-Methods", "*")
+	preflightHeaders.Set("Access-Control-Max-Age", strconv.FormatInt(int64(12*time.Hour/time.Second), 10))
+
+	return func(c *gin.Context) {
+		origin := c.Request.Header.Get("Origin")
+		if origin == "" {
+			return
+		}
```
nono: TODO for later: check that the `Origin` header matches a subdomain of the instance or a registered domain for an OAuth2 app/device.

web/router.go
nono: OK, you should note this in a comment. It could be helpful later to remember that

PR #116, opened 2016-11-25
Add possibility to specify fs-url in dev script

jinroh:
Add a `-f` option to the dev script in order to specify the `--fs-url` option of cozy-stack.

### DIFF
--- a/scripts/cozy-app-dev.sh
+++ b/scripts/cozy-app-dev.sh
@@ -80,6 +80,8 @@ do_start() {
 	do_start_proxy
 	check_hosts
 
+	echo "starting cozy-stack with ${vfsdir}..."
+
 	echo ""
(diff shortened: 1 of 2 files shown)


scripts/cozy-app-dev.sh
```diff
@@ -282,5 +288,11 @@ if [ ! -d ${appdir} ]; then
 	exit 1
 fi
 
+if [ -z "${vfsdir}" ]; then
+	vfsdir=$(echo "$(pwd)/storage")
```
nono: I should be missing something. Why the `echo` and not just this?

```bash
vfsdir="$(pwd)/storage"
```

jinroh: you're not missing anything... i fixed it

PR #119, opened 2016-11-29
Login

nono:


jinroh:
:shipit:

nono:
@aenario I have mixed feelings about this:

- It would be nice to just have a binary to deploy, and that go get can work out of the box with no options to pass.
- If we customize the login page for a partner, we might want to modify the HTML too (the link to cozy can be transformed to a link to the partner).
- It might be risky to let a client-side app overload the CSS & image. If an attacker can convince a user to install his app, it might confuse the user (probably not that much for login, but for oauth2 interactions that are more complex, it could).
- Having assets in a folder (HTML, CSS, images and later email templates) could be a way to customize them by a partner offering to host cozy instances.
- So, maybe a good solution could be to have a default version of the assets in the binary, and let the `--assets` options to overload it.

aenario:
@nono : agree this is delicate, my .02$

- The idea in [onboarding doc](https://github.com/cozy/cozy-stack/blob/aa372de477527e3e368dddf71e64f7bc5241f2a6/docs/onboarding.md) was to customize onboarding app for partners by branches / fork and eventually by some flags to be passed from cozy instances add to onboarding app. Having three separate processes (onboarding fork, CLI arguments, assets folder) for customization of onboarding & login seems unoptimal.
- Whereas we can imagine `home` app being customizable by the user (preference between dashboard / magic search bar / smartphone style icon grid), onboarding app is too security important: it handles the user password on registering.

So, while we keep it as an app in code because it has the same function (display HTML, has a token, changes data), it makes sense for `onboarding` to be locked and only modifiable by server admin and not instance owner.

If we go this way, we can even imagine login page to be handled by onboarding app too. This will brings onboarding app modes to 4 : 
- `register.html` with a registerToken, ask the user to pick a password and call `/register`
- `onboarding.html` with a contextToken, perform the follow up onboarding steps (initial config)
- `login.html` with a csrfLoginToken, ask the user for her password in a form that points to `/login`
- `oauth.html` with a csrfLoginToken & oauth scopes, display scopes & ask the user for her password in a form that points to `/login`

This way, all customization is done in the same way,  but it can make onboarding app overly complex...

--------

Alternatively, we'd need to make a list of everything that can change on the login page have these provided by CLI arguments to cozy-instances-add : **--login-help-url** **--login-logo-url**, ... Which can eventually be extended to **--login-index-page /path/to/file.html**.

All these customization SHOULD be optional, so that `cozy instances add xxxx` provides a functional cozy with sane defaults.

PR #121, opened 2016-11-30
Unify usage of folder vs directory and use directory

jinroh:
This PR propose to unify the usage of the words folder vs directory and uses the word directory only.

The breaking changes of this PR are:
  - the remove of the notion of `io.cozy.folders` for the creation API of files and directories. It uses for the query parameter `Type` the values `file` or `directory`
  - the renaming (and fixed doc) of the some directory IDs: `io.cozy.files.root-dir` and `io.cozy.files.trash-dir` (instead of `io.cozy.files.rootdir` and `io.cozy.files.trashdir`), to make it clearer for the next custom directories we'll have to introduce.

Also, for all the go code, the work `Dir` is used instead of `Directory` which was still hanging around in some function names, even if `Dir` was already mostly used.

----

If you prefer the word folder, I can change to that direction too. The main point of this PR is to unify things.

### DIFF
--- a/docs/instance.md
+++ b/docs/instance.md
@@ -53,10 +53,7 @@ and creates the proper databases ($PREFIX/$DOCTYPE) for these doctypes:
 
 Then, it creates the following indexes for these doctypes :
 
-- `io.cozy.apps : ["slug"]`
-- `io.cozy.manifest : ["url"]`
-- `io.cozy.files: ["dir_id", "name"], ["directory", "updated_at"]`
-- **TODO :** complete this list
+- **TODO :** complete this list of indexes
 
 Then, it creates some directories:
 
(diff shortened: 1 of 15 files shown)


docs/instance.md
```diff
@@ -48,32 +48,30 @@ and creates the proper databases ($PREFIX/$DOCTYPE) for these doctypes:
...
 - `io.cozy.files`
-- `io.cozy.folders`
 - `io.cozy.notifications`
 - `io.cozy.settings`
 
 Then, it creates the following indexes for these doctypes :
 
 - `io.cozy.apps : ["slug"]`
 - `io.cozy.manifest : ["url"]`
-- `io.cozy.files & io.cozy.folders: ["FolderId", "name"], ["FolderId", "modifiedDate"]`
```
aenario: should be ["dir_id", "updated_at"] or let's just remove it as we dont create it for now !

jinroh:
@aenario me too TBH. thanks [wikipedia](https://en.wikipedia.org/wiki/Directory_(computing)#Folder_metaphor)

PR #131, opened 2016-12-05
Use echov3 instead of gin

jinroh:


web/server.go
```diff
@@ -0,0 +1,217 @@
...
+
+func (r *renderer) ServeHTTP(res http.ResponseWriter, req *http.Request) {
+	r.h.ServeHTTP(res, req)
+}
+
+func createRenderer(conf *Config) (*renderer, error) {
+	// By default, use the assets packed in the binary
+	if conf.Assets != "" {
+		if !path.IsAbs(conf.Assets) {
+			return nil, fmt.Errorf("path %s is not absolute", conf.Assets)
```
nono: Please, can we keep the option to use a relative path for assets? It's handy when working on the CSS (`go run main.go serve --assets ./assets`)

web/server.go
```diff
@@ -0,0 +1,214 @@
...
+func Create(conf *Config) (*echo.Echo, error) {
+	appsRouter := echo.New()
+	apisRouter := conf.Router
+
+	r, err := createRenderer(conf)
+	if err != nil {
+		return nil, err
+	}
+
+	apisRouter.GET("/assets/*", echo.WrapHandler(http.StripPrefix("/assets/", r)))
```
nono: I think that _routing_ the `/assets` path should be in the router package

web/server.go
```diff
@@ -0,0 +1,214 @@
...
+	apisRouter.Renderer = r
```
nono: I don't see the purpose of this line? Is it something like a default renderer? But shouldn't the default to be a 404 error rendered by the HTTPErrorHandler?

web/server.go
jinroh: I don't really see how to do that without introducing some unnecessary complexity between the api routing and the asset handling... I could move all the asset handling in the `router` package though.

jinroh:
@aenario thank ! should've fixed your remarks in 302a7baa5b2cc14a1e5be34a09b3078efe3e7c53.

I think the `jsonapi.AbortWithError` was already removed.

jinroh:
Regarding the jsonapi errors, we should also update the data api that is not correct anymore. I can do that in another PR.

PR #132, opened 2016-12-06
Add redirect input on auth/login form

jinroh:


web/auth/auth.go
```diff
@@ -65,6 +85,39 @@ func logout(c echo.Context) error {
...
+	instance := middlewares.GetInstance(c)
+	parts := strings.SplitN(u.Host, ".", 2)
+	if len(parts) != 2 || parts[1] != instance.Domain || parts[0] == "" {
+		return "", echo.NewHTTPError(http.StatusBadRequest,
+			"bad url: should be subdomain")
+	}
+
+	// to protect against stealing authorization code with redirection, the
+	// fragment is always overriden.
+	u.Fragment = ""
```
nono: 2 things:

1. Do you know what is the scenario of such an attack? I know I have written that we should override the fragment to protect the authorization code, but I don't remember what are exactly the steps that can conduct to this attack (the authorization code is in the query string, not the fragment).

2. I don't think that setting the fragment to an empty string is enough. When the user clicks on a link `https://example.org/foo#bar`, his browser sends a request to the server with just `https://example.org/foo`. If the server redirects it to `https://example.org/qux`, the browser will then add the fragment and show `https://example.org/qux#bar` to the user.

PR #134, opened 2016-12-06
Split handlers PATCH handlers by path and by ID

jinroh:
`PATCH /files/metadata` was broken because of the way echo works.

I splitted the code in two handlers for `PATCH /files/metadata` and `PATCH /files/:file-id` which I think is cleaner.


nono:
In fact, I think it was we wanted to do initially, but gin router couldn't handle that.

PR #135, opened 2016-12-07
Add _changes API

aenario:
This is the first step in allowing couch protocol replication from and to cozy-stack.

There is no documentation for it as the only guarantee we make on this API for now is "It will allow a couchdb or pouchdb to replicate from a cozy" 

What differs from exploration : 
I added style=all_docs support, as couchdb 2 seems to use it in replication

### DIFF
--- a/couchdb/changes_test.go
+++ b/couchdb/changes_test.go
@@ -1,0 +1,1 @@
+package couchdb
--- a/couchdb/changes_test.go
+++ b/couchdb/changes_test.go
@@ -1,1 +1,0 @@
-package couchdb
--- a/web/data/data.go
+++ b/web/data/data.go
@@ -2,7 +2,6 @@
 package data
 
 import (
-	"fmt"
 	"net/http"
 	"strconv"
 
@@ -49,7 +48,7 @@ func createDoc(c echo.Context) error {
 	doctype := c.Get("doctype").(string)
 	instance := middlewares.GetInstance(c)
 
-	var doc = couchdb.JSONDoc{Type: doctype}
+	doc := couchdb.JSONDoc{Type: doctype}
 	if err := c.Bind(&doc.M); err != nil {
 		return jsonapi.NewError(http.StatusBadRequest, err)
 	}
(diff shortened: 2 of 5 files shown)


web/data/data.go
```diff
@@ -205,9 +207,60 @@ func findDocuments(c echo.Context) error {
...
+		return jsonapi.NewError(http.StatusBadRequest, err)
+	}
+
+	feedStyle, err := couchdb.ValidChangesStyle(c.QueryParam("style"))
+	if err != nil {
+		return jsonapi.NewError(http.StatusBadRequest, err)
+	}
+
+	var limitString = c.QueryParam("limit")
+	var limit = 0
```
jinroh: nit: why `var` and not `:=`

web/data/data.go
```diff
@@ -205,9 +207,60 @@ func findDocuments(c echo.Context) error {
...
+		}
+	}
+
+	results, err := couchdb.GetChanges(instance, &couchdb.ChangesRequest{
+		DocType: c.Get("doctype").(string),
+		Feed:    feed,
+		Style:   feedStyle,
+		Since:   c.QueryParam("since"),
+		Limit:   limit,
+	})
```
jinroh: missing err handling

web/data/data.go
```diff
@@ -205,9 +207,60 @@ func findDocuments(c echo.Context) error {
...
+}
+
+func changesFeed(c echo.Context) error {
+	instance := middlewares.GetInstance(c)
+
+	// Drop a clear error for parameters not supported by stack
+	for key := range c.QueryParams() {
+		if !allowedChangesParams[key] {
+			err := fmt.Errorf("Unsuported query parameter '%s'", key)
+			return jsonapi.NewError(http.StatusBadRequest, err)
```
jinroh: the `jsonapi.NewError` will `Sprintf` for you. You can do:

```go
jsonapi.NewError(http.StatusBadRequest, "Unsuported query parameter '%s'", key)
```

couchdb/changes_test.go
```diff
@@ -0,0 +1 @@
+package couchdb
```
jinroh: unwanted file (i think)

PR #136, opened 2016-12-07
OAuth2 Dynamic Client Registration

nono:


crypto/jwt.go
```diff
@@ -0,0 +1,14 @@
+package crypto
+
+import jwt "github.com/dgrijalva/jwt-go"
```
jinroh: maybe use `gopkg.in/dgrijalva/jwt-go.v2` instead to not rely on a github repo here

web/auth/client.go
```diff
@@ -0,0 +1,154 @@
...
+
+	var err error
+	c.ClientSecret, err = crypto.NewJWT(i.HmacSecret, jwt.StandardClaims{
+		Audience: "client_secret",
+		Issuer:   i.Domain,
+		IssuedAt: time.Now().Unix(),
+		Subject:  c.CouchID,
+	})
+	if err != nil {
+		return &ClientRegistrationError{
```
aenario: Error swallowing : we should at least log the original error

web/auth/client.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+	}
+	c.RegistrationToken, err = crypto.NewJWT(i.HmacSecret, jwt.StandardClaims{
+		Audience: "registration",
+		Issuer:   i.Domain,
+		IssuedAt: time.Now().Unix(),
+		Subject:  c.CouchID,
+	})
+	if err != nil {
+		return &ClientRegistrationError{
```
aenario: error swallowing

web/auth/client.go
```diff
@@ -0,0 +1,154 @@
...
+			Code:        http.StatusBadRequest,
+			Error:       "invalid_redirect_uri",
+			Description: "redirect_uris is mandatory",
+		}
+	}
+	for _, redirectURI := range c.RedirectURIs {
+		u, err := url.Parse(redirectURI)
+		if err != nil ||
+			(u.Scheme != "http" && u.Scheme != "https") ||
```
aenario: Wont we need non-http schema for Desktop & Mobile clients ?

web/auth/auth.go
```diff
@@ -130,6 +130,24 @@ func checkRedirectParam(c echo.Context, defaultRedirect string) (string, error)
 	return u.String() + "#", nil
 }
 
+func registerClient(c echo.Context) error {
+	// TODO add rate-limiting to prevent DOS attacks
+	if c.Request().Header.Get("Content-Type") != "application/json" {
+		return c.JSON(http.StatusBadRequest, echo.Map{
```
aenario: You can use a echo.HTTPError ?

nono: I don't see how to respond a JSON with an `echo.HTTPError` (and not jsonapi)

web/auth/client.go
nono: For desktop, it will be something like `http://localhost:3000/oauth/callback`, but for mobile, you are right! I'm going to fix this

web/auth/client.go
nono: OK

web/auth/auth.go
aenario: ok let's leave it at that and add a Debt card on cleanup handling of error with jsonapi vs non-jsonapi once we have more points.

PR #139, opened 2016-12-08
Add redirect on login (POST /auth/login)

jinroh:
This PR adds the same redirect mechanism as #132 on the login part (`POST /auth/login`)


jinroh:
@nono sorry for adding stuff after your review but I saw that the redirection was not effective for an already logged in user, d48a286433a873e263902309368d70d9aed79dab should fix it.

I also removed the `HasRedirect` which is not useful and adds unnecessary complexity IMHO.

PR #140, opened 2016-12-08
Add jobs doc

jinroh:


docs/jobs.md
```diff
@@ -0,0 +1,432 @@
+Cozy Jobs
+=========
+
+Jobs are designed to represent asynchronous tasks that your cozy can execute. These tasks can be scheduled in advance, recurring or suddenn and provide various services.
```
nono: s/suddenn/sudden/

docs/jobs.md
```diff
@@ -0,0 +1,432 @@
...
+
+At the time, we do not provide "generic" and programmable tasks. This list of available workers is part of the defined API.
+
+The job queue can be either local the cozy-stack when used as a monolitic instance (self-hosted for instance) or distributed via a redis server for distributed infrastructures.
```
nono: local **to** the cozy-stack?

docs/jobs.md
```diff
@@ -0,0 +1,432 @@
...
+
+
+### Launcher rate limitation
+
+Every launcher has a rate limitation policy to prevent launchers from spawning too many jobs at the same time. The specific rules are not yet decided, but they should work along these two properties:
+
+  - *parallel limitations*: each launcher will have a limited number of workers it is allowed to queue in parallel
+  - *back pressure*: each launcher should have a backpressure policy to drop job spawning when not necessary. For instance: 
+    - in the case of an `@event` launcher, a job doing an external API call should not be spawned if another one is already running for another close event.
+    - when no worker is availabe, or the queue has too many elements, it can decide to drop the job action (given some informations)
```
nono: s/availabe/available/

docs/jobs.md
```diff
@@ -0,0 +1,432 @@
...
+  - `io.cozy.launchers` for launchers
+
+
+Launchers
+---------
+
+Jobs can be launched by three types of launchers:
+
+  - `@cron` to schedule recurring jobs, either periodic or scheduled at specific times
+  - `@interval` to schedule jobs executed at a given fix interval
```
nono: My first impression is that `@cron` and `@interval` are 2 ways to do the same thing. What are the differences between them?

docs/jobs.md
```diff
@@ -0,0 +1,432 @@
...
+
+
+### `@interval` syntax
+
+The `@interval` launcher uses the same syntax as golang's `time.ParseDuration`:
+
+```
+A duration string is a possibly signed sequence of decimal numbers, each with
+optional fraction and a unit suffix, such as "300ms", "1.5h" or "2h45m". Valid
+time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
```
nono: It looks dangerous to accept units below the second.

docs/jobs.md
```diff
@@ -0,0 +1,432 @@
...
+  - *parallel limitations*: each launcher will have a limited number of workers it is allowed to queue in parallel
+  - *back pressure*: each launcher should have a backpressure policy to drop job spawning when not necessary. For instance: 
+    - in the case of an `@event` launcher, a job doing an external API call should not be spawned if another one is already running for another close event.
+    - when no worker is availabe, or the queue has too many elements, it can decide to drop the job action (given some informations)
+
+
+Error Handling
+--------------
+
+Jbos can fail to execute their task. We have two ways to parametrize how they should behave in suche cases.
```
nono: s/Jbos/Jobs/

docs/jobs.md
```diff
@@ -0,0 +1,432 @@
...
+
+### Retry
+
+A retry count can be optionnaly specified to ask the worker to re-exute the task if it has failed.
```
nono: s/exute/execute/

docs/jobs.md
```diff
@@ -0,0 +1,432 @@
...
+
+Each retry is executed after a configurable delay. The try count is part of the attributes of the job. Also, each occuring error is kept in the `errors` field containing all the errors that may have happened.
+
+### Timeout
+
+A worker may never end. To prevent this, a configurable timeout value is specified with the job.
+
+If a job does not end after the specified amount of time, it will be aborted. A timeout is just like another error from the worker and can provoke a retry if specified.
+
```
nono: I think we should have sensible defaults for both the retries and the timeout. Maybe, something like 3 retries and a timeout of 1 minute.

docs/jobs.md
```diff
@@ -0,0 +1,432 @@
...
+Job Data
+--------
+
+Example and description of the attributes of a `io.cozy.jobs`:
+
+```js
+{
+  "worker": "sendmail",    // worker type name
+  "worker_id": "123123",   // worker id, if any
```
nono: I think the `worker` key is enough

docs/jobs.md
```diff
@@ -0,0 +1,432 @@
...
+  "launcher": "@cron",     // "@cron", "@interval", "@event" or ""
+  "launcher_id": "1234",   // launcher id, if any
```
nono: I'm not sure to see why you put the launcher here. Can you explain?

docs/jobs.md
```diff
@@ -0,0 +1,432 @@
...
+  "options": {
+    "priority": 3,         // priority from 1 to 100
+    "timeout": 60,         // timeout value in seconds
+    "retry_max_count": 3,  // maximum number of retry
+    "retry_delay": 10,     // retry delay in seconds between each try
+    "argument": {},        // arguments message
```
nono: I'd have put arguments outside of the options. But, not a big deal.

docs/jobs.md
```diff
@@ -0,0 +1,432 @@
...
+  },
+  "state": "running",      // queued, running, errored
+  "worker_id": "123456",   // or unknown if in the queue
```
nono: s/worker_id/job_id/ ? Or is it something else?

docs/jobs.md
```diff
@@ -0,0 +1,432 @@
...
+      "rev": "1-12334",
+      "attributes": {},
+      "links": {
+        "self": "/jobs/launchers/123123"
+      }
+    },
+  ]
+}
+```
+
```
nono: We should probably also have a way to remove a launcher

docs/jobs.md
```diff
@@ -0,0 +1,432 @@
...
+  - `io.cozy.jobs` for jobs
+  - `io.cozy.launchers` for launchers
+
+
+Launchers
+---------
+
+Jobs can be launched by three types of launchers:
+
+  - `@cron` to schedule recurring jobs, either periodic or scheduled at specific times
```
nono: Is it possible to schedule a one-time job? Like send me a notification for this event in my calendar 10 minutes before it start. I don't see how to do that with the `@cron` syntax below.

docs/jobs.md
nono: s/argument/arguments/ ?

docs/jobs.md
```diff
@@ -0,0 +1,432 @@
...
+      "self": "/jobs/123123"
+    }
+  }
+}
+```
+
+
+### GET /jobs/queue
+
+List the jobs in the queue.
```
nono: I think that I'd want to have the number of jobs in queue, before having a list of paginated jobs.

docs/jobs.md
```diff
@@ -0,0 +1,432 @@
...
+### Launcher rate limitation
+
+Every launcher has a rate limitation policy to prevent launchers from spawning too many jobs at the same time. The specific rules are not yet decided, but they should work along these two properties:
+
+  - *parallel limitations*: each launcher will have a limited number of workers it is allowed to queue in parallel
+  - *back pressure*: each launcher should have a backpressure policy to drop job spawning when not necessary. For instance: 
+    - in the case of an `@event` launcher, a job doing an external API call should not be spawned if another one is already running for another close event.
+    - when no worker is availabe, or the queue has too many elements, it can decide to drop the job action (given some informations)
+
+
```
nono: How do you see the stack taking the jobs? We have n goroutines that can process jobs, and when one of them is free, we take the jobs with the highest priority (or the first arrived in case of egality)? Or we have something more complex with a limit by worker (like no more than 3 thumbnails generation in parallel because it takes a lot of RAM)?

docs/jobs.md
```diff
@@ -0,0 +1,432 @@
...
+Example and description of the attributes of a `io.cozy.jobs`:
+
+```js
+{
+  "worker": "sendmail",    // worker type name
+  "worker_id": "123123",   // worker id, if any
+  "launcher": "@cron",     // "@cron", "@interval", "@event" or ""
+  "launcher_id": "1234",   // launcher id, if any
+  "options": {
+    "priority": 3,         // priority from 1 to 100
```
nono: What is the highest priority? 1 or 100?

nono:
Good job! I have some questions & remarks, but it's very good base to construct on.

docs/jobs.md
```diff
@@ -0,0 +1,432 @@
...
+
+The job queue can be either local the cozy-stack when used as a monolitic instance (self-hosted for instance) or distributed via a redis server for distributed infrastructures.
+
+This doc introduces two cozy types:
+
+  - `io.cozy.jobs` for jobs
+  - `io.cozy.launchers` for launchers
+
+
+Launchers
```
aenario: I feel triggers is the more used term for this.

docs/jobs.md
aenario: parameterize  such

docs/jobs.md
```diff
@@ -0,0 +1,432 @@
...
+
+The `@event` syntax is not determined yet. Its main purpose will be to describe job scheduling after a filesystem or database modification.
+
+
+### Launcher rate limitation
+
+Every launcher has a rate limitation policy to prevent launchers from spawning too many jobs at the same time. The specific rules are not yet decided, but they should work along these two properties:
+
+  - *parallel limitations*: each launcher will have a limited number of workers it is allowed to queue in parallel
+  - *back pressure*: each launcher should have a backpressure policy to drop job spawning when not necessary. For instance:
```
aenario: This also brings the idea of leading / trailing debounce with `@event`
ie. a job that do some action after any contact update, but should only be triggered once after several contacts are updated.

docs/jobs.md
aenario: Privacy complication : an app has the permission sendMail, it does not mean it should be able to see all mails in the queue to be sent.

jinroh: Agreed, we should probably introduce a queue for each application. I'll probably add a "Permissions" paragraph to discuss that.

docs/jobs.md
jinroh: Thanks ! Added a comment on throttling / debouncing which is more explicit.

docs/jobs.md
jinroh: Added a paragraph on defaults.

docs/jobs.md
jinroh: This is more a "debuggability" / "inspection" feature. I think it may be useful to known how / why the job has been launched.

docs/jobs.md
jinroh: I would like the provide the list associated with metadata about its total size. How would you spec that with JSON-API... ?

nono: You can use `meta`:

```json
{
  "links": {
    "next": "/jobs/queue?page[cursor]=123123"
  },
  "data": [
    {
      "type": "io.cozy.jobs",
      "id": "123123",
      "rev": "1-12334",
      "attributes": {},
      "links": {
        "self": "/jobs/123123"
      },
      "meta": {
          "count": 456
       }
    },
  ]
}
```

nono:
I still don't see how to schedule a one-time job in the future:

> Is it possible to schedule a one-time job? Like send me a notification for this event in my calendar 10 minutes before it start. I don't see how to do that with the @cron syntax below.

docs/jobs.md
```diff
@@ -0,0 +1,477 @@
...
+
+
+Triggers
+--------
+
+Jobs can be launched by three different types of triggers:
+
+  - `@cron` to schedule recurring jobs scheduled at specific times
+  - `@interval` to schedule periodic jobs executed at a given fix interval
+  - `@event` to launch a job after a change in the cozy
```
nono: This week-end, I thought of a new type of trigger: `@webhook`. It would create an unguessable URL that we can give to external services for pushing a job. For example, we could give a webhook URL to github to update on our cozy the app we are working on after each git push.

jinroh: This is nice, I'll add it the spec.

docs/jobs.md
jinroh: which mean we would have to duplicate the meta informations about the list in each of its elements ?

nono: No, we should probably put in the global meta. Like this:

```json
{
  "links": {
    "next": "/jobs/queue?page[cursor]=123123"
  },
  "data": ["..."],
  "meta": {
      "count": 456
   }
}
```

docs/jobs.md
```diff
@@ -0,0 +1,509 @@
...
+This doc introduces two cozy types:
+
+  - `io.cozy.jobs` for jobs
+  - `io.cozy.triggers` for triggers
+
+
+Triggers
+--------
+
+Jobs can be launched by three different types of triggers:
```
nono: s/three/six/

docs/jobs.md
```diff
@@ -0,0 +1,509 @@
...
+
+  - `@cron` to schedule recurring jobs scheduled at specific times
+  - `@interval` to schedule periodic jobs executed at a given fix interval
+  - `@at` to schedule a one-time job executed after at a specific time in the future
+  - `@in` to schedule a one-time job executed after a specific amount of time
+  - `@event` to launch a job after a change in the cozy
+  - `@webhook` to launch a job by requesting an unguessable URL
+
+These three triggers have specific syntaxes to describe when jobs should be scheduled. See below for more informations.
```
nono: s/three/six/

PR #141, opened 2016-12-09
Oauth2 show the authorize page to the user

nono:
It includes the #136 PR


nono:
Yes, I should add tests

PR #143, opened 2016-12-12
Add HMACed sessions ids as session cookies

jinroh:
Use a message authentication code to validate the integrity and authenticity of our session cookies. The MAC is generated with a secret key unique to the user with an HMAC-SHA256.

~~The message's value is encoded in JSON.~~ The message itself is encoded in Base64.

The issued time is added to the message and MACed against. The name of the cookie is *not* part of the message but is also MACed against.


nono:
I'm a bit puzzled by your PR. I have some questions to help me understand your choices:

- Why did you choose to not use JWT?
- Why did you choose to HMAC-SHA256 over HMAC-SHA512 (which is used for OAuth2)? Is it to have smaller cookies?
- Why did you not reuse the `HmacSecret`?
- Did you find a source about what should be the length of secrets when using HMAC-SHA256 (or 512)?

It's not an attack on your work. It's just that this part is critical and I really want to be sure that we are making the right choices security-wise.

crypto/mac.go
```diff
@@ -0,0 +1,179 @@
...
+	MaxAge int64
+	MaxLen int
+}
+
+func assertMACConfig(c *MACConfig) {
+	if c.Key == nil {
+		panic("hash key is not set")
+	}
+	if len(c.Key) < 16 {
+		panic("hash key should not long enough")
```
nono: s/should/is/ ?

crypto/mac.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+}
+
+// EncodeAuthMessage encodes a value in JSON and associates it with a message
+// authentication code for integrity and authenticity.
+//
+// If the value, when JSON encoded and base64 encoded with a fixed size header
+// is shorter than the configured maximum length, it will panic.
```
nono: s/shorter/longer/ ?

instance/instance_test.go
```diff
@@ -57,12 +57,12 @@ func TestGetCorrectInstance(t *testing.T) {
...
 
-func TestInstancehasHmacSecret(t *testing.T) {
+func TestInstancehasOAuthSecret(t *testing.T) {
 	instance, err := Get("test.cozycloud.cc")
 	if assert.NoError(t, err) {
 		assert.NotNil(t, instance)
-		assert.NotNil(t, instance.HmacSecret)
-		assert.Equal(t, len(instance.HmacSecret), 64)
+		assert.NotNil(t, instance.OAuthSecret)
+		assert.Equal(t, len(instance.OAuthSecret), 64)
```
nono: s/64/oauthSecretLen/

web/auth/session.go
```diff
@@ -141,17 +137,44 @@ func (s *Session) Delete(i *instance.Instance) *http.Cookie {
...
+//
+// About MaxLength, for a session of size 100 bytes
+//
+//       8 bytes time
+//   +  32 bytes HMAC-SHA256
+//   + 100 bytes session
+//   + base64 encoding (4*n/3 + 2 padding)
+//   < 200 bytes
+//
+// 2048 bytes should be sufficient enough to support any type of session.
```
nono: IMHO, 2048 is too much. A TCP packet is less than 1500 bytes, and cookies are sent often (less with http2, but still). I think we should limit to 256 or 512 bytes.

jinroh:
@nono No problem !

To put a little context I wanted to first use the [securecookie](https://github.com/gorilla/securecookie) package from gorilla I found it overly complex for what I wanted to do (not saying their implementations is flawd):

  - their encoding rely on serialization+base64 and base64 again because they use string separators, which is a overly complex and makes payload biggers
  - they provide encryption of the cookie which adds complexity without more security for us: we should not have sensistive informations stored in the cookie itself, and it should not be replayed thanks to SSL, not an encryption of its content
  - their API provide a Decode + Encode with the same context which is a little odd
  - they have IMHO [overly complex errors](https://github.com/gorilla/securecookie/blob/master/securecookie.go#L57-L113)
  - their implementation is >= 600LOC while mine is ~180LOC.

Hope this gives a little context around this PR to make you believe this is not NIH or "I want to do my own crypto".

Here is my reasonings for your questions:

> Why did you choose to not use JWT?

I don't think JWT is a good candidate here. We want to authenticate a session with a signed-cookie for a long-lived session state. JWT are better canditates for single-use authorization token. More reasonings in [this article](http://cryto.net/~joepie91/blog/2016/06/13/stop-using-jwt-for-sessions/)

I've chosen JSON encoding in order to allow us to store more elements in the session cookie. But if we think we'll only store an ID, we could remove serialization completely and have a simple byte slice as value only.

> Why did you choose to HMAC-SHA256 over HMAC-SHA512 (which is used for OAuth2)? Is it to have smaller cookies?

We could use SHA-512 but I'm not sure it would add more security. I have not seen informations stating that SHA-256 was broken and SHA-512 should be used to increase security.

This also was the default hash function use for the securecookie HMAC.

The size of the cookie does not really matter here I think.

> Why did you not reuse the HmacSecret?

I hesitated to do so I finally decided not to reuse the key for the same reasons I chose not to use JWT. The OAuth secret key is used to authenticate a one-time exchange while the session secret is used to authenticate long-term sessions.

We may need to have a different key rotation scheme for OAuth and sessions handling. Also we may have different security requirements for these two secrets since they are not used in the same contexts at all.

> Did you find a source about what should be the length of secrets when using HMAC-SHA256 (or 512)?

http://security.stackexchange.com/a/96176

> It's not an attack on your work. It's just that this part is critical and I really want to be sure that we are making the right choices security-wise.

No problem ! I think this is good we have this kind of PRs with the code inside the PR instead of hidden in an external dependency (event if its copy/paste) because we have these kind of discussions.

jinroh:
@nono using SHA512 may be a good idea if we know we'll run on 64bits architecture on which it may be  faster that SHA256. Although this would require some measurements on said machines.

jinroh:
I removed the JSON serialisation of the message's value since we don't really need it as long as we only store an ID in the message.

nono:
> Hope this gives a little context around this PR to make you believe this is not NIH or "I want to do my own crypto".

Well maybe :wink:

> I don't think JWT is a good candidate here. We want to authenticate a session with a signed-cookie for a long-lived session state. JWT are better canditates for single-use authorization token. More reasonings in this article

You have made something that is really close to what is called Stateful JWT in this article (persisted in a cookie). Let's take the drawbacks from the article:

>> They take up more space

Maybe, but I'm not sure your solution is really better than JWT on that front. You have JSON serialisation and base64 encoding too.

>> They are less secure

Not relevant. As the article says, _When storing your JWT in a cookie, it's no different from any other session identifier_, and it is what we would have use if we were going to JWT.

>> You cannot invalidate individual JWT tokens

Yes, but you can't invalidate a cookie too. In both case, you invalidate the session on the server. It's the same.

>> Data goes stale

Not for stateful JWT (well, it's exactly the same as a cookie).

>> Implementations are less battle-tested or non-existent

Echo doesn't have a battle-tested implementation of signed cookies. And your implementation is less battle-tested than the JWT package.

nono:
You have removed JSON serialization, so let's keep your implementation. It's a question of simplicity+smaller cookies on one side, versus a package a little more battle-tested. And I favor simplicity.

nono:
>> Did you find a source about what should be the length of secrets when using HMAC-SHA256 (or 512)?

> http://security.stackexchange.com/a/96176

Well, it said that 128 bits should be enough and it doesn't discourage to use more. And, in fact the 128 bits secret is just pad with zeroes to a 512 bits value. I don't a good reason to not use a larger secret.

aenario:
I may be mistaken, but I think we are mixing up 2 different things here : 

A session cookie is just supposed to be the unique key to a session object stored server-side. 

A gorilla/securecookie (a.k.a stateless cookie) aims to avoid the need for sever-side session storing and instead put session data into the cookie, this is usefull for hyperscale, but of little use to us as we wont scale more than couchdb.

They are 3 main attacks we need to protect against :

**Session hijacking** : where an attacker **knows** an user sessionID. We assume we are protected against this by SSL and browser security. There is not much we can do except not leaking session ID into links (cf  :eggplant:  @pierrevdk)

**Session prediction**  : where the user **guess** what is going to be an user sessionID, this is what we were afraid with using couchdb `_id` as a session ID. We can protect against this either by signing the cookie or by using a cryptorandom value. The former use CPU on each request.

**Session fixation** : where the attacker force the user's sessionID to a known value, then get her to authenticate and then  use the known sessionID. We are protected against this as users only get a sessionID when they are connected.

nono:
@aenario this PR is about the second attack, **session prediction**. We were using UUID for the session id and they can be guessed/brute-forced by another user on the same cozy-stack server. With this PR, the cookie will be signed. Do you think that using a crypto-random value for session ID would be better?

jinroh:
> Session hijacking : where an attacker knows an user sessionID. We assume we are protected against this by SSL and browser security. There is not much we can do except not leaking session ID into links (cf 🍆 @pierrevdk)

SSL prevents the reuse of a signed cookie. Here, even if we leak a session ID, the attacker would have to sign it with a secret he cannot guesses.

> Session prediction : where the user guess what is going to be an user sessionID, this is what we were afraid with using couchdb _id as a session ID. We can protect against this either by signing the cookie or by using a cryptorandom value. The former use CPU on each request.

This PR should also prevent that. Although we could also choose to generate session IDs randomly but then we would have to be careful to not leak them. The ID becomes the secret which may not be obvious.

jinroh:
Using the session ID as the secret (basically not relying on a MAC but only on a random value only known by the server) would make it a bit more complicated to make session rotation I think...

aenario:
> Do you think that using a crypto-random value for session ID would be better?

It would be simpler. If there is no other advantages to HMAC cookie (not saying there is not, just making sure), simpler is better.

> SSL prevents the reuse of a signed cookie. Here, even if we leak a session ID, the attacker would have to sign it with a secret he cannot guesses.

IMHO, this is the only advantage the HMAC approach has over cryptorandom : it protects us from escalation of other security issues. **However,** it only does so in the very limited case where the attacker can see the session couchdb.Doc but not the instance couchdb.Doc.

> Using the session ID as the secret (basically not relying on a MAC but only on a random value only known by the server) would make it a bit more complicated to make session rotation I think...

I am not sure I understand this point : the session _id belongs to 1 device, when it connects, we just set another sessionid in the Cookie. Whereas the hmacSecret is shared by all devices, so rotating it implies that all devices need to connect at least once per rotation.

nono:
When a user thinks his passphrase was stolen, he will change it. It invalidates all the sessions (and probably creates a new one for the current device). To invalidate all the sessions, you can remove them in CouchDB one by one (and maybe in redis if we want to cache them), or you can change the secret used to sign all the sessions.

aenario:
:+1: OK, got it, let's go with HMAC then.

web/auth/auth.go
```diff
@@ -224,9 +229,15 @@ func authorizeForm(c echo.Context) error {
 }
 
 // IsLoggedIn returns true if the context has a valid session cookie.
-func IsLoggedIn(c echo.Context) bool {
+func IsLoggedIn(c echo.Context) (bool, error) {
```
nono: I'm not sure why to change this. We don't care if there is an error. Maybe it's just the secret that has been rotated. We just want to know if the user is logged in, or not.

jinroh: I agree. `GetSession` should be sufficient in case we need the error when checking if `LoggedIn`.

PR #145, opened 2016-12-12
Add administration api and rely on it for instances cli

jinroh:
This PR adds an administration endpoint on a separate port (6060 by default).

This administration offers for now the possibility to handle instances through an HTTP API:

- `POST /instances/?Domain=...`
- `GET /instances/`
- `DELETE /instances/:domain`

The `instances` cli command uses has been changed to use this API.

### DIFF
--- a/cmd/instances.go
+++ b/cmd/instances.go
@@ -4,11 +4,14 @@ import (
 	"bufio"
 	"encoding/hex"
 	"fmt"
+	"net/url"
 	"os"
 	"strings"
 
 	log "github.com/Sirupsen/logrus"
+	"github.com/cozy/cozy-stack/config"
 	"github.com/cozy/cozy-stack/instance"
+	"github.com/howeyc/gopass"
 	"github.com/spf13/cobra"
 )
 
@@ -28,7 +31,7 @@ process can manage several instances.
 
 Each instance has a separate space for storing files and a prefix used to
 create its CouchDB databases.
-	`,
+`,
 	Run: func(cmd *cobra.Command, args []string) { cmd.Help() },
 }
 
@@ -38,7 +41,7 @@ var addInstanceCmd = &cobra.Command{
 	Long: `
 cozy-stack instances add allows to create an instance on the cozy for a
 given domain.
-	`,
+`,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		if len(args) == 0 {
 			return cmd.Help()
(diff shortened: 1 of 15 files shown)


cmd/instances.go
```diff
@@ -27,6 +29,7 @@ For example, bob.cozycloud.cc is the instance of Bob. A single cozy-stack
 process can manage several instances.
 
 Each instance has a separate space for storing files and a prefix used to
+Each instance has a separate space for storing files and a prefix used to
```
nono: Duplicate line

cmd/instances.go
```diff
@@ -0,0 +1,254 @@
...
+	"github.com/cozy/cozy-stack/instance"
+	"github.com/howeyc/gopass"
+	"github.com/spf13/cobra"
+)
+
+var flagLocale string
+var flagApps []string
+
+// instanceCmdGroup represents the instances command
+var instanceCmdGroup = &cobra.Command{
```
aenario: [nitpick] we are already in the "instances" file, there is no need to put the word instance in all private variable names.

instance/instance.go
```diff
@@ -68,6 +69,21 @@ func (i *Instance) Rev() string { return i.DocRev }
 // SetRev implements couchdb.Doc
 func (i *Instance) SetRev(v string) { i.DocRev = v }
 
+// SelfLink is used to generate a JSON-API link for the directory
```
aenario: directory -> instance

cmd/instances.go
```diff
@@ -0,0 +1,254 @@
...
+		return nil, err
+	}
+	return doc.Data, nil
+}
+
+func init() {
+	instanceCmdGroup.AddCommand(addInstanceCmd)
+	instanceCmdGroup.AddCommand(lsInstanceCmd)
+	instanceCmdGroup.AddCommand(destroyInstanceCmd)
+	instanceCmdGroup.AddCommand(passwdInstanceCmd)
```
aenario: passwd should not be a children of instanceCmdGroup

web/instances/instances.go
```diff
@@ -0,0 +1,65 @@
+package instances
```
aenario: may be we should move this file to a separate web/admin/instances hierarchy

cmd/instances.go
```diff
@@ -0,0 +1,254 @@
...
+type instancesAPIData struct {
+	Data []*instanceData `json:"data"`
+}
+
+func instancesClient() *client {
+	var password []byte
+
+	if !config.IsDevRelease() {
+		var err error
+		fmt.Printf("Password:")
```
aenario: If the user starting the cmd instances can read the AdminConfigFile, it is not necessary to ask the user to enter it manually.

cmd/instances.go
jinroh: but we are the the `cmd` package

cmd/instances.go
jinroh: yes I agree its a bit weird under the instances command... you think it should be a root command ?

cmd/instances.go
jinroh: I'm not sure I understand your comment. What config file are you talking about ? The file containing the hashed password ?

cmd/instances.go
aenario: ok, nevermind

cmd/instances.go
aenario: I did not pay attention to the fact the password was hashed. Nevermind.

jinroh:
rebased and squashed

PR #147, opened 2016-12-13
OAuth POST /auth/authorize

nono:


web/auth/auth.go
```diff
@@ -231,12 +291,22 @@ func IsLoggedIn(c echo.Context) bool {
 
 // Routes sets the routing for the status service
 func Routes(router *echo.Group) {
+	noCSRF := middleware.CSRFWithConfig(middleware.CSRFConfig{
+		TokenLookup:    "form:csrf_token",
+		CookieMaxAge:   3600, // 1 hour
+		CookieHTTPOnly: true,
+		CookieSecure:   true,
+	})
```
jinroh: could it be interesting to store the CSRF in the session instead of using another cookie ?

nono: Well, I think another cookie is fine. I don't see a pro for using the session for CSRF, and there is a con: the echo middleware won't support that out of the box.

PR #149, opened 2016-12-14
OAuth access token endpoint

nono:


web/auth/auth.go
```diff
@@ -288,6 +289,101 @@ func authorize(c echo.Context) error {
...
+		}
+		out.Scope = accessCode.Scope
+		out.Refresh, err = client.CreateJWT(instance, RefreshTokenAudience, out.Scope)
+		if err != nil {
+			return c.JSON(http.StatusInternalServerError, echo.Map{
+				"error": "Can't generate refresh token",
+			})
+		}
+		// Delete the access code, it can be used only once
+		couchdb.DeleteDoc(instance, accessCode)
```
aenario: you should at least log the error if any.

web/auth/auth.go
```diff
@@ -288,6 +289,101 @@ func authorize(c echo.Context) error {
...
+			"error": "the client must be registered",
+		})
+	}
+	if subtle.ConstantTimeCompare([]byte(clientSecret), []byte(client.ClientSecret)) == 0 {
+		return c.JSON(http.StatusBadRequest, echo.Map{
+			"error": "invalid client_secret",
+		})
+	}
+
+	var err error
```
aenario: err is not set in the  switch, you might replace it by `:=` at line 377

web/auth/auth.go
jinroh: @nono is this deletion critical or not ? maybe it should be retried if it is..

web/auth/auth.go
nono: What about line 353?

web/auth/auth.go
nono: It's not critical. It's just one trick between many to avoid attacks. But even if the deletion fails, it remains very difficult to operate this attack: an attacker still have to steal the authorization code (difficulty: medium) AND the client_secret (difficulty: very hard).

web/auth/auth.go
aenario: my bad

PR #152, opened 2016-12-15
Make replication compatible with pouch

aenario:


.gitignore
```diff
@@ -4,3 +4,5 @@ swagger-ui
 /.cozy
 *.log
 local.env
+integrations-tests/pouchdb/node_modules
+integrations-tests/pouchdb/replication-target
```
jinroh: integration should not be plural here

integrations-tests/pouchdb/tests.js
```diff
@@ -0,0 +1,112 @@
...
+
+require('longjohn')
+
+var DOCTYPE = 'io.cozy.pouchtestobject'
+var SOURCE = 'http://localhost:8080/data/' + DOCTYPE + '/'
+
+function stackRequest (method, path, body, expect, callback) {
+  var opts = {
+    'method': method,
+    'port': process.env.STACK_PORT || 8080,
```
jinroh: may use `COZY_STACK_PORT` to use the same as `cozy-app-dev.sh`

scripts/integration.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+go run main.go instances add dev
```
jinroh: maybe use a name other than dev to not conflict. that way we can also do:

```sh
echo "yes" | cozy-stack instances destroy integration-tests
cozy-stack instances add integration-tests
```

jinroh: oh... but that would mean you could not hit "localhost:8080"... meh, nevermind if its too complex

aenario:
@jinroh made the changes

I agree it will be nice to cleanup instances management during tests and use /etc/hosts & :80 -> :8080 iptable to have test env closer to prod. But we should probably make that in another PR !

PR #155, opened 2016-12-15
Fix the OAuth redirectionss for a user not yet logged-in

nono:
If the user is not yet logged in when he starts the OAuth2 authorization code flow, he will be redirectered to the login page with a `redirect` parameter that will be rejected. This PR fixes this and allow the user to complete the OAuth2 flow.


aenario:
TestShowLoginPageWithRedirectBadURL is failing in travis

PR #161, opened 2016-12-19
Add a route to change passphrase

nono:


pkg/instance/instance.go
```diff
@@ -376,6 +378,27 @@ func (i *Instance) RegisterPassphrase(pass []byte, tok []byte) error {
 
 	i.RegisterToken = nil
 	i.PassphraseHash = hash
+	i.SessionSecret = crypto.GenerateRandomBytes(sessionSecretLen)
+
+	return couchdb.UpdateDoc(couchdb.GlobalDB, i)
+}
+
+// UpdatePassphrase replace the instance registerToken by a passphrase
```
aenario: bad comment

pkg/instance/instance.go
```diff
@@ -376,6 +378,27 @@ func (i *Instance) RegisterPassphrase(pass []byte, tok []byte) error {
...
+	if err := crypto.CompareHashAndPassphrase(i.PassphraseHash, current); err != nil {
+		return ErrInvalidPassphrase
+	}
+
+	hash, err := crypto.GenerateFromPassphrase(pass)
+	if err != nil {
+		return err
+	}
+
+	i.PassphraseHash = hash
```
aenario: These two lines are repeated in registerPassphrase & updatePassphrase, they will also appears in recoverPassphrase when we implement it. It could also grow more complex in the future (generating encryption key, ect ...). I's recommend we put them in a (*Instance) SetPassphraseAndSecret

PR #162, opened 2016-12-19
Fix pouchdb integration tests

aenario:
I just saw the integration does not pass, but because the bash script was not properly passing its exit code, travis was green. 

This should both fix error reporting & the bug in travis.

### DIFF
--- a/scripts/integration.sh
+++ b/scripts/integration.sh
@@ -6,6 +6,9 @@ sleep 3
 cd integration-tests/pouchdb
 npm install
 npm run test
+testresult=$?
 
 pidstack=$(jobs -pr)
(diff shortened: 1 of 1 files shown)


scripts/integration.sh
```diff
@@ -1,11 +1,15 @@
...
-go run main.go instances add dev
 go run main.go serve &
-sleep 3
+sleep 5
+go run main.go instances add dev
+
 cd integration-tests/pouchdb
 npm install
 npm run test
+testresult=$?
```
nono: maybe `npm install && npm run test`, to avoid running the tests if the npm install fails

PR #166, opened 2016-12-19
[TBD] Remove statik/statik from repo, it is built by go generate

aenario:
**I am not sure if this the proper fix**

The problem : I could not just run scripts/build, as it will fail on "statik/statik.go already exists", so I have to delete it every time, but then the build is marked as dirty as the generated file is not exactly the same.


nono:
I've added it to the git repository to make it _go gettable_ (`go get` doesn't execute `go generate`)

PR #170, opened 2016-12-20
Allow to update a git application

jinroh:
This PR adds the possibility to upgrade an application. The version comparison between to manifest is done by a simple string comparison (no order).

The modifications of the application files is still clunky since we do not have yet a proper `Remove` method implemented. It only adds and/or override the files, without deleting removed ones.

Testing is done with a local git daemon.

### DIFF
--- a/pkg/apps/git.go
+++ b/pkg/apps/git.go
@@ -117,8 +117,36 @@ func (g *gitFetcher) pull(appdir, gitdir string, src *url.URL) error {
 		return err
 	}
 
-	// TODO: remove all files from the tree before copying new ones. or just try
-	// to remove/update only the modified files.
+	// TODO: permanently remove application files instead of moving them to the
+	// trash
+	err = vfs.Walk(c, appdir, func(name string, dir *vfs.DirDoc, file *vfs.FileDoc, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if name == appdir {
+			return nil
+		}
+		if name == gitdir {
+			return vfs.ErrSkipDir
+		}
+
+		if dir != nil {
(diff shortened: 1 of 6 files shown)


pkg/apps/git.go
```diff
@@ -16,103 +16,116 @@ import (
 	git "gopkg.in/src-d/go-git.v4"
 	gitObj "gopkg.in/src-d/go-git.v4/plumbing/object"
 	gitSt "gopkg.in/src-d/go-git.v4/storage/filesystem"
-	gitFS "gopkg.in/src-d/go-git.v4/utils/fs"
+	gitFS "srcd.works/go-billy.v1"
```
nono: Is it normal to have this import path?

pkg/apps/installer_test.go
```diff
@@ -13,6 +21,74 @@ import (
...
+	}
+
+	// "git daemon --reuseaddr --base-path=./ --export-all ./.git"
+	localGitCmd = exec.Command("git", "daemon", "--reuseaddr", "--base-path=./", "--export-all", "./.git")
+	localGitCmd.Dir = localGitDir
+	if out, err := localGitCmd.CombinedOutput(); err != nil {
+		fmt.Println(string(out))
+		panic(err)
+	}
+}
```
nono: :+1:

pkg/apps/git.go
jinroh: @nono I don't know why they did not use `gopkg.in`. I can use the [github address](https://github.com/src-d/go-billy) instead, if you prefer...

nono: OK, we can keep it

PR #171, opened 2016-12-20
Move global consts in their own package

jinroh:
This PR adds a `pkg/consts` package containing global identifiers that we use throughout the projet. Mainly the `io.cozy.*` doctypes. Moving them in a separate package has two main advantages:

  - prevent circular dependencies just to import a const string
  - prevent importing the whole package just to import small consts (for a client for instance)

### DIFF
--- a/pkg/couchdb/errors.go
+++ b/pkg/couchdb/errors.go
@@ -40,7 +40,7 @@ import (
 // 		{"error":"conflict","reason":"Document update conflict."}
 // 412 Precondition Failed : The request headers from the client and the
 // 		capabilities of the server do not match.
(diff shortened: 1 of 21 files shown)


pkg/couchdb/errors.go
```diff
@@ -40,7 +40,7 @@ import (
 // 		{"error":"conflict","reason":"Document update conflict."}
 // 412 Precondition Failed : The request headers from the client and the
 // 		capabilities of the server do not match.
-// 415 Bad Content Type : The content types supported, and the content type of
+// 415 Bad Content Type : The content consts.supported, and the content type of
```
nono: ?

PR #176, opened 2016-12-27
Inject a token when serving the index of a context inside an app

nono:


web/apps/apps.go
```diff
@@ -3,21 +3,82 @@
...
+	content, err := i.FS().Open(name)
+	if err != nil {
+		return err
+	}
+	defer content.Close()
+	buf, err := ioutil.ReadAll(content)
+	if err != nil {
+		return err
+	}
+	tmpl, err := template.New(file).Parse(string(buf))
```
jinroh: Having to load the entire file just to see if it is a template is unfortunate. How could we improve this situation ?

nono: IMHO, it's fine. It should be a template, and even if it is not a template, loading the file is not lost, we still use it to respond.

PR #178, opened 2016-12-28
Apps access control

nono:


web/server.go
```diff
@@ -20,14 +21,17 @@ func splitHost(host string) (instanceHost string, appSlug string) {
...
 		// TODO(optim): minimize the number of instance requests
 		if parent, slug := splitHost(c.Request().Host); slug != "" {
-			if _, err := instance.Get(parent); err == nil {
+			if i, err := instance.Get(parent); err == nil {
 				if serveApps != nil {
-					return serveApps(c, parent, slug)
+					c.Set("instance", i)
+					c.Set("slug", slug)
+					handler := middlewares.LoadSession(serveApps)
+					return handler(c)
```
jinroh: seems like the handler creation could be moved out of the closure

nono: Good idea. It's done in e6e39af

PR #184, opened 2016-12-29
Add support for flat subdomains (https://user-app.domain/)

nono:


pkg/sessions/code.go
```diff
@@ -0,0 +1,65 @@
...
+//   - just for one application.
+type Code struct {
+	Value     string
+	SessionID string
+	AppHost   string
+	ExpiresAt int64
+}
+
+// The list of pending codes
+var codes []*Code
```
jinroh: Wouldn't it be possible to use an authenticated message with expiration instead of a random value for each code ?

That would offload the burden to store all the codes on our side (which will become even more tedious at scale where they couldn't be stored in-memory) and make it constant time to check instead of linear to the number of codes we are asked to generate.

nono: Nope, it's really better from a security point of view to make sure that a code can't be reused, and this is only possible with a state on the server.

At scale, we will use redis for them (and it will be a constant time check).

PR #185, opened 2016-12-30
Subdomains docs

nono:


docs/client-app-dev.md
```diff
@@ -69,3 +71,62 @@ $ docker run --rm -it \
...
+```js
+const main = document.getElementById("container")
+cozy.init({
+  url: main.data("cozy-stack"),
+  token: main.data("ctx-token")
+})
+cozy.bar(main)
+
+// ...
+```
```
m4dz: For performance, with defered assets, it's better to rely on DOM loading events, so the JS file is probably more something like:

```js
'use strict'

import cozy from 'cozy-client-js'

document.addEventListener('DOMContentLoaded', () => {
  const app = document.querySelector('[role=application]')
  cozy.init({
    url: app.dataset.cozyStack,
    token: app.dataset.ctxToken
  })
  cozy.bar(app)
})
```

docs/client-app-dev.md
```diff
@@ -69,3 +71,62 @@ $ docker run --rm -it \
...
+    <script src="my-app.js"></script>
+  </body>
+</html>
+```
+
+And `my-app.js`:
+
+```js
+const main = document.getElementById("container")
+cozy.init({
```
m4dz: We should probably require `cozy` from the `cozy-client-js` package, nope?

docs/client-app-dev.md
```diff
@@ -69,3 +71,62 @@ $ docker run --rm -it \
...
+    <link rel="stylesheet" src="{{.Domain}}/settings/theme.css">
+    <link rel="stylesheet" src="my-app.css">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+  </head>
+  <body>
+    <main id="container" data-ctx-token="{{.CtxToken}}" data-cozy-stack="{{.Domain}}">
+    </main>
+    <script src="{{.Domain}}/assets/cozy-client.js"></script>
+    <script src="{{.Domain}}/assets/cozy-bar.js"></script>
+    <script src="{{.Domain}}/assets/cozy-ui.js"></script>
```
m4dz: My minds: `cozy-bar.js` should be injected by the stack as an external asset, but some resources like `cozy-client-js`, `cozy-ui.js` (if/when available), etc, should be packaged in the `app.js` build (because they will probably be version dependent)

docs/client-app-dev.md
```diff
@@ -69,3 +71,62 @@ $ docker run --rm -it \
...
+
+- `{{.CtxToken}}` will be replaced by the token for the application and its
+  context.
+- `{{.Domain}}` will be replaced by the stack hostname.
+
+So, the `index.html` should probably looks like:
+
+```html
+<!DOCTYPE html>
+<html lang="en">
```
m4dz: Actually in v3 front apps, we rely on the `lang` attribute on the `<html/>` tag to instanciate the localization engine with the correct locale specified by the user's settings. Should it be possible to have a variabled value like token, something like:

```html
<html lang="{{.Locale}}">
```

docs/client-app-dev.md
nono: I don't think import is supported by browsers

docs/client-app-dev.md
```diff
@@ -69,3 +71,62 @@ $ docker run --rm -it \
...
+  <body>
+    <main id="container" data-ctx-token="{{.CtxToken}}" data-cozy-stack="{{.Domain}}">
+    </main>
+    <script src="{{.Domain}}/assets/cozy-client.js"></script>
+    <script src="{{.Domain}}/assets/cozy-bar.js"></script>
+    <script src="{{.Domain}}/assets/cozy-ui.js"></script>
+    <script src="my-app.js"></script>
+  </body>
+</html>
+```
```
nono: Well, I don't think it's faster to omit `<head>`, `<body>`, and `</div>` (it is still 1 TCP packet). And I prefer to have something a bit more explicit. But your example will be perfect for cozy-template.

docs/client-app-dev.md
nono: For cozy-ui, there is no JS for the moment, and there are probably some cases where it make sense to have different versions.

For cozy-client-js, I think it's better to have one version that matches the stack.

PR #191, opened 2017-01-02
Jobs

jinroh:
This PR add a small abstraction layer to define the basis of our job system as well as an implementation of a memory based job queues targeting self-hosting audience.

### DIFF
--- a/pkg/instance/instance.go
+++ b/pkg/instance/instance.go
@@ -13,6 +13,7 @@ import (
 	"github.com/cozy/cozy-stack/pkg/couchdb"
 	"github.com/cozy/cozy-stack/pkg/couchdb/mango"
 	"github.com/cozy/cozy-stack/pkg/crypto"
+	"github.com/cozy/cozy-stack/pkg/jobs"
 	"github.com/cozy/cozy-stack/pkg/settings"
 	"github.com/cozy/cozy-stack/pkg/vfs"
 	"github.com/cozy/cozy-stack/web/jsonapi"
@@ -121,6 +122,11 @@ func (i *Instance) FS() afero.Fs {
 		}
 	}
 	return i.storage
+}
+
+// JobsBroker returns the jobs broker associated with the instance
+func (i *Instance) JobsBroker() jobs.Broker {
+	return jobs.NewMemBroker(i.Domain, jobs.GetWorkersList())
 }
 
 // Prefix returns the prefix to use in database naming for the
--- a/pkg/jobs/jobs.go
+++ b/pkg/jobs/jobs.go
@@ -1,0 +1,120 @@
+package jobs
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"math/rand"
+	"time"
+)
+
+const (
+	// Queued state
+	Queued State = "queued"
+	// Running state
+	Running = "running"
+	// Errored state
+	Errored = "errored"
+)
+
+const (
+	// JSONEncoding is a JSON encoding message type
+	JSONEncoding = "json"
+)
+
+var idRand *rand.Rand
+
+func init() {
+	idRand = rand.New(rand.NewSource(time.Now().UTC().UnixNano()))
+}
+
+type (
+	// Queue interface is used to represent a asynchronous queue of jobs from
+	// which it is possible to enqueue and consume jobs.
+	Queue interface {
+		Enqueue(*Job) error
+		Consume() (*Job, error)
+		Len() int
+		Close()
+	}
+
+	// Broker interface is used to represent a job broker associated to a
+	// particular domain. A broker can be used to create jobs that are pushed in
+	// the job system.
+	Broker interface {
+		Domain() string
+		PushJob(*JobRequest) (*Job, error)
+	}
+
+	// State represent the state of a job.
+	State string
+
+	// Message is a byte slice representing an encoded job message type.
+	Message struct {
+		Data []byte
+		Type string
+	}
+
+	// Job struct encapsulates all the parameters of a job.
+	Job struct {
+		ID         string    `json:"id"`
+		WorkerType string    `json:"worker_type"`
+		Message    *Message  `json:"-"`
+		State      State     `json:"state"`
+		QueuedAt   time.Time `json:"queued_at"`
+	}
+
+	// JobRequest struct is used to represent a new job request.
+	JobRequest struct {
+		WorkerType string
+		Message    *Message
+	}
+)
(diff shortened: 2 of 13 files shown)


pkg/jobs/jobs.go
```diff
@@ -0,0 +1,128 @@
...
+
+func makeID() string {
+	const (
+		idLen         = 16
+		letters       = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
+		letterIdxMask = 1<<6 - 1 // 6 bits to represent a letter index
+	)
+	b := make([]byte, idLen)
+	for i := 0; i < idLen; {
+		if idx := int(rand.Int63() & letterIdxMask); idx < len(letters) {
```
nono: And shouldn't we run `rand.Seed` to avoid having the same ids when the cozy-stack is restarted?

pkg/jobs/jobs.go
```diff
@@ -0,0 +1,128 @@
...
+	// GobEncoding is a Gob encoding message type
+	GobEncoding = "gob"
+)
+
+type (
+	// Queue interface is used to represent a asynchronous queue of jobs from
+	// which it is possible to enqueue and consume jobs.
+	Queue interface {
+		Enqueue(*Job) error
+		Consume() (bool, *Job, error)
```
nono: What does the `bool` in the return mean?

pkg/instance/instance.go
```diff
@@ -123,6 +124,11 @@ func (i *Instance) FS() afero.Fs {
 	return i.storage
 }
 
+// JobsBroker returns the jobs broker associated with the instance
+func (i *Instance) JobsBroker() (jobs.Broker, error) {
```
jinroh: I don't know.. For the case of the in-memory queue I would tend to agree but for the case of an external queue this getter may initialize a connection with an external system.

Although I agree these are problems we may not want to resolve yet...

pkg/jobs/jobs.go
jinroh: To alert the consumer that the queue is closed. Might not be really idiomatic though.

pkg/jobs/jobs.go
```diff
@@ -0,0 +1,128 @@
...
+		Foreach(func(*Job) error) error
```
jinroh: I'll remove it. It is indeed of no use for now.

In the jobs doc I documented the possibility for a user to fetch all the element of the queue but I'm not sure we really want to allow that.

pkg/jobs/jobs.go
```diff
@@ -0,0 +1,128 @@
...
+		Len() int
+		Close()
+	}
+
+	// Broker interface is used to represent a job broker associated to a
+	// particular domain from which it is possible to create jobs that a pushed
+	// in the job system.
```
jinroh: No. However for the case of the memory queue it is the case, but it is particular.

I could add an explicit interface representing the responsibility to start the workers... but I'm not sure how to do it for now.

pkg/jobs/jobs.go
```diff
@@ -0,0 +1,128 @@
...
+
+	// JobRequest struct is used to represent a new job request.
+	JobRequest struct {
+		WorkerType string
+		Message    *Message
+	}
+)
+
+// NewMessage returns a new Message encoded in the specified format.
+func NewMessage(enc string, data interface{}) (*Message, error) {
```
jinroh: It is not used for now. Should be used by triggers.

pkg/jobs/jobs.go
```diff
@@ -0,0 +1,128 @@
...
+	var b []byte
+	var err error
+	switch enc {
+	case JSONEncoding:
+		b, err = json.Marshal(data)
+	case GobEncoding:
```
jinroh: Gob is a bit more compact and could be useful for long payload. If you think it is of no use, I'll remove it.

pkg/jobs/jobs.go
jinroh: I did not know the seed was always the same on the global random source. I'll fix that.

pkg/instance/instance.go
nono: Yes, we can add that later

pkg/jobs/jobs.go
nono: At least, you should add a comment to explain that

pkg/jobs/jobs.go
nono: The only use case I think about this is debugging. As a developer of a client-side applications, I may want to see what are the jobs in a queue. But, I'm not sure about it.

pkg/jobs/jobs.go
nono: Without adding an explicit interface, a comment can be enough

pkg/jobs/jobs.go
nono: Not sure, but I think it is the case

pkg/jobs/jobs.go
nono: I think it is a bit prematured to add this

pkg/jobs/jobs.go
```diff
@@ -113,16 +113,15 @@ func makeQueueName(domain, workerType string) string {
 
 func makeID() string {
 	const (
-		idLen         = 16
-		letters       = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
-		letterIdxMask = 1<<6 - 1 // 6 bits to represent a letter index
+		idLen   = 16
+		letters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
 	)
+	r := rand.New(rand.NewSource(time.Now().UTC().UnixNano()))
```
nono: I'd have created a new source for the whole jobs package, not one per id. But it should work too

jinroh:
Squashed into one commit

pkg/jobs/jobs.go
```diff
@@ -0,0 +1,120 @@
...
+	default:
+		return ErrUnknownMessageType
+	}
+}
+
+func makeQueueName(domain, workerType string) string {
+	return fmt.Sprintf("%s/%s", domain, workerType)
+}
+
+func makeID() string {
```
aenario: may be we could move this function into a separate package (crypto.NonCryptoRandomID ?)

jinroh: some kind of `utils` package for instance ?

PR #199, opened 2017-01-06
Fix race while using an rng source in RandomString

jinroh:


pkg/utils/utils.go
```diff
@@ -9,20 +9,18 @@ import (
 	"time"
 )
 
-var urand *rand.Rand
-
-func init() {
-	urand = rand.New(rand.NewSource(time.Now().UTC().UnixNano()))
```
nono: What about using the global random source and just calling `rand.Seed(time.Now().UTC().UnixNano())` in `init`?

pkg/utils/utils_test.go
```diff
@@ -1,15 +1,41 @@
...
+	n := 10000
+	var wg sync.WaitGroup
+	wg.Add(n)
+	for i := 0; i < n; i++ {
+		go func() {
+			RandomString(10)
+			wg.Done()
+		}()
+	}
+	wg.Wait()
```
nono: I think we should test that all the 10000 generated strings are uniques

PR #201, opened 2017-01-09
Permissions and sharing by link

nono:


docs/permissions.md
```diff
@@ -0,0 +1,419 @@
...
+}
+```
+
+And the other is for the events inside the calendar:
+
+```json
+{
+  "type": "io.cozy.events",
+  "verbs": "GET",
+  "selector": "calendar-id",
```
aenario: How about just using mango selector (or a subset of the syntax without operator)  and merging the _id case ?

ie. 
```
{
   "type": "io.cozy.calendars",
  "verbs": "GET",
  "selector": {"_id": "1355812c-d41e-11e6-8467-53be4648e3ad"}
 }
{
   "type": "io.cozy.events",
  "verbs": "GET",
  "selector": {"calendar_id": "1355812c-d41e-11e6-8467-53be4648e3ad"}
 }
```

docs/permissions.md
```diff
@@ -0,0 +1,419 @@
...
+        }
+      }
+    }
+  }
+}
+```
+
+### GET /permissions/:id
+
+Return the informations about a set of permissions
```
aenario: (details) the set should be a subset created by the current token?

docs/permissions.md
```diff
@@ -0,0 +1,419 @@
...
+Content-Type: application/vnd.api+json
+```
+
+```json
+{
+  "data": {
+    "type": "io.cozy.permissions",
+    "id": "5a9c1844-d427-11e6-ab36-2b684d437b0d",
+    "attributes": {
+      "application-id": "4cfbd8be-8968-11e6-9708-ef55b7c20863",
```
aenario: Using application-id and not application-slug means all permissions would be erased upon uninstallation / reinstallation

nono: Yes, I think it is safer. The slug can be reused by another application. You are OK with that?

docs/permissions.md
nono: I'm not sure but I think so. Do you have an opinion on that?

docs/permissions.md
nono: Well, first, I wanted to use mango selectors. But they are too complicated (lots of things to code in go, hard to serialize to OAuth2 scopes, etc.). So, I tried to limit the scopes and find a format that works for both JSON documents stored in couchdb and the JSON-API stuff (jobs, files, etc.). Using this selector breaks the symmetry between them.

docs/permissions.md
aenario: yes, it should

PR #202, opened 2017-01-09
Jobs eventstream and queue size

jinroh:
Includes commits from #200 

This PR adds the possibility to watch jobs state from `POST /jobs/queue/:worker-type` with and `Accept: text/event-stream` header.

It also adds the endpoint `GET /jobs/queue/:worker-type` returning the number of elements currently in the queue.

### DIFF
--- a/pkg/jobs/worker.go
+++ b/pkg/jobs/worker.go
@@ -15,10 +15,6 @@ var (
 	defaultMaxExecTime  = 60 * time.Second
 	defaultRetryDelay   = 60 * time.Millisecond
 	defaultTimeout      = 10 * time.Second
-
-	maxMaxExecCount = 5
-	maxMaxExecTime  = 5 * time.Minute
-	maxTimeout      = 1 * time.Minute
 )
 
 type (
--- a/pkg/jobs/worker.go
+++ b/pkg/jobs/worker.go
@@ -86,37 +86,32 @@ func (w *Worker) work(workerID string) {
 
 func (w *Worker) defaultedConf(opts *JobOptions) *WorkerConfig {
 	c := &(*w.Conf)
-	if opts != nil {
-		if opts.MaxExecCount != 0 {
-			c.MaxExecCount = opts.MaxExecCount
-		}
-		if opts.MaxExecTime > 0 {
-			c.MaxExecTime = opts.MaxExecTime
-		}
-		if opts.Timeout > 0 {
-			c.Timeout = opts.Timeout
-		}
-	}
 	if c.Concurrency == 0 {
 		c.Concurrency = uint(defaultConcurrency)
 	}
 	if c.MaxExecCount == 0 {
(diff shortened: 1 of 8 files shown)


web/jobs/jobs.go
```diff
@@ -58,7 +61,39 @@ func pushJob(c echo.Context) error {
...
 
-	return jsonapi.Data(c, http.StatusAccepted, &apiJob{job}, nil)
+	accept := c.Request().Header.Get("Accept")
+	if accept != TypeTextEventStream {
+		return jsonapi.Data(c, http.StatusAccepted, &apiJob{job}, nil)
+	}
+
+	w := c.Response().Writer
+	w.Header().Set("Content-Type", TypeTextEventStream)
+	w.WriteHeader(200)
```
nono: Are you doing Server Sent Events on a POST request? I didn't know that it was possible to use that from a browser!

pkg/jobs/worker.go
```diff
@@ -15,6 +15,10 @@ var (
 	defaultMaxExecTime  = 60 * time.Second
 	defaultRetryDelay   = 60 * time.Millisecond
 	defaultTimeout      = 10 * time.Second
+
+	maxMaxExecCount = 5
+	maxMaxExecTime  = 5 * time.Minute
+	maxTimeout      = 1 * time.Minute
```
nono: What is `MaxExecTime`? It's not the same thing as `Timeout`?

nono: I think that the max limits are a bit too low. Some jobs may have a good reason to have an higher timeout (a big import from a konnector that is ran once a month).

jinroh: The MaxExecTime is the maximum total amount of time the worker can take working on the same job (retries and retries delay counted).

I can increase the defaults. One other option would be to have these "global" defaults kept quite low, and the ability on the worker config side (hard-coded) to have greater defaults so that the worker configuration takes precedence over the "global" if informed.

web/jobs/jobs.go
```diff
@@ -58,12 +88,45 @@ func pushJob(c echo.Context) error {
...
+	accept := c.Request().Header.Get("Accept")
+	if accept != typeTextEventStream {
+		return jsonapi.Data(c, http.StatusAccepted, &apiJob{job}, nil)
+	}
+
+	w := c.Response().Writer
+	w.Header().Set("Content-Type", typeTextEventStream)
+	w.WriteHeader(200)
+	if err := streamJob(job, w); err != nil {
+		return nil
```
aenario: [not sure] I am not sure how go channels work but wont an error on job serialization leave the connection indefinitively open and block the channnel?

jinroh: The sender routine is asynchronous so that it drop the message is there is no listener: https://github.com/cozy/cozy-stack/blob/master/pkg/jobs/mem_jobs.go#L203-L212

PR #207, opened 2017-01-10
Make it easier to set/update passphrase from a client-side app

nono:


jinroh:
@nono you should update the `cozy-app-dev.sh` script that performs the passphrase registration on the old route using query params.

PR #220, opened 2017-01-13
Add a client package with oauth support

jinroh:
This PR includes a `client` package with `files` and `instances` support (still on `data`). It is used by the cmd package.

This package contains two subpackages:
  - `auth`: to handle the OAuth authizations
  - `request`: a useful http request wrapper

It is a self contained package (no dependencies on `pkg/vfs` or `pkg/instances`).

There are no tests for now and the OAuth is still missing "token refreshing" but its a good start if you want to give some inputs.


nono:
I think:

- we should have tests for the client package
- maybe, we can put the client package outside of pkg, to make it clear that it doesn't depend on any other package in pkg

cmd/root.go
```diff
@@ -38,6 +44,45 @@ profiles you.`,
 
 var cfgFile string
 
+func createClient(domain string) *client.Client {
+	return &client.Client{
+		Domain:        domain,
+		DisableSecure: config.IsDevRelease(),
+		AuthClient: &auth.Client{
+			RedirectURIs: []string{"http://foobar.com"},
```
nono: It looks like a weird default

pkg/client/auth/auth.go
```diff
@@ -0,0 +1,269 @@
...
+
+// AuthHeader implements the Tokener interface for the request
+func (r *Request) AuthHeader() string {
+	return r.token.AuthHeader()
+}
+
+// defaultClient defaults some values of the given client
+func defaultClient(c *Client) *Client {
+	c.SoftwareID = "github.com/cozy/cozy-stack"
+	c.SoftwareVersion = "0.0"
```
nono: I'm not convinced that `0.0` is a sensible default

pkg/client/client.go
```diff
@@ -0,0 +1,187 @@
...
+	if c.Retries == 0 {
+		c.Retries = 3
+	}
+	if c.Timeout == 0 {
+		c.Timeout = 30 * time.Second
+	}
+	if c.Transport == nil {
+		c.Transport = &http.Transport{
+			Proxy:               http.ProxyFromEnvironment,
+			MaxIdleConnsPerHost: 1024,
```
nono: Why this setting?

pkg/client/instances.go
```diff
@@ -0,0 +1,78 @@
...
+	Rev   string `json:"rev"`
+	Attrs struct {
+		Domain         string `json:"domain"`
+		Locale         string `json:"locale"`
+		StorageURL     string `json:"storage"`
+		Dev            bool   `json:"dev"`
+		PassphraseHash []byte `json:"passphraseHash,omitempty"`
+		RegisterToken  []byte `json:"registerToken,omitempty"`
+		SessionSecret  []byte `json:"sessionSecret,omitempty"`
+		OAuthSecret    []byte `json:"oauthSecret,omitempty"`
```
nono: Why are SessionSecret and OAuthSecret here?

aenario:
+1 on lack of test and moving the client to a more separate hierarchy.

PR #223, opened 2017-01-16
Secure headers

jinroh:
Add secure headers HSTS, CSP and XFrameOption support for applications.


web/middlewares/secure.go
```diff
@@ -0,0 +1,112 @@
...
+	XFrameSameOrigin              = "SAMEORIGIN"
+	XFrameAllowFrom               = "ALLOW-FROM"
+
+	CSPSrcSelf CSPSource = iota
+	CSPSrcParent
+	CSPSrcParentSubdomains
+	CSPSrcAny
+)
+
+func Secure(conf *SecureConfig) echo.MiddlewareFunc {
```
nono: I don't see where this middleware is used for the main domain (`https://alice.example.com`)

web/middlewares/secure.go
```diff
@@ -0,0 +1,112 @@
...
+			if in := c.Get("instance"); in != nil && in.(*instance.Instance).Dev {
+				hsts = false
+			}
+			h := c.Response().Header()
+			if hsts && hstsHeader != "" {
+				h.Set(echo.HeaderStrictTransportSecurity, hstsHeader)
+			}
+			if xFrameHeader != "" {
+				h.Set(echo.HeaderXFrameOptions, xFrameHeader)
+			}
```
nono: We don't set the `X-Content-Type-Options` header?

web/middlewares/secure.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func makeCSPHeader(host, header string, sources []CSPSource) string {
+	headers := make([]string, len(sources))
+	for i, src := range sources {
+		switch src {
+		case CSPSrcSelf:
+			headers[i] = "'self'"
+		case CSPSrcParent:
+			headers[i] = parentHost(host)
```
nono: If `CSPSrcParent` means the main domain, `parentHost` will only work for self-hosted users. For the big instances, it's more complicated.

See https://github.com/cozy/cozy-stack/commit/b770af495059b1c02a3565c929c502449cd84280, https://github.com/cozy/cozy-stack/commit/79efe2c865abd71641e49dedf51eed143a20bad9 and https://github.com/cozy/cozy-stack/blob/master/web/server.go#L15-L28

web/server.go
```diff
@@ -30,24 +32,31 @@ func splitHost(host string) (instanceHost string, appSlug string) {
...
 // host of the request.
 func Create(router *echo.Echo, serveApps echo.HandlerFunc) (*echo.Echo, error) {
-	appsHandler := middlewares.LoadSession(serveApps)
+	appsHandler := middlewares.Compose(serveApps,
+		middlewares.Secure(&middlewares.SecureConfig{
+			HSTSMaxAge:    365 * 24 * time.Hour, // 1 year
+			CSPScriptSrc:  []middlewares.CSPSource{middlewares.CSPSrcSelf, middlewares.CSPSrcParent},
+			CSPConnectSrc: []middlewares.CSPSource{middlewares.CSPSrcSelf, middlewares.CSPSrcParent},
+			CSPFrameSrc:   []middlewares.CSPSource{middlewares.CSPSrcParent},
+			XFrameOptions: middlewares.XFrameDeny,
```
nono: Just to be sure, XFrameDeny is temporary. When we will have intents, we will allow the apps/routes with an intent to be embedded in an iframe for inter-apps. Right?

web/middlewares/secure.go
```diff
@@ -0,0 +1,112 @@
...
+	"strings"
+	"time"
+
+	"github.com/cozy/cozy-stack/pkg/instance"
+	"github.com/labstack/echo"
+)
+
+type (
+	XFrameOption string
+	CSPSource    int
```
nono: Gometalinter says that we should add comments

web/middlewares/secure.go
```diff
@@ -0,0 +1,112 @@
...
+		xFrameHeader = string(XFrameSameOrigin)
+	case XFrameAllowFrom:
+		xFrameHeader = fmt.Sprintf("%s %s", XFrameAllowFrom, conf.XFrameAllowed)
+	}
+
+	return func(next echo.HandlerFunc) echo.HandlerFunc {
+		return func(c echo.Context) error {
+			hsts := true
+			if in := c.Get("instance"); in != nil && in.(*instance.Instance).Dev {
+				hsts = false
```
nono: I'm not sure we should not also disable CSP for dev instances. I wonder if it will interfer with debugging tools. Maybe we can start like that and see if it causes trouble.

web/middlewares/secure.go
```diff
@@ -0,0 +1,112 @@
...
+			if hsts && hstsHeader != "" {
+				h.Set(echo.HeaderStrictTransportSecurity, hstsHeader)
+			}
+			if xFrameHeader != "" {
+				h.Set(echo.HeaderXFrameOptions, xFrameHeader)
+			}
+			var cspHeader string
+			host := c.Request().Host
+			if len(conf.CSPScriptSrc) > 0 {
+				cspHeader += makeCSPHeader(host, "script-src", conf.CSPScriptSrc)
```
nono: @CPatchane should we allow inline scripts?

web/middlewares/secure.go
```diff
@@ -0,0 +1,112 @@
...
+			}
+			if len(conf.CSPFrameSrc) > 0 {
+				cspHeader += makeCSPHeader(host, "frame-src", conf.CSPFrameSrc)
+			}
+			if len(conf.CSPConnectSrc) > 0 {
+				cspHeader += makeCSPHeader(host, "connect-src", conf.CSPConnectSrc)
+			}
```
nono: I think we should also consider `font-src`, `img-src`, `manifest-src`, `media-src`, `object-src`, `style-src`, and `worker-src`.

web/middlewares/secure.go
CPatchane: I think we should allow inline-script because it seems to be very restrictive without that when I tested it. Maybe for us we can figured out (it would be great to not need them at all) but that could complicate the app development by the community. What's your opinion about that? We could also try for now without and see if we have troubles.

web/middlewares/secure.go
CPatchane: I agree with @nono. Do you use `default-src`? IMHO, it could be better to set all default values and then do specific changes to have a kind of whitelist.

web/middlewares/secure.go
CPatchane: I agree with @nono, maybe we can use CSP in the development in order to have a dev environment as close as possible to the production environment.

web/middlewares/secure.go
CPatchane: Is it useful here? It seems to be not supported by all safari versions.

web/server.go
jinroh: Yes, I chose to start with the most drastic option that we will losen as we go and need to for explicit cases.

web/middlewares/secure.go
jinroh: Yes, this is basically the option I chose here: start with the most strict options and losen if necessary with clear context where it should be losen.

web/middlewares/secure.go
jinroh: Why would you need 'inline-script' ? I do not see how it is so restrictive ? It is one of the most useful rule to forbid to prevent XSS.

web/middlewares/secure.go
jinroh: I'll add all of them. @CPatchane good idea of the default-src, I'll use it.

web/middlewares/secure.go
jinroh: @nono It is not for now.

web/middlewares/secure.go
CPatchane: Agree. I think we could let it disabled and see if we need them or not.

nono:
@jinroh I'd have put the spec packages inside the package that it tests (like `web/auth/auth_spec`)

web/middlewares/secure_test.go
```diff
@@ -0,0 +1,98 @@
...
+	"time"
+
+	"github.com/labstack/echo"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSecureMiddlewareHSTS(t *testing.T) {
+	e := echo.New()
+	req, _ := http.NewRequest(echo.GET, "http://app.cozy.local/", nil)
+	req.TLS = &tls.ConnectionState{} // fors IsTls to return true
```
nono: s/fors/for/

web/middlewares/secure.go
```diff
@@ -0,0 +1,152 @@
...
+	case XFrameSameOrigin:
+		xFrameHeader = string(XFrameSameOrigin)
+	case XFrameAllowFrom:
+		xFrameHeader = fmt.Sprintf("%s %s", XFrameAllowFrom, conf.XFrameAllowed)
+	}
+
+	return func(next echo.HandlerFunc) echo.HandlerFunc {
+		return func(c echo.Context) error {
+			hsts := true
+			if !c.IsTLS() {
```
nono: I'm not sure about this condition. When we have a reverse-proxy in front of cozy-stack, there will an https connexion between the client/browser and the reverse-proxy, but the connexion between the reverse-proxy and the stack can be http or https (and if the reverse-proxy is on the same server that the stack, it will be probably http, not https). The reverse-proxy can set the `X-Forwarded-Protocol` header, but the question is: does `c.IsTLS()` checks this header? From what I've read, I think it is not the case.

So, I think we should remove this condition for the moment. And add it again later if we have some issues, and testing how it behaves with a reverse-proxy.

jinroh:
@aenario i'll fix that and merge it myself. Thanks :)

PR #227, opened 2017-01-18
Permissions first step

aenario:
Extract permissions from jwt Bearer and add a route to check permissions


pkg/permissions/permissions_test.go
```diff
@@ -0,0 +1,158 @@
...
+	assert.NoError(t, err)
+	assert.Len(t, s, 2)
+	assert.Equal(t, "io.cozy.contacts", s[0].Type)
+	assert.Equal(t, "io.cozy.files", s[1].Type)
+	assert.Len(t, s[1].Verbs, 1)
+	assert.Equal(t, GET, s[1].Verbs[0])
+	assert.Len(t, s[1].Values, 1)
+	assert.Equal(t, "io.cozy.files.music-dir", s[1].Values[0])
+
+	rule, err := UnmarshalRuleString("io.cozy.events:GET:calendar-id:mygreatcalendar")
```
nono: In my mind, the selector was at the end: `type:verbs:values:selector`

pkg/permissions/rule.go
```diff
@@ -0,0 +1,83 @@
...
+	if len(r.Verbs) != 0 {
+		out += partSep + r.Verbs.String()
+	}
+	if len(r.Selector) != 0 {
+		out += partSep + r.Selector
+	}
+	if len(r.Values) != 0 {
+		out += partSep + strings.Join(r.Values, valueSep)
+	}
+	return out, nil
```
nono: If a rule comes from JSON, it can have a selector and values but not verbs. See https://github.com/aenario/cozy-stack/blob/b7bbb6f1fdbc00c1090e245f7dc764d7f9aade28/pkg/permissions/permissions_test.go#L68-L73 for an example. In that case, we should insert the `ALL` for verbs

pkg/permissions/rule.go
```diff
@@ -0,0 +1,83 @@
...
+	case 4:
+		out.Type = parts[0]
+		out.Verbs = VerbSplit(parts[1])
+		out.Selector = parts[2]
+		out.Values = strings.Split(parts[3], valueSep)
+
+	default:
+		return out, fmt.Errorf("Too many parts in %s", in)
+
+	}
```
nono: If we take the `type:verbs:values:selector` format, it can be rewritten nicely to:

```go
switch len(parts) {
case 4:
    out.Selector = parts[3]
    fallthrough
case 3:
    out.Values = strings.Split(parts[2], valueSep)
    fallthrough
case 2:
    out.Verbs = VerbSplit(parts[1])
    fallthrough
case 1:
    if parts[0] == "" {
        return out, errors.New("the type is mandatory for a permissions rule")
    }
    out.Type = parts[0]
default:
    return out, fmt.Errorf("Too many parts in %s", in)
}
```

web/permissions/permissions.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/labstack/echo"
+	jwt "gopkg.in/dgrijalva/jwt-go.v3"
+)
+
+var readable = true
+var none = false
+
+var blackList = map[string]bool{
+	consts.Sessions:  none,
+	consts.Files:     readable,
```
nono: I'm not sure what you mean by blacklisting here. A permission like `io.cozy.files:ALL:io.cozy.files.music-dir` must be acceptable. It allows to use all the methods/verbs on `/files` for the music directory. On `/data`, it is still read-only, but it is not the permissions that enforce that.

web/permissions/permissions.go
```diff
@@ -0,0 +1,158 @@
...
+	consts.Instances: readable,
```
nono: I don't think that instances should be readable. They contain some secrets.

web/permissions/permissions.go
aenario: Yes, I was mixing concerns (blackList of operations from `/data/` endpoint vs permissions for everything). Removed and will be properly reintroduced in next PR.

web/permissions/permissions.go
aenario: Noted for next PR.

PR #229, opened 2017-01-18
Add sendmail worker

jinroh:
This PR adds a `sendmail` and a `sendmail-template` workers to send mails.

The sendmail-template accepts a template html string along with values. It may be not really useful as is, but could become interesting with a preset of predefined HTML templates.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -74,7 +74,7 @@ func init() {
 	flags.Int("couchdb-port", 5984, "couchdbdb port")
 	checkNoErr(viper.BindPFlag("couchdb.port", flags.Lookup("couchdb-port")))
 
-	flags.String("mail-host", "", "mail smtp host")
+	flags.String("mail-host", "localhost", "mail smtp host")
 	checkNoErr(viper.BindPFlag("mail.host", flags.Lookup("mail-host")))
 
 	flags.Int("mail-port", 465, "mail smtp port")
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -74,6 +74,21 @@ func init() {
 	flags.Int("couchdb-port", 5984, "couchdbdb port")
 	checkNoErr(viper.BindPFlag("couchdb.port", flags.Lookup("couchdb-port")))
 
+	flags.String("mail-host", "", "mail smtp host")
+	checkNoErr(viper.BindPFlag("mail.host", flags.Lookup("mail-host")))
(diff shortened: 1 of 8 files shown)


jinroh:
Depends on https://github.com/cozy/gomail/pull/1

cmd/root.go
```diff
@@ -74,6 +74,21 @@ func init() {
 	flags.Int("couchdb-port", 5984, "couchdbdb port")
 	checkNoErr(viper.BindPFlag("couchdb.port", flags.Lookup("couchdb-port")))
 
+	flags.String("mail-host", "", "mail smtp host")
```
nono: Why not `localhost` for the default? Is it because the TLS certificate won't match in that case?

pkg/jobs/workers/mail.go
```diff
@@ -0,0 +1,126 @@
...
+	Name string `json:"name"`
+	Mail string `json:"mail"`
+}
+
+type mailOptions struct {
+	From    *MailAddress          `json:"from"`
+	To      []*MailAddress        `json:"to"`
+	Subject string                `json:"subject"`
+	Dialer  *gomail.DialerOptions `json:"dialer,omitempty"`
+	Date    *time.Time            `json:"date"`
```
nono: What is the purpose of the `Date` field?

pkg/jobs/workers/mail.go
```diff
@@ -0,0 +1,126 @@
...
+	return sendMail(ctx, opts.mailOptions, "text/html", b.String())
+}
+
+func sendMail(ctx context.Context, opts *mailOptions, bodyType, body string) error {
+	if bodyType == "" {
+		bodyType = "text/plain"
+	}
+	if bodyType != "text/plain" && bodyType != "text/html" {
+		return fmt.Errorf("Unknown body content-type %s", bodyType)
+	}
```
nono: If I understand right, there is no multipart mails, with both plain text and html

cmd/root.go
jinroh: I could add that, just did not think of it.

pkg/jobs/workers/mail.go
jinroh: For tests purposes mostly...

aenario:
(restarted build now that  cozy/gomail#1 is merged)

PR #232, opened 2017-01-19
Handle workers that may panic

jinroh:
Recover in case a worker function does panic.

### DIFF
--- a/pkg/jobs/mem_jobs_test.go
+++ b/pkg/jobs/mem_jobs_test.go
@@ -184,7 +184,7 @@ func TestRetry(t *testing.T) {
 	w.Wait()
 }
 
-func TestPanic(t *testing.T) {
+func TestPanicRetried(t *testing.T) {
 	var w sync.WaitGroup
 
 	maxExecCount := 4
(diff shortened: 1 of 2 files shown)


pkg/jobs/mem_jobs_test.go
```diff
@@ -184,6 +184,33 @@ func TestRetry(t *testing.T) {
...
+	})
+
+	w.Add(maxExecCount)
+	_, _, err := broker.PushJob(&JobRequest{
+		WorkerType: "panic",
+		Message:    nil,
+	})
+
+	assert.NoError(t, err)
+	w.Wait()
```
nono: What I'd have done is a worker that panics when given an odd number and works fine when given an even number, then give it a job with an odd number, and then another job with an even number, and check that the even number has been correctly processed (ie the panic on the odd number has not interrupted the worker). This doesn't mean that your test is wrong, just that I don't understand it.

jinroh: This test checks that we the job that has paniced is retried, which is what a really wanted to test.

I could indeed add a test checking that the worker is still available even after a panic.

PR #233, opened 2017-01-19
Add GET _all_docs

nono:
The `POST _all_docs` works, but it's better to also have `GET _all_docs` for when we will have permissions on /data.

### DIFF
--- a/web/data/data_test.go
+++ b/web/data/data_test.go
@@ -668,3 +668,25 @@ func TestUnsupportedOption(t *testing.T) {
 	assert.Equal(t, "400 Bad Request", res.Status, "should get a 400")
 	assert.NoError(t, err)
 }
+
+func TestGetAllDocs(t *testing.T) {
+	url := ts.URL + "/data/" + Type + "/_all_docs?include_docs=true"
+	req, _ := http.NewRequest("GET", url, nil)
+	req.Header.Add("Host", Host)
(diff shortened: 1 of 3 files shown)


web/data/data_test.go
```diff
@@ -668,3 +668,25 @@ func TestUnsupportedOption(t *testing.T) {
...
+
+func TestGetAllDocs(t *testing.T) {
+	url := ts.URL + "/data/" + Type + "/_all_docs?include_docs=true"
+	req, _ := http.NewRequest("GET", url, nil)
+	req.Header.Add("Host", Host)
+	out, res, err := doRequest(req, nil)
+	assert.Equal(t, "200 OK", res.Status, "should get a 200")
+	assert.NoError(t, err)
+	totalRows := out["total_rows"].(float64)
+	assert.Equal(t, float64(3), totalRows)
```
sebn: Why do you cast to `float64`? I first thought it could be some requirement of the assert package, but looking at the doc, it doesn't seem so... Just out of curiosity :)

nono: `assert.Equal` wants the same type for both values. And `totalRows` comes a JSON unserialized to a `map[string]interface{}`, for which go chooses a float64 for number.

sebn: Oh yeah, forgot that js/json didn't have an int type, only Number. Thanks!

PR #234, opened 2017-01-19
Downloading a zip archive

nono:


web/files/files.go
```diff
@@ -315,6 +315,24 @@ func ReadFileContentFromPathHandler(c echo.Context) error {
...
 
+// ArchiveHandler handles requests to /files/archive and creates on the fly
+// zip archives with the given files and folders.
+func ArchiveHandler(c echo.Context) error {
+	archive := &vfs.Archive{}
+	if _, err := jsonapi.Bind(c.Request(), archive); err != nil {
+		return err
+	}
+	if len(archive.Files) == 0 {
+		return c.JSON(http.StatusNotFound, "Can't create an archive with no files")
```
jinroh: probably more a `BadRequest` than `NotFound`

pkg/vfs/archive.go
```diff
@@ -0,0 +1,62 @@
...
+type Archive struct {
+	Name  string   `json:"name"`
+	Files []string `json:"files"`
+}
+
+// Serve creates on the fly the zip archive and streams in a http response
+func (a *Archive) Serve(c Context, w http.ResponseWriter) error {
+	header := w.Header()
+	header.Set("Content-Type", ZipMime)
+	header.Set("Content-Disposition", "attachment; filename="+a.Name+".zip")
```
jinroh: I'd expect golang `net/http` to escape `\n`s and `\r`s from headers but this is worth a check (and commentary) and it makes me wonder if we shouldn't sanitize this input.

jinroh: At least we should escape `"` i guess

jinroh: Apparently [it does](https://github.com/golang/go/blob/master/src/net/http/header.go#L155) so it seems fine. Although, we should probably disallow [newlines in filenames](https://github.com/cozy/cozy-stack/blob/master/pkg/vfs/vfs.go#L47)...

PR #235, opened 2017-01-20
Permissions settings

aenario:
I recommend reviewing commit by commit.

- Introduces the Validable interface and some utility function for validating documents
- settings/settings.go was getting too big, splitted it into 1 file/route in the same package. 
- Add permissions validation for settings/instance & settings/disk-usage. 
- No permissions for passphrase actions, they require the actual passphrase.

- **Debatable** no permissions for theme: It will be used by the login page, so we cant really protect it


pkg/permissions/verbs.go
```diff
@@ -18,36 +21,87 @@ const (
 	DELETE = Verb("DELETE")
 )
 
+var allVerbsOrder = []Verb{GET, POST, PUT, PATCH, DELETE}
+
 // VerbSet is a Set of Verbs
-type VerbSet []Verb
+type VerbSet map[Verb]bool
```
nono: Do you use the value of these maps? Else, using `map[Verb]struct{}` is a more compact representation of a set

pkg/permissions/verbs.go
```diff
@@ -18,36 +21,87 @@ const (
...
 	}
 	return out[1:]
 }
 
+// MarshalJSON implements json.Marshaller on VerbSet
+// the VerbSet is converted to a json array
+func (vs VerbSet) MarshalJSON() ([]byte, error) {
+	s := make([]string, len(vs))
+	i := 0
+	for _, v := range allVerbsOrder {
```
nono: You don't iter on `vs` directly to preserve the order of the verbs (GET always POST) ?

nono:
> Debatable no permissions for theme

I'm OK with that

web/settings/instance.go
```diff
@@ -38,8 +39,13 @@ func getInstance(c echo.Context) error {
 	if err != nil {
 		return err
 	}
+	doc.Type = consts.Settings
 	doc.M["locale"] = instance.Locale
 
+	if err = permissions.Allow(c, permissions.GET, doc); err != nil {
```
nono: `getInstance` should also be accessible with the register token. Can you add this case?

web/settings/disk_usage.go
```diff
@@ -37,5 +38,11 @@ func diskUsage(c echo.Context) error {
 		return err
 	}
 
-	return jsonapi.Data(c, http.StatusOK, &apiDiskUsage{used}, nil)
+	var result = &apiDiskUsage{used}
+
+	if err = permissions.Allow(c, permissions.GET, result); err != nil {
+		return err
+	}
```
nono: I'd prefer to check the permission before calling `vfs.DiskUsage`. Something like that:

```go
func diskUsage(c echo.Context) error {
    var result = &apiDiskUsage{}
    if err = permissions.Allow(c, permissions.GET, result); err != nil {
        return err
    }

    used, err := vfs.DiskUsage(instance)
    if err != nil {
        return err
    }

    result.Used = used
    return jsonapi.Data(c, http.StatusOK, result, nil)
}
```

web/permissions/permissions.go
```diff
@@ -49,46 +63,95 @@ const ContextClaims = "token_claims"
...
-			err = crypto.ParseJWT(token, keyPicker(instance), &claims)
-		} else {
-			// no token is provided, hopefully the next handler does not need one
-			return next(c)
-		}
 
-		if err != nil {
+		_, _, err := extract(c)
+		if err != nil && err != ErrNoToken {
+			// no token is provided, hopefully the next handler does not need one
```
nono: The comment is misleading, no? When no token is provided, we don't enter in the if branch

pkg/permissions/verbs.go
jinroh: An even more compact representation, since we known all the element that can compose the set, would be to use [a struct of 5 bools](http://golang-sizeof.tips/?t=Ly8gU2FtcGxlIGNvZGUKc3RydWN0IHsKCUdldCBib29sCglQb3N0IGJvb2wKCVB1dCBib29sCglQYXRjaCBib29sCglEZWxldGUgYm9vbAp9Cg==). You also would not have to worry about the order.

aenario: I am going to go with map[verb]struct as I feel we might neet more later.

pkg/permissions/verbs.go
aenario: yes, order of map iteration is random

web/permissions/permissions.go
```diff
@@ -98,7 +107,9 @@ func extract(c echo.Context) (*permissions.Claims, *permissions.Set, error) {
 	}
 
 	var pset permissions.Set
-	if pset, err = claims.PermissionsSet(); err != nil {
+	if claims.Audience == permissions.RegistrationTokenAudience {
+		pset = registerTokenPermissions
```
nono: I think there is a confusion between the registration token used in OAuth 2 dynamic client registration/management protocol, and the registerToken used for the onboarding. The registerToken for onboarding is not a JWT, and the registration token for OAuth 2 should not give the permissions to read/write the instance!

aenario: Correct, I did mix the two. Thanks !

nono:
Let's merge this one. Works will continue in #238

PR #238, opened 2017-01-23
Permissions persistence

aenario:
Basically, we add a Permission doctype
- Upon install we create it by copying from the manifest
- The permission Set can now come from 3 sources : A - the app's permission doc, B - the oauth client JWT scope C - For registerToken, a hard-coded "io.cozy.settings:GET:instance" 
- The permission type already has the codes & expires which we will use for subpermissions routes in docs/permission.md

Still lacking some tests, working on it now.


web/permissions/permissions.go
```diff
@@ -183,6 +185,10 @@ func displayPermissions(c echo.Context) error {
 	return c.JSON(200, set)
 }
 
+func constantTimeEqual(a, b string) bool {
+	return 1 ==
```
nono: It looks like something you forgot to remove :wink:

web/permissions/permissions.go
```diff
@@ -119,7 +120,8 @@ func extract(c echo.Context) (*permissions.Claims, *permissions.Set, error) {
 	} else if claims.Audience == permissions.AccessTokenAudience {
 		// Oauth token, extract permissions from JWT-encoded scope
 		pset, err = permissions.UnmarshalScopeString(claims.Scope)
-	} else if claims.Audience == permissions.RegistrationTokenAudience {
+	} else if registerToken := []byte(c.QueryParam("registerToken"));
+			subtle.ConstantTimeCompare(registerToken, instance.RegisterToken) == 1 {
```
nono: 2 things:

1. This test is too late. The registerToken is used when there is no JWT token, and so, we have returned at line 99.
2. We should do the constant time comparison only when there is a token. No need to burn CPU cycles when there is no token at all.

web/permissions/permissions.go
```diff
@@ -49,50 +74,143 @@ const ContextClaims = "token_claims"
...
+	}
+}
+
+func extractJWTClaims(c echo.Context, instance *instance.Instance) (*permissions.Claims, error) {
+	var claims permissions.Claims
+	var err error
+	if token := getBearerToken(c); token != "" {
+		err = crypto.ParseJWT(token, keyPicker(instance), &claims)
+	} else if token := getQueryToken(c); token != "" {
+		err = crypto.ParseJWT(token, keyPicker(instance), &claims)
```
jinroh: This is not an issue raised by this PR, but the `ErrInvalidAudience` and `ErrInvalidToken`, and maybe some other errors that would be returned by `ParseJWT`, should be treated so that they do not cause an HTTP error `500` (probably a `401`...)

pkg/apps/installer.go
```diff
@@ -235,9 +236,24 @@ func (i *Installer) Poll() (man *Manifest, done bool, err error) {
...
+
+	if err := permissions.Destoy(db, man.Slug); err != nil {
+		return err
+	}
+
+	if err := couchdb.UpdateDoc(db, man); err != nil {
+		return err
+	}
+
+	return permissions.Create(db, man.Slug)
```
nono: When we will have optional permissions, we will have to do something more complex here to preserve them when an app is updated. But for now, it's OK.

pkg/permissions/permissions_doc.go
```diff
@@ -0,0 +1,105 @@
...
+
+// Permission is a storable object containing a set of rules and
+// several codes
+type Permission struct {
+	pID           string
+	pRev          string
+	ApplicationID string            `json:"application_id"`
+	Permissions   Set               `json:"permissions, omitempty"`
+	ExpiresAt     int               `json:"expires_at, omitempty"`
+	Codes         map[string]string `json:"codes, omitempty"`
```
nono: Just a small thing: in the rest of our codebase, we didn't put a space between `omitempty`

aenario: And we were right to do so as it does not works XD

PR #241, opened 2017-01-24
Jobs scheduler and @at, @in, @interval triggers

jinroh:
Adds a in-memory scheduler system usable by self-hosted instances, with the necessary interface abstraction to foresee our job scheduler with distributed stacks.

The storage used as persistence of the triggers, with this in-mem scheduler, is CouchDB, using the io.cozy.triggers doctype. On startup, the jobs broker and jobs scheduler associated to each instance are started.


pkg/jobs/trigger_at.go
```diff
@@ -0,0 +1,89 @@
...
+import (
+	"time"
+
+	"github.com/cozy/cozy-stack/web/jsonapi"
+)
+
+// AtTrigger implements the @at trigger type. It schedules a job at a specified
+// time in the future.
+type AtTrigger struct {
+	typ  string
```
nono: Why `typ`? It is the same as `in.Type`, no?

pkg/jobs/jobs.go
```diff
@@ -121,8 +121,71 @@ type (
...
+		Options    *JobOptions `json:"options"`
+		Message    *Message    `json:"message"`
+	}
+
+	// TriggerRequest struct contains the paramameters to create a new trigger.
+	TriggerRequest struct {
+		Type      string          `json:"type"`
+		Arguments json.RawMessage `json:"arguments"`
+		Options   *WorkerConfig   `json:"options"`
+	}
```
nono: This struct seems to be used nowhere. Is it for a future usage?

cmd/serve.go
```diff
@@ -13,6 +15,20 @@ var serveCmd = &cobra.Command{
...
+		// instances on each startup
+		ins, err := instance.List()
+		if err != nil && !couchdb.IsNoDatabaseError(err) {
+			return err
+		}
+		for _, in := range ins {
+			if err := in.StartJobSystem(); err != nil {
+				return err
+			}
+		}
```
nono: I'd prefer if we can move this logic to the jobs package. Maybe something like:

```go
jobs.StartSystem()
defer jobs.StopSystem()
return web.ListenAndServe()
```

pkg/jobs/jobs.go
```diff
@@ -121,8 +121,71 @@ type (
...
+		Schedule() <-chan *JobRequest
+		// Unschedule should be used to clean the trigger states and should close
+		// the returns jobs channel.
+		Unschedule()
+	}
+
+	// TriggerStorage interface is used to represent a persistent layer on which
+	// triggers are stored.
+	TriggerStorage interface {
+		GetAll() ([]Trigger, error)
```
nono: [Optim] maybe we can add a `GetInfos()` method to return a list of TriggerInfos. It would more efficient than to return a list of triggers for the listing of the triggers in the web API, as it avoids to allocate a channel for each trigger.

pkg/jobs/trigger_interval.go
```diff
@@ -0,0 +1,69 @@
...
+// fixed period of time.
+type IntervalTrigger struct {
+	du   time.Duration
+	in   *TriggerInfos
+	done chan struct{}
+}
+
+// NewIntervalTrigger returns a new instace of IntervalTriven given the
+// specified options.
+func NewIntervalTrigger(infos *TriggerInfos) (*IntervalTrigger, error) {
```
nono: To be clear, for the moment, we have no persistence of when the last job was executed, right?

If I have a konnector that is launched each month, but I restart the stack every weeks (for updates maybe), the trigger will never be fired. I think it is something we should address, but it can be later. A todo can be enough for the moment.

pkg/jobs/trigger_at.go
```diff
@@ -0,0 +1,89 @@
...
+	go func() {
+		select {
+		case <-time.After(-duration):
+			req := &JobRequest{
+				WorkerType: a.in.WorkerType,
+				Message:    a.in.Message,
+				Options:    a.in.Options,
+			}
+			ch <- req
+			close(ch)
```
nono: Maybe we should remove the `@at/@in` trigger from the database after it has been executed.

web/jobs/jobs_test.go
```diff
@@ -89,6 +91,195 @@ func TestCreateJobWithEventStream(t *testing.T) {
...
+	req4, err := http.NewRequest("DELETE", ts.URL+"/jobs/triggers/"+triggerID, nil)
+	if !assert.NoError(t, err) {
+		return
+	}
+
+	res4, err := http.DefaultClient.Do(req4)
+	if !assert.NoError(t, err) {
+		return
+	}
+	assert.Equal(t, http.StatusOK, res4.StatusCode)
```
jinroh: done b272bd4c63e8736f433ac616aefc322121a36df2

web/jobs/jobs.go
```diff
@@ -23,11 +23,20 @@ type (
 		Arguments json.RawMessage  `json:"arguments"`
 		Options   *jobs.JobOptions `json:"options"`
 	}
-
 	apiQueue struct {
 		Count      int `json:"count"`
 		workerType string
 	}
+	apiTrigger struct {
```
jinroh: Indeed I could. Though the `triggerDoc` does only implement the `couchdb.Doc` interface, so I would have to add some methods.

However I don't want to fix that in this PR since I would like to introduce a more general approach to fix this issue: by having a simple struct in the `jsonapi` that would implement the `jsonapi.Object` and that we could create on the fly for these type of objects that do not implement the `jsonapi.Object` interface.

web/jobs/jobs.go
```diff
@@ -108,10 +131,72 @@ func pushJob(c echo.Context) error {
 	return nil
 }
 
+func newTrigger(c echo.Context) error {
+	instance := middlewares.GetInstance(c)
+	scheduler := instance.JobsScheduler()
+	req := &apiTriggerRequest{}
```
jinroh: > Why do you use an intermediate apiTriggerRequest and can't deserialize directly to a jobs.TriggerInfos?

They do not share the exact same properties regarding the worker arguments. This allow to have a `bytes.RawMessage` and not having to parse the worker arguments on trigger creation and delay it at execution.

> You don't use JSON-API for the requests, only for the responses?

Indeed I need to fix that. The lazy part of myself hid it from me.

web/jobs/jobs.go
```diff
@@ -108,10 +131,72 @@ func pushJob(c echo.Context) error {
...
+	instance := middlewares.GetInstance(c)
+	scheduler := instance.JobsScheduler()
+	ts, err := scheduler.GetAll()
+	if err != nil {
+		return wrapJobsError(err)
+	}
+	objs := make([]jsonapi.Object, 0, len(ts))
+	for _, t := range ts {
+		objs = append(objs, &apiTrigger{t})
+	}
```
jinroh: I found this trick recently. I don't think there is a difference. However I can fallback on the basic solution for coherency with the rest of the code.

jinroh: done 883dc19bc6492a59edbff46b699cd5b9e70bffd3

pkg/jobs/trigger_at.go
jinroh: Because `type` is a reserved keyword and `Type()` is a method is part of the `Trigger` interface

pkg/jobs/trigger_at.go
jinroh: [It is the case](https://github.com/jinroh/cozy-stack/blob/883dc19bc6492a59edbff46b699cd5b9e70bffd3/pkg/jobs/mem_jobs.go#L353-L362).

And I have tests for it, checking that the trigger does not exist anymore after it triggered.

pkg/jobs/jobs.go
jinroh: I'll try that. Thanks

pkg/jobs/jobs.go
jinroh: Good spot ! It's a mistake. Removed 960c864aaade183c7e60c5e95b3f4de7574c2be5.

pkg/jobs/trigger_at.go
nono: My question is: why do we need this field? I think that the `Type()` method can be implemented as `return a.in.Type`

pkg/jobs/trigger_at.go
nono: In that, if `duration >=0` should fire a job request

pkg/jobs/trigger_interval.go
jinroh: Yes... indeed. I should fix that or make it clear on the documentation that the `@interval` has no state.

However, I think the `@cron` scheduler should be more appropriate for these type of recurring tasks.

pkg/jobs/trigger_at.go
jinroh: Oh indeed. I did not think of that. done d6e801db5e104d8973e9f212eda49781bc8a67cf

pkg/jobs/trigger_interval.go
nono: We have the same thing for `@in`. If you add a trigger to send you a reminder by email in one month, and that the stack is restarted between, it will reset the counter.

Maybe we should transform `@in` triggers in `@at` triggers, and `@interval` triggers in `@cron`.

jinroh: @nono agreed. `@in` triggers are already `@at` triggers. I'll probably remove the `@interval` implementation for now.

pkg/jobs/trigger_at.go
jinroh: done eb25cf908c1eced5250a8f2fe6c04d187a0eeef1

pkg/jobs/jobs.go
jinroh: done eb25cf908c1eced5250a8f2fe6c04d187a0eeef1

pkg/jobs/trigger_interval.go
jinroh: done b5ffffb2da3b45f3e14eed385e35cf0f393cb8f6

cmd/serve.go
jinroh: You sure about that. I put it in `serve.go` knowing that it is temporary.

The thing is that we rely on one job system per instance since they are namespaced by domains (broker and queues...).

pkg/jobs/mem_storage.go
```diff
@@ -0,0 +1,66 @@
+package jobs
```
jinroh: indeed. done be923dc5621be6cf1bff83f8dffd20b250b70dac

web/jobs/jobs.go
jinroh: done c1e2f079fdc46a1a021e4f8784ea22a03b19af8e

PR #243, opened 2017-01-25
Add documentation on documents referenced by a file

nono:


docs/references-docs-in-vfs.md
```diff
@@ -0,0 +1,301 @@
...
+
+## Implementation
+
+The references are persisted in the `io.cozy.files` documents in CouchDB. A
+mango index is used to fetch all the files that are associated to a given
+document (for `GET /files/:type/:doc-id/relationships/references`).
+
+For request to update or move to trash a file, it is easy to fetch its
+CouchDB document to see if it has a reference. But it is more difficult when
+moving a folder to trash. To do that, we will add on each folder a counter of
```
aenario: Not sure about this tradeof

**Option A : keep counter of "referenced" in folder doc**
 we add two couchdb request GET folder + PUT folder to every stack request for references changes and file change. The many changes on folder doc will cause noise & bandwith to be replicated to devices 

**Option B : only check for "referenced" when trashing a folder**
we need two requests to fetch the number of references in a folder
- get all descendant folders from a given folder (1 map view) 
- get total number of "referenced" file for this folders list (1 mapreduce?keys view)

docs/references-docs-in-vfs.md
```diff
@@ -0,0 +1,301 @@
...
+
+### Modification of a referenced file
+
+Before an application updates a file, it can check if the file has some
+references. If it is the case, it may offer to the user two choices:
+
+- update the file with the new version (the albums and playlists will use the
+  new version)
+- save the new version as a new file and preserve the old file (the old file
+  may be moved to a `originals` directory).
```
aenario: We should define now this folder that will have more usage than just unreferencing docs.

May be something along the lines of 
> A special VFS folder Id = `io.cozy.attachments`, Path=`nil`/`_cozy_hidden` is used to store files that are referenced by documents but should not appears in the VFS user view. It contains one VFS folder by type, which each contains on folder by ID of the referencing doc.

docs/references-docs-in-vfs.md
```diff
@@ -0,0 +1,301 @@
...
+#### Response
+
+```http
+HTTP/1.1 204 No Content
+Content-Type: application/vnd.api+json
+```
+
+### DELETE /files/:type/:doc-id/relationships/references
+
+This bulk deletion of refences on many files can be useful when an album or
```
aenario: s/refences/references

docs/references-docs-in-vfs.md
```diff
@@ -0,0 +1,301 @@
...
+```http
+POST /files/io.cozy.playlists/e9308dc2-e2e3-11e6-b685-fb88662613d4/relationships/references HTTP/1.1
+Content-Type: application/vnd.api+json
+Accept: application/vnd.api+json
+```
+
+```json
+{
+  "data": [
+    { "type": "io.cozy.files", "id": "417c4e58-e2e4-11e6-b7dc-2b68ed7b77f4" },
```
aenario: [nitpick] HTTP400 if the given id is a folder ? Guarantee of atomicity?

docs/references-docs-in-vfs.md
```diff
@@ -0,0 +1,301 @@
...
+```
+
+#### Response
+
+```http
+HTTP/1.1 204 No Content
+Content-Type: application/vnd.api+json
+```
+
+### GET /files/:type/:doc-id/relationships/references
```
aenario: [nitpick] While it makes sense in echo groups structure to have these routes as child of /files/, it seems weird from an user/rest POV. `/data/:type/:doc-id/relationships/reference` ?

docs/references-docs-in-vfs.md
```diff
@@ -0,0 +1,301 @@
...
+    "relationships": {
+      "parent": {
+        "links": {
+          "related": "/files/fce1a6c0-dfc5-11e5-8d1a-1f854d4aaf81"
+        },
+        "data": {
+          "type": "io.cozy.files",
+          "id": "fce1a6c0-dfc5-11e5-8d1a-1f854d4aaf81"
+        }
+      }
```
aenario: you don't actually list references here

docs/files.md
```diff
@@ -297,6 +297,14 @@ Location: http://cozy.example.com/files/9152d568-7e7c-11e6-a377-37cbfb190b4b
           "type": "io.cozy.files",
           "id": "fce1a6c0-dfc5-11e5-8d1a-1f854d4aaf81"
         }
+      },
+      "references": {
```
aenario: [naming] `referenced_by` ? `references` would be the invert relationship, from doc to files.

docs/references-docs-in-vfs.md
nono: @aenario how do you build the map view for all descendants folder from a given folder? My option A was because I didn't find how to do that. But if it is possible, option B is clearly better.

docs/files.md
nono: @jinroh what do you think of that?

docs/references-docs-in-vfs.md
nono: oops

docs/references-docs-in-vfs.md
nono: Yes, let's go for that

docs/references-docs-in-vfs.md
nono: Yes, 400 for a directory.

docs/references-docs-in-vfs.md
nono: I didn't want to give too much details now. I think our PO may have some opinions on that subject. In particular, it should be still possible to find how the disk space is used.

cc: @jsilvestre @poupotte

docs/references-docs-in-vfs.md
nono: For option A, I think we will have to make the GETs even without the references, just to be sure that we didn't create a loop (the parent of folder A is B and the parent of folder B is A).

For the PUTs, we can use a local document in couchdb instead of putting this information in the main documents (IIRC, documents those _id starts with an `_` are not replicated in couchdb).

docs/files.md
jinroh: `referenced_by` is clearer to me here.

docs/references-docs-in-vfs.md
aenario: While files only have a ref to their parent directory, folders have the whole path, so a view&query like this should give us all folders descendant of a parent.
```
map = function(doc) { if(doc.type === "folder") emit(doc.path) }
query = {  
    starkey: parent_folder_path + "/",
    endkey: parent_folder_path +"/\uFFFF"
}
```
Then a view like

```
map = function(doc) { if(doc.referenced != null) emit(doc.parentID) }
reduce = count
query = {keys: [list from above]}
```

will give us the number of referenced in the parent (with a race condition to mutex).

docs/references-docs-in-vfs.md
jsilvestre: So, the idea is that we want that kind of things but we haven't any clue of what to do it technically. My input here would be that you can organize the folders/original files like you want, as long as we can do what we want with the related permissions: the app Photos should be allowed to access the files it can restores, the app Music too, but they shouldn't be allowed to mess around with files they aren't related to. What do you think, is it something to keep in mind?

nono: OK, I will keep this in mind

PR #245, opened 2017-01-25
Fix the deletion of an instance with : in its domain

nono:


jinroh:
What about domains with characters that should be escaped in a url ?
Shouldn't we unescape the `c.Param("domain")`, or pass the domain via a query parameter ?

cmd/instances.go
```diff
@@ -21,6 +21,10 @@ var flagEmail string
 var flagApps []string
 var flagDev bool
 
+func validDomain(domain string) bool {
+	return !strings.ContainsAny(domain, " /?#@\t\r\n")
+}
```
jinroh: why not also put these validations on the [backend side](https://github.com/cozy/cozy-stack/blob/master/pkg/instance/instance.go#L317-L326) ?

nono: Done with a1aa1bdfabda79d9b0d9bbd3a4fb839582aade93

PR #246, opened 2017-01-25
Do query unescape in DELETE /instances/xxx route

aenario:


nono:
I prefer my version: #245

PR #248, opened 2017-01-25
Add modes noreply and from to sendmail worker

jinroh:
Splits the `sendmail` worker into two different modes:

  - `noreply` to send notification to the user's mail
  - `from` to send mail from the user's mail to any other mail addresses

Also add a workers.md document with some documentation on the sendmail worker.


docs/workers.md
```diff
@@ -0,0 +1,78 @@
...
+
+A part can be represented by two different object types:
+
+  - `type` string of the content type: either `text/html` or `text/plain`
+  - `body` string of the actual body content of the part
+
+or
+
+  - `template` string of the [html template](https://golang.org/pkg/html/template/)
+  - `values` object containing the key/values to fill the template with
```
nono: The template can only be used with html? I'd have put the values outside of the parts and use them in both the html and text parts. Maybe something like this:

```js
{
    "mode": "from",
    "to": [
        {"name": "John Doe 1", "email":"john1@doe"},
        {"name": "John Doe 2", "email":"john2@doe"}
    ],
    "subject": "Hey !",
    "parts": [
        {"type": "text/html", "template":"<h1>{{.Title}}</h1>"},
        {"type": "text/plain", "body": "{{.Title}} !"}
    ],
    "values": {"Title": "Hello!"}
}
```

pkg/jobs/workers/mail.go
```diff
@@ -53,12 +58,61 @@ type MailPart struct {
...
+	err = couchdb.GetDoc(in, consts.Settings, consts.InstanceSettingsID, doc)
+	if err != nil {
+		return nil, err
+	}
+	email, ok := doc.M["email"].(string)
+	if !ok {
+		return nil, fmt.Errorf("Domain %s has no email in its settings", domain)
+	}
+	return &MailAddress{
+		Name:  "", // TODO: no name settings for an instance ?
```
nono: You can take `public_name` in the same settings document

docs/workers.md
jinroh: @nono of course we need to do that ! I'll change the API.

pkg/jobs/workers/mail.go
jinroh: @nono is this used anywhere else ? I can't find this string in the projet..

docs/workers.md
jinroh: done 4a081cb3350c03cf74a634f7a2c788d336826408

pkg/jobs/workers/mail.go
nono: @jinroh https://github.com/cozy/cozy-stack/blob/master/docs/settings.md#get-settingsinstance

It's not yet used. It will be set by the onboarding app later.

jinroh: done 697cbda00e977c7d644bd920702e91b6dcb7f87f

aenario:
My understanding was that the "from" option will not work (always ending up in spams) unless we have the user's SMTP account?

Otherwise, good to merge.

jinroh:
@aenario yes I think so too.. This is why we might need to introduce an SMTP configuration per instance for self-hoster at some point (instead of a global SMTP configuration).

PR #251, opened 2017-01-26
Add a "instances token" command to generate application tokens

jinroh:
Adds a http endpoint and CLI command to generate an application token.


jinroh:
I'll add the possibility to generate more varieties of token (not just applications token) I'll reopen the PR when ready.

Sorry for the flood.

PR #252, opened 2017-01-27
Commands app-token and oauth-token

jinroh:
Adds a generic endpoint `GET /instances/token` to generate different kinds of tokens together with two commands:

  - `cozy-stack instances token-app [domain] [slug]`
  - `cozy-stack instances token-oauth [domain] [clientid] [scopes]`

### DIFF
--- a/cmd/instances.go
+++ b/cmd/instances.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"encoding/hex"
 	"fmt"
+	"io/ioutil"
 	"net/url"
 	"os"
 	"strings"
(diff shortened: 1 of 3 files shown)


nono:
Just to be clear, the token given by `token-app` can only be used if an app with the same slug is installed, right? Maybe, we also need a command `cozy-stack apps install [domain] [repository]` to allow client-side apps to use such a token for their tests.

cc: @aenario

jinroh:
@nono yes, an application with the specified slug name should be installed with a set of permissions for this token to be useful.

I agree that without your proposed CLI, the `token-app` command is not that useful for tests purposes.

PR #255, opened 2017-01-27
Simpler image docker with cozy-stack build outside of docker

jinroh:
This PR adds a `docker-dev` command to the `build.sh` script in order to create the `cozy-app-dev` docker image.

Using this command, a new cozy-stack image is built *outside of docker* which has advantages:

  - the `Dockerfile` is much simpler and does not need to install a Go dependency (`caddy` is installed by downloading a binary release)
  - docker images are much smaller and the diff between the images versions should be greatly optimized
  - docker images are much much faster to build
  - it allows to package a docker image with the cozy-stack version of our choice


scripts/build.sh
```diff
@@ -22,11 +22,12 @@ echo_err() {
...
 	echo -e "\nCommands:\n"
-	echo -e "  release  builds a release of the current working-tree"
-	echo -e "  install  builds a release and install it the GOPATH"
-	echo -e "  deploy   builds a release of the current working-tree and deploys it"
-	echo -e "  assets   move and download all the required assets (see: ./assets/externals)"
-	echo -e "  clean    remove all generated files from the working-tree"
+	echo -e "  release     builds a release of the current working-tree"
+	echo -e "  install     builds a release and install it the GOPATH"
+	echo -e "  deploy      builds a release of the current working-tree and deploys it"
+	echo -e "  docker-dev  builds a cozy-app-dev docker image"
```
nono: Can you update https://github.com/cozy/cozy-stack/blob/master/docs/docker.md#building-the-cozy-app-dev-image with this new command?

jinroh: done 6d1e5668cffc843678aec3eccaec8bd48cc1f5ee

jinroh:
@nono I opted for the last commit date of the `externals` file instead of its mod time since they are not preserved by git and so were not consistent across clones

PR #256, opened 2017-01-27
Introduce download-store & 2 steps download

aenario:


docs/files.md
```diff
@@ -592,7 +592,7 @@ Location: http://cozy.example.com/files/9152d568-7e7c-11e6-a377-37cbfb190b4b
 
 ### POST /files/archive
 
-Create an archive and download it. The body of the request lists the files and
+Create an archive. The body of the request lists the files and
```
nono: Would it be difficult to have content negotiation on that endpoint? When the client asks for `application/vnd.api+json` (JSON-API), the stack gives it a link for downloading the archive. But if it asks anything else, the stack sends the archive directly. I think it can be easier for apps that are not in a browser to have the possibility to download the archive in one request and not two.

docs/files.md
```diff
@@ -623,13 +623,52 @@ Content-Type: application/vnd.api+json
...
+    "type": "io.cozy.files.archives",
+    "id": "4521DC87",
+    "attributes": {
+      "href": "/files/archive/4521DC87/project-X.zip"
+    }
+  }
+}
+```
+
+### GET /files/archive/:key/:name?
```
nono: Why the `?`

pkg/vfs/download_store.go
```diff
@@ -0,0 +1,92 @@
...
+	Archives map[string]*Archive
+	Files    map[string]*fileRef
+}
+
+func (s *memStore) makeSecret() string {
+	return hex.EncodeToString(crypto.GenerateRandomBytes(16))
+}
+
+func (s *memStore) AddFile(f string) (string, error) {
+	fref := &fileRef{
```
nono: Maybe before adding a file, we should remove all the filesRef that have expired in `s.Files` to avoid memory leak

pkg/vfs/download_store.go
```diff
@@ -0,0 +1,92 @@
...
+		Path:      f,
+		ExpiresAt: time.Now().Add(downloadStoreTTL),
+	}
+	key := s.makeSecret()
+	s.Files[key] = fref
+	return key, nil
+}
+
+func (s *memStore) AddArchive(a *Archive) (string, error) {
```
nono: And the same for archive

docs/files.md
```diff
@@ -62,7 +62,7 @@ Location: http://cozy.example.com/files/6494e0ac-dfcb-11e5-88c1-472e84a9cbee
     "type": "io.cozy.files",
     "id": "6494e0ac-dfcb-11e5-88c1-472e84a9cbee",
     "meta": {
-      "rev": "1-ff3beeb456eb"
+      "rev": "1-ff3beeb456eb",
```
nono: In JSON, the trailing comma is not allowed...

docs/files.md
```diff
@@ -297,14 +297,6 @@ Location: http://cozy.example.com/files/9152d568-7e7c-11e6-a377-37cbfb190b4b
...
           "id": "fce1a6c0-dfc5-11e5-8d1a-1f854d4aaf81"
         }
-      },
-      "referenced_by": {
-        "links": {
-          "self": "/files/fce1a6c0-dfc5-11e5-8d1a-1f854d4aaf81/relationships/references"
-        },
-        "data": [
-          { "type": "io.cozy.playlists", "id": "94375086-e2e2-11e6-81b9-5bc0b9dd4aa4" }
-        ]
```
nono: Why did you remove the example of `referenced_by`?

docs/files.md
```diff
@@ -317,9 +309,6 @@ Location: http://cozy.example.com/files/9152d568-7e7c-11e6-a377-37cbfb190b4b
 **Note**: for an image, the links section will also include a link called
 `thumbnail` to the thumbnail URL of the image.
 
-**Note**: see [references of documents in VFS](references-docs-in-vfs.md) for
-more informations about the references field.
-
```
nono: Why did you remove this note?

docs/files.md
aenario: bad merge I suppose. fixing now

docs/files.md
```diff
@@ -623,13 +622,52 @@ Content-Type: application/vnd.api+json
...
+Download a previously created archive. The name parameter is not used in
+the stack but aims to allow setting a name even for browser / downloader that do not support Content-Disposition filename.
+
+**This route does not require Basic Authentification**
+
+
+```http
+GET /files/archive/4521DC87 HTTP/1.1
+Accept: application/zip
++Content-Length: 12345
```
nono: There is a `+` to remove

pkg/vfs/download_store.go
```diff
@@ -0,0 +1,128 @@
...
+type fileRef struct {
+	Path      string
+	ExpiresAt time.Time
+}
+
+// downloadStoreTTL is the time an Archive stay alive
+const downloadStoreTTL = 1 * time.Hour
+const downloadStoreCleanInterval = 1 * time.Hour
+
+var storeStore map[string]*memStore
```
jinroh: `maps` are not thread safe, you should have a mutex around this one since it is used concurrently

pkg/vfs/download_store.go
```diff
@@ -0,0 +1,128 @@
...
+		Archives: make(map[string]*Archive),
+		Files:    make(map[string]*fileRef),
+	}
+	return storeStore[domain]
+}
+
+type memStore struct {
+	Archives map[string]*Archive
+	Files    map[string]*fileRef
+}
```
jinroh: this memStore should also have at least one mutex to protect read/writes from `Archives` and `Files` fields

pkg/vfs/download_store.go
```diff
@@ -0,0 +1,144 @@
...
+	s.Mutex.Lock()
+	cleanStore(s)
+	key := s.makeSecret()
+	s.Archives[key] = a
+	s.Mutex.Unlock()
+	return key, nil
+}
+
+func (s *memStore) GetFile(k string) (string, error) {
+	s.Mutex.Lock()
```
jinroh: use defer s.Mutex.Unlock() to avoid a deadlock on error branches where the lock is not unlocked.

pkg/vfs/download_store.go
```diff
@@ -0,0 +1,144 @@
...
+	if time.Now().After(f.ExpiresAt) {
+		delete(s.Files, k)
+		return "", nil
+	}
+	s.Mutex.Unlock()
+	return f.Path, nil
+}
+
+func (s *memStore) GetArchive(k string) (*Archive, error) {
+	s.Mutex.Lock()
```
jinroh: idem

pkg/vfs/download_store.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	for k, a := range s.Archives {
+		if now.After(a.ExpiresAt) {
+			delete(s.Archives, k)
+		}
+	}
+}
+
+func cleanDownloadStore() {
+	storeStoreMutex.Lock()
```
jinroh: would use a `defer Unlock()` on all locking methods (even the correct ones), for clarity

pkg/vfs/download_store.go
```diff
@@ -0,0 +1,144 @@
...
+			Archives: make(map[string]*Archive),
+			Files:    make(map[string]*fileRef),
+		}
+		storeStore[domain] = store
+	}
+	return store
+}
+
+type memStore struct {
+	Mutex    *sync.Mutex
```
jinroh: `sync.Mutex` are [meant to be used as non-pointers](https://golang.org/src/sync/mutex.go?s=579:825#L14) that way you do not need to initialize them when creating a var mutex or struct containing one.

jinroh: More infos in [this post](https://dave.cheney.net/2013/01/19/what-is-the-zero-value-and-why-is-it-useful).

aenario: Done in 2689490 Thanks for the link !

PR #258, opened 2017-01-30
Revoke OAuth2 client

nono:
It includes #257

### DIFF
--- a/web/settings/clients.go
+++ b/web/settings/clients.go
@@ -1,0 +1,31 @@
+package settings
+
+import (
+	"net/http"
+
+	"github.com/cozy/cozy-stack/pkg/consts"
+	"github.com/cozy/cozy-stack/pkg/oauth"
+	"github.com/cozy/cozy-stack/web/jsonapi"
+	"github.com/cozy/cozy-stack/web/middlewares"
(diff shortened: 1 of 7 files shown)


web/settings/clients.go
```diff
@@ -0,0 +1,50 @@
...
+	for i, d := range clients {
+		objs[i] = jsonapi.Object(d)
+	}
+	return jsonapi.DataList(c, http.StatusOK, objs, nil)
+}
+
+func revokeClient(c echo.Context) error {
+	instance := middlewares.GetInstance(c)
+
+	if err := permissions.AllowWholeType(c, permissions.DELETE, consts.OAuthClients); err != nil {
```
aenario: Making client a `Validable Validate(){ return false }` and checking permissions with `permissions.Allow` after getting the client would allow us to give each client a scope to revoke itself, which would be useful in the future.

nono: @aenario a client can already revoke itself via the OAuth 2.0 Dynamic Management Protocol (cf https://cozy.github.io/cozy-stack/auth.html#delete-authregisterclient-id). So I took the shortcut to not make client a validable

PR #260, opened 2017-01-31
Add a `serve-appdir` command

jinroh:
Add a `serve-appdir` command allowing to serve an application directory via the stack, withouth installing it directly into the VFS.

Most of the code is shared which allow to have the same behaviour as for an installed application.

The `ServeAppFile` method takes a implementation of the `AppFileServer` interface which allow to have some freedom on how we want to serve these files.


jinroh:
This PR is waiting for the split and release of cozy-bar into a css *and* js assets. After that it should be ok to merge. (cc @m4dz)

nono:
-> #271

PR #267, opened 2017-02-01
Fix - the register token is hex encoded in the querystring

nono:


gregorylegarec:
Works for me, thanks @nono !

PR #268, opened 2017-02-02
Add references basics

aenario:


pkg/vfs/file.go
```diff
@@ -97,19 +99,45 @@ func (f *FileDoc) SelfLink() string {
...
+		"parent":        f.parentRelationShip(),
+		"referenced_by": f.referencedByRelationShip(),
+	}
+}
+
+// HideFields returns a jsonapi.Object which serialize like the original
+// file but without the ReferencedBy field
+func (f *FileDoc) HideFields() jsonapi.Object {
+	return &struct {
+		ReferencedBy []jsonapi.ResourceIdentifier `json:"referenced_by,omitempty"`
```
nono: Does it work with?

```go
ReferencedBy []jsonapi.ResourceIdentifier `json:"-"`
```

I think it is more explicit

pkg/vfs/file.go
```diff
@@ -32,7 +32,7 @@ type FileDoc struct {
 	// File name
 	Name string `json:"name"`
 	// Parent directory identifier
-	DirID       string `json:"dir_id"`
+	DirID       string `json:"dir_id,omitempty"`
```
nono: Why `omitempty` here? Can we have files without a parent directory?

pkg/vfs/file.go
aenario: I agree with you, but unfortunately, it does not work.

PR #280, opened 2017-02-08
Fix the permissions for an app token with serve-appdir

nono:


gregorylegarec:
Works well on my environment, thanks ! 👍

PR #288, opened 2017-02-10
Support Date header on POST /files/:folder-id

nono:


docs/files.md
```diff
@@ -29,16 +29,23 @@ given, the directory is created at the root of the virtual file system.
...
+----------|-------------------
 Type      | `directory`
 Name      | the directory name
 Tags      | an array of tags
 
+#### HTTP headers
+
+Parameter | Description
+----------|---------------------------------------
+Date      | The modification date of the directory
```
sebn: Maybe make it clear it will set both the creation and the modification dates?

PR #292, opened 2017-02-13
Improve serve appdir command

nono:


cmd/serve.go
```diff
@@ -24,6 +25,15 @@ Use the --port and --host flags to change the listening option.
...
 to mount a directory as the application with the 'app' slug.
 `,
+	Example: `The most often, this command is used in its simple form:
+
+	$ cozy-stack serve
+
+But if you want to develop two apps in local (to test their interactions for
+example), you can use the --appsdir flag like this:
+
+	$ cozy-stack serve --appsdir appone:/path/to/app_one,apptwo:/path/to/app_two
```
gregorylegarec: @nono Does it still works with another app (like Files) installed directly via a POST on `/apps/files` ?

nono: @gregorylegarec yes, it works. Don't forget to install the build branch

gregorylegarec: @nono  You're right ! Thanks it works great.

PR #295, opened 2017-02-15
Add permissions to /files

aenario:
- [x] Create function to check permission with hierarchy rules 
- [x] Use said function to validate every route in /files
- [x] Use said function to validate archive routes

Still lacking some test but reviewable

### DIFF
--- a/web/permissions/permissions.go
+++ b/web/permissions/permissions.go
@@ -251,8 +251,8 @@ func AllowVFS(c echo.Context, v permissions.Verb, o vfs.Validable) error {
 	if err != nil {
 		return err
 	}
-	err = vfs.Allows(instance, *pset, v, o)
-	if err != nil {
+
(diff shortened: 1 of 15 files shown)


web/permissions/permissions.go
```diff
@@ -244,6 +244,20 @@ func AllowTypeAndID(c echo.Context, v permissions.Verb, doctype, id string) erro
...
 
+// AllowVFS validates a vfs.Validable against the context permission set
+func AllowVFS(c echo.Context, v permissions.Verb, o vfs.Validable) error {
+	instance := middlewares.GetInstance(c)
+	pset, err := getPermission(c)
+	if err != nil {
+		return err
+	}
+	err = vfs.Allows(instance, *pset, v, o)
+	if err != nil {
```
nono: You can squash these two lines in one

nono:
Can you check that `cozy-stack files` CLI still works with your PR please?

PR #296, opened 2017-02-15
[feat] Use UI form cozy-proxy for login page

gregorylegarec:
At this time, I only wrote a static version of the existing jade templates in cozy-proxy:

* I tried to clean the CSS to keep the minimum ruleset 
* I use only a small file of JS to manage password visibility
* This does not handle i18n, on purpose.

In the future, it could be interesting to be able to put the login asset from another repo. I think we may need more JS features for resetting password for example and i18n.

Thanks @aenario to have a look on the go code. First lines ever for me.


aenario:
You will need to change these tests https://github.com/cozy/cozy-stack/blob/606a0e849b5520b0674c6f03470a95659292f414/web/auth/auth_test.go#L74

assets/fonts/fonts.css
```diff
@@ -0,0 +1,47 @@
...
+ * - font-style: italic / font-weight: bold
+ */
+@font-face {
+    font-family: 'Source Sans Pro';
+    font-style: normal;
+    font-weight: 400;
+    src: local('Source Sans Pro'),
+         local('SourceSansPro-Regular'),
+         url('./sourcesanspro-regular.woff2') format('woff2'),
+         url('./sourcesanspro-regular.woff') format('woff');
```
nono: Do we need woff? I thought that woff2 was supported by all browsers now

assets/scripts/password-visibility.js
```diff
@@ -0,0 +1,11 @@
...
+  var passwordVisibility = false
+  var passwordInput = d.getElementById('password')
+  var passwordVisibilityButton = d.getElementById('password-visibility-button')
+  passwordVisibilityButton.addEventListener('click', function (event) {
+    event.preventDefault()
+    passwordVisibility = !passwordVisibility
+    passwordInput.type = passwordVisibility ? 'text' : 'password'
+    passwordVisibilityButton.attributes['aria-pressed'] = passwordVisibility
+  })
+})(window.document)
```
nono: :+1: for KISS

assets/sprites/icon-cozy.svg
```diff
@@ -0,0 +1,3 @@
...
+</svg>
```
nono: It's nice to have separated SVG files. Maybe we can rename this directory something else than `sprites`. `icon` or `images` maybe

assets/templates/login.html
```diff
@@ -1,46 +1,68 @@
-<!DOCTYPE html>
+  <!DOCTYPE html>
```
nono: Can we remove the spaces before the doctype? I'm not sure but I think that the doctype has to start from the first byte to be well supported

assets/templates/login.html
```diff
@@ -1,46 +1,68 @@
...
 <html lang="fr">
   <head>
     <meta charset="utf-8">
     <title>Cozy</title>
     <meta name="viewport" content="width=device-width, initial-scale=1">
-    <link rel="stylesheet" href="/settings/theme.css">
-    <link rel="stylesheet" href="/assets/styles/stack.css">
+    <link rel="stylesheet" href="/assets/fonts/fonts.css">
+    <link rel="stylesheet" href="/assets/styles/login.css">
```
nono: Can we have a name more generic than `login.css`? I'd like to use the same CSS for the page where we ask the user to give permissions to a device (OAuth2)

assets/templates/login.html
```diff
@@ -1,46 +1,68 @@
...
...3,26.9957683 552.32118,27.0000592 C552.035908,27.0054228 551.809368,27.2467844 551.814612,27.5364185 C551.81671,27.5750363 551.831393,28.0792139 552.066323,28.6735 C548.949302,31.6942753 544.051427,31.698566 540.928113,28.6917363 C541.169336,28.0888684 541.185068,27.576109 541.185068,27.5374911 C541.190312,27.2478572 540.964821,27.0086409 540.681646,27.0011319 C540.401618,26.9925502 540.163541,27.2264027 540.154102,27.5160368 C540.154102,27.5589455 540.11215,28.6370275 539.234308,29.2216592 C538.995183,29.3825669 538.92806,29.7097461 539.08433,29.9532532 C539.182917,30.1077246 539.346529,30.1924694 539.516434,30.1924694 C539.612923,30.1924694 539.710461,30.1645787 539.797512,30.1066519 C540.023003,29.9564713 540.211786,29.7848363 540.370154,29.6024742 C542.104862,31.2008247 544.296845,32 546.488828,32 C548.686055,32 550.883282,31.1976066 552.621136,29.5917471 C552.780553,29.7762546 552.971434,29.9521804 553.203218,30.1066519 C553.289219,30.1645787 553.387806,30.1924694 553.484295,30.1924694 C553.652102,30.1924694 553.816763,30.1066519 553.916399,29.9521804 C554.07162,29.7076006 554.004497,29.3793488 553.766421,29.2205864 L553.766421,29.2227318 Z" transform="translate(-520)"></path>
```
nono: Before, the logo was different for dev environments and production. Well, I will find a way to take this back later

assets/templates/login.html
```diff
@@ -1,46 +1,68 @@
...
+            </header>
             <div role="region">
-              <h1>Welcome</h1>
-              <p id="login-passphrase-tip" class="help">Please enter your passphrase to access to your Cozy</p>
-              <label for="login-passphrase" aria-describedby="login-passphrase-tip" class="with-input">
-                <input id="login-passphrase" name="passphrase" placeholder="Passphrase" autofocus="" type="password">
-              </label>
+              <form id="login-form" method="POST" action="/auth/login" class="login auth">
+                <input type="hidden" name="redirect" value="{{.Redirect}}" />
+                <p class="help" id="login-password-tip">Enter your password to access your Cozy</p>
```
nono: I think we have decided to call it a passphrase, and not a password, to make it clears that it has to more robust that just a few characters.

cc: @poupotte

web/auth/auth.go
```diff
@@ -84,7 +84,20 @@ func loginForm(c echo.Context) error {
...
+	}
+
+	publicName, ok := doc.M["public_name"].(string)
+
+	if !ok {
+		publicName = ""
+	}
+
 	return c.Render(http.StatusOK, "login.html", echo.Map{
+		"PublicName":        publicName,
```
nono: :+1:

assets/templates/login.html
gregorylegarec: My  goal was to separate the two concerns (login an authorization) until we update authorization, to avoid conflict and regression. When the authorization styles will be up-to-date, we will be able to merge all styles in one file. What do you think ?

assets/fonts/fonts.css
gregorylegarec: I don't know, I picked it based on the font name. Totally open to remove it. @m4dz ?

assets/templates/login.html
nono: OK, if it is just a temporary thing until we redo the authorization page, I'm totally fine with that

assets/templates/login.html
poupotte: After discussion with @kelukelu , @pierrevdk and @jsilvestre , we've decided to keep 'password' for user. 'Passphrase' seems too much technical.

assets/fonts/fonts.css
m4dz: Yep, we can safely remove them, woff2 is largely supported on our browsers scope

PR #299, opened 2017-02-16
OAuth token revocation & expiration

nono:


pkg/oauth/client.go
```diff
@@ -287,6 +287,8 @@ func (c *Client) CreateJWT(i *instance.Instance, audience, scope string) (string
 }
 
 // ValidToken checks that the JWT is valid and returns the associate claims
+// It is expected to be used for registration token and refresh token, and
+// it doesn't check when they was issued as they don't expire.
```
aenario: s/was/were

web/permissions/permissions.go
```diff
@@ -36,6 +41,11 @@ func keyPicker(i *instance.Instance) jwt.Keyfunc {
 		case permissions.AppAudience:
 			return i.SessionSecret, nil
 		case permissions.RefreshTokenAudience, permissions.AccessTokenAudience:
+			// An OAuth2 token is only valid if the client has not been revoked
```
aenario: I feel this should probably be in extractJWTClaims : 
- keyPicker responsibility is to choose the key (it handle rotation for instance)
- extractJWTClaims does the token validation (domain, ...)

aenario:
I may have used Oauth Token without associated client for testing permissions :fearful:

PR #302, opened 2017-02-17
Add sharing creation

Gara64:
This adds the possibility to create a new sharing through a `/sharings/` route.
A new sharing structure is defined, as well as a recipient structure and the link between them.
Some tests are added for the `web` and `pkg` part.

Note that no verification is done on the shared documents, I think it should be done in the permission part (check the documents exist, etc).

### DIFF
--- a/pkg/consts/consts.go
+++ b/pkg/consts/consts.go
@@ -27,6 +27,11 @@ const (
 	// Doctypes doc type for doctype list
 	Doctypes = "io.cozy.doctypes"
 
+	// Sharings doc type for sharing descriptors
+	Sharings = "io.cozy.sharings"
+	// Recipients doc type for sharing recipients
+	Recipients = "io.cozy.recipients"
+
 	// OAuthClients doc type for OAuth2 clients
 	OAuthClients = "io.cozy.oauth.clients"
 	// OAuthAccessCodes doc type for OAuth2 access codes
@@ -52,4 +57,13 @@ const (
 	DiskUsageID = "io.cozy.settings.disk-usage"
 	// InstanceSettingsID is the id of settings document for the instance
 	InstanceSettingsID = "io.cozy.settings.instance"
+)
+
+const (
+	// OneShotSharing is a sharing with no continuous updates
+	OneShotSharing = "one-shot"
+	// MasterSlaveSharing is a sharing with unilateral continuous updates
+	MasterSlaveSharing = ""
+	// MasterMasterSharing is a sharing with bilateral continuous updates
+	MasterMasterSharing = ""
 )
--- a/pkg/consts/consts.go
+++ b/pkg/consts/consts.go
@@ -27,7 +27,7 @@ const (
 	// Doctypes doc type for doctype list
 	Doctypes = "io.cozy.doctypes"
 
-	// Sharings doc type for sharing descriptors
+	// Sharings doc type for document and file sharing
 	Sharings = "io.cozy.sharings"
 	// Recipients doc type for sharing recipients
 	Recipients = "io.cozy.recipients"
@@ -63,7 +63,16 @@ const (
 	// OneShotSharing is a sharing with no continuous updates
 	OneShotSharing = "one-shot"
 	// MasterSlaveSharing is a sharing with unilateral continuous updates
-	MasterSlaveSharing = ""
+	MasterSlaveSharing = "master-slave"
 	// MasterMasterSharing is a sharing with bilateral continuous updates
-	MasterMasterSharing = ""
+	MasterMasterSharing = "master-master"
+)
+
+const (
+	// PendingStatus is the sharing pending status
+	PendingStatus = "pending"
+	// RefusedStatus is the sharing refused status
+	RefusedStatus = "refused"
+	// AcceptedStatus is the sharing accepted status
+	AcceptedStatus = "accepted"
 )
--- a/pkg/consts/consts.go
+++ b/pkg/consts/consts.go
@@ -69,10 +69,10 @@ const (
 )
 
 const (
-	// PendingStatus is the sharing pending status
-	PendingStatus = "pending"
-	// RefusedStatus is the sharing refused status
-	RefusedStatus = "refused"
-	// AcceptedStatus is the sharing accepted status
-	AcceptedStatus = "accepted"
+	// PendingSharingStatus is the sharing pending status
+	PendingSharingStatus = "pending"
+	// RefusedSharingStatus is the sharing refused status
+	RefusedSharingStatus = "refused"
+	// AcceptedSharingStatus is the sharing accepted status
+	AcceptedSharingStatus = "accepted"
 )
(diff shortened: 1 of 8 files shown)


pkg/consts/consts.go
```diff
@@ -53,3 +58,21 @@ const (
...
+	MasterMasterSharing = "master-master"
+)
+
+const (
+	// PendingStatus is the sharing pending status
+	PendingStatus = "pending"
+	// RefusedStatus is the sharing refused status
+	RefusedStatus = "refused"
+	// AcceptedStatus is the sharing accepted status
+	AcceptedStatus = "accepted"
```
nono: I think you can keep them in the `pkg/sharing` package. Or rename them to something that make it clear that they are related to the sharing, like `PendingSharingStatus`

pkg/sharings/recipients.go
```diff
@@ -0,0 +1,42 @@
...
+func (r *Recipient) DocType() string { return consts.Recipients }
+
+// SetID changes the recipient qualified identifier
+func (r *Recipient) SetID(id string) { r.RID = id }
+
+// SetRev changes the recipient revision
+func (r *Recipient) SetRev(rev string) { r.RRev = rev }
+
+// Relationships implements jsonapi.Doc
+func (r *Recipient) Relationships() jsonapi.RelationshipMap { return nil }
```
nono: It seems that you are implementing the jsonapi interfaces but you don't use it. Is it for future work?

pkg/sharings/recipients.go
```diff
@@ -0,0 +1,42 @@
...
+
+// Included implements jsonapi.Doc
+func (r *Recipient) Included() []jsonapi.Object { return nil }
+
+// Links implements jsonapi.Doc
+func (r *Recipient) Links() *jsonapi.LinksList {
+	return &jsonapi.LinksList{Self: "/recipients/" + r.RID}
+}
```
nono: We often add a compile check when a struct implements an interface. It is allow us to find easily what to change when we have to modify the interface. You can find an example here:

https://github.com/cozy/cozy-stack/blob/master/pkg/settings/settings.go#L89-L91

pkg/sharings/sharings.go
```diff
@@ -0,0 +1,140 @@
...
+
+	Permissions *permissions.Set    `json:"permissions,omitempty"`
+	SRecipients []*SharingRecipient `json:"recipients,omitempty"`
+}
+
+// SharingRecipient contains the information about a recipient for a sharing
+type SharingRecipient struct {
+	Status       string
+	AccessToken  string
+	RefreshToken string
```
nono: No JSON declaration for these fields?

pkg/sharings/sharings.go
```diff
@@ -0,0 +1,140 @@
...
+	SID         string `json:"_id,omitempty"`
+	SRev        string `json:"_rev,omitempty"`
+	Type        string `json:"type,omitempty"`
+	Owner       bool   `json:"owner"`
+	Desc        string `json:"desc,omitempty"`
+	SharingID   string `json:"sharing_id,omitempty"`
+	SharingType string `json:"sharing_type"`
+
+	Permissions *permissions.Set    `json:"permissions,omitempty"`
+	SRecipients []*SharingRecipient `json:"recipients,omitempty"`
```
nono: It's a bit tricky, but if the recipients are listed in the the relationship of JSON-API, we should avoid them in the attributes.

@aenario you have done something like this for the relations between VFS and DS. Can you show us the code?

pkg/sharings/sharings.go
```diff
@@ -0,0 +1,140 @@
...
+	return nil
+}
+
+// Create inserts a Sharing document in database
+func Create(db couchdb.Database, doc *Sharing) (*Sharing, error) {
+	err := couchdb.CreateDoc(db, doc)
+	if err != nil {
+		return nil, err
+	}
+	return doc, nil
```
nono: You don't need to return the doc:

```go
func Create(db couchdb.Database, doc *Sharing) error {
	return couchdb.CreateDoc(db, doc)
}
```

pkg/sharings/sharings.go
Gara64: It is not needed for this PR, as those fields are never modified throught json request. But it will be the case for the next steps, so it is worth nothing to add it now.

pkg/consts/consts.go
Gara64: Ok, I'll rename in `PendingSharingStatus`, I prefer to keep everything in one place for now.

pkg/sharings/recipients.go
Gara64: Yes, but it's not obvious when or if it will be needed. Should I remove it for now?

pkg/sharings/sharings.go
Gara64: Here, the recipients id and type are listed in relationship, while the content is in included attributes. 
Isn't it the same thing than for [files](https://github.com/cozy/cozy-stack/blob/master/docs/files.md#get-filesfile-id)?

pkg/sharings/recipients.go
nono: Yes, it's better to not have dead code

pkg/sharings/sharings.go
nono: For files, we have:

- ids and types in `data.relationships`
- full content in `included`

For sharing, I think we currently have:

- ids and types in `data.relationships`
- full content in `included`
- but also full content in `data.attributes.recipients`

If it is the case, we should avoid having `data.attributes.recipients`.

Gara64: Actually, it's a bit tricky, what we have for sharing is:

- Recipients ids and types in `data.relationships`
- Recipients full content in `included`
- SharingRecipient full content in `data.attributes.recipients`

A SharingRecipient contains data only relevant for the sharing, e.g. the recipient `status` (accepted/denied), the `access_token`, and the reference to the Recipient (`RefRecipient` field), so its id and type.

aenario: The code you were speaking about I guess is this https://github.com/cozy/cozy-stack/blob/749c557e354f134cb2b359ebf5736b88683d363b/pkg/vfs/file.go#L132-L142

In files, there is a "referenced_by" field, which is stored in couchdb, but we removed them from jsonapi (where they are relations instead).  

May be we mostly have a naming issue
- Recipient -> Contact
- SharingRecipient -> RecipientStatus

Not sure what is the proper JSON-API representation of relationships meta data (something with "meta" along side "links" within "relationships"."data" ???)

pkg/sharings/sharings.go
```diff
@@ -24,9 +24,9 @@ type Sharing struct {
 
 // SharingRecipient contains the information about a recipient for a sharing
 type SharingRecipient struct {
-	Status       string
-	AccessToken  string
-	RefreshToken string
+	Status       string `json:"status,omitempty"`
+	AccessToken  string `json:"status,omitempty"`
+	RefreshToken string `json:"status,omitempty"`
```
nono: @Gara64 `status` 3 times is probably not what you want

PR #303, opened 2017-02-17
Improve login page with JS

gregorylegarec:
As requested by @poupotte, I added asynchronous login validation to keep the same behavior as current login on cozy-proxy.

My goal was to quickly have a look to see if it was possible and how. It was, but as usual it took me a little more time than expected.

The main problems that I encountered and that drove me to implement two dirty hack :

  * The `/auth/login` is not expected to be called by an XHR, so I based the login success only on response status value. But as the response is supposed to redirect the browser, response.status is equal to zero. That the test I do to conclude that the passphrase was ok (first hack).

  * As the `/auth/login` done via XHR does not _actually_ log the user in, I have to submit manually the form when response is "ok" (second and most shameful hack).

  I also spent some time to request installed apps to find files and redirect to it, like it is done in onboarding-v3, but I finally added a new constant in the stack, and change the default redirect from `home` to `files`.


gregorylegarec:
@nono I think I can close this one now ?

PR #304, opened 2017-02-20
Improve login page with JS

nono:
Makes #303 less a hack


assets/scripts/async-submit.js
```diff
@@ -20,6 +20,7 @@
     }
 
     errorPanel.innerHTML = `<p>${error}</p>`
+    submitButton.removeAttribute('disabled')
```
gregorylegarec: 👍

web/auth/auth.go
```diff
@@ -97,34 +94,50 @@ func loginForm(c echo.Context) error {
...
-		"Redirect":          redirect,
+		"PublicName":       publicName,
+		"CredentialsError": nil,
+		"Redirect":         redirect,
 	})
 }
 
 func login(c echo.Context) error {
 	instance := middlewares.GetInstance(c)
+	wantsJSON := c.Request().Header.Get("Accept") == "application/json"
```
gregorylegarec: :+1:

assets/scripts/async-submit.js
```diff
@@ -31,25 +31,27 @@
...
-      } else {
-        throw new Error('The credentials you entered are incorrect, please try again.')
-      }
+      const loginSuccess = response.status < 400
+      response.json().then((body) => {
+        if (loginSuccess) {
+          submitButton.innerHTML = '<i class="fa fa-check"></i>'
+          submitButton.classList.add('btn-success')
+          if (body.redirect) {
+            window.location = body.redirect
```
gregorylegarec: Faaaaaar better 🎉

assets/scripts/async-submit.js
```diff
@@ -31,25 +31,27 @@
...
+          } else {
+            form.submit()
```
gregorylegarec: What are the cases where the `form.submit()` is called ?

pkg/apps/apps.go
```diff
@@ -43,7 +43,7 @@ const (
 const (
 	OnboardingSlug = "onboarding"
 	HomeSlug       = "home"
-	FilesSlug       = "files"
+	FilesSlug      = "files"
```
gregorylegarec: So sorry about this one. :(

(Great example of what kind of typos aligned values may cause, I am not a big fan but I swear it was an accident)

gregorylegarec:
Do we wait for @m4dz review ? I think he is pretty much busy today. If we merge it now, it could still be validated by the product team today.

pkg/apps/apps.go
nono: The golang developers have gofmt called when they save a file, and it does this stuff for us. It is really cool in golang to not have to think about code style nitpicks like how many spaces for indentation, and just let gofmt does this work for us!

To be honest, we don't use gofmt, but goimports. It is the same, but it also adds the imports automatically.

nono:
Let's merge!

pkg/apps/apps.go
gregorylegarec: Thanks for the tips !

PR #305, opened 2017-02-20
FileDoc.UpdatedAt should be set to user-provided timestamp on update

sebn:


sebn:
Tests pass and this make https://github.com/cozy/cozy-client-js/pull/78 work.
I'm still encountering weird issues on cozy-desktop, I'd like to make sure I didn't introduce a bug before merging.

But early feedback is welcome!

sebn:
The more I think about it, the more I believe I should've just removed the [`newdoc.UpdatedAt = time.Now()`](https://github.com/cozy/cozy-stack/pull/305/commits/46c44476ebb8cadd54504903434b07c8c78ce8c7#diff-bb746a71c02bde0fea11873c6e5f8524L348) line, since this is already done in [web/files.fileDocFromReq()](https://github.com/sebn/cozy-stack/blob/46c44476ebb8cadd54504903434b07c8c78ce8c7/web/files/files.go#L622).

sebn:
I reverted and just stopped overwriting `UpdatedAt` on every update (see comment above).
After discussion, I also removed my changes to prevent swallowing date parsing errors.

I added a test in `web/files/files_test.go` for the `Date` case.
As `uploadMod()` doesn't take a date param, and passing empty strings everywhere doesn't feel like an improvement, I used `doUploadOrMod()` directly.

Rebased on master, build should be green, :ship:

sebn:
I'm still getting weird issues on cozy-desktop, please wait before merging...

sebn:
After investigating, I don't believe my change introduces a new bug.
It may eventually make an existing one visible, but I'm not even sure of this.

:ship:

PR #307, opened 2017-02-21
Update Cozy-bar to 3.0.0-beta7

m4dz:


nono:
I've taken your first commit, and updated cozy-client-js too before packing assets. Thanks!

m4dz:
> I've taken your first commit

About it: do you prefer we only commit assets refs and let you pack the stack yourself?

nono:
You can do both. In general, I will merge the PR. It's just this time I also wanted to update cozy-client-js at the same time.

m4dz:
@nono is there a new version of the sack w/ those assets?

PR #308, opened 2017-02-21
Study about konnectors isolation solutions

doubleface:
To much detail, not enough. Remarks are welcome!

### DIFF
--- a/docs/konnectors.md
+++ b/docs/konnectors.md
@@ -214,13 +214,13 @@ The short list of tools which will be tested to isolate connectors is Rkt and Ns
 
 ### NsJail
 
-NsJail is a lightweight process isolation tool, making use of Linux namespaces and seccomp-bpf syscall filters. It is not a container tool like docker. Its features are quiet extensive regarding isolation. The [README](https://github.com/google/nsjail) gives the full list of options available. Although available in the google github, it is not an official google tool.
+NsJail is a lightweight process isolation tool, making use of Linux namespaces and seccomp-bpf syscall filters. It is not a container tool like docker. Its features are quite extensive regarding isolation. The [README](https://github.com/google/nsjail) gives the full list of available options. Although available in the google github, it is not an official google tool.
 
 NsJail is:
 
-- easy to install : juste a make away with standard build tools
+- easy to install : just a make away with standard build tools
 - offers a full list or isolation tools
-- lightly documented the only documentation is nsjail -h (also available in the main github page) and it is quiet cryptic for a non-sysadmin         
+- lightly documented the only documentation is nsjail -h (also available in the main github page) and it is quite cryptic for a non-sysadmin         
 like me. I could not find any help in any search engine. Some examples are available to run a back in an isolated process and work but I could not run a full nodejs (only nodejs -v worked)
 - The konnectors will need a full nodejs installed on the host
 - Is still actively maintained
@@ -249,7 +249,7 @@ Note: the --insecure-options param is to avoid the check of the image signature 
 
 ### Choice
 
-The best choice would be Rkt for it's ease of use (which is good for contribution) and wide range of isolation features + access to the big docker echosystem without beeing a burden for the host administrator.
+The best choice would be Rkt for it's ease of use (which is good for contribution) and wide range of isolation features + access to the big docker ecosystem without beeing a burden for the host administrator.
 Note : the limitation of NsJail I saw might be due to my lack of knowledge regarding system administration.
 
 ### Proposed use of rkt regarding connectors
@@ -266,7 +266,7 @@ To create an ACI file image, you just need to run a docker image one time :
     rkt export --app=nodeslim `cat uuid` nodeslim.aci && rm uuid
 
 
-The node:slim image weights 84M at this time. The node:alpine image also exists and is way lighter (19M) but I had problems with DNS with this, which could be solved with more time.
+The node:slim image weights 84M at this time. The node:alpine image also exists and is way lighter (19M) but I had problems with DNS with this, and alpine can cause some nasty bugs that are difficult to track.
 
 #### Running a connector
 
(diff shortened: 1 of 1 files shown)


docs/konnectors.md
```diff
@@ -208,6 +208,114 @@ single thread or application. But NaCl is [no longer
 maintained](https://bugs.chromium.org/p/chromium/issues/detail?id=239656#c160)
 and ZeroVM has some severe limitations, so, it won't be used.
 
+## Konnector isolation study
+
+The short list of tools which will be tested to isolate connectors is Rkt and NsJail which on paper better fullfill our needs.
+
+### NsJail
+
+NsJail is a lightweight process isolation tool, making use of Linux namespaces and seccomp-bpf syscall filters. It is not a container tool like docker. Its features are quiet extensive regarding isolation. The [README](https://github.com/google/nsjail) gives the full list of options available. Although available in the google github, it is not an official google tool.
```
nono: s/quiet/quite/

s/options available/available options/

docs/konnectors.md
```diff
@@ -208,6 +208,114 @@ single thread or application. But NaCl is [no longer
...
+
+NsJail is:
+
+- easy to install : juste a make away with standard build tools
```
nono: s/juste/just/

docs/konnectors.md
```diff
@@ -208,6 +208,114 @@ single thread or application. But NaCl is [no longer
...
+- Is still actively maintained
+
+### Rkt
+
+Rkt is very similar to docker. It can even directly run docker images from the docker registry, which gives us a lot of existing images to use, even if we want to be able to use other languages than node. For example, we could have also a container dedicated to weboob, another container could use phantomjs or casper and without forcing self-hosted users to do complicated installation procedures.
+
+Rkt is :
+- easy to install : debian, rpm package available, archlinux community package : https://github.com/coreos/rkt/releases
+- has network isolation like docker
+- offers CPU, memory limitation, seccomp isolation (but the set of rules to use is out of my understanding)
```
nono: Is it possible to customize the set of seccomp rules?

docs/konnectors.md
```diff
@@ -208,6 +208,114 @@ single thread or application. But NaCl is [no longer
...
+- is well [documented](https://coreos.com/rkt/docs/latest/), complete man pages, but not as well known as docker, then there is not a lot of things to find outside the official documentation.
+- can use docker image directly or can convert them to one runnable aci file with one simple cli command (rkt export)
+- is in active developpement but relatively stable regarding core features.
+- container images can be easily signed and the signature is checked by default when running a container.
+
+I managed to run a nodejs container with just the following commands :
+
+    rkt run --insecure-options=image --interactive docker://node:slim --name nodeslim -- -v
```
nono: Can a non-root user launch `rkt run`?

docs/konnectors.md
```diff
@@ -208,6 +208,114 @@ single thread or application. But NaCl is [no longer
...
+    rkt list   # to get the container uuid
+    rkt export --app=nodeslim <uuid> nodeslim.aci
+    rkt run --insecure-options=image --interactive nodeslim.aci -- -v  # to run node -v in the new container
+    
+Note: the --insecure-options param is to avoid the check of the image signature to ease the demonstration
+
+### Choice
+
+The best choice would be Rkt for it's ease of use (which is good for contribution) and wide range of isolation features + access to the big docker echosystem without beeing a burden for the host administrator.
```
nono: s/echosystem/ecosystem/

docs/konnectors.md
```diff
@@ -208,6 +208,114 @@ single thread or application. But NaCl is [no longer
...
+
+#### Image creation
+
+To create an ACI file image, you just need to run a docker image one time :
+    
+    rkt run  --uuid-file-save=$PWD/uuid --insecure-options=image --interactive docker://node:slim --name nodeslim -- -v
+    rkt export --app=nodeslim `cat uuid` nodeslim.aci && rm uuid
+
+
+The node:slim image weights 84M at this time. The node:alpine image also exists and is way lighter (19M) but I had problems with DNS with this, which could be solved with more time.
```
nono: Alpine is not a good idea. It uses a libc that is not the glibc, and it has some nasty side effects: it introduces some bugs that are difficult to track.

cc: @aeris for more on this subject

docs/konnectors.md
```diff
@@ -208,6 +208,114 @@ single thread or application. But NaCl is [no longer
...
+    rkt run --net=host --uuid-file-save=$PWD/uuid --volume data,kind=host,source=$PWD/container_dir --insecure-options=image nodeslim.aci --cpu=100m --memory=128M --name rktnode --mount volume=data,target=/usr/src/app --exec node -- /usr/src/app/$1 $2 &
+    # the container will handle itself the communication with the stack
+    sleep 60
+    rkt stop --force --uuid-file=uuid
+    rkt rm --uuid-file=uuid
+    rm -rf ./container_dir
+
+This script can be run like this :
+
+    ./rkt.sh mynewkonnector.js COZY_STACK_CREDENTIALS
```
nono: The credentials must be passed in env variables, not on the command-line. Else, they appear on `ps aux` and can be easily stolen by another user on the server.

docs/konnectors.md
```diff
@@ -208,6 +208,114 @@ single thread or application. But NaCl is [no longer
...
+    rm -rf ./container_dir
+
+This script can be run like this :
+
+    ./rkt.sh mynewkonnector.js
+
+If the mynewkonnector.js file is available in the container_dir_template directory.
+
+Pros : no breach regarding localhost network, more secure
+Cons: a file communication protocol with the stack need to be defined (any other idea ?), it may be more difficult to migrate the existing connectors.
```
nono: I'm not sure to understand this part. The cozy-syack has a REST API for files, and the konnectors will use it the same way they will use the REST API for data.

docs/konnectors.md
nono: We should also give the URL of the cozy instance to the konnector.

docs/konnectors.md
```diff
@@ -208,6 +208,114 @@ single thread or application. But NaCl is [no longer
...
+
+    ./rkt.sh mynewkonnector.js COZY_STACK_CREDENTIALS
+
+If the mynewkonnector.js file is available in the container_dir_template directory.
+
+Pros: the container can directly the stack to make its own updates, like before-> less work on migrating konnectors
+Cons : must forbid acces to port 5984 and 6060 + SMTP server
+
+For both solutions, the limitation of time, CPU and memory will avoid most DOS attacks (to my knowledge). For memory use, I still don't see a way to prevent the excessive use of swap from the container.
+To prevent the connectors from listening to each other, they should be run one by one in their own container and not at the same time.
```
nono: Does it mean that we can't parallelize the konnectors by running several containers at the same time? If yes, it's a very serious inconvenient.

docs/konnectors.md
doubleface: Yes, if they belong to the rkt group. I will add the information in the document

docs/konnectors.md
doubleface: Ok, I will add this information about alpine

docs/konnectors.md
doubleface: Ok, I was just trying to propose a solution to more completely isolate the konnectors. If the konnectors must use the REST-API, then the second solution is the only one. I will remove the first solution.

docs/konnectors.md
doubleface: Ok, will update the script
@doubleface

docs/konnectors.md
doubleface: No, we can absolutely run many containers at the same time. But in that case, they could spy other konnectors to get some credentials, except if the other connectors use https.

PR #312, opened 2017-02-22
Add link sharing creation

aenario:


pkg/apps/apps_test.go
```diff
@@ -5,8 +5,8 @@ import (
 
 	"github.com/cozy/cozy-stack/pkg/crypto"
 	"github.com/cozy/cozy-stack/pkg/instance"
-	jwt "github.com/dgrijalva/jwt-go"
 	"github.com/stretchr/testify/assert"
+	jwt "gopkg.in/dgrijalva/jwt-go.v3"
```
nono: Good catch!

pkg/permissions/permissions.go
```diff
@@ -1,28 +1,70 @@
...
+
+// Views are athe necessary views for this package
+// used in instance creation
+var Views = couchdb.Views{
+	shareByCView: couchdb.View{
+		Map: `
+function(doc){
+	if(doc.type === "share" && doc.codes)
+		Object.keys(doc.codes).forEach(function(k){ emit(doc.codes[k]) })
+}`,
```
nono: It's not possible to use a mango index here?

web/auth/auth.go
```diff
@@ -148,9 +148,8 @@ func login(c echo.Context) error {
 
 func logout(c echo.Context) error {
 	instance := middlewares.GetInstance(c)
-	claims, ok := c.Get("token_claims").(*permissions.Claims)
-	if !ok || claims.Audience != permissions.AppAudience {
-		return c.Redirect(http.StatusSeeOther, instance.SubDomain(apps.FilesSlug).String())
+	if !webpermissions.AllowLogout(c) {
+		return c.Redirect(http.StatusSeeOther, instance.SubDomain(apps.HomeSlug).String())
```
nono: It should be `FilesSlug`, not `HomeSlug`, until we have an `home` app

pkg/permissions/permissions.go
aenario: Impossible to iterate over the keys with mango, we would need a sharing doc / recipient

pkg/permissions/rule.go
```diff
@@ -99,12 +99,18 @@ func (r Rule) ValuesValid(o Validable) bool {
...
-	for _, v := range r.Values {
-		if v == value {
-			return true
+// ValuesContain returns true if all the values are in r.Values
+func (r Rule) ValuesContain(values ...string) bool {
+	for _, value := range values {
+		valueOK := false
+		for _, v := range r.Values {
+			if v == value {
+				valueOK = true
```
jinroh: nit: a little suboptimal... even if it is not critical for now, it may be worth to add a `TODO(optim)` since the number of values in a rule is a user input without a hard limit.

PR #322, opened 2017-02-24
Fix cozy-stack instances oauth-token --expire

nono:


aenario:
This will conflict with #312.

I applied the same change there (https://github.com/cozy/cozy-stack/pull/312/commits/638f503a72f4001d50408513cbd521781748a572). If we do mrge #312, It will be simpler to close this one. Otherwise, I will rebase #312 .

nono:
OK, let's do that in #312

PR #323, opened 2017-02-24
Fix login page

nono:
The checkmark after a successful login can be displayed as garbage if the loading of font awesome is too slow. I've added preloading to fix that.


assets/scripts/async-submit.js
```diff
@@ -35,7 +35,7 @@
     fetch('/auth/login', {
       method: 'POST',
       headers: headers,
-      body: `passphrase=${passphrase}&redirect=${redirect}`,
+      body: `passphrase=${passphrase}&redirect=${encodeURIComponent(redirect)}`,
```
jinroh: `encodeURIComponent` should also be used for `passphrase`

assets/scripts/async-submit.js
```diff
@@ -57,4 +57,7 @@
 
   passphraseInput.focus()
   submitButton.removeAttribute('disabled')
+
+  // Preload font awesome
+  try { document.fonts.load('14px FontAwesome') } catch (e) {}
```
gregorylegarec: Is supposed to fix issue on Linux environnement (at least @poupotte's one ?)

gregorylegarec:
Nice job @nono 👍

PR #324, opened 2017-02-24
Sharing request

Gara64:
Add a sharing request reception handler. 
It covers the case where a recipient received a mail with an URL containing the sharing informations in the query-string and clicked on this URL (the actual sending mail PR will come soon).

A sharing document is then created in the recipient database, which will be later used to ask the recipient if he accepts the sharing or not.

### DIFF
--- a/pkg/sharings/sharings.go
+++ b/pkg/sharings/sharings.go
@@ -106,14 +106,51 @@ func GetRecipient(db couchdb.Database, recID string) (*Recipient, error) {
 	return doc, err
 }
 
-// CheckSharingCreation initializes and check some sharing fields at creation
-func CheckSharingCreation(db couchdb.Database, sharing *Sharing) error {
-
-	sharingType := sharing.SharingType
+//CheckSharingType returns an error if the sharing type is incorrect
+func CheckSharingType(sharingType string) error {
 	if sharingType != consts.OneShotSharing &&
 		sharingType != consts.MasterSlaveSharing &&
 		sharingType != consts.MasterMasterSharing {
 		return ErrBadSharingType
+	}
+	return nil
+}
+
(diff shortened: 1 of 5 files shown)


web/sharings/sharings_test.go
```diff
@@ -93,6 +179,19 @@ func postJSON(u string, v echo.Map) (*http.Response, error) {
 	return http.Post(ts.URL+u, "application/json", bytes.NewReader(body))
 }
 
+func requestGET(v url.Values, u string) (*http.Response, error) {
+	reqURL := v.Encode()
+	fmt.Printf("url : %v\n", reqURL)
```
aenario: debug info ?

Gara64: Right, I removed

pkg/sharings/sharings.go
```diff
@@ -106,15 +106,53 @@ func GetRecipient(db couchdb.Database, recID string) (*Recipient, error) {
...
-
-	sharingType := sharing.SharingType
+//CheckSharingType returns an error if the sharing type is incorrect
+func CheckSharingType(sharingType string) error {
 	if sharingType != consts.OneShotSharing &&
 		sharingType != consts.MasterSlaveSharing &&
 		sharingType != consts.MasterMasterSharing {
 		return ErrBadSharingType
 	}
+	return nil
```
nono: FYI, it can also be written with a switch case, like this:

```go
switch sharingType {
case consts.OneShotSharing, consts.MasterSlaveSharing, consts.MasterMasterSharing:
    return nil
}
return ErrBadSharingType
```

Gara64: Thanks, I switched it ;)

PR #328, opened 2017-02-28
Client

jinroh:
Re-opening of the #220 PR.

The PR adds a client package containing a Go (OAuth) client. A basic testing of the package has been added.

It also makes the `cmd` package using this client implementation. Which allows to not have the CLI directly fetch data from CouchDB and only rely on the stack API.


client/auth/auth.go
```diff
@@ -0,0 +1,274 @@
...
+		c.SoftwareID = "github.com/cozy/cozy-stack"
+	}
+	if c.SoftwareVersion == "" {
+		c.SoftwareVersion = config.Version
+	}
+	if c.ClientName == "" {
+		c.ClientName = "Cozy Go client"
+	}
+	if c.ClientKind == "" {
+		c.ClientKind = "desktop"
```
nono: Maybe we an use a `command-line` kind

pkg/permissions/claims.go
```diff
@@ -26,20 +25,24 @@ const (
...
+	return time.Unix(claims.IssuedAt, 0)
+}
+
 // Expired returns true if a Claim is expired
 func (claims *Claims) Expired() bool {
-	now := crypto.Timestamp()
-	validUntil := claims.IssuedAt + TokenValidityDuration
-	return validUntil < now
+	validUntil := claims.IssuedAtTime().Add(TokenValidityDuration)
+	return validUntil.Before(time.Now())
```
nono: You are sure about the timezones? I'd have expected to see `.UTC()`

cmd/root.go
```diff
@@ -39,54 +42,59 @@ profiles you.`,
...
-	checkNoErr(viper.BindPFlag("subdomains", flags.Lookup("subdomains")))
-
-	flags.String("assets", "", "path to the directory with the assets (use the packed assets by default)")
-	checkNoErr(viper.BindPFlag("assets", flags.Lookup("assets")))
-
-	flags.String("admin-host", "localhost", "administration server host")
-	checkNoErr(viper.BindPFlag("admin.host", flags.Lookup("admin-host")))
-
-	flags.Int("admin-port", 6060, "administration server port")
-	checkNoErr(viper.BindPFlag("admin.port", flags.Lookup("admin-port")))
```
nono: We should add `admin-host` and `admin-port` options the `cozy-stack instances` command

pkg/permissions/claims.go
jinroh: I think we are OK since we compare `time.Time` of the same locations. I'll force the `UTC()` though to be clearer on the `IssuedAtTime()` method (renamed to `IssuedAtUTC()`)

client/auth/auth.go
jinroh: I defaulted to `unknown` if not defined. OK with that ?

PR #329, opened 2017-03-01
Konnectors installation description

doubleface:
Here is the konnector install and execution description.

### DIFF
--- a/docs/konnectors.md
+++ b/docs/konnectors.md
@@ -220,7 +220,7 @@ NsJail is:
 
 - easy to install : just a make away with standard build tools
 - offers a full list or isolation tools
-- lightly documented the only documentation is nsjail -h (also available in the main github page) and it is quite cryptic for a non-sysadmin         
+- lightly documented the only documentation is nsjail -h (also available in the main github page) and it is quite cryptic for a non-sysadmin
 like me. I could not find any help in any search engine. Some examples are available to run a back in an isolated process and work but I could not run a full nodejs (only nodejs -v worked)
 - The konnectors will need a full nodejs installed on the host
 - Is still actively maintained
@@ -244,7 +244,7 @@ I managed to run a nodejs container with just the following commands :
     rkt list   # to get the container uuid
     rkt export --app=nodeslim <uuid> nodeslim.aci
     rkt run --insecure-options=image --interactive nodeslim.aci -- -v  # to run node -v in the new container
-    
+
 Note: the --insecure-options param is to avoid the check of the image signature to ease the demonstration
 
 ### Choice
@@ -261,7 +261,7 @@ As stated before, rkt is easy to install. It may also be possible to make it ava
 #### Image creation
 
 To create an ACI file image, you just need to run a docker image one time :
-    
+
     rkt run  --uuid-file-save=$PWD/uuid --insecure-options=image --interactive docker://node:slim --name nodeslim -- -v
     rkt export --app=nodeslim `cat uuid` nodeslim.aci && rm uuid
 
@@ -299,7 +299,7 @@ To prevent the connectors from listening to each other, they should be run in co
 
 The container must be started in bridged mode. With that, the container still has access to localhost but through a specific IP address visible with ifconfig. That way, the host can have iptable rules to forbid access to specified ports to the bridge.
 
-To connect a container in bridge mode : 
+To connect a container in bridge mode :
 
 On the host create the file /etc/rkt/net.d/10-containers.conf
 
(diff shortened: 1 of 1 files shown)


docs/konnectors.md
```diff
@@ -320,18 +320,40 @@ On the host create the file /etc/rkt/net.d/10-containers.conf
 
 and run your container with the "--net=bridge" option. That way, a new interface is available in the container and gives you access to the host.
 
+## Konnector install and run details
+
+### Install
+
```
nono: When is installed a konnector? When the user says he/she wants to use it, or just before running it? In the first case, how it will work with several servers?

docs/konnectors.md
```diff
@@ -320,18 +320,40 @@ On the host create the file /etc/rkt/net.d/10-containers.conf
...
+The konnectors will be installed in the .cozy_konnectors directory (not in VFS ?) using npm install
```
nono: Do we use `yarn` or `npm` to install it? If `npm`, is it installed with the `-g` flag, or do we need to create a stub `package.json`?

docs/konnectors.md
```diff
@@ -320,18 +320,40 @@ On the host create the file /etc/rkt/net.d/10-containers.conf
...
+(with the ```--ignore-scripts``` option). This command will give the possibility to install konnectors
+from npm, github, gitlab, bitbucket, tar.gz url. The stack will have to check that the given package
+does not use a "file:" path.
```
nono: For a konnector, do we have one install for all users, one install per user? If one installs for all users, how do we differentiate the same konnector installed from two source? For example, if user `toto` asks to use the trainline konnector from npm, and user `titi` asks to install the tranline konnector too, but from git

docs/konnectors.md
```diff
@@ -320,18 +320,40 @@ On the host create the file /etc/rkt/net.d/10-containers.conf
...
+
+In the .cozy_konnectors directory, a ```npm ls -depth 0``` will be able to give the list of installed
+konnectors (npm outdated can also be used) or yarn equivalent.
```
nono: How are managed updates?

docs/konnectors.md
```diff
@@ -320,18 +320,40 @@ On the host create the file /etc/rkt/net.d/10-containers.conf
...
+
+### Details about running a konnector
+
```
nono: - How are passed the credentials for talking to the cozy-stack to the konnectors? And the cozy instance?
- How does the konnectors know which account to use?

docs/konnectors.md
jinroh: I would advise against using `npm` (or `yarn`) to install konnectors on production. During development it may be useful to use such dependency managements systems, however I think they would cause issues on our systems in production:

  - they make tons of I/O requests and are usually resource intensive
  - they do not provide a real reproducible build (even if `yarn` seems to be a little better than `npm` on this regard)
  - they do not provide enough security regarding the code that is actually downloaded (leftpadgate) (even if yarn is a also little better in this regard with the `yarn.lock` file)
  - requires a new dependency on quite complex programs, which mean we have to package the `cozy-stack` with a specific version `npm` or `yarn` if we want to stay dep free
  - we cannot provide a simple signing mechanism for our contributors to sign and verify the integrity of their applications
  - we are more coupled with the javascript ecosystem and it will be more difficult to allow a konnector to be written in another language in the future

I would argue that konnectors contributors should provide a versioned and packaged archive containing all the code to run their application.

However, I agree that without such tools, we have to develop a new hosting and versioning solution. Maybe our marketplace should provide these possibilities.

docs/konnectors.md
doubleface: I supposed it would be installed when the user says he wants to use it and each user would have its own installation of konnectors and I suppose it forces use to install it in VFS to make it work with several servers. 

But this raises the problem of running a konnector in VFS. Couldn't we copy the repository in a kind of sandbox directory which is not in VFS for each konnector run ?

docs/konnectors.md
doubleface: Using npm packages is mainly to get the openness with the many possibilities of hosting (github, gitlab, bitbucket, gist, etc.... and it also offers the versioning. It seems that yarn could be the response to the points you raise. And nodejs and yarn which would be used to get and update the packages would not need to be updated really often, it would not be it, which would run konnectors. We could even imagine using the rkt container to fetch the packages (then no need install nodejs with cozy-stack).

Especially for the konnectors, we need to ease the development and hosting to keep the very active community of konnectors developers. A github repo is still (to my view) the easiest way of hosting a konnector and it is compatible with yarn. But I suppose using only github repos as the only source of packages is still OK and would also allow the versioning as for the applications.

docs/konnectors.md
doubleface: See https://github.com/cozy/cozy-stack/pull/329#discussion_r103718920 which tries to address this question.

docs/konnectors.md
doubleface: Updates was supposed to be a reinstall of the same package, using npm or yarn or git does not change anything for this.

If you talk about update notification, I see that npm our yarn are not of any help if packages are git repositories or anything not an npm package. The only way would be to check version of each package.json for each source which will be a pain.

doubleface:
Updated the document, more details are given about how the konnector is run and I removed reference to npm or yarn for konnector installation since it has more cons than pros.

docs/konnectors.md
```diff
@@ -320,25 +320,56 @@ On the host create the file /etc/rkt/net.d/10-containers.conf
...
+To run a given konnector, the stack will copy this connector in a "run" directory, which is not in
+the VFS. This directory will be given to the rocket container as the current working directory
+with full read and write access on it. This is where the container will put its logs and any temp
+file needed. There will be also cozy-client.js and the shared libraries in a lib directory inside
+this directory.
+
+The konnector will be run with the following environment variables :
+
+- COZY_CREDENTIALS
```
nono: s/COZY_CREDENTIALS/COZY_TOKEN/

docs/konnectors.md
```diff
@@ -320,25 +320,56 @@ On the host create the file /etc/rkt/net.d/10-containers.conf
...
+possibility to upgrade the konnector to the latest version if any.
+
+To update a given konnector, a ```git pull``` command is run on the konnector.
+
+### Details about running a konnector
+
+To run a given konnector, the stack will copy this connector in a "run" directory, which is not in
+the VFS. This directory will be given to the rocket container as the current working directory
+with full read and write access on it. This is where the container will put its logs and any temp
+file needed. There will be also cozy-client.js and the shared libraries in a lib directory inside
```
nono: What are the shared libraries? Does it mean that we will have some nodejs dependencies that are chosen by the cozy team and installed on the servers, and the konnectors can use them? And if a konnector want to use another dependency, it will have to include it in its git repository?

doubleface: The shared libraries are the tools, specific to the konnectors to help developer to write its connector. It corresponds to the actual content of the [lib](https://github.com/cozy-labs/konnectors/tree/master/server/lib) directory. Anyway, it is mandatory to allow easy migration of existing konnectors at least.

For the second part, yes, the konnectors will have to contain their own dependencies (npm packages, etc) in their git repository.

nono: The node modules are tons of small files. The VFS is bad for managing this workload. We should test, but I fear that even copying a konnector with some node modules from the VFS to a server can take a lot longer than executing it.

doubleface: I suppose we will test it, but another possibility is, like the current implementation of konnectors, propose a predefined list of node_modules which will be the only ones usable by the konnectors. But konnector developers will have to make a pullrequest to cozy-stack to add a dependency. Or something between these two extreme propositions.

docs/konnectors.md
```diff
@@ -320,25 +320,57 @@ On the host create the file /etc/rkt/net.d/10-containers.conf
...
+directory](https://github.com/cozy-labs/konnectors/tree/master/server/lib).
+
+The konnector will be run with the following environment variables :
+
+- COZY_TOKEN
+- COZY_URL : to know what instance is running the konnector
+- KONNECTOR_CREDENTIALS : as a json string with all the values from the account associated to the
+konnector.
+
+In the end of the konnector execution (or timeout), the logs are read in the log.txt file and added
```
aenario: If rocket has something aking to docker logs, it might be easier to just use the process stdout.

doubleface: There is no rkt logs but it is possible to access logs directly using journalctl. And it is also possible to get logs from stdout.

PR #330, opened 2017-03-01
Apps cli

jinroh:
Based on #328 

This PR adds new CLI for application installation and update:
  - `cozy-stack apps install --domain localhost:8080 "slug" "git://github.com/...`
  - `cozy-stack apps update --domain localhost:8080 "slug"`

It also makes some changes to the `/apps/` API:
  - separate install (`POST`) from update (`PUT`)
  - adds the possibility for these endpoints to be synchronous (via `Accept: text/event-stream`)

### DIFF
--- a/pkg/apps/installer_test.go
+++ b/pkg/apps/installer_test.go
@@ -131,14 +131,6 @@ func TestInstallBadSlug(t *testing.T) {
 
 func TestInstallBadAppsSource(t *testing.T) {
 	_, err := NewInstaller(c, &InstallerOptions{
-		Slug:      "app2",
-		SourceURL: "",
-	})
-	if assert.Error(t, err) {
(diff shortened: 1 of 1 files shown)


docs/apps.md
```diff
@@ -243,6 +243,55 @@ like this:
...
+  }]
+}
+```
+
+#### Status codes
+
+* 202 Accepted, when the application installation has been accepted.
+* 400 Bad-Request, when the manifest of the application could not be processed (for instance, it is not valid JSON).
+* 404 Not Found, when the application with the specified slug was not found or when the manifest or the source of the application is not reachable.
+* 422 Unprocessable Entity, when the sent data is invalid (for example, the slug is invalid or the Source parameter is not a proper or supported url)
```
nono: I think we have no `Source` parameter for update

PR #331, opened 2017-03-01
Implements @event trigger for jobs with a realtime hub.

aenario:


pkg/jobs/trigger_event.go
```diff
@@ -0,0 +1,54 @@
...
+		c := realtime.MainHub().Subscribe(t.infos.Arguments)
+		for {
+			select {
+			case <-c.Read():
+				ch <- &JobRequest{
+					WorkerType: t.infos.WorkerType,
+					Message:    t.infos.Message,
+					Options:    t.infos.Options,
+				}
+			case <-t.unscheduled:
```
jinroh: `ch` should probably be closed here

pkg/jobs/trigger_event.go
```diff
@@ -0,0 +1,54 @@
...
+				return
+			}
+		}
+	}()
+	return ch
+}
+
+// Unschedule implements the Unschedule method of the Trigger interface.
+func (t *EventTrigger) Unschedule() {
+	go func() { t.unscheduled <- struct{}{} }()
```
jinroh: using `close(t.unscheduled)` should be enough. it is not blocking.

pkg/realtime/memhub.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+type instancehub struct {
+	instance string
+	mainHub  *hub
+}
+
+func (ih *instancehub) Publish(e *Event) {
+	ih.mainHub.Lock()
+	defer ih.mainHub.Unlock()
```
jinroh: I would add `getInstanceTopics/getGlobalTopics` on the `hub` struct and handle the mutex (un)locking from there. Also it looks like we could maybe benefit from a `RWMutex` (since Publish is read-only) and maybe one mutex per map.

PR #337, opened 2017-03-03
Send emails after the creation of a sharing

Ljinod:
This PR introduces the following behavior: when the user creates a new
sharing, "invitation emails" are automatically sent to the recipients.

To send the emails a job is queued for each recipient.

This email contains a button that, once clicked, will trigger the next
step of the sharing process.

A route (PUT /sharings/:id/sendMails) was also introduced in case the
email could not be sent to one or more of the recipients. In addititon
the status of the recipients for whom the email was not sent are changed
to "error".

To allow all this, it was necessary to modify the structure of the
former `sharingRecipient` so that one could access the underlying
`Recipient`. As a consequence, and to add clarity, it was renamed
`RecipientStatus`.


pkg/consts/consts.go
```diff
@@ -69,10 +69,12 @@ const (
...
-	// RefusedSharingStatus is the sharing refused status
-	RefusedSharingStatus = "refused"
-	// AcceptedSharingStatus is the sharing accepted status
-	AcceptedSharingStatus = "accepted"
+	// PendingStatus is the sharing pending status
+	PendingStatus = "pending"
+	// RefusedStatus is the sharing refused status
+	RefusedStatus = "refused"
+	// AcceptedStatus is the sharing accepted status
+	AcceptedStatus = "accepted"
```
Gara64: Be careful, this is a regression: I changed the *Status by *SharingStatus in #302.

pkg/sharings/send_mails.go
```diff
@@ -0,0 +1,222 @@
...
+// filled through the `mailTemplateValues` structure.
+const (
+	mailTemplateEn = `
+        <h2>Hey {{.RecipientName}}!</h2>
+        <p>{{.SharerPublicName}} wants to share a document with you! You will only be able to view it.</p>
+
+        <p>The description given is: {{.Description}}.</p>
+
+        <form action="{{.OAuthQueryString}}">
+            <input type="submit" value="See this document" />
```
Gara64: We discussed to change this description by something like "Click here to accept this sharing", instead of "See this document", which is confusing.

pkg/sharings/send_mails.go
```diff
@@ -0,0 +1,222 @@
...
+    `
+
+	mailTemplateFr = `
+        <h2>Bonjour {{.RecipientName}} !</h2>
+        <p>{{.SharerPublicName}} veut partager un document avec vous ! Vous pourrez seulement le consulter.</p>
+
+        <p>La description associée est : {{.Description}}.</p>
+
+        <form action="{{.OAuthQueryString}}">
+            <input type="submit" value="Voir le document" />
```
Gara64: Même remarque que le commentaire précédent

pkg/sharings/send_mails.go
```diff
@@ -0,0 +1,222 @@
...
+        </form>
+    `
+)
+
+// The sharing-dependent information: the recipient's name, the sharer's public
```
Gara64: dependent -> dependant

pkg/sharings/send_mails.go
```diff
@@ -0,0 +1,222 @@
...
+
+	errorOccurred := false
+	for _, rs := range recipientsStatus {
+		recipient := rs.recipient
+
+		// Generate recipient specific OAuth query string.
+		recipientOAuthQueryString, err := generateOAuthQueryString(s, recipient)
+		if err != nil {
+			errorOccurred = logErrorAndSetRecipientStatus(rs, err)
+			break
```
Gara64: Shouldn't be `continue` instead of `break`?

pkg/sharings/send_mails.go
```diff
@@ -0,0 +1,222 @@
...
+		}
+
+		// Augment base values with recipient specific information.
+		mailValues.RecipientName = recipient.Email
+		mailValues.OAuthQueryString = recipientOAuthQueryString
+
+		sharingMessage, err := generateMailMessage(s, recipient, mailValues)
+		if err != nil {
+			errorOccurred = logErrorAndSetRecipientStatus(rs, err)
+			break
```
Gara64: Same question

pkg/sharings/send_mails.go
```diff
@@ -0,0 +1,222 @@
...
+		// We ask to queue a new mail job.
+		// XXX Do we have to do something with the returned values?
+		_, _, err = instance.JobsBroker().PushJob(&jobs.JobRequest{
+			WorkerType: "sendmail",
+			Options:    nil,
+			Message:    sharingMessage,
+		})
+		if err != nil {
+			errorOccurred = logErrorAndSetRecipientStatus(rs, err)
+			break
```
Gara64: Same question :)

pkg/sharings/send_mails.go
```diff
@@ -0,0 +1,222 @@
...
+	log.Error(`[Sharing] An error occurred while trying to send
+        the email invitation`, err)
+
+	rs.Status = consts.ErrorStatus
+	return true
+}
+
+// generatemailMessage will extract and compute the relevant information
+// from the sharing to generate the mail we will send to the recipient
+// specified.
```
Gara64: generatemailMessage -> generateMailMessage
recipient specified -> specified recipient

pkg/sharings/send_mails.go
```diff
@@ -0,0 +1,222 @@
...
+	mailPartFr := workers.MailPart{
+		Type: "text/html",
+		Body: mailTemplateFr}
+
+	mailParts := []*workers.MailPart{&mailPartEn, &mailPartFr}
+
+	// The address of the recipient.
+	// XXX In addition to testing if the email was provided, do we test if it
+	// actually corresponds to an email or do we "wait" for the error triggered
+	// by the mail job?
```
Gara64: ping @nono for the question in the comment

pkg/sharings/send_mails.go
```diff
@@ -0,0 +1,222 @@
...
+		mailValues.OAuthQueryString = recipientOAuthQueryString
+
+		sharingMessage, err := generateMailMessage(s, recipient, mailValues)
+		if err != nil {
+			errorOccurred = logErrorAndSetRecipientStatus(rs, err)
+			break
+		}
+
+		// We ask to queue a new mail job.
+		// XXX Do we have to do something with the returned values?
```
Gara64: ping @nono for the question in comment.

pkg/sharings/send_mails.go
Ljinod: Yup totally, my very bad

pkg/sharings/send_mails.go
nono: @jinroh knows this part better than me

jinroh: the mail sender is asynchronous so testing that the mail has been send can take an unknown amount of time.

if the mail are not "reliables", we should test their well formatting here. although the formatting of an email does of course not tell if the mail address is actually a good one.

pkg/sharings/send_mails.go
jinroh: Not necessarily.

The first argument contains the initial job informations built synchronously and the second one is a channel on which you can listen for the advanced of the job.

PR #341, opened 2017-03-06
Add route to list permissions for docs by IDs

aenario:


aenario:
Let's merge as-is... More complex permissions (for albums) will come in another PR.

docs/permissions.md
```diff
@@ -418,3 +418,46 @@ Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3O
 ```http
 HTTP/1.1 204 No Content
 ```
+
+### POST /permissions/exists
```
nono: Why a `POST` and not a `GET`?

aenario: The list of IDs we want to fetch the sharing status for can be quite long, so we have to pass it in the body. GET request cannot have URLS...

aenario:
@jinroh ?

PR #342, opened 2017-03-06
client: use https by default and fallback on http if possible

jinroh:


client/request/request.go
```diff
@@ -89,10 +95,7 @@ func (b *BearerAuthorizer) AuthHeader() string {
 // Req performs a request with the specified request options.
 func Req(opts *Options) (*http.Response, error) {
 	var scheme string
-	if opts.DisableSecure {
-		if !config.IsDevRelease() {
-			panic("Should not disable HTTPs")
-		}
+	if atomic.LoadInt32(&httpFallback) == 1 {
```
aenario: [curiosity] Why do we need an atomic here? Which cases does it prevent against?

jinroh: This is just to be extra-careful to keep the `Req` method thread-safe, since the `httpFallback` var is global.

PR #343, opened 2017-03-06
Jobs permissions

nono:


web/jobs/jobs.go
```diff
@@ -153,6 +171,11 @@ func newTrigger(c echo.Context) error {
 	if err != nil {
 		return wrapJobsError(err)
 	}
+
+	if err = permissions.Allow(c, permissions.GET, t); err != nil {
```
aenario: why GET ?

nono: You know, I let some errors so that people reviewing my code are not bored :wink:

PR #350, opened 2017-03-07
Apps: uninstall

jinroh:
This PR adds the possibility to remove/uninstall an application from the HTTP API and CLI.

Tests are on there way.

### DIFF
--- a/pkg/apps/installer.go
+++ b/pkg/apps/installer.go
@@ -121,6 +121,23 @@ func (i *Installer) Update() {
 	return
 }
 
+// Delete will remove the application linked to the installer.
+func (i *Installer) Delete() (*Manifest, error) {
+	if i.man == nil {
+		return nil, ErrNotFound
+	}
+	if state := i.man.State; state != Ready && state != Errored {
+		return nil, ErrBadState
+	}
+	if err := deleteManifest(i.ctx, i.man); err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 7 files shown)


pkg/apps/installer.go
```diff
@@ -121,6 +121,23 @@ func (i *Installer) Update() {
...
+	if state := i.man.State; state != Ready && state != Errored {
+		return nil, ErrBadState
+	}
+	if err := deleteManifest(i.ctx, i.man); err != nil {
+		return nil, err
+	}
+	if err := vfs.RemoveAll(i.ctx, i.appDir()); err != nil {
+		return nil, err
+	}
+	return i.man, nil
```
nono: IIRC, we said that we have to remove some permissions when uninstalling an app. Probably the reverse of `permissions.CreateAppSet`

jinroh: Looks like the `permissions.DestroyApp` method (called in `deleteManifest`) does the work. can you confirm @aenario ?

aenario: Yes, I changed it when introducing sub-permissions.

pkg/apps/installer.go
```diff
@@ -121,6 +121,23 @@ func (i *Installer) Update() {
 	return
 }
 
+// Delete will remove the application linked to the installer.
+func (i *Installer) Delete() (*Manifest, error) {
```
aenario: [~nitpick~ troll] Installer is now a weird name for this class...

PR #351, opened 2017-03-07
Install apps on instances add

nono:


pkg/instance/instance.go
```diff
@@ -558,6 +589,24 @@ func Destroy(domain string) (*Instance, error) {
 	return i, nil
 }
 
+// BuildAppToken is used to build a token to identify the app for requests made
+// to the stack
+func (i *Instance) BuildAppToken(m *apps.Manifest) string {
```
aenario: [factorisation] we could use the MakeJWT function from down this file.

PR #352, opened 2017-03-07
Add sharing authorization

Gara64:
When a recipient receive a sharing request throught an URL sent by mail, he is redirected to the OAuth authorize page, based on the information specified in the query string. 
This is done in addition to the sharing document creation.

### DIFF
--- a/web/sharings/sharings.go
+++ b/web/sharings/sharings.go
@@ -2,36 +2,14 @@ package sharings
 
 import (
 	"net/http"
-	"net/url"
 
 	"github.com/cozy/cozy-stack/pkg/consts"
 	"github.com/cozy/cozy-stack/pkg/couchdb"
-	"github.com/cozy/cozy-stack/pkg/instance"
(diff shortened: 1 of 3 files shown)


web/sharings/sharings.go
```diff
@@ -2,17 +2,39 @@ package sharings
...
+
+	values := url.Values{
+		"state":         {state},
+		"client_id":     {clientID},
+		"redirect_uri":  {redirectURI},
+		"scope":         {scope},
+		"response_type": {responseType},
+	}
+
+	redirectAuthorize := instance.PageURL("/auth/authorize", values)
```
nono: You can shorten this function to:

```go
redirectAuthorize := instance.PageURL("/auth/authorize", c.QueryParams())
return c.Redirect(http.StatusSeeOther, redirectAuthorize)
```

Gara64: Right, this makes the function pointless. I removed it.

PR #359, opened 2017-03-08
Cross platform err checking for connection refused error

jinroh:


client/request/request.go
```diff
@@ -152,11 +154,23 @@ func Req(opts *Options) (*http.Response, error) {
...
+			switch t := err.(type) {
+			case *net.OpError:
+				if t.Op == "read" {
+					isConnRefused = true
+				}
+			case syscall.Errno:
+				if t == syscall.ECONNREFUSED {
+					isConnRefused = true
+				}
+			}
```
nono: It looks complicated and I'm not sure that we are not missing some cases. For example, if we connect in https to the admin port (http), I think we will another error to say that tle TLS handshake has failed.

PR #360, opened 2017-03-08
Move indexes and views declarations in const package

jinroh:
This PR is a little refactoring to store all the views and indexes declarations into the `consts` package. It should offer a clearer view of all the indexes and views that are defined by the stack.

The `ExecuteView` method takes a `View` pointer which make it also less error-prone.

The `instance.Create` function has been inlined to so that it is easier to see all the creation steps it goes though.


jinroh:
rebased on master

PR #366, opened 2017-03-09
Konnectors installation on VFS study

doubleface:
Here is a study about different means of installation of konnectors on the VFS without putting too many files on it.


nono:
Looks good to me. Can you rebase this PR to fix the conflict?

doubleface:
@nono Done

PR #372, opened 2017-03-09
Handle refused sharing answer

Gara64:
This handles a refused sharing answer on the sharer side to eventually update the recipient sharing status.
For this end, a mango index is used to retrieve the sharing document from the state parameter, and then the recipient from the client_id parameter.

### DIFF
--- a/web/sharings/sharings.go
+++ b/web/sharings/sharings.go
@@ -15,7 +15,6 @@ import (
 func SharingAnswer(c echo.Context) error {
 
 	var err error
-	var sharingAccepted bool
 
 	state := c.FormValue("state")
 	clientID := c.FormValue("client_id")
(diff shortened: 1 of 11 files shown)


web/sharings/sharings.go
```diff
@@ -11,6 +11,40 @@ import (
...
+	accessCode := c.FormValue("access_code")
+
+	instance := middlewares.GetInstance(c)
+
+	// The sharing is refused if there is no access code or scope
+	if scope == "" || accessCode == "" {
+		sharingAccepted = false
+	} else {
+		sharingAccepted = true
+	}
```
nono: You can shorten this block to:

```go
sharingAccepted = scope != "" && accessCode != ""
```

Gara64: Yes, I used this notation first but I wasn't sure the shortness wouldn't harm the readability, but maybe it is just because I'm not quite used to it

PR #375, opened 2017-03-10
Realtime fix

jinroh:
Fixes on the realtime package:

  - goroutine leak in the topic loop (not released)
  - asynchrounous send with fixed size buffer for `subscribe`/`unsubscribe` channels (instead of goroutines)
  - close/send race on the `send` channel, guarded with an atomic state

### DIFF
--- a/pkg/realtime/memsub.go
+++ b/pkg/realtime/memsub.go
@@ -1,6 +1,9 @@
 package realtime
 
-import "sync/atomic"
+import (
+	"errors"
+	"sync/atomic"
+)
 
 // Subscription to multiple hub channels.
 type sub struct {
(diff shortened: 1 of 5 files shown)


pkg/realtime/memsub.go
```diff
@@ -1,16 +1,17 @@
 package realtime
 
+import "sync/atomic"
+
 // Subscription to multiple hub channels.
 type sub struct {
 	topics []*topic
-	closed chan bool
 	send   chan *Event
+	c      uint32 // mark whether or not the sub is closed
```
nono: A `bool` is not enough?

jinroh: atomic does not have operations on bool. i think it is pretty common in go to use a uint32 for that purpose (see [once](https://github.com/golang/go/blob/964639cc338db650ccadeafb7424bc8ebb2c0f6c/src/sync/once.go) or [exec](https://github.com/golang/go/blob/964639cc338db650ccadeafb7424bc8ebb2c0f6c/src/os/exec.go#L29-L35)).

PR #381, opened 2017-03-15
Add realtime to couchdb package

aenario:
Turns out we dont currently support Couchdb -> Stack replication, so we dont have couchdb.Proxy routes that write data...

### DIFF
--- a/pkg/couchdb/couchdb.go
+++ b/pkg/couchdb/couchdb.go
@@ -345,40 +345,25 @@ func ResetDB(db Database, doctype string) error {
 	return CreateDB(db, doctype)
 }
 
-// Delete destroy a document by its doctype and ID .
+// DeleteDoc deletes a struct implementing the couchb.Doc interface
 // If the document's current rev does not match the one passed,
 // a CouchdbError(409 conflict) will be returned.
-// This functions returns the tombstone revision as string
-func Delete(db Database, doctype, id, rev string) (string, error) {
-	var err error
-	id, err = validateDocID(id)
-	if err != nil {
-		return "", err
-	}
-	var res updateResponse
-	qs := url.Values{"rev": []string{rev}}
-	url := docURL(db, doctype, id) + "?" + qs.Encode()
-	err = makeRequest("DELETE", url, nil, &res)
-	if err != nil {
-		return "", fixErrorNoDatabaseIsWrongDoctype(err)
-	}
-	rtevent(db, realtime.EventDelete, doctype, id, res.Rev)
-	return res.Rev, nil
-}
-
-// DeleteDoc deletes a struct implementing the couchb.Doc interface
 // The document's SetRev will be called with tombstone revision
 func DeleteDoc(db Database, doc Doc) error {
 	id, err := validateDocID(doc.ID())
 	if err != nil {
 		return err
 	}
-	tombrev, err := Delete(db, doc.DocType(), id, doc.Rev())
+
+	var res updateResponse
+	qs := url.Values{"rev": []string{doc.Rev()}}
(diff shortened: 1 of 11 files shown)


pkg/couchdb/couchdb.go
```diff
@@ -47,6 +48,15 @@ func SimpleDatabasePrefix(prefix string) Database {
 	return &simpleDB{prefix}
 }
 
+func rtevent(db Database, evtype, doctype, id, rev string) {
+	realtime.InstanceHub(db.Prefix()).Publish(&realtime.Event{
+		Type:    evtype,
+		DocType: doctype,
+		DocID:   id,
+		DocRev:  rev,
+	})
```
nono: The whole document is not sent to the realtime package? How `include_docs` of [the websocket API](https://cozy.github.io/cozy-stack/realtime.html) will work?

pkg/jobs/trigger_event.go
```diff
@@ -2,21 +2,30 @@ package jobs
 
 import (
 	"github.com/cozy/cozy-stack/pkg/consts"
+	"github.com/cozy/cozy-stack/pkg/permissions"
 	"github.com/cozy/cozy-stack/pkg/realtime"
+	"github.com/labstack/gommon/log"
```
nono: We are using `github.com/Sirupsen/logrus` for logs, not `github.com/labstack/gommon/log`.

I've added a check here: https://github.com/cozy/cozy-stack/blob/master/scripts/integration.sh#L18 but it missed it...

pkg/jobs/trigger_event.go
```diff
@@ -44,18 +53,55 @@ func (t *EventTrigger) Valid(key, value string) bool {
...
-			case <-c.Read():
-				ch <- &JobRequest{
-					WorkerType: t.infos.WorkerType,
-					Message:    t.infos.Message,
-					Options:    t.infos.Options,
+			case e := <-c.Read():
+				if t.interestedBy(e) {
+					msg, err := addEventToMessage(e, t.infos.Message)
+					if err != nil {
+						log.Error(err)
```
nono: You still continue even if there was an error?

pkg/permissions/verbs.go
```diff
@@ -105,6 +105,8 @@ func (vs *VerbSet) Merge(verbs *VerbSet) {
 }
 
 // VerbSplit parse a string into a VerbSet
+// Note : this does not check if Verbs are proper HTTP Verbs
```
nono: No space before `:` in english

pkg/jobs/trigger_event.go
```diff
@@ -44,18 +53,56 @@ func (t *EventTrigger) Valid(key, value string) bool {
...
+	if v, ok := e.Doc.(permissions.Validable); ok {
+		return t.mask.ValuesValid(v)
+	}
+
+	return false
+}
+
+func addEventToMessage(e *realtime.Event, base *Message) (*Message, error) {
+	var basemsg interface{}
+	base.Unmarshal(&basemsg)
```
jinroh: Couldn't we use the `Data` field directly instead of unmarshalling/marshalling the message ? Like so:

```go
return NewMessage(JSONEncoding, map[string]interface{}{
  "message": json.RawMessage(base.Data),
  "event":   e,
})
```

jinroh: Oh.. that would work only if the base message if of Type `json`, well...

aenario: I agree this part feel hackish.

I also thought about adding a TriggerMessage to the whole JobRequest / JobInfos / Job / WorkerFunc chain, but it seemed too complicated.

PR #387, opened 2017-03-16
Sharing recipient creation 

Gara64:
This allows to create a sharing recipient throught a `POST sharings/recipient` route. 

It triggers the creation of the recipient in the sharer database, and the OAuth registration of the sharer to the Cozy's recipient.
Note that the` public_name` option in the instance settings has been added, as it is needed for the OAuth `client_name` parameter.

### DIFF
--- a/pkg/instance/instance.go
+++ b/pkg/instance/instance.go
@@ -94,12 +94,13 @@ type Instance struct {
 
 // Options holds the parameters to create a new instance.
 type Options struct {
-	Domain   string
-	Locale   string
-	Timezone string
-	Email    string
-	Apps     []string
-	Dev      bool
+	Domain     string
+	Locale     string
+	Timezone   string
+	Email      string
+	PublicName string
+	Apps       []string
+	Dev        bool
 }
 
 // DocType implements couchdb.Doc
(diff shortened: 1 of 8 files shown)


pkg/instance/instance.go
```diff
@@ -134,8 +135,9 @@ func (i *Instance) Included() []jsonapi.Object {
 
 // settings is a struct used for the settings of an instance
 type instanceSettings struct {
-	Timezone string `json:"tz,omitempty"`
-	Email    string `json:"email,omitempty"`
+	Timezone   string `json:"tz,omitempty"`
+	Email      string `json:"email,omitempty"`
+	PublicName string `json:"public_name,omitempty"`
```
jinroh: Would be interesting to add the possibility to specify this public name in the:
  - `/instance` API (defined in `web/instance` via a `?PublicName` query param)
  - in the client (`client/instances#CreateInstance` via a `PublicName` field in the `InstanceOptions` struct)
  - in the CLI (`cmd/instances#addInstanceCmd`, via a `--public-name` flag for instance)

Gara64: Agreed, but I think it should be another PR to avoid to make this one too big.

jinroh: @Gara64 no problem

web/sharings/sharings_test.go
```diff
@@ -187,21 +210,29 @@ func TestCreateSharingSuccess(t *testing.T) {
...
 	}
-
 	clientOAuth, _ = setup.GetTestClient("")
 	clientID = clientOAuth.ClientID
 
 	ts = setup.GetTestServer("/sharings", Routes)
+	ts2 = setup2.GetTestServer("/auth", auth.Routes)
+
 	os.Exit(setup.Run())
+	os.Exit(setup2.Run())
```
aenario: setup2 will never run, (os.Exit closes the process), so the second Instance will not be cleaned up.
It is a bit ugly, but you can add a public Cleanup to test_utils
```
func (c *TestSetup) Cleanup() {
	c.cleanup()
}
```

and then link the two testSetup
```
setup.AddCleanup(setup2.Cleanup)
```

PR #389, opened 2017-03-16
Set http prefix to url in sharing mail

Ljinod:
If the url specified in the recipient's structure is not prefixed by
either "http" or "https" the button generated in the invitation email
cannot be correctly handled.

This PR fixes this problem by prefixing it by the appropriate value,
depending on `instance.Scheme()`. If the url is already prefixed then
nothing is changed.

### DIFF
--- a/pkg/sharings/send_mails.go
+++ b/pkg/sharings/send_mails.go
@@ -1,8 +1,8 @@
 package sharings
 
 import (
-	"fmt"
 	"net/url"
+	"strings"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/cozy/cozy-stack/pkg/consts"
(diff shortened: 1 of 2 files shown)


pkg/sharings/send_mails.go
```diff
@@ -154,20 +154,42 @@ func generateOAuthQueryString(s *Sharing, r *Recipient) (string, error) {
...
+	baseURL := r.URL + "/sharings/request"
+	// The link/button we put in the email has to have an http:// or https://
+	// prefix, otherwise it cannot be open in the browser.
+	baseURL = setHTTPPrefixToURL(baseURL, scheme)
+
+	oAuthQuery, err := url.Parse(baseURL)
+	if err != nil {
+		return "", err
+	}
+	// We use url.encode to safely escape the query parameters.
```
aenario: If you wait until after url.Parse, you dont need to manipulate the url string as much.

if(oAuthQUery.Scheme != "http" && oauthQuery.Scheme != "https") { oauthQuery.Scheme = scheme }

Ljinod: Indeed and it is much better! I changed it.

PR #390, opened 2017-03-17
[VFS] Pagination

jinroh:
This PR adds the pagination possibilities to the VFS directories children.

### DIFF
--- a/docs/files.md
+++ b/docs/files.md
@@ -105,8 +105,8 @@ Contents is paginated. By default, only the 30 first entries are given.
 
 Parameter    | Description
 -------------|--------------------------------------
-Page[cursor] | the last id of the results
-Page[limit]  | the number of entries (30 by default)
+page[cursor] | the last id of the results
+page[limit]  | the number of entries (30 by default)
 
 #### Request
 
@@ -125,7 +125,7 @@ Content-Type: application/vnd.api+json
 ```json
 {
   "links": {
-    "next": "/files/fce1a6c0-dfc5-11e5-8d1a-1f854d4aaf81?Page[cursor]=9152d568-7e7c-11e6-a377-37cbfb190b4b"
+    "next": "/files/fce1a6c0-dfc5-11e5-8d1a-1f854d4aaf81?page[cursor]=9152d568-7e7c-11e6-a377-37cbfb190b4b"
   },
   "data": {
     "type": "io.cozy.files",
@@ -709,8 +709,8 @@ List the files inside the trash. It's paginated.
 
 Parameter    | Description
 -------------|---------------------------------------
-Page[cursor] | the last id of the results
-Page[limit]  | the number of entries (100 by default)
+page[cursor] | the last id of the results
+page[limit]  | the number of entries (100 by default)
 
 #### Request
 
--- a/docs/files.md
+++ b/docs/files.md
@@ -99,14 +99,14 @@ Location: http://cozy.example.com/files/6494e0ac-dfcb-11e5-88c1-472e84a9cbee
 ### GET /files/:file-id
 
 Get a directory or a file informations. In the case of a directory, it contains the list of files and sub-directories inside it.
-Contents is paginated. By default, only the 100 first entries are given.
+Contents is paginated. By default, only the 30 first entries are given.
 
 ### Query-String
 
 Parameter    | Description
--------------|---------------------------------------
-page[cursor] | the last id of the results
-page[limit]  | the number of entries (100 by default)
+-------------|--------------------------------------
+Page[cursor] | the last id of the results
+Page[limit]  | the number of entries (30 by default)
 
 #### Request
 
@@ -125,7 +125,7 @@ Content-Type: application/vnd.api+json
 ```json
 {
   "links": {
-    "next": "/files/fce1a6c0-dfc5-11e5-8d1a-1f854d4aaf81?page[cursor]=9152d568-7e7c-11e6-a377-37cbfb190b4b"
+    "next": "/files/fce1a6c0-dfc5-11e5-8d1a-1f854d4aaf81?Page[cursor]=9152d568-7e7c-11e6-a377-37cbfb190b4b"
   },
   "data": {
     "type": "io.cozy.files",
@@ -709,8 +709,8 @@ List the files inside the trash. It's paginated.
 
 Parameter    | Description
 -------------|---------------------------------------
-page[cursor] | the last id of the results
-page[limit]  | the number of entries (100 by default)
+Page[cursor] | the last id of the results
+Page[limit]  | the number of entries (100 by default)
 
 #### Request
 
--- a/docs/files.md
+++ b/docs/files.md
@@ -708,9 +708,9 @@ List the files inside the trash. It's paginated.
 ### Query-String
 
 Parameter    | Description
--------------|---------------------------------------
+-------------|--------------------------------------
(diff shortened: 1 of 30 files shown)


docs/files.md
```diff
@@ -99,14 +99,14 @@ Location: http://cozy.example.com/files/6494e0ac-dfcb-11e5-88c1-472e84a9cbee
...
 
 ### Query-String
 
 Parameter    | Description
--------------|---------------------------------------
-page[cursor] | the last id of the results
-page[limit]  | the number of entries (100 by default)
+-------------|--------------------------------------
+Page[cursor] | the last id of the results
+Page[limit]  | the number of entries (30 by default)
```
nono: You don't want to respect JSON-API? Cf http://jsonapi.org/format/#fetching-pagination

pkg/couchdb/mango/query.go
```diff
@@ -152,13 +152,27 @@ func Between(field string, v1 interface{}, v2 interface{}) Filter {
 	}}
 }
 
+// Inside returns a filter that check if v1 < field < v2
+func Inside(field string, v1 interface{}, v2 interface{}) Filter {
```
nono: I'd have called this function `Between`

docs/files.md
```diff
@@ -709,8 +709,8 @@ List the files inside the trash. It's paginated.
 
 Parameter    | Description
 -------------|---------------------------------------
-page[cursor] | the last id of the results
-page[limit]  | the number of entries (100 by default)
+Page[cursor] | the last id of the results
+Page[limit]  | the number of entries (100 by default)
```
nono: Maybe s/100/30/ like above

pkg/vfs/iter.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/cozy/cozy-stack/pkg/couchdb/mango"
+)
+
+// ErrIteratorDone is returned by the Next() merthod of the iterator when
+// the iterator is actually done.
+var ErrIteratorDone = errors.New("No more element in the iterator")
+
+// IteratorDefaultFetchSize is the default number of elements fetched from
+// couchdb on each iteration.
+const IteratorDefaultFetchSize = 100
```
nono: Shouldn't it be 30 like in doc?

pkg/vfs/iter.go
```diff
@@ -0,0 +1,100 @@
...
+
+import (
+	"errors"
+
+	"github.com/cozy/cozy-stack/pkg/consts"
+	"github.com/cozy/cozy-stack/pkg/couchdb"
+	"github.com/cozy/cozy-stack/pkg/couchdb/mango"
+)
+
+// ErrIteratorDone is returned by the Next() merthod of the iterator when
```
nono: s/merthod/method/

pkg/couchdb/mango/query.go
jinroh: Between already exist and is `v1 < fields <= v2`...

pkg/vfs/iter.go
jinroh: This is the number of elements fetched from the database on each roundtrip of the iterator. The iterator is not only used by the pagination of the `web/files` package, it is sometime used to iterate over all the elements of the directory.

In the paginated case, where we potentially stop iterating before `ErrIteratorDone`, the `ByFetch` is less than this default and the pagination is only one roundtrip with `couchdb`. See [here](https://github.com/jinroh/cozy-stack/blob/040664d4d210d509751fabc9d058bdb0b122f86c/web/files/paginated.go#L46-L51).

jinroh:
I've just seen that the named document are not taken well into account with the [`AfterID` selector](https://github.com/jinroh/cozy-stack/blob/040664d4d210d509751fabc9d058bdb0b122f86c/pkg/couchdb/mango/query.go#L170-L174)... regardless of the id string given the trash directory is always returned as the first element of the list.

I am not sure yet how to fix that and I'm looking into it (inputs are welcome).

pkg/couchdb/mango/query.go
```diff
@@ -152,13 +152,27 @@ func Between(field string, v1 interface{}, v2 interface{}) Filter {
...
 
 // StartWith returns a filter that check if field's string value start with prefix
 func StartWith(field string, prefix string) Filter {
 	return Between(field, prefix, prefix+uFFFF)
 }
 
+// AfterID returns a filter that can be used to get the elements after a
+// specific id.
+func AfterID(id string) Filter {
+	return Inside("_id", id, uFFFF)
```
nono: Why do you use `Inside`, and not `Gt("_id", id)`?

nono:
@jinroh try to add `_id` as the last field of your mango index. My supposition is that documents are sorted by sequence number and not ids when there is a tie.

jinroh:
@nono indeed, that was the problem. I added the use of the `use_index` fields for our find request to make sure we use the right index.

The PR is quite big now... sorry about that

nono:
@jinroh FYI, there is a conflit on sharings_test.go

jinroh:
@nono thx, rebased on master.

pkg/couchdb/mango/index.go
```diff
@@ -26,9 +26,12 @@ type Index struct {
 
 // IndexOnFields constructs a new Index
 // it lets couchdb defaults for index & designdoc names.
-func IndexOnFields(doctype string, fields ...string) *Index {
+func IndexOnFields(doctype, name string, fields []string) *Index {
 	return &Index{
 		Doctype: doctype,
-		Request: &IndexRequest{Index: IndexFields(fields)},
+		Request: &IndexRequest{
+			DDoc:  name,
```
aenario: Using DDoc value here means index cand be duplicated or broken by developpers mistake.

Example : 
`IndexOnFields("io.cozy.files", "by-mime-type", []string{"mimetype"})` vs `IndexOnFields("io.cozy.files", "by-type", []string{"mimetype"})` vs `IndexOnFields("io.cozy.files", "by-type", []string{"type"})`

While it will probably not occurs as all indexes are defined in consts, why do we need to name design docs?

web/files/paginated.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+	return jsonapi.DataList(c, statusCode, included, nil)
+}
+
+// newFile creates an instance of file struct from a vfs.FileDoc document.
+// Warning: the given document is mutated, the ReferencedBy field is nilified
+// so that it not dumped in the json document.
+func newFile(doc *vfs.FileDoc) *file {
+	ref := doc.ReferencedBy
+	doc.ReferencedBy = nil
```
aenario: [curious] Why not do it in file.MarshalJSON ?
```go
func (f *file) MarshalJSON() ([]byte, error) {
    f.doc.Referencedby = nil
    res, err := json.Marshal(f.doc)
    f.doc.Referencedby = f.ref
    return res, err
}
```

pkg/couchdb/mango/index.go
jinroh: The reason I used DDoc here is that it is how I succeeded to referer to them using the `use_index` field (the `name` did not work). I chose to name them so that we use this `use_index` parameter in our `FindDocs` requests to be more explicit about the index associated with the request.

Is there is a way to tell couchdb not to duplicate an index ?

web/files/paginated.go
jinroh: That's better indeed. I'll make the change.

PR #391, opened 2017-03-17
Fix doc with current @event behaviour

aenario:


jinroh:
@aenario would it be possible to add more examples ? or maybe make clear that it is still work in progress ?

nono:
I'd also like to know what the job will have as arguments when such a job is triggered: the doctype? the `_id`? the `_rev`? the event (create / update / delete)?

PR #395, opened 2017-03-21
Sharing acceptation handler

Gara64:
This allows to handle a recipient's sharing answer on the sharer side.

From the user point of view, he is redirected to his own Cozy, after being redirected to the sharer Cozy.
Behind the hoods, the answer triggers an `access_token` request to the recipient and an update in database with the new credentials and recipient status.

Note the `auth/authorize` redirection has been completed with a the `client_id`, as it is required by the sharer to identify the recipient.

### DIFF
--- a/pkg/sharings/recipients.go
+++ b/pkg/sharings/recipients.go
@@ -48,16 +48,13 @@ func (r *Recipient) Links() *jsonapi.LinksList {
 
 // ExtractDomain returns the recipient's domain without the scheme
 func (r *Recipient) ExtractDomain() (string, error) {
-	var rURL string
 	if r.URL == "" {
 		return "", ErrRecipientHasNoURL
 	}
-	tokens := strings.Split(r.URL, "://")
-	rURL = r.URL
-	if len(tokens) > 1 {
-		rURL = tokens[1]
+	if tokens := strings.Split(r.URL, "://"); len(tokens) > 1 {
+		return tokens[1], nil
 	}
-	return rURL, nil
+	return r.URL, nil
 }
 
 // GetAccessToken sends an access_token requests to the recipient, using the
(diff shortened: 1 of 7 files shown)


pkg/sharings/recipients.go
```diff
@@ -45,21 +46,51 @@ func (r *Recipient) Links() *jsonapi.LinksList {
...
+	var rURL string
+	if r.URL == "" {
+		return "", ErrRecipientHasNoURL
+	}
+	tokens := strings.Split(r.URL, "://")
+	rURL = r.URL
+	if len(tokens) > 1 {
+		rURL = tokens[1]
+	}
+	return rURL, nil
```
nono: You can shorten this function to:

```go
if r.URL == "" {
    return "", ErrRecipientHasNoURL
}
if tokens := strings.Split(r.URL, "://"); len(tokens) > 1 {
    return tokens[1], nil
}
return r.URL, nil
```

web/sharings/sharings.go
```diff
@@ -13,31 +13,27 @@ import (
...
+	state := c.QueryParam("state")
+	clientID := c.QueryParam("client_id")
+	accessCode := c.QueryParam("access_code")
 
 	instance := middlewares.GetInstance(c)
 
-	// The sharing is refused if there is no access code or scope
-	sharingAccepted := scope != "" && accessCode != ""
+	// The sharing is refused if there is no access code
+	sharingAccepted := accessCode != ""
```
nono: I'm not sure if it is easy useful to use a local variable for `sharingAccepted`. The comment could be enough:

```go
// The sharing is refused if there is no access code
if accessCode != "" {
```

If you want to keep the local variable to describe the condition, you can do it like this:

```go
if sharingAccepted := accessCode != ""; sharingAccepted {
```

Gara64: Right for both remarks, I changed it.

PR #404, opened 2017-03-23
Intents

nono:


y-lohse:
@aenario thanks for the feedback!

> I am not sure why we need to store the intent in the stack, except if Service Resolution also creates a one-time token to open the frame with X-Frame-Option: ALLOW-FROM clientURL to avoid clickjacking

Actually, the main purpose of the storing is to allow the service to retrieve the client's URL. The service needs this information in order to trust messages coming through [`postMessage`](https://developer.mozilla.org/en/docs/Web/API/Window/postMessage) -- see the section about the dispatched event's origin.

> If not for security purpose, action & type can simply be passed as query parameter to the service url, as they will be short.

`action` and `type` are not very sensitive, so yes, they could be another query parameter. That being said, the service *needs* to call the stack anyway, and returning `action` and `type` as part of that response creates less friction for the developper, while protecting these informations.

> Regarding the EDIT=crop example, it is not clear how the client pass its specs to the service (like crop size)

Thanks, this is a very good point we had somehow overlooked.
I can see 2 immediate ways to solve this:

1. The image is passed as a [data URL](https://github.com/cozy/cozy-stack/pull/404/files#diff-875f49e9c2f27001a8862f6d60c2d6dcR148), and data URL's support query string parameters. So the full intent might look something like: `cozy.startIntent('EDIT', 'image/png', 'data:image/png;base64,iVBORw...?width=50&height=50'})`. (By the way, the data URL base64 encoded format is being added to the proposal.)
2. When dealing with intents that have a MIME type as `type`, use JSON to structure the data, and define a standard field in the json such as `content` for the actual file representation. This allows to pass extra data in other fields, even when dealing with MIME types.

What do you think? Any other ideas?

IMO, the problem with option 2 is that it essentially relies on a gentleman's agreement to put the file in the `content` field and never elsewhere. On the other hand, it also means that communication is always in JSON, and we can remove the part about infering a communication format, which would make things simpler. So I like this one a bit more.

aenario:
@y-lohse Thanks for the big reply !!

:+1:  for the intent giving the service the clientURL for postMessage, still think we could do some X-Frame-Option for increased security, but no rush.

:+1:  for the new description of message format


New question (mostly curious), why the `{{.Intent}}` and not always `?intent=xxxx` or even `?cozy_intent=xxx`  ?

docs/intents.md
```diff
@@ -0,0 +1,404 @@
...
+
+- [Overview](#overview)
+- [Glossary](#glossary)
+- [Proposal](#proposal)
+    - [1. Manifest](#1-manifest)
+    - [2. Intent Start](##2-intent-start)
+    - [3. Service Resolution](#3-service-resolution)
+    - [4. Handshake](#4-handshake)
+    - [5. Processing & Terminating](#5-processing--terminating)
+- [Annex](#annex)
```
enguerran: Should you use a plural form: annexes?

y-lohse:
Short answer: developer flexibility.

Long answer: depending on the js framework (or lack thereof), it might be easier to get the token as a GET parameter, or part of the route (`/view/12234`), or whatever. And we already use the `{{.ReplaceMe}}` technique elsewhere, so it made sense to stay consistent with that.
But yeah, `?intent=xxx` is the fallback and I actually expect it to be be sufficient most of the time.

docs/intents.md
```diff
@@ -0,0 +1,404 @@
...
+    - [Discarded Ideas](#discarded-ideas)
+
+
+## Overview
+
+A typical Cozy Cloud runs multiple applications, but most of these applications are focused on one task and interact with one particular type of data.
+
+However, Cozy Cloud especially shines when data is combined across apps to provide an integrated experience. This is made difficult by the fact that apps have no dedicated back-end and that they have restricted access to documents.
+
+This document outlines a proposal for apps to rely on each other to accomplish certain tasks and gain access to new documents, in a way that hopefully is neither painful for the users or the developers.
```
enguerran: s/neither painful for the users or the developers/neither painful for the users nor the developers/

PR #407, opened 2017-03-23
Adding support for a OpenStack Swift VFS

jinroh:


web/routing.go
```diff
@@ -220,10 +220,12 @@ func CreateSubdomainProxy(router *echo.Echo, serveApps echo.HandlerFunc) (*echo.
 
 // setupRecover sets a recovering strategy of panics happening in handlers
 func setupRecover(router *echo.Echo) {
-	recoverMiddleware := middleware.RecoverWithConfig(middleware.RecoverConfig{
-		StackSize:         1 << 10, // 1 KB
-		DisableStackAll:   !config.IsDevRelease(),
-		DisablePrintStack: !config.IsDevRelease(),
-	})
-	router.Use(recoverMiddleware)
+	if !config.IsDevRelease() {
```
nono: Why this change?

pkg/instance/instance.go
```diff
@@ -468,19 +456,7 @@ func List() ([]*Instance, error) {
...
-			Limit:    1,
-		}
-		if err = couchdb.FindDocs(couchdb.GlobalDB, consts.Instances, req, &instances); err != nil {
-			return nil, err
-		}
-		if len(instances) == 0 {
-			return nil, ErrNotFound
-		}
-		i = instances[0]
+		return nil, err
```
nono: Can you rebase master?

pkg/vfs/vfs.go
```diff
@@ -39,35 +39,31 @@ const (
...
-	DirByPath(name string) (*DirDoc, error)
-	FileByID(fileID string) (*FileDoc, error)
-	FileByPath(name string) (*FileDoc, error)
-	DirOrFileByID(fileID string) (*DirDoc, *FileDoc, error)
-	DirOrFileByPath(name string) (*DirDoc, *FileDoc, error)
-	DirIterator(doc *DirDoc, opts *IteratorOptions) DirIterator
-
+	// CreateDir is used to create a new directory from its document.
 	CreateDir(doc *DirDoc) error
+	// UpdateDir is used to update the document of a directory. It takes the new
```
nono: The comment is about `UpdateDir`, but the function just below is `CreateFile`

pkg/vfs/vfsswift/impl.go
```diff
@@ -0,0 +1,352 @@
...
+		return nil, fmt.Errorf("vfsswift: specified domain is empty")
+	}
+	return &swiftVFS{
+		Indexer: index,
+		c:       conn,
+		domain:  domain,
+	}, nil
+}
+
+func confOrEnv(val string) string {
```
nono: I'm a bit puzzled by this function. As I understand it, it's possible to set a value in the config file that starts with a `$`, and it means to take the environment variable. But if the config says nothing, it stays a blank value. Am I right?

pkg/vfs/vfsswift/impl.go
```diff
@@ -0,0 +1,352 @@
...
+	return sfs.Indexer.CreateDirDoc(doc)
+}
+
+func (sfs *swiftVFS) CreateFile(newdoc, olddoc *vfs.FileDoc) (vfs.File, error) {
+	if olddoc != nil {
+		newdoc.SetID(olddoc.ID())
+		newdoc.SetRev(olddoc.Rev())
+		newdoc.CreatedAt = olddoc.CreatedAt
+	}
+	objName := newdoc.DirID + "/" + newdoc.DocName
```
nono: Won't it be simple to use the document ID as the object name for swift?

web/routing.go
jinroh: Because I feel that in dev, a panic should crash the stack. That way we do not miss it and also it is very useful to have the whole stacktraces dump of the go program when developing. With this middleware, the traces are truncated (only 1 KB) and the json formatting is frustrating.

nono: It's OK for me. If @aenario is also OK, we should update the lines below:

```go
DisableStackAll: true,
DisablePrintStack: true,
```

aenario: +1 for crash & burn during tests / dev

pkg/vfs/vfsswift/impl.go
jinroh: Yes, the idea is to reuse our URL notation in our configuration file, but allowing to specify env variables. For instance:

```yaml

fs:
  # file system url - flags: --fs-url
  # default url is the directory relative to the binary: ./storage

  url: >
    swift://openstack/
    ?UserName=$OS_USERNAME
    &Password=$OS_PASSWORD
    &ProjectName=$OS_PROJECT_NAME
    &UserDomainName=$OS_USER_DOMAIN_NAME
```

jinroh: Another solution would be to introduce swift specific fields in out `fs` options.

jinroh:
@nono @aenario In 6a9b38fe406fd4c11f11d133a01ca8ea43660836, I added the possibility to use templating in the configuration file like so, for instance:

```yaml
fs:
  # file system url - flags: --fs-url
  # default url is the directory relative to the binary: ./storage

  url: >
    swift://openstack/
    ?UserName={{ .Env.OS_USERNAME }}
    &Password={{ .Env.OS_PASSWORD }}
    &ProjectName={{ .Env.OS_PROJECT_NAME }}
    &UserDomainName={{ .Env.OS_USER_DOMAIN_NAMEE }}
```

PR #411, opened 2017-03-27
Send documents after sharing acceptation

Gara64:
This allows a sharing recipient to receive the documents in his Cozy after he gave the authorization.

Note that it doesn't use the CouchDB replication protocol as initially planned, because some work is still needed at the stack level.

### DIFF
--- a/pkg/sharings/sharings.go
+++ b/pkg/sharings/sharings.go
@@ -217,10 +217,8 @@ func sendDoc(docType, id string, doc *couchdb.JSONDoc, recStatus *RecipientStatu
 		return err
 	}
 
-	client := new(http.Client)
 	_, err = request.Req(&request.Options{
 		Domain: domain,
-		Client: client,
 		Method: "PUT",
 		Path:   path,
 		Headers: request.Headers{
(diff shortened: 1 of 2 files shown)


pkg/sharings/sharings.go
```diff
@@ -198,6 +199,66 @@ func findSharingRecipient(db couchdb.Database, sharingID, clientID string) (*Sha
...
+	delete(doc.M, "_rev")
+	body, err := request.WriteJSON(doc.M)
+	if err != nil {
+		return err
+	}
+
+	client := new(http.Client)
+	_, err = request.Req(&request.Options{
+		Domain: domain,
+		Client: client,
```
aenario: I think you only need to give a value to Client if you want to do something special (like it's done in client package tests). Otherwise, the default value of Client is http.DefaultClient which should work. Confirmation @jinroh ?

jinroh: Confirmed. You can remove the Client field declaration in this case.

Gara64: Al right, I removed it.

PR #417, opened 2017-03-28
Fix refused sharing answer

Gara64:
The refused sharing answer handler on the recipient side used to make a POST to the sharer. 
This changes the request to follow the same behaviour than the accepted sharing answer, i.e. an HTTP redirect to the sharer endpoint.

### DIFF
--- a/pkg/sharings/recipients.go
+++ b/pkg/sharings/recipients.go
@@ -83,8 +83,8 @@ func (r *Recipient) Register(instance *instance.Instance) error {
 		Domain:     rURL,
 		HTTPClient: client,
 	}
-	clientURI := instance.Scheme() + "://" + instance.Domain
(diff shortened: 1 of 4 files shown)


pkg/sharings/recipients.go
```diff
@@ -83,7 +83,8 @@ func (r *Recipient) Register(instance *instance.Instance) error {
 		Domain:     rURL,
 		HTTPClient: client,
 	}
-	redirectURI := instance.Scheme() + "://" + instance.Domain + "/sharings/answer"
+	clientURI := instance.Scheme() + "://" + instance.Domain
+	redirectURI := clientURI + "/sharings/answer"
```
jinroh: Try using the `instance.PageURL` helper instead

Gara64: Ok, I changed it

PR #418, opened 2017-03-28
Pagination helpers

aenario:
This PR includes 2 changes related to pagination

- For mango requests results, we add some documentation and a `next` and `limit` to requests answer 
- For /data/type/id/references, we use jsonapi pagination with cursor


docs/mango.md
```diff
@@ -116,3 +116,49 @@ Content-Type: application/json
...
+
+```json
+{
+    "limit": 100,
+    "next": true,
+    "docs": [ "... first hundred docs ..."],
+}
+```
+
+If there is the number of docs is lower or equal to the limit, these fields are not present
```
jinroh: `s/there is/the/` ?

pkg/couchdb/cursor.go
```diff
@@ -0,0 +1,46 @@
...
+		req.Limit = c.Limit + 1
+	}
+
+	return req
+}
+
+// GetNextCursor returns a cursor to the end of a ViewResponse
+// it removes the last item from the response to create a Cursor
+func GetNextCursor(res *ViewResponse) *Cursor {
+	next := res.Rows[len(res.Rows)-1]
```
jinroh: shouldn't we have a check when `res.Rows` is of length 0 ?

web/data/data.go
```diff
@@ -295,13 +297,28 @@ func findDocuments(c echo.Context) error {
 		return err
 	}
 
+	limit, hasLimit := findRequest["limit"].(float64)
+	if !hasLimit || limit > maxMangoLimit {
+		limit = 100
+	}
+
+	// add 1 so we know if there is more.
+	findRequest["limit"] = limit + 1
```
jinroh: couchdb does not provide a better way than that ?

web/data/data.go
```diff
@@ -295,13 +297,28 @@ func findDocuments(c echo.Context) error {
 		return err
 	}
 
+	limit, hasLimit := findRequest["limit"].(float64)
```
jinroh: why `float64` and not `int` ?

docs/mango.md
```diff
@@ -116,3 +116,49 @@ Content-Type: application/json
...
+
+```json
+{
+    "docs": [ "... less than a hundred docs ..."],
+}
+```
```
jinroh: Maybe I'm wrong, but I would better see an api where we always put the `next` and `limit` field (even if they are `false` and the default value) to be more explicit ?

Also IMHO, it is better than having to handle `undefined`s.

docs/mango.md
```diff
@@ -116,3 +116,49 @@ Content-Type: application/json
...
+Pagination of mango query should be handled by the client.
+The stack will always limit query results to maximum 100 docs.
+
+If the limit cause some docs to not be returned, the response will have a
+limit & next values.
+
+```json
+{
+    "limit": 100,
+    "next": true,
```
nono: In `next`, I'd have put the URL for asking the next documents (if it is not too complicated to compute)

web/data/data.go
nono: Go use `float64` when deserializing a JSON to a `map[string]interface{}`

web/data/data.go
aenario: nope http://docs.couchdb.org/en/2.0.0/couchapp/views/pagination.html#paging-alternate-method

docs/mango.md
aenario: good idea, will do.

docs/mango.md
aenario: It is a bit complicated to compute : we would need to have the index definition to see which field we can/should use as a cursor but the /data/xxxx/_find route does not have it.

nono: OK, let's forgot it

PR #419, opened 2017-03-28
Fix a segfault when trying to install an app with no source URL

nono:


jinroh:
We should probably handle this new error as a 400 in `pkg/apps#wrapAppsError`

jinroh:
Also, this line should not be necessary anymore: https://github.com/cozy/cozy-stack/pull/419/files#diff-cbf2fe94745e349ebd63ee84a69231f4R75

PR #421, opened 2017-03-28
Sharing worker creation

Gara64:
A sharing worker is triggered on documents updates specified by the sharing permissions.
The trigger creation is done just after the sharing accept.

Note that the worker does nothing yet, this will be implemented in a future PR.

### DIFF
--- a/pkg/sharings/sharings.go
+++ b/pkg/sharings/sharings.go
@@ -219,7 +219,10 @@ func addTrigger(instance *instance.Instance, rule permissions.Rule, sharingID st
 	type TriggerMessage struct {
 		SharingID string `json:"sharing_id"`
 	}
-	m := TriggerMessage{sharingID}
+	m := struct {
+		SharingID string `json:"sharing_id"`
+	}{sharingID}
+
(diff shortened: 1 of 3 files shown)


pkg/sharings/sharings.go
```diff
@@ -207,6 +211,34 @@ func findSharingRecipient(db couchdb.Database, sharingID, clientID string) (*Sha
...
 
+// addTrigger creates a new trigger on the updates of the shared documents
+func addTrigger(instance *instance.Instance, rule permissions.Rule, sharingID string) error {
+	scheduler := instance.JobsScheduler()
+
+	eventArgs := rule.Type + ":UPDATED:" + strings.Join(rule.Values, ",")
+	type TriggerMessage struct {
+		SharingID string `json:"sharing_id"`
+	}
+	m := TriggerMessage{sharingID}
```
nono: If you use a type only once, you can avoid to give it a name:

```go
m := struct {
    SharingID string `json:"sharing_id"
} {sharingID}
```

Gara64: Neat, thanks for the tip

PR #424, opened 2017-03-30
Add /intents service

nono:


web/intents/intents.go
```diff
@@ -0,0 +1,128 @@
...
+	intent.SetID("")
+	intent.SetRev("")
+	intent.Services = nil
+	if err = intent.Save(instance); err != nil {
+		return wrapIntentsError(err)
+	}
+	if err = intent.FillServices(instance); err != nil {
+		return wrapIntentsError(err)
+	}
+	if err = intent.Save(instance); err != nil {
```
aenario: Why save in 2 steps?

pkg/intents/intents.go
```diff
@@ -0,0 +1,95 @@
...
+		return ret
+	}
+	return u.String() + buf.String()
+}
+
+// FillServices looks at all the application that can answer this intent
+// and save them in the services field
+func (in *Intent) FillServices(instance *instance.Instance) error {
+	var res []apps.WebappManifest
+	err := couchdb.GetAllDocs(instance, consts.Apps, &couchdb.AllDocsRequest{}, &res)
```
aenario: We could use a couchdb view to find intent-corresponding apps.

web/intents/intents.go
nono: We need an id for building the href of services, and we need later to persist the services

pkg/intents/intents.go
nono: Not really, we can do a first filtering with a couchdb view, but not have the intents because of the possible jokers (`*`)

PR #426, opened 2017-03-30
One OAuth client per sharing

Ljinod:
Before this PR the behavior regarding OAuth in C2C sharing was to
register a sharer once at each recipient and then ask for an access
token for every new share using the same OAuth client. We also thought
that a recipient could revoke an access token directly, which is not
true. A recipient can only revoke an OAuth client, which means all the
access tokens and this is not what we want.

To give the recipient the possibility to revoke a sharing we have to
have one OAuth client per sharing. This is what this PR introduces.

The modifications can be roughly summed up as following:
* Refine the sharing status to include errors when the sharer could not
register or when the email was not sent.
* Move the `RecipientStatus` code in the "Recipients.go" file.
* Add `auth.Client` and `auth.AccessToken` structures in
`RecipientStatus` so as to access the OAuth information later on.
* Creating a recipient no longer triggers the registration process.
* Creating a sharing now triggers the registration process for each
recipient. Their status is updated accordingly.
* Fix lots of tests that these modifications broke.
* Fix gometalinter numerous warnings…


pkg/sharings/recipients.go
```diff
@@ -57,46 +71,109 @@ func (r *Recipient) ExtractDomain() (string, error) {
...
+	// without "http://" or "https://") so we parse it.
+	recipientURL, err := url.Parse(rs.recipient.URL)
 	if err != nil {
 		return nil, err
+	} else if recipientURL.Host == "" {
+		// Caveat: if the URL does not have a scheme then url.Parse returns an
+		// empty url object. If that is the case we will assume that the URL
+		// declared in the Recipient structure is ok…
+		recipientURL.Host = rs.recipient.URL
 	}
```
Gara64: Is there a reason you removed the `ExtractDomain` method? It seems to do the same thing, and it also used in `pkg/sharings`

pkg/sharings/recipients.go
```diff
@@ -57,46 +71,109 @@ func (r *Recipient) ExtractDomain() (string, error) {
...
+
+	// If the recipient has no URL there is no point in registering.
+	if rs.recipient.URL == "" {
+		return ErrRecipientHasNoURL
 	}
-	clientURI := instance.PageURL("", nil)
-	redirectURI := instance.PageURL("/sharings/answer", nil)
 
-	// Get the Cozy's public name
+	// We get the instance document to extract the public name.
```
Gara64: You say the same thing, but longer :)

pkg/sharings/recipients.go
```diff
@@ -105,19 +182,30 @@ func (r *Recipient) Register(instance *instance.Instance) error {
...
-	resClient, err := req.RegisterClient(authClient)
+	recipientURL, err := url.Parse(rs.recipient.URL)
 	if err != nil {
 		return err
+	} else if recipientURL.Host == "" {
+		// Caveat: if the URL does not have a scheme then url.Parse returns an
+		// empty url object. If that is the case we will assume that the URL
+		// declared in the Recipient structure is ok…
+		recipientURL.Host = rs.recipient.URL
 	}
```
Gara64: Same remark for the `ExtractDomain`. The code is duplicated here.

pkg/sharings/sharings.go
```diff
@@ -303,47 +275,72 @@ func ShareDoc(instance *instance.Instance, sharing *Sharing, recStatus *Recipien
...
+
+	// Update the document for later usage.
 	err = couchdb.UpdateDoc(instance, sharing)
 	if err != nil {
 		return "", err
 	}
+
 	// Share all the documents with the recipient
+	// err = ShareDoc(instance, sharing, recStatus)
 	err = ShareDoc(instance, sharing, recStatus)
```
Gara64: You forgot to remove the comment

web/sharings/sharings.go
```diff
@@ -49,16 +49,12 @@ func AddRecipient(c echo.Context) error {
...
-	}
-
 	return jsonapi.Data(c, http.StatusCreated, recipient, nil)
 }
 
-// SharingRequest handles a sharing request from the recipient side
-// It creates a temporary sharing document and redirects to the authorize page
+// SharingRequest handles a sharing request from the recipient side.
+//
+// It creates a temporary sharing document and redirects to the authorize page.
```
Gara64: The extra line is not needed

pkg/sharings/recipients.go
```diff
@@ -57,46 +70,104 @@ func (r *Recipient) ExtractDomain() (string, error) {
...
 
-// GetAccessToken sends an access_token requests to the recipient, using the
-// given authorization code
-func (r *Recipient) GetAccessToken(code string) (*auth.AccessToken, error) {
-	client := new(http.Client)
-	rURL, err := r.ExtractDomain()
+// CreateRecipient inserts a Recipient document in database
+func CreateRecipient(db couchdb.Database, doc *Recipient) error {
+	// Sanity check: we need an URL…
+	if doc.Email == "" {
```
nono: You have swapped the checks in code, or in the comments

pkg/sharings/recipients.go
```diff
@@ -57,46 +70,104 @@ func (r *Recipient) ExtractDomain() (string, error) {
...
+	if rs.recipient.URL == "" {
+		return nil, ErrRecipientHasNoURL
+	}
+	if rs.Client.ClientID == "" {
+		return nil, ErrNoOAuthClient
+	}
+
+	// The structure `auth.Request` expects a domain WITHOUT a scheme (i.e.
+	// without "http://" or "https://") so we parse it.
+	recipientURL, err := rs.recipient.ExtractDomain()
```
nono: You can call it a `recipientDomain` instead of `recipientURL` if it is just the domain

pkg/sharings/send_mails.go
```diff
@@ -40,34 +41,30 @@ func SendSharingMails(instance *instance.Instance, s *Sharing) error {
...
 	errorOccurred := false
-	for _, rs := range recipientsStatus {
+	for _, rs := range s.RecipientsStatus {
 		recipient := rs.recipient
 
 		// Generate recipient specific OAuth query string.
-		recipientOAuthQueryString, err := generateOAuthQueryString(s, recipient, instance.Scheme())
-		if err != nil {
-			errorOccurred = logErrorAndSetRecipientStatus(rs, err)
+		recipientOAuthQueryString, errGenOAuth := generateOAuthQueryString(s,
```
nono: You should use shorter names for local variables

pkg/sharings/sharings.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/cozy/cozy-stack/pkg/permissions"
 	"github.com/cozy/cozy-stack/pkg/utils"
 	"github.com/cozy/cozy-stack/web/jsonapi"
+	"github.com/labstack/gommon/log"
```
nono: We are using `github.com/Sirupsen/logrus` for logging

pkg/sharings/send_mails.go
```diff
@@ -76,15 +72,30 @@ func SendSharingMails(instance *instance.Instance, s *Sharing) error {
...
 		}
+
+		// Job was created, we set the status to "pending".
+		rs.Status = consts.PendingSharingStatus
+	}
+
+	// Persist the modifications in the database.
+	err = couchdb.UpdateDoc(instance, s)
+	if err != nil {
+		if errorOccurred {
```
jinroh: `s/errorOccured/err/` ?

Ljinod: The idea here is to return an error if one of the email could not be sent, and that error must say so. Now that I think about it this is probably not the best way to do it as it doesn't give much information (doesn't say for whom), and as one error was already logged before.

jinroh: OK I got it. Maybe not the best way but still pretty good.

PR #428, opened 2017-03-30
[WIP] allow listing permissions for a doctype

aenario:
Still lacking doc.

### DIFF
--- a/pkg/couchdb/couchdb.go
+++ b/pkg/couchdb/couchdb.go
@@ -10,6 +10,7 @@ import (
 	"net/url"
 	"strings"
 	"time"
+	"unicode"
 
 	log "github.com/Sirupsen/logrus"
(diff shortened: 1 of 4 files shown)


pkg/couchdb/couchdb.go
```diff
@@ -19,6 +20,10 @@ import (
 	"github.com/labstack/echo"
 )
 
+// InfiniteString is the unicode character "\uFFFF", useful in query as
+// a upperbound for string.
+const InfiniteString = string(unicode.MaxRune)
```
nono: Is is defined as `uFFFF` here: https://github.com/cozy/cozy-stack/blob/master/pkg/couchdb/mango/query.go#L155. We should probably use the same name to be consistent

aenario:
replaced by #440

PR #431, opened 2017-03-31
Fix session issue with onboarding and flat subdomains

nono:


gregorylegarec:
@nono Why was it working locally but not in production ?

nono:
@gregorylegarec the route for the onboarding is public, so with flat subdomains, the session was not set on the onboarding domain, and the app token was empty

PR #435, opened 2017-03-31
Use cozy.tools instead of cozy.local (no need to edit /etc/hosts)

nono:


scripts/Dockerfile
```diff
@@ -70,7 +70,6 @@ EXPOSE 8080 8025 5984
 
 CMD couchdb 2>/dev/null 1>/dev/null & \
     MailHog 2>/dev/null 1>/dev/null & \
-    echo "0.0.0.0 cozy.local app.cozy.local" >> /etc/hosts && \
     /usr/bin/cozy-app-dev.sh \
       -d /data/cozy-app \
       -f /data/cozy-storage
```
jinroh: shouldn't we remove the `cozy-stack instances add localhost:8080` ? does not seem necessary anymore

nono: Well, it can be useful to have 2 instances (for playing with the sharing for example) without having to type advanced docker commands

PR #436, opened 2017-03-31
VFS locking

jinroh:
This PR introduces a global RW lock on the VFS to make concurrent access to the files deterministic.

### DIFF
--- a/pkg/instance/instance.go
+++ b/pkg/instance/instance.go
@@ -163,13 +163,14 @@ func (i *Instance) VFS() vfs.VFS {
 
 func (i *Instance) makeVFS() error {
 	fsURL := config.FsURL()
+	mutex := vfs.NewMemLock()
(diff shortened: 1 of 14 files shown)


pkg/instance/instance.go
```diff
@@ -163,13 +163,14 @@ func (i *Instance) VFS() vfs.VFS {
 
 func (i *Instance) makeVFS() error {
 	fsURL := config.FsURL()
+	mutex := vfs.NewMemLock()
```
nono: As I understand, it creates a new mutex for each request. So, two requests for the same domain will use two different locks, can mess things. I think we should have a map for domain -> mutex (its-self protected by a mutex)

jinroh: Wow, sorry I missed that... Fixing it.

PR #437, opened 2017-04-01
Add a worker to share data and deal with updates

Gara64:
This PR mainly consists in 3 things:

- Adds a share data worker to send the documents asynchronously
- Supports shared document updates
- Fix a permission check for doc updates in `web/data/data.go`

Note this PR relies on #426, which must be merged first.


pkg/jobs/workers/sharing_updates.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	return doc, err
+}
+
+// ExtractDomain returns the recipient's domain without the scheme
+func ExtractDomain(u string) (string, error) {
+	if u == "" {
+		return "", ErrRecipientHasNoURL
+	}
+	if tokens := strings.Split(u, "://"); len(tokens) > 1 {
```
jinroh: nit: why not using `url.Parse` ? it would be a little more precise error wise

pkg/jobs/workers/share_data.go
```diff
@@ -0,0 +1,125 @@
...
+				"Content-Type":  "application/json",
+				"Accept":        "application/json",
+				"Authorization": "Bearer " + rec.Token,
+			},
+			Body: body,
+		})
+		if errSend != nil {
+			log.Error("[sharing] An error occurred while trying to share "+
+				"data : ", errSend)
+		}
```
jinroh: The response should be closed. You can close it yourself or use the `NoResponse` field in `request.Options` for that.

pkg/jobs/workers/sharing_updates.go
Gara64: No particular reason. I switched to url.Parse

PR #439, opened 2017-04-03
Decrease iframe security policy to allow intents

gregorylegarec:
To allow intents we need to be able to display iframes providing resource from another subdomain of the same domain. Example : onboarding.cozy.tools may display an intent from konnectors.cozy.tools.


nono:
As explained on IRC, it won't work. I will open a new PR with something better.

PR #441, opened 2017-04-03
Allow to open iframes for intents

nono:
Replace #439

### DIFF
--- a/web/apps/serve.go
+++ b/web/apps/serve.go
@@ -2,6 +2,7 @@ package apps
 
 import (
 	"bytes"
+	"fmt"
 	"html/template"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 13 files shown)


y-lohse:
Works for me. 
This was half the reason why the href was allowed and why intents were in their own part of the manifest, so if we get rid of this we might consider moving the intent manifest in the `routes` part. Anyway, topi for another day.

web/apps/serve.go
```diff
@@ -57,6 +59,27 @@ func onboarding(c echo.Context) bool {
...
+		}
+	}
+	if !allowed {
+		return
+	}
+	from := i.SubDomain(intent.Client).Host
+	hdr := fmt.Sprintf("%s %s", middlewares.XFrameAllowFrom, from)
+	c.Response().Header().Set(echo.HeaderXFrameOptions, hdr)
+}
+
```
gregorylegarec: To be sure that I understand clearly : this means that to open an intent, an app need to open first an iframe on itself, and in this iframe the X-frame-options will allow to add a second iframe to the real service ?

nono: No, you just open the iframe on the new app. `X-Frame-Options` is set on the HTML page inside the iframe and control if the origin of the parent.

For example, you are in the contacts app and you want to pick a photo for an avatar of one contact. You will make a request to the stack to create the intent. In the response, you will have an href for a service. You can then open the iframe on this URL.

gregorylegarec: Ok, thanks. My bad, I misunderstood what was the role of the `X-Frame-Options` header. I thought that it was a header for the main opener window. Actually it's not, it is just used for embedded ones. And you deal the CSP and url's siblings thanks to your `splitHost` method. Now I get the whole, thing, thanks !

PR #442, opened 2017-04-03
Improve error handling of apps install/update

jinroh:
This PR improves the error handling of the installation / update of applications. On the server-side the installer is given on "Operation" type to better check the manifest existence only if necessary (update/delete). On the API/Client side, the error are always returned in a `text/event-stream` format if required by the client.

### DIFF
--- a/pkg/apps/installer.go
+++ b/pkg/apps/installer.go
@@ -14,6 +14,14 @@ import (
 
 var slugReg = regexp.MustCompile(`^[A-Za-z0-9\-]+$`)
 
+type Operation int
+
+const (
+	Install Operation = iota + 1
+	Update
+	Delete
+)
+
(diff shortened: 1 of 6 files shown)


pkg/apps/installer.go
```diff
@@ -35,6 +46,7 @@ type Installer struct {
 // source URL.
 type InstallerOptions struct {
 	Type      AppType
+	Operation Operation
```
aenario: [nitpick] The Installer interface gets a bit weird with these changes.

 A distracted developper can do  `NewInstaller(.... { Operation : Update }).Delete()`  
How about Installer->AppManager & .Update/Install/Delete -> .Run ?

jinroh: I agree with you, the design needs a rework. We can do that later though...

PR #447, opened 2017-04-04
Implement redis caching for domain->instance

aenario:


pkg/config/config.go
```diff
@@ -113,6 +115,21 @@ func CouchURL() string {
 	return config.CouchDB.URL
 }
 
+var redisClient *redis.Client
+
+// RedisClient returns the global redis client or
```
jinroh: This method should be thread-safe.

Also I am not sure it is the good place to store this global handler.

pkg/config/config.go
```diff
@@ -70,6 +71,7 @@ type Config struct {
 	AdminPort  int
 	Fs         Fs
 	CouchDB    CouchDB
+	Redis      *redis.Options
```
nono: We should have one redis URL per usage: one for cache now, and later one for locks and one for pub/sub

nono: And it should be configurable from command-line (`cozy-stack serve --redis-cache redis://localhost:6379/0`)

pkg/instance/cache.go
```diff
@@ -0,0 +1,90 @@
...
+var _ instanceCache = &redisCache{}
+
+type noCache struct{}
+
+func (*noCache) Get(string) *Instance  { return nil }
+func (*noCache) Set(string, *Instance) {}
+func (*noCache) Revoke(string)         {}
+
+type redisCache struct {
+	mu     sync.RWMutex
```
nono: Do we really need a mutex? I'd have think that the redis client have a pool of connections and be goroutine-safe.

pkg/instance/instance.go
```diff
@@ -392,13 +392,28 @@ func Create(opts *Options) (*Instance, error) {
 
 // Get retrieves the instance for a request by its host.
 func Get(domain string) (*Instance, error) {
+	cached := getCache().Get(domain)
+	if cached != nil {
+		return cached, nil
```
nono: We don't need to call `makeVFS` here?

pkg/cache/cache.go
```diff
@@ -0,0 +1,72 @@
...
+func (*noCache) Set(string, interface{}) {}
+func (*noCache) Del(string)              {}
+
+type jsonCache struct {
+	namespace  string
+	expiration time.Duration
+	client     subRedisInterface
+}
+
+func (c *jsonCache) Get(d string, out interface{}) {
```
nono: I'd have expected `Get` to return a bool (or an err) to say if it is a hit or a miss

pkg/config/config.go
```diff
@@ -249,6 +272,9 @@ fs:
 couchdb:
     url: http://localhost:5984/
 
+cache:
+    url: redis://localhost:6379
```
nono: Can we make it explicit that we use the database 0, like this: `redis://localhost:6379/0`?

pkg/cache/cache.go
```diff
@@ -0,0 +1,72 @@
...
+	err = c.client.Set(c.namespace+d, bytes, c.expiration).Err()
+	if err != nil {
+		log.Error("unable to cache instance", i, err)
+	}
+}
+
+func (c *jsonCache) Del(d string) {
+	err := c.client.Del(c.namespace + d).Err()
+	if err != nil {
+		log.Error("unable to revoke cache", err)
```
nono: If the cache has expired, do we log an error?

pkg/config/config.go
```diff
@@ -119,6 +126,19 @@ func CouchURL() string {
...
 }
 
+// CacheOptions returns the options for caching redis
+func CacheOptions() *redis.Options {
+	if config.Cache.URL == "" {
+		return nil
+	}
+	opts, err := redis.ParseURL(config.Cache.URL)
+	if err != nil {
+		log.Errorf("cant parse cache.URL(%s), ignoring", config.Cache.URL)
```
nono: s/cant/can't/

pkg/instance/cache.go
```diff
@@ -0,0 +1,39 @@
...
+}
+func (ic *instanceCache) Revoke(d string) {
+	ic.base.Del(d)
+}
+
+var globalCache *instanceCache
+
+func getCache() *instanceCache {
+
+	if globalCache == nil {
```
nono: You are setting a global variable from a function that can be called from several goroutines -> you need a mutex.

pkg/instance/instance.go
```diff
@@ -419,13 +419,30 @@ func Create(opts *Options) (*Instance, error) {
 
 // Get retrieves the instance for a request by its host.
 func Get(domain string) (*Instance, error) {
+	cache := getCache()
+	var err error
+	var i = cache.Get(domain)
```
nono: s/var i =/i :=/

pkg/cache/cache.go
aenario: I prefer logging excesively for now so we can see problem with bad keys quickly. We can remove log if there is too much in production later. 

cache.Del is called from Update & Delete, which will always happen in the same request than a GetInstance, so the cache should almost never be a miss. (unless expiration in the ms between beginning and end of request)

PR #448, opened 2017-04-04
Move apps/konnectors data out of the VFS

jinroh:
This PR moves the apps and konnectors storage out of the VFS index.

For retro-compatibility reason, the `.cozy-apps` and `.cozy-konnectors` are stilled used and stored in the VFS directory of the user.

The `swift` filesystem is not yet implemented.

### DIFF
--- a/pkg/instance/instance.go
+++ b/pkg/instance/instance.go
@@ -200,6 +200,7 @@ func (i *Instance) hiddenFS(dirname string) afero.Fs {
 	case "swift":
 		panic("Not implemented")
 	}
+	return nil
 }
 
 // StartJobSystem creates all the resources necessary for the instance's job
--- a/pkg/instance/instance.go
+++ b/pkg/instance/instance.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
+	"path"
 	"strings"
 	"time"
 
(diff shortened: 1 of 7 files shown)


pkg/instance/instance.go
```diff
@@ -177,6 +179,30 @@ func (i *Instance) makeVFS() error {
...
+}
+
+func (i *Instance) hiddenFS(dirname string) afero.Fs {
+	fsURL := config.FsURL()
+	switch fsURL.Scheme {
+	case "file", "mem":
+		return afero.NewBasePathFs(afero.NewOsFs(),
+			path.Join(fsURL.Path, i.Domain, dirname))
+	case "swift":
+		panic("Not implemented")
```
nono: Do you think it will complicated to have a swift implementation for this? In particular, do you think that having a `dirname` here can make things harder?

web/apps/serve.go
```diff
@@ -133,6 +133,12 @@ func ServeAppFile(c echo.Context, i *instance.Instance, fs AppFileServer, app *a
 		return err
 	}
 	defer content.Close()
+
+	ext := path.Ext(file)
+	if ext != ".html" && ext != ".htm" {
+		return fs.ServeFileContent(c.Response(), c.Request(), modtime, slug, route.Folder, file)
+	}
```
nono: Why are you doing this? We already have a check a few lines before to use templates only for the index file of a route. Do you see a case where this index file won't be HTML?

jinroh: My bad, did expect to apply this in this PR. It is not relevant anyway like you said... removed it f0660a4292d05eb3602ab60dca22e03a312aef94

pkg/instance/instance.go
jinroh: Not that much, but we have to think it through.

For instance I was wondering if we really wanted to store konnectors and apps in the same way. The konnector be seen as one single application blob/code (could be stored in a tarball) while an app contains multiple assets that can be served independently.

That said, the work should be relatively straightforward.

PR #457, opened 2017-04-05
[VFS] Maximum disk space and usage

jinroh:


cmd/instances.go
```diff
@@ -235,6 +247,7 @@ func init() {
 	addInstanceCmd.Flags().StringVar(&flagTimezone, "tz", "", "The timezone for the user")
 	addInstanceCmd.Flags().StringVar(&flagEmail, "email", "", "The email of the owner")
 	addInstanceCmd.Flags().StringVar(&flagPublicName, "public-name", "", "The public name of the owner")
+	addInstanceCmd.Flags().StringVar(&flagDiskSpace, "disk-space", "", "The size of the allowed")
```
nono: Maybe we can call this disk-quota instead of disk-space

cmd/instances.go
```diff
@@ -235,6 +247,7 @@ func init() {
 	addInstanceCmd.Flags().StringVar(&flagTimezone, "tz", "", "The timezone for the user")
 	addInstanceCmd.Flags().StringVar(&flagEmail, "email", "", "The email of the owner")
 	addInstanceCmd.Flags().StringVar(&flagPublicName, "public-name", "", "The public name of the owner")
+	addInstanceCmd.Flags().StringVar(&flagDiskQuota, "disk-space", "", "The size of the allowed")
```
aenario: Lack a word there, no ?

pkg/vfs/errors.go
```diff
@@ -35,4 +35,6 @@ var (
 	ErrDirNotEmpty = errors.New("Directory is not empty")
 	// ErrWrongCouchdbState is given when couchdb gives us an unexpected value
 	ErrWrongCouchdbState = errors.New("Wrong couchdb reduce value")
+	// ErrFileTooBig is used when there is no more space left on the filesystem
+	ErrFileTooBig = errors.New("The file is too big")
```
aenario: May be we can make this error a bit more clear "Adding file would exceed quota", as it is now it makes me think more of 413 error, aka "we will never accept a file bigger than 10Go"

docs/settings.md
```diff
@@ -74,6 +76,8 @@ Content-type: application/vnd.api+json
     "type": "io.cozy.settings",
     "id": "io.cozy.settings.disk-usage",
     "attributes": {
+      "is_limited": true,
+      "total_size": "123456789",
```
nono: I'd have called it `quota` too

nono: And I'd have not put a `is_limited` field. If true, there is a `quota`/`total_size`. If false, we just omit this field.

pkg/instance/instance.go
```diff
@@ -72,6 +72,8 @@ type Instance struct {
 	Locale string `json:"locale"`         // The locale used on the server
 	Dev    bool   `json:"dev"`            // Whether or not the instance is for development
 
+	BytesDiskQuota int64 `json:"disk_quota,string"` // The total size in bytes allowed to the user
```
nono: `omitempty`?

pkg/instance/instance.go
```diff
@@ -191,6 +195,11 @@ func (i *Instance) AppsFS(appsType apps.AppType) afero.Fs {
 	panic(fmt.Errorf("Unknown application type %s", string(appsType)))
 }
 
+// DiskQuota returns the number of bytes allowed on the disk to the user.
+func (i *Instance) DiskQuota() (int64, error) {
+	return i.BytesDiskQuota, nil
```
nono: Why do you return an error if it is always nil?

jinroh: It implements an interface `DiskThresholder` which defines a `DiskQuota() (int64, error)`, allowing the possibility to have an error... maybe that is pretty useless for now...

pkg/instance/instance.go
jinroh: why not. done.

pkg/instance/instance.go
nono: Yes, but the implementation in the test also return nil for the error

jinroh: I feel it is premature now that you noted that. I removed the error return from the function's signature.

jinroh:
rebased on master

PR #458, opened 2017-04-06
Shellcheck on our scripts

jinroh:


scripts/cozy-app-dev.sh
```diff
@@ -69,7 +69,7 @@ do_start() {
 		echo "ok"
 	fi
 
-	trap "trap - SIGTERM && kill 2>&1 > /dev/null -- -${$}" SIGINT SIGTERM EXIT
+	trap 'trap - SIGTERM && kill 2>&1 > /dev/null -- -${$}' SIGINT SIGTERM EXIT
```
nono: I don't understand this line. So maybe my question is stupid, but changing the quotes doesn't change the interpolation of `${$}`?

jinroh: The reason is that the expand of the variable is done when the trap is being executed. In this case `${$}` is a constant so it does not have an impact, but it is good practice to do that when using `trap`.

PR #459, opened 2017-04-06
Thumbnails

nono:


pkg/jobs/workers/thumbnail.go
```diff
@@ -27,30 +38,93 @@ func init() {
...
+
+func generateThumb(ctx context.Context, in io.Reader, out io.Writer, format string) error {
+	args := []string{"-", "-limit", "Memory", "2GB", "-thumbnail", format, "jpg:-"}
+	cmd := exec.CommandContext(ctx, "convert", args...)
+	cmd.Stdin = in
+	cmd.Stdout = out
+	return cmd.Run()
+}
+
+func removeThumbnails(i *instance.Instance, img vfs.FileDoc) error {
```
jinroh: `vfs.FileDoc` is quite a big struct. could be better to pass a pointer no ?

pkg/jobs/workers/thumbnail.go
```diff
@@ -27,30 +38,93 @@ func init() {
...
+	if err != nil {
+		return err
+	}
+	defer in.Close()
+	largeName := thumbName(img, "large")
+	large, err := fs.OpenFile(largeName, flags, 0640)
+	if err != nil {
+		return err
+	}
+	defer large.Close()
```
jinroh: since this serie of `generateThumb` calls can take quite some time, I would have the `generateThumb` method do the call to the `Close()` method in order to have less file handlers laying around.

PR #460, opened 2017-04-06
Fix X-Frame-Options by adding the scheme for allow-from

nono:


web/apps/serve.go
```diff
@@ -75,7 +75,7 @@ func handleIntent(c echo.Context, i *instance.Instance, slug, intentID string) {
 	if !allowed {
 		return
 	}
-	from := i.SubDomain(intent.Client).Host
+	from := i.SubDomain(intent.Client).String()
```
gregorylegarec: Thanks @nono, it works better. Another problem is that when an intent is save in DB, the Client value is set to something like `io.cozy.apps/test-app`, the  `io.cozy.apps/` appears in the header and make the iframe loading fail, for Security reason.

nono: Oh!

gregorylegarec: Apparently somthing is done here to split the Client value, but then the previous value is re-affected https://github.com/cozy/cozy-stack/blob/master/web/intents/intents.go#L54

web/apps/serve.go
```diff
@@ -75,7 +76,11 @@ func handleIntent(c echo.Context, i *instance.Instance, slug, intentID string) {
 	if !allowed {
 		return
 	}
-	from := i.SubDomain(intent.Client).String()
+	parts := strings.SplitN(intent.Client, "/", 2)
```
gregorylegarec: Works far better ! Thanks @nono !

Still have a warning message from Chrome about the `ALLOW-FROM` directive, but from what I read on the web, it is a known problem.

PR #461, opened 2017-04-06
[refacto] Clean package graph 

aenario:
Remove almost all dependency from "/pkg" to "/web".

Only remaining dependency is in pkg/sharings which needs to setup a fake remote cozy to communicate with.

Use the apiXXX technique to remove dependency to jsonapi & a new couchdb.DocReference which replace jsonapi.ResourceIdentifier.

### DIFF
--- a/pkg/jobs/trigger_at.go
+++ b/pkg/jobs/trigger_at.go
@@ -4,7 +4,6 @@ import (
 	"time"
 
 	"github.com/cozy/cozy-stack/pkg/consts"
-	"github.com/cozy/cozy-stack/web/jsonapi"
 )
 
 // maxPastTriggerTime is the maximum duration in the past for which the at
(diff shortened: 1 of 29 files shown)


pkg/jobs/trigger_at.go
```diff
@@ -24,7 +23,7 @@ type AtTrigger struct {
 func NewAtTrigger(infos *TriggerInfos) (*AtTrigger, error) {
 	at, err := time.Parse(time.RFC3339, infos.Arguments)
 	if err != nil {
-		return nil, jsonapi.BadRequest(err)
+		return nil, ErrMalformedTigger
```
jinroh: just a type: s/ErrMalformedTigger/ErrMalformedTrigger/

aenario: Thinking of doubling down on my  mistake and remove the second g instead ...

PR #466, opened 2017-04-06
Add share files capability

Gara64:
This allows to share files, and not only JSON documents anymore.
The mails and sharings workers have been moved in separate packages, to avoid dependency cycles.

Note this relies on #437 which must be merged first.


aenario:
@jinroh This PR's tests fail with  https://travis-ci.org/cozy/cozy-stack/jobs/219613043#L1002 which appears to be linked with the konnector worker, any idea what is it?

pkg/jobs/workers/sharings/sharing_updates.go
```diff
@@ -0,0 +1,178 @@
...
+	err := m.Unmarshal(&event)
+	if err != nil {
+		return err
+	}
+	sharingID := event.Message.SharingID
+	docType := event.Message.DocType
+	docID := event.Event.Doc.M["_id"].(string)
+
+	// Get the sharing document
+	db := couchdb.SimpleDatabasePrefix(domain)
```
aenario: SimpleDatabasePrefix was supposed to be used ony in tests. 

An `instances.Get()` seems more close to what you wants to do here.

Gara64: All right, `SimpleDatabasePrefix` was also used in `share_data.go`, so I changed both and fixed the tests accordingly.

Gara64:
@aenario @jinroh The tests fail because I started to move the workers in separate packages (for mails and sharing). In the meantime, the konnector workers have been added, that no longer work because it lacks a `TestMain` method. Should I undo the separate packages or move the konnectors in their own package ?

nono:
@Gara64 FYI, I have also made changes on the workers in this PR: #459

aenario:
@Gara64 move the konnector into its own package.

pkg/sharings/send_mails.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/cozy/cozy-stack/pkg/couchdb"
 	"github.com/cozy/cozy-stack/pkg/instance"
 	"github.com/cozy/cozy-stack/pkg/jobs"
-	"github.com/cozy/cozy-stack/pkg/jobs/workers"
+	"github.com/cozy/cozy-stack/pkg/jobs/workers/mails"
```
nono: It's weird, I should be missing something. You are still using workers at line 128 of this file, but the import is for mails now

Gara64: It's weird indeed, `workers.` works to use mails, but not `mails.`... I changed that with `mails "github.com/cozy/cozy-stack/pkg/jobs/workers/mails"` as a workaround.

Gara64:
@nono I'm not sure I understand your comment about konnectors. I never use it, so where should I import it?

nono:
@Gara64 Oh, I see. You let `package workers` at the top of the "mails" package (same for konnectors)

nono:
@Gara64 for konnectors, you have moved it from a package that is imported in the build to a new package that is never imported. So, as if, it's not possible to build cozy-stack and use a konnector job in it. You have to import this package from somewhere. I suggest to do like for others workers: https://github.com/cozy/cozy-stack/blob/master/web/jobs/jobs.go#L12

PR #468, opened 2017-04-07
Update the /data blacklist

nono:


jinroh:
@nono is wanted to not have `consts.DocTypes` as a `readable` ?

nono:
@jinroh I don't know. We have no couchdb database for this. I think we can let the request respond a 404 when trying to read doctypes, and it can be the right thing to do if we later add a database with the list of the doctypes.

PR #469, opened 2017-04-10
Add a CSP whitelist, with Open Street Map and Mapbox

nono:


aenario:
merging, red travis was a NPM problem

PR #473, opened 2017-04-11
Dynamic sharing support

Gara64:
This adds the possibility to define a dynamic sharing throught the use of `selector`, for JSON documents.
This is done at the trigger creation, once the sharing is accepted.

Note this PR relies on #466.

### DIFF
--- a/pkg/sharings/sharings.go
+++ b/pkg/sharings/sharings.go
@@ -157,7 +157,13 @@ func findSharingRecipient(db couchdb.Database, sharingID, clientID string) (*Sha
 func addTrigger(instance *instance.Instance, rule permissions.Rule, sharingID string) error {
 	scheduler := instance.JobsScheduler()
 
-	eventArgs := rule.Type + ":UPDATED:" + strings.Join(rule.Values, ",")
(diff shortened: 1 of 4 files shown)


pkg/sharings/sharings.go
```diff
@@ -157,7 +157,13 @@ func findSharingRecipient(db couchdb.Database, sharingID, clientID string) (*Sha
 func addTrigger(instance *instance.Instance, rule permissions.Rule, sharingID string) error {
 	scheduler := instance.JobsScheduler()
 
-	eventArgs := rule.Type + ":UPDATED:" + strings.Join(rule.Values, ",")
+	var eventArgs string
+	if rule.Selector != "" {
+		eventArgs = rule.Type + ":CREATED,UPDATED,DELETED:" + rule.Values[0] + ":" + rule.Selector
```
aenario: The syntax is designed so there can also be a selector by field = one of multiple value, not sure if you removed that on purpose.

Gara64: I didn't, I thought it was an equality check for only value. I'll change that.

PR #474, opened 2017-04-12
Add supports for skip pagination on /files/

aenario:
This adds a meta:count to the contents  relationships of a folder. 

We also introduce a separate route to fetch more of this relationships.

Improved couchdb.Cursor to  support both skip & cursor, this will allows supporting both kind of pagination for all routes using couchdb.Cursor.

Note @nono I added a couple "TODO change me for flag hidden" which will require revision when introducing `hidden` flag.

### DIFF
--- a/cmd/instances.go
+++ b/cmd/instances.go
@@ -49,7 +49,6 @@ create its CouchDB databases.
 var cleanInstanceCmd = &cobra.Command{
 	Use:   "clean [domain]",
 	Short: "Clean badly removed instances",
-	Long:  ``,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		if len(args) == 0 {
 			return cmd.Help()
--- a/cmd/instances.go
+++ b/cmd/instances.go
@@ -10,6 +10,7 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/cozy/cozy-stack/client"
+	"github.com/cozy/cozy-stack/pkg/couchdb"
 	"github.com/cozy/cozy-stack/pkg/instance"
 	humanize "github.com/dustin/go-humanize"
 	"github.com/spf13/cobra"
@@ -42,6 +43,22 @@ create its CouchDB databases.
 `,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		return cmd.Help()
+	},
+}
+
+var cleanInstanceCmd = &cobra.Command{
+	Use:   "clean [domain]",
+	Short: "Clean badly removed instances",
+	Long:  ``,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(args) == 0 {
+			return cmd.Help()
+		}
+
+		domain := args[0]
+		i := couchdb.SimpleDatabasePrefix(domain)
+		return couchdb.DeleteAllDBs(i)
+
 	},
 }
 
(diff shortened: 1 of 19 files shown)


cmd/instances.go
```diff
@@ -45,6 +46,22 @@ create its CouchDB databases.
 	},
 }
 
+var cleanInstanceCmd = &cobra.Command{
+	Use:   "clean [domain]",
+	Short: "Clean badly removed instances",
+	Long:  ``,
```
nono: You can remove this line

pkg/consts/views.go
```diff
@@ -55,6 +55,19 @@ function(doc) {
 }`,
 }
 
+// FilesByParentView is the view used for fetching files referenced by a
+// given document
+// TODO change me for flag hidden
+var FilesByParentView = &couchdb.View{
+	Name:    "by-parent-type-name",
```
nono: Is the `dir-file-child` index still used?

pkg/vfs/couchdb_indexer.go
```diff
@@ -271,3 +272,60 @@ func (c *couchdbIndexer) DirOrFileByPath(name string) (*DirDoc, *FileDoc, error)
...
+		}
+		docs[i] = doc
+	}
+
+	return docs, nil
+}
+
+func (c *couchdbIndexer) DirLength(doc *DirDoc) (int, error) {
+
+	// TODO change me for flag hidden, depending if doc is Trash
```
nono: I don't see what to change in this method for the hidden flag. Can you enlighten me?

web/files/paginated.go
```diff
@@ -108,50 +79,73 @@ func dirData(c echo.Context, statusCode int, doc *vfs.DirDoc) error {
...
 func dirDataList(c echo.Context, statusCode int, doc *vfs.DirDoc) error {
-	included := make([]jsonapi.Object, 0)
 
-	count, iterOpts, err := paginationConfig(c)
+	cursor, err := jsonapi.ExtractPaginationCursor(c, defPerPage)
+	if err != nil {
+		return err
+	}
+
+	inst := middlewares.GetInstance(c)
```
nono: It looks like you are just using the VFS on inst, so we can be put in a variable:

```go
fs := middlewares.GetInstance(c).VFS()
```

pkg/consts/views.go
aenario: Yes, it is used for the iterator, which is used for all operation within the stack where we iterate over the children of a folder.

pkg/vfs/couchdb_indexer.go
aenario: Yep, did not think this through, thought we were going to need to include the hidden field in the index, so we can remove them from view result, but there is no reason to do it when all hidden files are descendant of io.cozy.trash-dir. Removing comments.

pkg/consts/views.go
nono: Won't it be more practical to use the new view for iteraring? If yes, we can keep this for later (just add a TODO, so that we don't forget)

jinroh: I also think it would be better to use the view for iterating too. Maybe a TODO and an issue to not forget.

pkg/couchdb/cursor.go
```diff
@@ -1,19 +1,96 @@
...
+type SkipCursor struct {
+	*baseCursor
+	// Skip is the number of elements to start from
+	Skip int
+}
+
+// ApplyTo applies the cursor to a ViewRequest
+// the transformed ViewRequest will retrieve elements from Cursor to
+// Limit or EndKey whichever comes first
+// /!\ Mutates req
```
jinroh: Maybe making the `ApplyTo` method not return a `ViewRequest` would make it clearer that it is a "mutating" method ?

PR #475, opened 2017-04-12
Centralized FS for storing applications and konnectors

jinroh:
This PR is an intermediate work into making complete execution pipeline for konnectors working. Since the code change is already quite important, I'm making this PR right now.

What it contains:

--------------------

1/ The first commit make the job system independent from the "domain" of the instance.

For now, we had one "job system" for each instance: each (domain, worker) couple had its own queue. This was mainly done because we had only one "in-memory" model of this job system for single nodes. In the case of multiple nodes, it will be more useful to have shared queues for workers where the domain information is stored on the job message, rather that having as many queues as users.

It also removes the code that had to iterate over all the instances on startup to initiate all the queues.

--------------------

2/ The second part of the commits implement a centralized/versioned storage for applications and konnectors

The implementation is not yet complete, and the swift storage is not yet tested.

It contains two new interfaces to abstract how application files are copied and served (for a webapp): `apps.Copier` and `app.FileServer`.

The Copier implementation copy application in a `${slug}/${version}/${file}` path so that we can de-duplicate application data by version. In the git fetcher, the version is patched with the shasum of the current commit: `${slug}/${version}-${shasum}/${file}`. I removed the storage of the `.git` directory which is now only temporary and always recreated with a fresh clone which simplify many things.

For konnectors, a tarball is created containing all the file and then copied on the application storage.

The `Installer` methods have been cleaned up also, by using a single `Run` method, rather that `Install/Update/Delete`

### DIFF
--- a/pkg/instance/instance.go
+++ b/pkg/instance/instance.go
@@ -154,30 +154,31 @@ func (i *Instance) makeVFS() error {
 	return err
 }
 
-// AppsFS returns the hidden filesystem associated with the specified
+// AppsCopier returns the application copier associated with the specified
 // application type
-func (i *Instance) AppsFS(appsType apps.AppType) afero.Fs {
-	switch appsType {
-	case apps.Webapp:
-		return i.hiddenFS(vfs.WebappsDirName)
-	case apps.Konnector:
-		return i.hiddenFS(vfs.KonnectorsDirName)
-	}
-	panic(fmt.Errorf("Unknown application type %s", string(appsType)))
+func (i *Instance) AppsCopier(appsType apps.AppType) apps.Copier {
+	// switch appsType {
+	// case apps.Webapp:
+	// 	return i.hiddenFS(vfs.WebappsDirName)
+	// case apps.Konnector:
+	// 	return i.hiddenFS(vfs.KonnectorsDirName)
+	// }
+	// panic(fmt.Errorf("Unknown application type %s", string(appsType)))
+	return nil
+}
+
+func (i *Instance) AppsFileServer(appsType apps.AppType) apps.FileServer {
+	return nil
 }
 
 // ThumbsFS returns the hidden filesystem for storing the thumbnails of the
 // photos/image
 func (i *Instance) ThumbsFS() afero.Fs {
-	return i.hiddenFS(vfs.ThumbsDirName)
-}
-
-func (i *Instance) hiddenFS(dirname string) afero.Fs {
 	fsURL := config.FsURL()
 	switch fsURL.Scheme {
 	case "file", "mem":
 		return afero.NewBasePathFs(afero.NewOsFs(),
-			path.Join(fsURL.Path, i.Domain, dirname))
+			path.Join(fsURL.Path, i.Domain, vfs.ThumbsDirName))
 	case "swift":
 		panic("Not implemented")
 	}
@@ -247,7 +248,7 @@ func (i *Instance) installApp(slug string) error {
 	if !ok {
 		return errors.New("Unknown app")
 	}
-	inst, err := apps.NewInstaller(i, i.AppsFS(apps.Webapp), &apps.InstallerOptions{
+	inst, err := apps.NewInstaller(i, i.AppsCopier(apps.Webapp), &apps.InstallerOptions{
 		Operation: apps.Install,
 		Type:      apps.Webapp,
 		SourceURL: source,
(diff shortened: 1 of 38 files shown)


pkg/instance/instance.go
```diff
@@ -26,6 +26,9 @@ import (
 	"github.com/leonelquinteros/gotext"
 	"github.com/spf13/afero"
 	jwt "gopkg.in/dgrijalva/jwt-go.v3"
+
+	// import the mail worker to test passphrase reset
+	_ "github.com/cozy/cozy-stack/pkg/jobs/workers/mails"
```
nono: If it is just for test, why don't you put it in the test file?

pkg/instance/instance.go
```diff
@@ -154,66 +157,44 @@ func (i *Instance) makeVFS() error {
...
-	switch fsURL.Scheme {
-	case "file", "mem":
-		return afero.NewBasePathFs(afero.NewOsFs(),
-			path.Join(fsURL.Path, i.Domain, dirname))
-	case "swift":
-		panic("Not implemented")
-	}
-	return nil
+	return afero.NewBasePathFs(afero.NewOsFs(),
+		path.Join(fsURL.Path, i.Domain, vfs.ThumbsDirName))
```
nono: Currently, we still support only local file system for apps, konnectors, and thumbs. Right?

pkg/jobs/couch_storage.go
```diff
@@ -21,24 +21,22 @@ func (t triggerDoc) MarshalJSON() ([]byte, error) {
 	return json.Marshal(t.t.Infos())
 }
 
-// CouchStorage implements the TriggerStorage interface and uses CouchDB as the
+// CouchStorage cmplements the TriggerStorage interface and uses CouchDB as the
```
nono: typo?

pkg/jobs/couch_storage.go
```diff
@@ -21,24 +21,22 @@ func (t triggerDoc) MarshalJSON() ([]byte, error) {
...
 // underlying storage for triggers.
-type CouchStorage struct {
-	db couchdb.Database
+type couchStorage struct {
 }
 
-// NewTriggerCouchStorage returns a new instance of CouchStorage using the
+// NewTriggerCouchStorage ceturns a new instance of CouchStorage using the
```
nono: typo?

pkg/jobs/couch_storage.go
```diff
@@ -21,24 +21,22 @@ func (t triggerDoc) MarshalJSON() ([]byte, error) {
...
 
-// GetAll implements the GetAll method of the TriggerStorage.
-func (s *CouchStorage) GetAll() ([]*TriggerInfos, error) {
+func (s *couchStorage) GetAll() ([]*TriggerInfos, error) {
 	var infos []*TriggerInfos
 	// TODO(pagination): use a sort of couchdb.WalkDocs function when available.
-	req := &couchdb.AllDocsRequest{Limit: 100}
-	if err := couchdb.GetAllDocs(s.db, consts.Triggers, req, &infos); err != nil {
+	req := &couchdb.AllDocsRequest{Limit: 1000}
+	if err := couchdb.GetAllDocs(couchdb.GlobalDB, consts.Triggers, req, &infos); err != nil {
```
nono: I don't really like the idea of putting the triggers from all instances in the same databases (for jobs, I care less). At least, we should use a different database than for instances (`couchdb.GlobalTriggersDB`)

pkg/instance/instance.go
jinroh: Yes, coming next and soon.

pkg/jobs/couch_storage.go
jinroh: I agree this is not ideal, and I'll make this specific DB. Note that this storage will probably only be used for self-hosted instances, since it is not suitable for distributed triggers.

aenario:
In the [konnector workflow doc](https://github.com/cozy/cozy-stack/blob/master/docs/konnectors_workflow_example.md), we use triggers as links between konnectors, accounts & recurrence `@cron` . 

This means :

A - We need to be able to list all triggers for an instance, this can be implemented as a special route.
B - When exporting / moving instance, we willl need to copy triggers as well.

jinroh:
@aenario:

A - is already available on `GET /jobs/triggers`
B - noted

aenario:
A - Oups, my bad

jinroh:
rebased on master

PR #476, opened 2017-04-12
A recipient can receive a sharing.

Ljinod:
This PR adds a route `/sharings/doc/:doctype/:id` that will handle the
documents sent by the sharer. Depending on the doctype either the
request is transmitted to `/data` or our custom handler (for the
creation of a file/dir with an id) is called.

This route was added because we need to modify the parent of an incoming
file/dir and set it to "Shared With Me" as we don't want to propagate
the entire hierarchy with a document.

We created a custom handler that heavily resembles that of files because
we need to be able to predefine the id of a file/dir for synchronisation
purposes. To do so we had to add two new functions to the interface of
the indexer and update the vfs implementation.

The list of changes is:
* consts.go: add `SharedWithMe` constant.
* share_data.go: the functions `SendDoc` and `SendFile` now call the
  correct route on the recipient's Cozy.
* sharings_test.go: it is no longer possible to check on the recipient's
  test server that the sharing went well. The route called to share a
  document is in web/sharings and we can't add it to the test server
  handler as it would create a dependency cycle. This is a *temporary*
  patch as fixing everything is out of the scope of this PR.
* couchdb_indexer.go: add the functions `CreateNamedFileDoc` and
  `CreateNamedDirDoc` to be able to create a file/dir with an id.
* vfs.go: update the interface of the indexer to include the functions
  `CreateNamedFileDoc` and `CreateNamedDirDoc`.
* vfsafera/impl.go: When creating a file or a directory if an ID is
  already present in the document then the corresponding `Named*Doc`
  function is called.
* vfsswift/impl.go: (same as above).
* web/files/files.go: to ease the creation of a file/dir with a given id
  the function `FileDocFromReq` was turned public so that I can access it
  from the package Sharings.
* web/sharings/files.go: creation of a file or dir with an id, plus
  creation of the dir "Shared With Me" if it doesn't exist.
* web/sharings/sharings.go: add the route `/sharings/doc/:doctype/:id`
  and its handler function `receiveDocument`.
* web/sharings/sharings_test.go: add tests for the newly added route.

### DIFF
--- a/pkg/sharings/sharings_test.go
+++ b/pkg/sharings/sharings_test.go
@@ -277,48 +277,54 @@ func acceptedSharing(t *testing.T, sharingType string, isFile, withSelector bool
 	assert.NotNil(t, recStatus.AccessToken.AccessToken)
 	assert.NotNil(t, recStatus.AccessToken.RefreshToken)
 
-	// TODO Refactoring necessary.
+	// TODO Refactoring necessary: it is no longer possible to actually share
+	// between two tests servers as the route to send the documents is declared
+	// in web/sharings. Hence checking if the documents do exist will only
+	// result in failures.
+	t.SkipNow()
+
 	// Check updates in case of master-* sharing
-	// if sharingType == consts.MasterSlaveSharing ||
-	// 	sharingType == consts.MasterMasterSharing {
+	if sharingType == consts.MasterSlaveSharing ||
+		sharingType == consts.MasterMasterSharing {
 
-	// 	// Wait for the document to arrive and check it
-	// 	time.Sleep(2000 * time.Millisecond)
+		// Wait for the document to arrive and check it
+		time.Sleep(2000 * time.Millisecond)
 
-	// 	if !isFile {
-	// 		if withSelector {
-	// 			var testDoc2 *couchdb.JSONDoc
-	// 			testDoc2, err = createTestDoc(t)
-	// 			assert.NoError(t, err)
-	// 			assert.NotNil(t, testDoc2)
-	// 			// Wait for the document to arrive and check it
-	// 			time.Sleep(2000 * time.Millisecond)
-	// 			recDoc := &couchdb.JSONDoc{}
-	// 			err = couchdb.GetDoc(recipientIn, testDocType, testDoc2.ID(), recDoc)
-	// 			assert.NoError(t, err)
-	// 			recDoc.Type = testDocType
-	// 			assert.Equal(t, testDoc2, recDoc)
-	// 		} else {
-	// 			updKey := "test"
-	// 			updVal := "update me!"
-	// 			updateTestDoc(t, testDoc, updKey, updVal)
-	// 			// Wait for the document to arrive and check it
-	// 			time.Sleep(2000 * time.Millisecond)
-	// 			recDoc := &couchdb.JSONDoc{}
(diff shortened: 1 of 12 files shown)


pkg/sharings/sharings_test.go
```diff
@@ -268,49 +277,48 @@ func acceptedSharing(t *testing.T, sharingType string, isFile, withSelector bool
...
-				Name: &newFileName,
-			}
-			updateTestFile(t, testDocFile, patch)
-
-			// TODO check the file on the recipient side when we'll be able
-			// to create files with fixed id
-		}
-
-	}
+	// if sharingType == consts.MasterSlaveSharing ||
```
aenario: It's better not to leave dead code in master. If you need to keep the previous version for future implementation, you can use a github link : https://github.com/cozy/cozy-stack/blob/99da7f41319aa51c606b3a0b24af0fca545ed000/pkg/sharings/sharings_test.go#L271-L313

web/sharings/sharings.go
```diff
@@ -211,6 +212,51 @@ func RecipientRefusedSharing(c echo.Context) error {
...
+	id := c.Param("id")
+
+	switch c.Param("doctype") {
+	case "":
+		err = echo.NewHTTPError(http.StatusBadRequest,
+			"Missing parameter: doctype")
+	case consts.Files:
+		err = creationWithIDHandler(c)
+	default:
+		// We call PUT /data/:doctype/:docid
```
aenario: Why not just   `err = data.updateDoc(c)` ?

pkg/sharings/sharings_test.go
nono: For a test, it's also possible to skip it: https://golang.org/pkg/testing/#T.Skip

pkg/vfs/vfs.go
```diff
@@ -94,6 +94,9 @@ type Indexer interface {
 
 	// CreateFileDoc creates and add in the index a new file document.
 	CreateFileDoc(doc *FileDoc) error
+	// CreateNamedFileDoc creates and add in the index a new file document with
+	// its id already set.
+	CreateNamedFileDoc(doc *FileDoc) error
```
jinroh: Rather than adding a new method in this interface (which is already pretty big...), why not handling the case of a document with a non-empty `ID()` in the `CreateFileDoc()` method directly ?

jinroh: Also, are we sure we won't have collisions between couchdb identifiers form two different instances ?

web/sharings/sharings.go
Ljinod: Because I thought it would be best not to change other parts of the code (the function `updateDoc` muste be turned public). If that's ok I'll do it, it is way easier indeed!

pkg/vfs/vfs.go
Ljinod: I asked what would be the best solution to do so and the reply I got was to augment the interface with two dedicated methods. Hence this.
I can also do what you just said both solutions are fine to me.

Regarding the collisions this is an issue I was not aware of: what you mean is that two instances cannot have a document with the same id? And the collision would occurr at which level? CouchDB or the underlying filesystem?

jinroh: Let's leave it like that, it's ok for me.

I was wrong about the collisions, CouchDB should provide IDs with small collision probabilities.

Ljinod: Okay!

PR #478, opened 2017-04-13
Returns the origin in the client field for intents

nono:


web/intents/intents_test.go
```diff
@@ -40,7 +40,7 @@ func checkIntentResult(t *testing.T, res *http.Response) {
 	attrs := data["attributes"].(map[string]interface{})
 	assert.Equal(t, "PICK", attrs["action"].(string))
 	assert.Equal(t, "io.cozy.files", attrs["type"].(string))
-	assert.Equal(t, "app.cozy.example.net", attrs["client"].(string))
+	assert.Equal(t, "https://app.cozy.example.net/", attrs["client"].(string))
```
gregorylegarec: Actually the `origin` property of a `MessageEvent` does not end with a slash. It something like `https://app.cozy.example.net`. Is it possible to strip it ? Everything should be ok afterward.

PR #483, opened 2017-04-13
Docs retrieval at dynamic sharing creation

Gara64:
At the dynamic sharing creation, one needs to create and request a mango index based on the selector to retrieve all the docs concerned by this sharing, before being sent to the recipients.

### DIFF
--- a/pkg/sharings/sharings.go
+++ b/pkg/sharings/sharings.go
@@ -207,7 +207,8 @@ func ShareDoc(instance *instance.Instance, sharing *Sharing, recStatus *Recipien
 		// Dynamic sharing
 		if rule.Selector != "" {
 			// Create index based on selector to retrieve documents to share
-			index := mango.IndexOnFields(docType, sharing.ID(), []string{rule.Selector})
+			indexName := "by-" + rule.Selector
+			index := mango.IndexOnFields(docType, indexName, []string{rule.Selector})
 			err := couchdb.DefineIndex(instance, index)
 			if err != nil {
 				return err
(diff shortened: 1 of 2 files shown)


pkg/sharings/sharings.go
```diff
@@ -204,8 +202,39 @@ func ShareDoc(instance *instance.Instance, sharing *Sharing, recStatus *Recipien
 			}
 		}
 
+		var values []string
+
+		// Dynamic sharing
+		if rule.Selector != "" {
+			// Create index based on selector to retrieve documents to share
+			index := mango.IndexOnFields(docType, sharing.ID(), []string{rule.Selector})
```
aenario: Use "by-"+rule.Selector as index name, otherwise it will create a lot of useless indexes.

pkg/sharings/sharings.go
```diff
@@ -204,8 +202,39 @@ func ShareDoc(instance *instance.Instance, sharing *Sharing, recStatus *Recipien
...
+			err := couchdb.DefineIndex(instance, index)
+			if err != nil {
+				return err
+			}
+
+			var docs []couchdb.JSONDoc
+
+			// Request the index for all values
+			for _, val := range rule.Values {
```
aenario: You can also use mango.Or(mango.Equal(), mango.Equal(), ...) to combine all values into a single a couchdb request.

Gara64: I tried that, but I receive a `CouchDB(no_usable_index): There is no index available for this selector.`.
I pass a `Filter` slice, as a parameter : `mango.Or(filters...)`, I don't know if I do something wrong.

pkg/sharings/sharings.go
Gara64: Right, but I guess I have to handle the case where the index already exists, which will return error.
Or maybe there is a way to query the existing indexes ?

aenario: Index creation is idempotent, it does not return an error if the index already exists.

pkg/sharings/sharings.go
aenario: Oups my bad, I thought couchdb was smart enough to transform a mango or into a keys request, looks like it does not... 
Well let's keep it like that, we can change it to a mapreduce request if it proves to be a bottleneck.

Gara64: Fair  enough, I added a note in the comment related to that

PR #485, opened 2017-04-14
Konnectors exec

jinroh:


pkg/jobs/workers/konnectors/konnector_test.go
```diff
@@ -84,6 +84,8 @@ func TestBadFileExec(t *testing.T) {
 }
 
 func TestSuccess(t *testing.T) {
+	t.Skip()
```
nono: I think you can use env variables (`TRAVIS=true` is set by default on travis) to disable the test only on travis

PR #487, opened 2017-04-14
Fix redirect value for apps redirect

jinroh:


web/apps/serve.go
```diff
@@ -104,8 +104,11 @@ func ServeAppFile(c echo.Context, i *instance.Instance, fs apps.FileServer, app
 		if file != "" {
 			return echo.NewHTTPError(http.StatusUnauthorized, "You must be authenticated")
 		}
+		reqURL := c.Request().URL
 		subdomain := i.SubDomain(slug)
-		subdomain.Path = c.Request().URL.String()
+		subdomain.Path = reqURL.Path
+		subdomain.RawQuery = reqURL.RawQuery
+		subdomain.Fragment = reqURL.Fragment
```
nono: FYI, the browsers don't send the fragment to the server, so the last line doesn't do anything (but it does no harm too)

PR #493, opened 2017-04-18
[bugfix] Wrong included on directory in some cases

aenario:


web/files/paginated.go
```diff
@@ -125,13 +129,16 @@ func dirDataList(c echo.Context, statusCode int, doc *vfs.DirDoc) error {
 		return err
 	}
 
-	included := make([]jsonapi.Object, len(children))
-	for i, child := range children {
+	included := make([]jsonapi.Object, 0)
```
jinroh: nit: you could use the `len(children)` as the cap of the `included` array since it is still a good estimate of its final size (for a +/- 1 offset)

PR #496, opened 2017-04-19
Make travis more resilient

aenario:


aenario:
Removing noise, will make another PR when there is time.

PR #504, opened 2017-04-24
Add CLI to list instaled applications

jinroh:


cmd/apps.go
```diff
@@ -154,6 +154,29 @@ var uninstallAppCmd = &cobra.Command{
 	},
 }
 
+var lsAppsCmd = &cobra.Command{
+	Use:     "ls",
+	Short:   "List the installed applications.",
+	Aliases: []string{"rm"},
```
nono: I don't think this is a good alias :wink:

PR #507, opened 2017-04-24
Allow launching cozy-app-dev without application dir

jinroh:


scripts/Dockerfile
```diff
@@ -56,22 +56,22 @@ ENV MAILHOG_VERSION 0.2.1
 ENV MAILHOG_SRC_URL https://github.com/mailhog/MailHog/releases/download/v$MAILHOG_VERSION/MailHog_linux_amd64
 ENV MAILHOG_SRC_SHA256 25b3cd56e7a1042b695fed8b32808eef778eafa8fb6bd583dde90a8fc2877e0e
 
+RUN curl -fsSL "$MAILHOG_SRC_URL" -o /usr/bin/MailHog \
```
nono: You have splitted the `RUN` in two. I think we had made the inverse to limit the number of docker layers in the past. It wasn't a good idea?

jinroh: You're right, but since we won't touch the MailHog versions as often as the cozy-stack one, it is a valid layer... Also that way we don't have to fetch it every time we rebuild it with a different stack.

PR #509, opened 2017-04-24
Beginning of a doc for redis usage in stack

aenario:
This PR aims to be a spot for persistent discussion on our several redis usage.

### DIFF
--- a/docs/redis.md
+++ b/docs/redis.md
@@ -1,0 +1,34 @@
+# Usage of Redis for cozy-stack
+
+
+It will be important to figure out if we use a cluster of Redis or only one.
+Assuming a cluster allows us to scale better in the future but for now we will focus on a single redis target for the stack.
+
+## As a cache / store (KV)
+
+Already implemented in https://github.com/cozy/cozy-stack/blob/master/pkg/cache/cache.go
+
+To use with sessions / download manager
+
+
+## As a simple lock
+
+Usage ?
+
+ref
+https://redis.io/topics/distlock
+(algorithm for cluster)
+
+
+## As a FS lock (range+RW)
+
+Using a Range+RW lock is an optimization, it would be acceptable to use a simple lock instead to start.
+
+Sortedset & lexical sort can be used to implement a simple range-lock
+More complex logic would allow to implement range+RW lock
+Lua can be used to ensure
+
+
+## As a jobs/trigger/event queue
+
+sortedset by timestamp ?


docs/redis.md
```diff
@@ -0,0 +1,34 @@
+# Usage of Redis for cozy-stack
+
+
+It will be important to figure out if we use a cluster of Redis or only one.
+Assuming a cluster allows us to scale better in the future but for now we will focus on a single redis target for the stack.
```
nono: We don't to shard the data in redis for now (and probably for a long time). What we need is high availability, and it is given by master/slave. From a developer point of view, it's the same as a single node. The work will be for our devops.

Also, I'd like to have configuration to use separate redis instances for different use cases:

- one redis instance (well, one master and one slave) for caching
- one for locks
- one for the jobs system
- etc.

docs/redis.md
```diff
@@ -0,0 +1,34 @@
...
+Usage ?
+
+ref
+https://redis.io/topics/distlock
+(algorithm for cluster)
+
+
+## As a FS lock (range+RW)
+
+Using a Range+RW lock is an optimization, it would be acceptable to use a simple lock instead to start.
```
nono: - Range: I don't want of this optimization. It looks like what we can gain doesn't worth its complexity. Except, of course, if a benchmark shows that I'm wrong. But let's forget it for now.

- RW: it looks like it's complicated to implement, but I think it has the potential to really improve the performances (mainly by allowing several readers at the same time). Let's keep it for later.

docs/redis.md
```diff
@@ -0,0 +1,34 @@
...
+
+Sortedset & lexical sort can be used to implement a simple range-lock
+More complex logic would allow to implement range+RW lock
+Lua can be used to ensure
+
+
+## As a jobs/trigger/event queue
+
+sortedset by timestamp ?
```
nono: Yes, there will be a sorted set by timestamps for the `@cron`/`@at`/`@in` triggers. But also, some lists for the waiting jobs, and some structs for runnings jobs/triggers (to retry after a timeout).

docs/redis.md
aenario: @nono for Range, the main gain I see is in extreme cases: I move my whole Picture folder and its 5k files to Trash. As an user, I expect the picture & trash folder to just give me a spinner for 5 minute. I do not expect I cant read my Documents folder. 

Anyway, I agree it will wait.

PR #515, opened 2017-04-25
Add Redis-based simple lock on VFS

aenario:


pkg/lock/simple_mem.go
```diff
@@ -0,0 +1,30 @@
+package lock
+
+import "sync"
+
+var locks map[string]*memLock
```
nono: What is the purpose of the `memlock` struct? `var locks map[string]sync.Mutex` is not enough?

pkg/lock/client.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/go-redis/redis"
+)
+
+type subRedisInterface interface {
+	SetNX(key string, value interface{}, expiration time.Duration) *redis.BoolCmd
+	Eval(script string, keys []string, args ...interface{}) *redis.Cmd
+}
+
+var mu sync.Mutex
+var globalRedisClient subRedisInterface
```
nono: Is is possible to initialize globalRedisClient in `init()`? If yes, we can avoid having a mutex, and `mu.Lock()` + `mu.Unlock()` a lot of times if we are not using redis.

pkg/lock/lock.go
```diff
@@ -0,0 +1,22 @@
+package lock
+
+// ReadWrite returns a read/write lock for the given name
+func ReadWrite(domain string) ErrorRWLocker {
+	if c := getClient(); c != nil {
```
nono: As said before, if redis is not configured, calling `getClient` each time is costly for nothing.

pkg/lock/simple_redis.go
```diff
@@ -0,0 +1,134 @@
...
+	WaitTimeout = 2 * time.Minute
+
+	// WaitRetry time to wait between retries
+	WaitRetry = 100 * time.Millisecond
+)
+
+var (
+	// ErrTooManyRetry is the error returned when despite several tries
+	// we never managed to get a lock
+	ErrTooManyRetry = errors.New("too many retry")
```
nono: ErrTooManyRetries?

pkg/lock/simple_redis.go
```diff
@@ -0,0 +1,134 @@
...
+)
+
+type redisLock struct {
+	client subRedisInterface
+	mu     sync.Mutex
```
nono: What is the purpose of this mutex?

pkg/lock/simple_redis.go
```diff
@@ -0,0 +1,134 @@
...
+		}
+		time.Sleep(WaitRetry)
+	}
+
+	return ErrTooManyRetry
+}
+
+func (rl *redisLock) Unlock() error {
+	rl.mu.Lock()
+	defer func() { rl.token = "" }()
```
nono: Why do you defer this?

pkg/vfs/vfsafero/impl.go
```diff
@@ -70,7 +71,10 @@ func New(index vfs.Indexer, disk vfs.DiskThresholder, mu vfs.Locker, fsURL *url.
 // Init creates the root directory document and the trash directory for this
 // file system.
 func (afs *aferoVFS) InitFs() error {
-	afs.mu.Lock()
+	lockerr := afs.mu.Lock()
+	if lockerr != nil {
+		return lockerr
+	}
```
nono: [Nitpick] This pattern is repeated a lot. I'd prefer a more compact form:

```go
if lockerr := afs.mu.Lock(); lockerr != nil {
    return err
}
```

pkg/lock/lock.go
```diff
@@ -0,0 +1,22 @@
...
+		return &fakeRWLock{getReadisReadWriteLock(c, domain)}
```
nono: getRe**a**disReadWriteLock

pkg/lock/simple_redis.go
```diff
@@ -0,0 +1,134 @@
...
+
+const luaRefresh = `if redis.call("get", KEYS[1]) == ARGV[1] then return redis.call("pexpire", KEYS[1], ARGV[2]) else return 0 end`
+const luaRelease = `if redis.call("get", KEYS[1]) == ARGV[1] then return redis.call("del", KEYS[1]) else return 0 end`
+
+type fakeRWLock struct {
+	ErrorLocker
+}
+
+func (l fakeRWLock) Lock() error    { return l.ErrorLocker.Lock() }
+func (l fakeRWLock) RLock() error   { return l.ErrorLocker.Lock() }
```
jinroh: It may be dangerous to have `RLock` behave like `Lock` since calling `RLock` recursively will deadlock while it is a perfectly valid use of a `RWLock`. I'm not sure that this would not cause a deadlock in our current code, it would require further analysis and tests.

One option would be to have `RLock` be a noop. The lock would prevent corruption of the FS by locking the writes but would not assure the reads to be fail-safe, which may be sufficient in our case.

pkg/lock/client.go
aenario: It is not possible as the config is not loaded in init. 
To be included in the RedisClient refactoring card.

pkg/lock/simple_mem.go
aenario: To implement the same ErrorLogger interface, with Lock() possibly returning an error.

pkg/lock/client.go
nono: OK

pkg/lock/simple_redis.go
aenario: Protecting the sharedState rl.token.

PR #516, opened 2017-04-25
Register the recipient for master-master sharing

Gara64:
In a Master-Master sharing context, the recipient must register an OAuth client to the sharer, as it will be necessary for the recipient to have access on the sharer docs.
This PR implements this registration, just after the sharing request is received on the recipient side. It then sends the generated `client_id` to the sharer, for him to save it in his sharing document.

Note this relies on #502

### DIFF
--- a/pkg/sharings/send_mails.go
+++ b/pkg/sharings/send_mails.go
@@ -43,54 +43,58 @@ func SendSharingMails(instance *instance.Instance, s *Sharing) error {
 
 	errorOccurred := false
 	for _, rs := range s.RecipientsStatus {
-		// Sanity check: recipient is private.
-		if rs.recipient == nil {
-			rs.recipient, err = GetRecipient(instance, rs.RefRecipient.ID)
-			if err != nil {
-				return err
+
+		// A non-empty status indicates a recipient already treated
+		if rs.Status != "" {
+			// Sanity check: recipient is private.
+			if rs.recipient == nil {
+				rs.recipient, err = GetRecipient(instance, rs.RefRecipient.ID)
+				if err != nil {
+					return err
+				}
 			}
-		}
 
-		// Generate recipient specific OAuth query string.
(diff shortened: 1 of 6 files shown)


pkg/sharings/send_mails.go
```diff
@@ -43,54 +43,58 @@ func SendSharingMails(instance *instance.Instance, s *Sharing) error {
...
-		// Sanity check: recipient is private.
-		if rs.recipient == nil {
-			rs.recipient, err = GetRecipient(instance, rs.RefRecipient.ID)
-			if err != nil {
-				return err
+
+		// A non-empty status indicates a recipient already treated
+		if rs.Status != "" {
+			// Sanity check: recipient is private.
+			if rs.recipient == nil {
```
aenario: this code is repeated, might be worth it to make it a method on rs.

pkg/sharings/sharings.go
```diff
@@ -408,10 +410,77 @@ func CreateSharingRequest(db couchdb.Database, desc, state, sharingType, scope,
...
 }
 
-// CreateSharingAndRegisterSharer checks the sharing, creates the document in
+// RegisterRecipient registers a sharing recipient
+func RegisterRecipient(instance *instance.Instance, rs *RecipientStatus) error {
+	err := rs.Register(instance)
+	if err != nil {
+		if rs.recipient != nil {
+			log.Error("[sharing] Could not register at "+
+				rs.recipient.URL+" ", err)
```
jinroh: nit: FYI, you can use `log.Errorf` for this kind of format.

pkg/sharings/send_mails.go
Gara64: Yup, I just added the method

pkg/sharings/sharings.go
Gara64: Okay, thanks for the tip

PR #519, opened 2017-04-26
Allows manipulating docs with slash in _id from /data

aenario:
Needed for the front to access data about a konnector or an app.


gregorylegarec:
🎉

PR #522, opened 2017-04-26
Create a download store backed by redis

jinroh:


pkg/vfs/download_store.go
```diff
@@ -28,119 +26,165 @@ const downloadStoreTTL = 1 * time.Hour
 // cleanup.
 const downloadStoreCleanInterval = 1 * time.Hour
 
-var storeStoreMutex sync.Mutex
-var storeStore map[string]*memStore
+var globalStoreMu sync.Mutex
+var globalMemStore *memStore
```
aenario: Why using 2 global variables with actual types instead of a DownloadStore ?

PR #523, opened 2017-04-26
Redis sessions code store

aenario:


pkg/sessions/code_storage.go
```diff
@@ -0,0 +1,98 @@
...
+
+type redisCodeStorage struct {
+	cl subRedisInterface
+}
+
+func (store *redisCodeStorage) Add(c *Code) error {
+	return store.cl.Set(c.Value+":"+c.AppHost, c.SessionID, CodeTTL).Err()
+}
+
+const luaGetAndDelete = `local v = redis.call("GET", KEYS[1]); redis.call("DEL", KEYS[1]); return v`
```
nono: FYI, it's possible to do that with MULTI/EXEC. But lua is OK for me too

PR #526, opened 2017-04-27
Dynamic sharing for files

Gara64:
This adds the possibility to define a permission on a `dir_id`, which is useful to share a directory through a selector.
Note this only handles one level of a hierarchy: full recursive permissions will probably require more work.


Gara64:
@nono Ok I missed that. How can we declare them?

aenario:
@Gara64 you just need to have the permissions on a folder and it will give permission for all its children. However, you need to use the vfs.Allows instead of permissions.Allow to validate them.

PR #541, opened 2017-05-02
Use a secret instead of a bearer token for thumbnails

nono:


nono:
@aenario yes, I think too. I have let the file-id in the thumbnail route, because an optimization could be to generate only one secret for all the thumbnails of the images in a folder

PR #542, opened 2017-05-02
Add album sharing support

Gara64:
This adds the possibility to define a `referenced_by` selector, which is useful for album sharing.
It sends the `references_by` attributes to the recipient, for him to check the associated permissions. 

Note that at the sharing creation, the `Title` permission field is used to know which doctype is referenced, needed to query the referenced view. I'm open to suggestions for a better way to get this doctype.

### DIFF
--- a/pkg/sharings/sharings.go
+++ b/pkg/sharings/sharings.go
@@ -210,30 +210,53 @@ func ShareDoc(instance *instance.Instance, sharing *Sharing, recStatus *Recipien
 
 		// Dynamic sharing
 		if rule.Selector != "" {
-			// Create index based on selector to retrieve documents to share
-			indexName := "by-" + rule.Selector
-			index := mango.IndexOnFields(docType, indexName, []string{rule.Selector})
-			err := couchdb.DefineIndex(instance, index)
-			if err != nil {
-				return err
-			}
 
-			var docs []couchdb.JSONDoc
+			// Particular case for referenced_by: use the existing view
+			if rule.Selector == "referenced_by" {
+				// The permission tilte is used to know the referenced doctype
+				refType := rule.Title
 
-			// Request the index for all values
-			// NOTE: this is not efficient in case of many Values
-			// We might consider a map-reduce approach in case of bottleneck
-			for _, val := range rule.Values {
-				err = couchdb.FindDocs(instance, docType, &couchdb.FindRequest{
-					UseIndex: indexName,
-					Selector: mango.Equal(rule.Selector, val),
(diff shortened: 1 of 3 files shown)


pkg/sharings/sharings.go
```diff
@@ -210,30 +210,53 @@ func ShareDoc(instance *instance.Instance, sharing *Sharing, recStatus *Recipien
...
-			err := couchdb.DefineIndex(instance, index)
-			if err != nil {
-				return err
-			}
 
-			var docs []couchdb.JSONDoc
+			// Particular case for referenced_by: use the existing view
+			if rule.Selector == "referenced_by" {
+				// The permission tilte is used to know the referenced doctype
+				refType := rule.Title
```
aenario: Not sure how it's done in sharing context, but in application context a rule's Title can be whatever the dev writes into it.

web/sharings/files.go
```diff
@@ -77,6 +79,14 @@ func createFileWithIDHandler(c echo.Context, fs vfs.VFS) error {
 	doc.SetID(c.Param("docid"))
 	doc.DirID = consts.SharedWithMeDirID
 
+	refBy := c.QueryParam("Referenced_by")
+	var refs = &[]couchdb.DocReference{}
+	b := []byte(refBy)
+	if err = json.Unmarshal(b, refs); err != nil {
```
aenario: Would be more readable as 
```
var refs = []couchdb.DocReference{}
json.Unmarshal(... , &refs)
doc.ReferencedBy = refs
```

pkg/sharings/sharings.go
Gara64: Yes I know, I'm not really satisfied by this 'trick', but I didn't see another way to link a reference doctype with the referenced_by selector. We could add a field in the Permissions, but I'm not sure it's worth it.

aenario: Mmmh, we could change the Values format for referenced_by rules. 
The rule would become VERB:io.cozy.files:io.cozy.album/45878584848485:referenced_by.

This will need handling both in sharing and vfs.File.Valid

Gara64:
@aenario Yep that would work. The work on the sharing part would be straightforward, but I don't know for the vfs validation part :)

aenario:
@Gara64 you just need to split the expected value at https://github.com/cozy/cozy-stack/blob/master/pkg/vfs/permissions.go#L150-L156 and check both parts.

Gara64:
@aenario Okay, if you agree with that, I will do this in another PR

PR #546, opened 2017-05-03
Add cron triggers with multi-stack, via redis

nono:
I'll write documentation and make another PR with it

### DIFF
--- a/pkg/scheduler/redis_scheduler.go
+++ b/pkg/scheduler/redis_scheduler.go
@@ -67,12 +67,13 @@ func redisKey(infos *TriggerInfos) string {
 func (s *RedisScheduler) Start(b jobs.Broker) error {
 	s.broker = b
 	go func() {
-		tick := time.Tick(pollInterval)
+		ticker := time.NewTicker(pollInterval)
 		for {
 			select {
 			case <-s.stopped:
+				ticker.Stop()
 				return
-			case <-tick:
+			case <-ticker.C:
 				now := time.Now().UTC().Unix()
 				if err := s.Poll(now); err != nil {
 					log.Warnf("[Scheduler] Failed to poll redis: %s", err)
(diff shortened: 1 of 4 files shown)


pkg/scheduler/redis_scheduler.go
```diff
@@ -22,91 +22,126 @@ const TriggersKey = "triggers"
...
-return res`
-
 // pollInterval is the time interval between 2 redis polling
 const pollInterval = 1 * time.Second
 
+// luaPoll returns the lua script used for polling triggers in redis.
+// If a trigger is in the scheduling key for more than 10 seconds, it is
+// an error and we can try again to schedule it.
+const luaPoll = `
+local w = KEYS[1] - 10
```
aenario: Might be cleaner to move the constant to a separate variable. Are 10s enough for a konnector to run?

pkg/scheduler/redis_scheduler.go
```diff
@@ -22,91 +22,126 @@ const TriggersKey = "triggers"
...
 
 // Start a goroutine that will fetch triggers in redis to schedule their jobs
 func (s *RedisScheduler) Start(b jobs.Broker) error {
-	s.Broker = b
+	s.broker = b
 	go func() {
-		for range time.Tick(pollInterval) {
-			if err := s.Poll(); err != nil {
-				log.Warnf("[Scheduler] Failed to poll redis: %s", err)
+		ticker := time.NewTicker(pollInterval)
```
jinroh: you could introduce and random delay before starting this ticker to avoid synchronous polling between stacks.

pkg/scheduler/redis_scheduler.go
nono: > Might be cleaner to move the constant to a separate variable.

Well, it's complicated to have a separate int variable and having luaPoll declared as a constant.

> Are 10s enough for a konnector to run?

The 10s are for putting a job in a queue and finding the next time to scheduler the trigger, not run a job.

pkg/scheduler/redis_scheduler.go
nono: Well, the stacks won't probably started at the same milliseconds. And even if it is the case, it's not really important if their pollings are synchronous.

PR #552, opened 2017-05-03
Fix meta extraction for large file

aenario:
In some cases (large files), image.DecodeConfig will return sooner, so we closes the reader part of the Pipe. which causes a io.ErrClosedPipe on next Write, which cause loss of metadata. This fixes that.


GoOz:
Thank you all 🙏

PR #557, opened 2017-05-04
Use node 7 on travis to have npm 3 (and less errors hopefully)

nono:


nono:
It looks like travis ignore the node version when the language is `go`.

PR #564, opened 2017-05-05
Remove outdated code used to determine the use of http or https

jinroh:


client/client.go
```diff
@@ -49,17 +50,14 @@ type Client struct {
...
 	auth   *auth.Request
 }
 
-func (c *Client) init() error {
-	c.initMu.Lock()
-	defer c.initMu.Unlock()
-	if c.inited {
-		return nil
+func (c *Client) init() {
+	if !atomic.CompareAndSwapInt32(&c.inited, 0, 1) {
```
nono: I'm a bit lost here. Do you expect that a client can be used from several goroutines?

If yes, I think we should keep the mutex to avoid race condition on using a client that has not finished to initialize to make requests.
If no, I don't see why we bother to use `sync/atomic`.

jinroh: You're right, I did a revert but the code wasn't right. Should be fixed 61932cf28ae20fe17fb56eb315a3b8a7afb2a703

PR #566, opened 2017-05-05
Add structure for konnectors oauth

aenario:


pkg/accounts/wellknown_type.go.txt
```diff
@@ -0,0 +1,14 @@
...
+	ClientID:              "xxxxx",
+	ClientSecret:          "xxxxx",
+	AuthEndpoint:          "https://accounts.google.com/o/oauth2/v2/auth",
+	TokenEndpoint:         "https://www.googleapis.com/oauth2/v4/token",
+	RegisteredRedirectURI: "_oauth_redirect.xxxxx",
+	ExtraAuthQuery: map[string]string{
+		"access_type":            "offline",
+		"include_granted_scopes": "true",
+	},
+}
```
nono: What is this file?

web/routing.go
```diff
@@ -156,6 +157,7 @@ func SetupRoutes(router *echo.Echo) error {
 		middlewares.LoadSession,
 	}
 	router.GET("/", auth.Home, mws...)
+	oauth.Routes(router.Group("/oauth", mws...))
```
nono: Can we mount these endpoints on something more specific like `/oauth-for-konnectors`? I'd like to avoid confusion with the /auth endpoints for OAuth where cozy is the service provider.

pkg/accounts/wellknown_type.go.txt
aenario: Oups, forgot to remove it.

jinroh:
@aenario I don't see any permissions limiting the access to these new routes ? Shouldn't we check at least a token from a konnector audience ?

web/routing.go
aenario: Finnaly went with `/accounts/`

aenario:
@jinroh The /redirect route is called from outside, so cant have check. The /start route checks the user is logged in and the /refresh route can be called by a konnector with GET:io.cozy.account!

PR #572, opened 2017-05-09
Add master-master sharing support for JSON documents

Gara64:
This implements an actual master-master sharing, i.e. the changes are synchronized from both the sharer and recipient.
This supports creation, updates, and deletes on JSON documents. Full files support will come later.

### DIFF
--- a/pkg/workers/sharings/share_data.go
+++ b/pkg/workers/sharings/share_data.go
@@ -158,12 +158,13 @@ func SendData(ctx context.Context, m *jobs.Message) error {
 // provided.
 func DeleteDoc(opts *SendOptions) error {
 	for _, rec := range opts.Recipients {
-		rev, err := getDocRevAtRecipient(nil, http.MethodDelete, opts.DocType, opts.DocID, rec)
+		doc, err := getDocAtRecipient(nil, opts.DocType, opts.DocID, rec)
 		if err != nil {
-			log.Error("[sharing] An error occurred while trying to send "+
-				"update : ", err)
+			log.Error("[sharing] An error occurred while trying to get "+
+				"remote doc : ", err)
 			continue
 		}
+		rev := doc.M["_rev"].(string)
 
 		_, errSend := request.Req(&request.Options{
 			Domain: rec.URL,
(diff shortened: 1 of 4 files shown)


pkg/workers/sharings/share_data.go
```diff
@@ -617,3 +637,44 @@ func getFileOrDirMetadataAtRecipient(id string, recInfo *RecipientInfo) (map[str
...
+	// Compare the md5 of the incoming doc and the existing one
+	newID := newDoc.M["_id"].(string)
+	newRev := newDoc.M["_rev"].(string)
+	delete(newDoc.M, "_id")
+	delete(newDoc.M, "_rev")
+	bNew, err := newDoc.MarshalJSON()
+	if err != nil {
+		return false, err
+	}
+	newChecksum := md5.Sum(bNew) // #nosec
```
aenario: Probably faster to use reflect.DeepEqual instead of json.Marshal+md5.Sum.

Gara64: Yep, I didn't know that, thanks.

pkg/workers/sharings/share_data.go
```diff
@@ -529,7 +543,7 @@ func generateDirOrFilePatch(dirDoc *vfs.DirDoc, fileDoc *vfs.FileDoc) (*jsonapi.
 
 // getDocAtRecipient returns the revision of the document at the given
 // recipient.
-func getDocRevAtRecipient(doctype, docID string, recInfo *RecipientInfo) (string, error) {
+func getDocRevAtRecipient(newDoc *couchdb.JSONDoc, method string, doctype, docID string, recInfo *RecipientInfo) (string, error) {
```
Ljinod: (probably nitpick) I find the name `getDocRevAtRecipient` to be no longer appropriate: this function doesn't just get the revision of the document at the recipient but also checks for changes in the context of an update… 
But that's for the sake of clarity.

Gara64: Right, I did a small refactoring to return the whole doc from this function and check the changes outside.

pkg/workers/sharings/share_data.go
```diff
@@ -218,12 +220,17 @@ func UpdateDoc(ins *instance.Instance, opts *SendOptions) error {
...
-		rev, err := getDocRevAtRecipient(opts.DocType, opts.DocID, rec)
+		remoteDoc, err := getDocAtRecipient(doc, opts.DocType, opts.DocID, rec)
 		if err != nil {
 			log.Error("[sharing] An error occurred while trying to get "+
-				"the revision of a document: ", err)
+				"remote doc : ", err)
+			continue
+		}
+		// No changes: nothing to do
+		if changes := docHasChanges(doc, remoteDoc); !changes {
```
nono: You can shorten this to:

```go
if !docHasChanges(doc, remoteDoc) {
```

pkg/workers/sharings/share_data.go
```diff
@@ -326,14 +333,22 @@ func UpdateOrPatchFile(ins *instance.Instance, opts *SendOptions, fileDoc *vfs.F
 	defer opts.closeFile()
 
 	for _, recipient := range opts.Recipients {
-		md5AtRec, rev, err := getFileMD5AndRevAtRecipient(opts.DocID, recipient)
+		// Get recipient data
+		data, err := getFileOrDirMetadataAtRecipient(opts.DocID, recipient)
 		if err != nil {
-			log.Error("[sharing] Could not get MD5 at "+recipient.URL+": ", err)
+			log.Error("[sharing] Could not get data at "+recipient.URL+": ", err)
```
nono: You can use `log.Errorf` instead of string concatenation

PR #575, opened 2017-05-10
feat: folders hierarchy in sharing

Ljinod:
This PR introduces the necessary code to propagate what we call the
"folders hierarchy": if Alice shares a folder containing subfolders then
the recipient should retrieve the same structure, and if changes are
made they should also be propagated.

In addition this PR introduces the possibility to create specific
folders other than "Shared With Me". It will be used for photos in
particular.

### DIFF
--- a/pkg/sharings/sharings.go
+++ b/pkg/sharings/sharings.go
@@ -287,6 +287,8 @@ func ShareDoc(instance *instance.Instance, sharing *Sharing, recStatus *Recipien
 
 			workerMsg, err := jobs.NewMessage(jobs.JSONEncoding, sharingWorker.SendOptions{
 				DocID:      val,
+				Selector:   rule.Selector,
+				Values:     values,
 				DocType:    docType,
 				Recipients: []*sharingWorker.RecipientInfo{rec},
 			})
--- a/pkg/sharings/sharings.go
+++ b/pkg/sharings/sharings.go
@@ -174,7 +174,7 @@ func AddTrigger(instance *instance.Instance, rule permissions.Rule, sharingID st
 
 	msg := sharingWorker.SharingMessage{
 		SharingID: sharingID,
-		DocType:   rule.Type,
+		Rule:      rule,
 	}
 	workerArgs, err := json.Marshal(msg)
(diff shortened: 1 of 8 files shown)


Ljinod:
Note for later: this PR will conflict with #572

pkg/sharings/sharings.go
```diff
@@ -287,6 +287,8 @@ func ShareDoc(instance *instance.Instance, sharing *Sharing, recStatus *Recipien
 
 			workerMsg, err := jobs.NewMessage(jobs.JSONEncoding, sharingWorker.SendOptions{
 				DocID:      val,
+				Selector:   rule.Selector,
+				Values:     values,
```
Gara64: Don't you need to pass those parameters in sharing_updates as well?

Gara64: Don't you need to pass those parameters in sharing_updates as well?

pkg/workers/sharings/share_data.go
```diff
@@ -496,6 +514,46 @@ func sendPatchToRecipient(patch *jsonapi.Document, opts *SendOptions, recipient
...
+// Two scenarii are possible:
+// 1. There is NO selector: the sharing is based on folders/files. If the file
+//    we are about to send has its id in the `values` declared in the
+//    permissions then it is one of the targets of this sharing. Its
+//    `dirID` must be set to `SharedWithMe`. If not then we don't modify it.
+// 2. There is a selector: the sharing is not based on folders/files. We change
+//    the `dirID` to `SharedWithMe`.
+func getParentDirID(opts *SendOptions, dirID string) (parentID string, err error) {
+	if opts.Selector == "" {
+		if opts.DocID == consts.RootDirID {
```
jinroh: Could have the same policy for the trash folder with ID `consts.TrashDirID` ?

Ljinod: It depends on whether we want to share files that were trashed or not. Right now the behavior is to stop the sharing if the target (i.e. what is shared) is trashed, but if at some point we want to "share" trashed files then yes we could have the same policy here.

PR #578, opened 2017-05-10
Allow to remove permissions with patch

aenario:


gregorylegarec:
@aenario Does it make sense to update the documentation ?

PR #579, opened 2017-05-10
chore: add piwik.js as an external lib

enguerran:
We will need `piwik.js` to get analytics from usage as explained in https://trello.com/c/zUe4D0KA/269-2-glob-7-traces-d-usage

We tought we could use `piwik.js` by implementing different solution:
- [ ] allow CSP to load `piwik.cozycloud.cc/piwik.js`
- [X] add `piwik.js` in external assets served by the stack
- [ ] bundle the library within the app bundle

Do you think we choose the best solution?


jinroh:
I'm OK with this solution !

It's not blocking, but I think the source URL should be versioned (ie: `https://piwik.cozycloud.cc/piwik-v1.2.3.js`)

PR #586, opened 2017-05-11
Support master-master sharing for files

Gara64:
This deals with PUT/PATCH/DELETE on files, both on the sharer and recipient side.
It is able to detect if a remote file is different after a local change, and mitigate the modification.

Note this is based on #575.

### DIFF
--- a/pkg/workers/sharings/share_data.go
+++ b/pkg/workers/sharings/share_data.go
@@ -355,7 +355,7 @@ func UpdateOrPatchFile(ins *instance.Instance, opts *SendOptions, fileDoc *vfs.F
 		// The MD5 didn't change: this is a PATCH
 		if md5 == md5AtRec {
 			// Check the metadata did change to do the patch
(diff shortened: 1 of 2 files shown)


pkg/workers/sharings/share_data.go
```diff
@@ -716,3 +706,26 @@ func docHasChanges(newDoc *couchdb.JSONDoc, doc *couchdb.JSONDoc) bool {
...
+	var obj *jsonapi.ObjectMarshalling
+	if err := json.Unmarshal(*doc.Data, &obj); err != nil {
+		return nil, err
+	}
+	if obj.Attributes != nil {
+		if err := json.Unmarshal(*obj.Attributes, &dirOrFileDoc); err != nil {
+			return nil, err
+		}
+	}
+	return dirOrFileDoc, nil
```
Gara64: It is not handled, I'll add it

PR #590, opened 2017-05-11
Piwik is using <img>, not <script>

nono:


web/middlewares/secure.go
```diff
@@ -63,17 +63,11 @@ const (
...
-	// CSPSrcTracker enables the whitelist of piwik
-	CSPSrcTracker
 
 	// CSPWhitelist is a whitelist of domains that are allowed in CSP. It's not
 	// permanent, this whitelist will be removed when we will have a more
 	// generic way to enable client-side apps to access some domains (proxy).
-	CSPWhitelist = "*.tile.openstreetmap.org *.tile.osm.org *.tiles.mapbox.com api.mapbox.com"
-
-	// CSPTracker is used to whitelist our usage tracker, based on piwik
-	// FIXME remove this as soon as we have a proper way to declare a proxy based on permissions
```
enguerran: We must keep that comment, don't we?

PR #592, opened 2017-05-12
Implement the @event trigger via redis

jinroh:


pkg/scheduler/redis_scheduler.go
```diff
@@ -25,6 +27,10 @@ const SchedKey = "scheduling"
 // pollInterval is the time interval between 2 redis polling
 const pollInterval = 1 * time.Second
 
+// eventLoopSize is the number of goroutines handling @events and triggering
+// jobs.
+const eventLoopSize = 50
```
nono: Why do we need more than 1 goroutine for that? And why 50?

pkg/scheduler/redis_scheduler.go
```diff
@@ -54,39 +60,103 @@ type RedisScheduler struct {
...
 
+func eventsKey(domain string) string {
+	return "events-" + domain
+}
+
 // Start a goroutine that will fetch triggers in redis to schedule their jobs
 func (s *RedisScheduler) Start(b jobs.Broker) error {
 	s.broker = b
+	s.stopped = make(chan struct{})
+	go s.eventDispatcher()
```
nono: s/go// ? `eventDispatcher` is just starting some goroutines (or we can refactor `eventDispatcher`)

pkg/scheduler/redis_scheduler.go
```diff
@@ -54,39 +60,103 @@ type RedisScheduler struct {
...
-	return nil
+	for i := 0; i < eventLoopSize; i++ {
+		go s.eventLoop(eventsCh)
+	}
+}
+
+func (s *RedisScheduler) eventLoop(ch <-chan *realtime.Event) {
+	for event := range ch {
+		key := eventsKey(event.Domain)
+		m, err := s.client.HGetAll(key).Result()
```
nono: As an optimization, I think we should not fetch the `@event` triggers for a domain more than once per second (cache)

pkg/scheduler/redis_scheduler.go
```diff
@@ -158,6 +228,11 @@ func (s *RedisScheduler) Add(t Trigger) error {
 func (s *RedisScheduler) addToRedis(t Trigger, prev time.Time) error {
 	var timestamp time.Time
 	switch t := t.(type) {
+	case *EventTrigger:
+		hKey := eventsKey(t.Infos().Domain)
+		hMap := make(map[string]interface{})
+		hMap[t.ID()] = t.Infos().Arguments
+		return s.client.HMSet(hKey, hMap).Err()
```
nono: Wouldn't be easier to use `HSet`, like this?

```go
hKey := eventsKey(t.Infos().Domain)
return s.client.HSet(hKey, t.ID(), t.Infos().Arguments).Err()
```

pkg/scheduler/redis_scheduler.go
```diff
@@ -54,39 +60,103 @@ type RedisScheduler struct {
...
+			if !eventMatchPermission(event, &rule) {
+				continue
+			}
+			t, err := s.Get(event.Domain, triggerID)
+			if err != nil {
+				log.Warnf("[scheduler] Could not fetch @event trigger %s %s: %s",
+					event.Domain, triggerID, err.Error())
+				continue
+			}
+			_, _, err = s.broker.PushJob(t.(*EventTrigger).Trigger(event))
```
nono: There will be some minor conflicts with my PR, but it should be easy to fix them

nono:
Travis is red, but only because of a failure of npm to fetch one module, I have restarted it

pkg/scheduler/redis_scheduler.go
jinroh: I think we can benefit from some level of concurrency to handle the global event channel to perform the work done in the event loop.

This handling of each event is quite important:
  - fetching the set of triggers form redis
  - parsing all the rules
  - fetching the matching trigger from couchdb
  - sending a new job request into the queue

Even with some caching, having some fixed size concurrency to handle that seems beneficial to me. Regarding the amount of go-routine, we could make it configurable.

pkg/scheduler/redis_scheduler.go
jinroh: You're right, no need to spawn a goroutine here. fixed b4dec7b720b114949a7daa2b31823ea84d5f8f57

pkg/scheduler/redis_scheduler.go
jinroh: Indeed, did not see this method, thanks. Fixed ea44ec272657c16f388c1efd35d6c7b096ee180b

pkg/scheduler/redis_scheduler.go
nono: OK, we should make proportional to `runtime.NumCPU()` (https://golang.org/pkg/runtime/#NumCPU), maybe 4x it.

PR #595, opened 2017-05-15
Add a "installed" state to apps and a permission page on first use

jinroh:
This PR adds an intermediate state `installed` meaning that an application has been successfully installed but is not yet `ready` because the user has not yet authorized the application.

A new page form on `/auth/authorize/app` has been added to authorize an `installed` app. The user is automatically redirected to this page on the first opening of the app.

On the CLI, the new flag `--deactivated` allow to install an application that should not be activated on installation.


cmd/apps.go
```diff
@@ -283,6 +286,7 @@ func foreachDomains(predicate func(*client.Instance) error) error {
 func init() {
 	webappsCmdGroup.PersistentFlags().StringVar(&flagAppsDomain, "domain", "", "specify the domain name of the instance")
 	webappsCmdGroup.PersistentFlags().BoolVar(&flagAllDomains, "all-domains", false, "work on all domains iterativelly")
+	installWebappCmd.PersistentFlags().BoolVar(&flagAppsDeactivated, "deactivated", false, "specify that the application should not be activated after installation")
```
nono: Maybe s/deactivated/ask-permissions/

PR #596, opened 2017-05-15
Send file when remote doc is not found

Gara64:
It has the side effect to support photo album adding. As this is viewed as an update on the file, with a new `referenced_by`, the sharing worker will try to fetch the file from the remote, fail, and thus POST the file.

### DIFF
--- a/pkg/workers/sharings/sharing_updates.go
+++ b/pkg/workers/sharings/sharing_updates.go
@@ -270,8 +270,11 @@ func ExtractHostAndScheme(fullURL string) (string, string, error) {
 	return host, scheme, nil
 }
 
+// isRecipientSide is used to determine whether or not we are on the recipient side.
+// A sharing is on the recipient side iff:
+// - the sharing type is master-master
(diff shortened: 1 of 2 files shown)


pkg/workers/sharings/sharing_updates.go
```diff
@@ -269,3 +269,15 @@ func ExtractHostAndScheme(fullURL string) (string, string, error) {
 	}
 	return host, scheme, nil
 }
+
+func isRecipientSide(sharing *Sharing) bool {
```
Ljinod: I think it's worth telling that a recipient will never enter this part of the code if the sharing is not of type Master-Master.
The function could lead us to believe that it could return false even if it actually is called by a recipient, if the sharing type is not Master-Master for instance (which is an impossible scenario…).

Gara64: Well, it is normally an impossible scenario, but we never know when a protocol misused could happen, so it actually might be called from the recipient side :)
But if you've been confused, I'll add some comments to clarify.

PR #598, opened 2017-05-15
Add route for mango query on files

aenario:


web/files/files.go
```diff
@@ -714,13 +716,72 @@ func DestroyFileHandler(c echo.Context) error {
...
+
+	var results []vfs.DirOrFileDoc
+	err := couchdb.FindDocsRaw(instance, consts.Files, &findRequest, &results)
+	if err != nil {
+		return err
+	}
+
+	var total int
+	if len(results) > int(limit) {
+		total = math.MaxInt32 - 1          // we dont know the actual number
```
nono: What is this magic mumber?

aenario: Agree this is not ideal:
- We dont and cant know the total number of results from a mango query.
- We want to let the client-side know when there is no more results, with JSON-APi convention
- This solutions returns a number which is higher than any possible number, so we cant simply do client-side `if(total > received)`

nono: And it's too complicated to give a `next` link to fetch the next results when there are more?

aenario: It's feasible, but a bit complicated: you have to parse the mango query selector and add a $gt condition on last sorting field.

nono: OK, let's keep it for another time

PR #602, opened 2017-05-15
Support photo suppression in shared album

Gara64:
This does several things:

- Change the `UpdateDoc` behaviour to get the old doc and emit through `rtevent`
- Add an `OldDoc` field in the `Event` struct
- Add a check in the `trigger_event` to see if the `OldDoc` is Valid
- Compare the references between the local and remote files, to know if the sharing worker needs to send a DELETE request to the remote.
- Add some tests

### DIFF
--- a/pkg/scheduler/trigger_event.go
+++ b/pkg/scheduler/trigger_event.go
@@ -123,6 +123,16 @@ func eventMatchPermission(e *realtime.Event, rule *permissions.Rule) bool {
 	}
 
 	if v, ok := e.Doc.(permissions.Validable); ok {
+		if rule.ValuesValid(v) {
+			// Particular case where the new doc is not valid but the old one was.
+			if e.OldDoc != nil {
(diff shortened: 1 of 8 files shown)


pkg/scheduler/trigger_event.go
```diff
@@ -123,6 +123,16 @@ func eventMatchPermission(e *realtime.Event, rule *permissions.Rule) bool {
...
 	if v, ok := e.Doc.(permissions.Validable); ok {
+		if rule.ValuesValid(v) {
+			// Particular case where the new doc is not valid but the old one was.
+			if e.OldDoc != nil {
+				if vOld, okOld := e.OldDoc.(permissions.Validable); okOld {
+					return rule.ValuesValid(vOld)
+				}
+			}
+		} else {
+			return true
```
nono: You sure you didn't invert the then and else clauses?

Gara64: You're right, well seen...

PR #613, opened 2017-05-17
Use ?sharecode=... for sharing codes

nono:


web/apps/serve.go
```diff
@@ -161,6 +161,8 @@ func ServeAppFile(c echo.Context, i *instance.Instance, fs apps.FileServer, app
 	token := "" // #nosec
```
aenario: gometalinter wants you to change this into var token string

enguerran:
[(2) [PHO-28 Partage d'un album par lien] Générer un lien de partage](https://trello.com/c/QFFEMBdU/280-2-pho-28-partage-d-un-album-par-lien-generer-un-lien-de-partage)

PR #628, opened 2017-05-18
Parse json message from konnector and fail job if an error occured

aenario:


pkg/workers/konnectors/konnector.go
```diff
@@ -162,33 +173,57 @@ func Worker(ctx context.Context, m *jobs.Message) error {
 	scanOut := bufio.NewScanner(cmdOut)
 	scanOut.Buffer(nil, 256*1024)
 
-	go doScanOut(jobID, scanOut, domain)
+	var msgChan = make(chan konnectorMsg)
+	var messages []konnectorMsg
```
jinroh: nit: instead of storing all messages, why not storing only the last message of type `konnectorMsgTypeError ` directly ?

aenario: I had the idea that in the future we will need to store progress & warning messages in konnectors.result.

pkg/workers/konnectors/konnector.go
```diff
@@ -162,33 +174,57 @@ func Worker(ctx context.Context, m *jobs.Message) error {
...
-		err := json.Unmarshal(scanner.Bytes(), &doc.M)
-		if err != nil {
-			log.Warnf("[konnector] %s: Could not parse Stdout as JSON: %s", jobID, err)
-			log.Warnf("[konnector] %s: Stdout: %s", jobID, scanner.Text())
-			continue
+		linebb := scanner.Bytes()
+		from := bytes.IndexByte(linebb, '{')
+		to := bytes.LastIndexByte(linebb, '}')
+		var msg konnectorMsg
+		log.Infof("[konnector] %s: Stdout: %s", jobID, string(linebb[from:to+1]))
```
nono: I don't think you can safely use `linebb[from:to+1]` before checking `from` and `to`

pkg/workers/konnectors/konnector.go
```diff
@@ -162,33 +174,57 @@ func Worker(ctx context.Context, m *jobs.Message) error {
...
+		if from > -1 && to > -1 {
```
nono: Maybe also check that `to > from`

PR #630, opened 2017-05-18
feat: revoke a sharing

Ljinod:
This PR is the first step towards a full revokation: the sharing
document is set to revoked and all associated OAuth clients are deleted.


pkg/sharings/sharings.go
```diff
@@ -597,6 +603,41 @@ func CreateSharing(instance *instance.Instance, sharing *Sharing) error {
 	return couchdb.CreateDoc(instance, sharing)
 }
 
+// RevokeSharing revokes the sharing and deletes all the OAuth client
+// associated with it.
+//
+// Revoking a sharing consists of setting the field `Revoked` to `true`.
+func RevokeSharing(ins *instance.Instance, sharing *Sharing) error {
+	sharing.Revoked = true
```
nono: You don't persist in db this change?

pkg/sharings/sharings.go
```diff
@@ -597,6 +603,41 @@ func CreateSharing(instance *instance.Instance, sharing *Sharing) error {
...
+	return err
+}
+
+func deleteOAuthClient(ins *instance.Instance, id string) error {
+	client, err := oauth.FindClient(ins, id)
+	if err != nil {
+		log.Errorf("[sharings] Could not find client %s: %v", id, err)
+		return err
+	}
+	crErr := client.Delete(ins)
```
nono: Why `crErr`? And not just `err = ...`?

pkg/sharings/sharings.go
Ljinod: I should indeed, thanks!

pkg/sharings/sharings.go
Ljinod: Because `client.Delete` returns a `ClientRegistrationError` structure not a simple error

PR #632, opened 2017-05-18
Increase konnector timeout & remove some artifacts

aenario:


jinroh:
@aenario you've commited an `app.tar` file in this commit

aenario:
@jinroh nope, I removed an app.tar that was commited before

PR #642, opened 2017-05-23
Add domain on whitelist for app LaMusiqueDeMesFilms

jacquarg:
This PR add a list of domains, required to fetch metadata around VoD and replay data from Orange, in the Orange app La Musique de Mes Films.

To make it work, I also add to add this sites to the img-src and default-src CSP headers.


nono:
Fixed by https://github.com/cozy/cozy-doctypes/pull/6

PR #655, opened 2017-05-27
feature: trash document when sharing ends

Ljinod:
With this PR when a document is no longer shared it is trashed. If the
document is still shared in another sharing it is kept.
To enable this feature the `Valid` method of the JSONDoc objet was
improved to handle `referenced_by` permissions.

Several other additions where also made:
* some logging was added;
* the test coverage was improved in pkg/workers/sharings;
* if we share a JSON document (not an update) a check is performed on
its doctype to see if the database already exists and, if it doesn't, to
create it;
* in web/files/referencedby.go the function `RemoveReferencedHandler`
returns `200` instead of `204` because the latter triggers
the error "http: request method or response status code does not allow
body".

### DIFF
--- a/pkg/consts/consts.go
+++ b/pkg/consts/consts.go
@@ -1,5 +1,9 @@
 package consts
 
+import (
+	"github.com/cozy/cozy-stack/pkg/couchdb"
+)
+
 // Instances doc type for User's instance document
 const Instances = "instances"
 
@@ -111,15 +115,15 @@ const (
 
 const (
 	// SelectorReferencedBy is the "referenced_by" selector.
-	SelectorReferencedBy = "referenced_by"
+	SelectorReferencedBy = couchdb.SelectorReferencedBy
 )
 
 const (
 	// QueryParamRev is the key for the revision value in a query string.
-	QueryParamRev = "rev"
+	QueryParamRev = "Rev"
 	// QueryParamDirID is the key for the `DirID` field of a vfs.FileDoc or
 	// vfs.DirDoc, in a query string.
-	QueryParamDirID = "dir_id"
+	QueryParamDirID = "Dir_id"
 	// QueryParamName is the key for the name value in a query string.
 	QueryParamName = "Name"
 	// QueryParamType is the key for the `type` value (file or directory) in
@@ -141,7 +145,7 @@ const (
 	QueryParamTags = "Tags"
 	// QueryParamReferencedBy is the key for the `referenced_by` values in a
 	// query string.
-	QueryParamReferencedBy = SelectorReferencedBy
+	QueryParamReferencedBy = "Referenced_by"
 )
 
 // AppsRegistry is an hard-coded list of known apps, with their source URLs
--- a/pkg/consts/consts.go
+++ b/pkg/consts/consts.go
@@ -110,6 +110,11 @@ const (
 )
 
 const (
+	// SelectorReferencedBy is the "referenced_by" selector.
+	SelectorReferencedBy = "referenced_by"
+)
+
+const (
 	// QueryParamRev is the key for the revision value in a query string.
 	QueryParamRev = "rev"
 	// QueryParamDirID is the key for the `DirID` field of a vfs.FileDoc or
@@ -129,14 +134,14 @@ const (
 	// QueryParamUpdatedAt is the key for the `Updated_at` value in a query
 	// string.
 	QueryParamUpdatedAt = "Updated_at"
-	// QueryParamReferencedBy is the key for the `referenced_by` values in a
-	// query string.
-	QueryParamReferencedBy = "Referenced_by"
 	// QueryParamRecursive is the key for the `recursive` value in a query
 	// string.
 	QueryParamRecursive = "Recursive"
 	// QueryParamTags is the key for the `tags` values in a query string.
 	QueryParamTags = "Tags"
+	// QueryParamReferencedBy is the key for the `referenced_by` values in a
+	// query string.
+	QueryParamReferencedBy = SelectorReferencedBy
 )
 
 // AppsRegistry is an hard-coded list of known apps, with their source URLs
--- a/pkg/couchdb/couchdb.go
+++ b/pkg/couchdb/couchdb.go
@@ -25,9 +25,9 @@ import (
 // a upperbound for string.
 const MaxString = mango.MaxString
 
-// This constant is used to evaluate permissions in the `Valid` method of a
-// JSONDoc object.
-const selectorReferencedBy = "referenced_by"
+// SelectorReferencedBy is the string constant for the references in a JSON
+// document.
+const SelectorReferencedBy = "referenced_by"
 
 // Doc is the interface that encapsulate a couchdb document, of any
 // serializable type. This interface defines method to set and get the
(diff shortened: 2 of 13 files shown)


pkg/couchdb/couchdb.go
```diff
@@ -25,6 +25,10 @@ import (
 // a upperbound for string.
 const MaxString = mango.MaxString
 
+// This constant is used to evaluate permissions in the `Valid` method of a
+// JSONDoc object.
+const selectorReferencedBy = "referenced_by"
```
nono: It is duplicated between `consts` and `couchdb`. Can we define it in one package and just reference it from the other?

pkg/consts/consts.go
```diff
@@ -129,14 +134,14 @@ const (
...
-	// query string.
-	QueryParamReferencedBy = "Referenced_by"
 	// QueryParamRecursive is the key for the `recursive` value in a query
 	// string.
 	QueryParamRecursive = "Recursive"
 	// QueryParamTags is the key for the `tags` values in a query string.
 	QueryParamTags = "Tags"
+	// QueryParamReferencedBy is the key for the `referenced_by` values in a
+	// query string.
+	QueryParamReferencedBy = SelectorReferencedBy
```
nono: In JSON-API, only some reserved query parameters are allowed to be in lowercase. For most of them, it should start with an upper case.

pkg/couchdb/couchdb.go
```diff
@@ -192,8 +196,37 @@ func (j JSONDoc) Get(key string) interface{} {
...
+//     {"type": "doctype1", "id": "id1"},
+//     {"type": "doctype2", "id": "id2"},
+// ]
 func (j JSONDoc) Valid(field, value string) bool {
+	if field == selectorReferencedBy {
+		rawReferences := j.Get(field)
+		if rawReferences == nil {
+			return false
+		}
+		references := rawReferences.([]interface{})
```
nono: It would be better to check the type assertion, like this:

```go
references, ok := rawReferences.([]interface{})
if !ok {
    return false
}
```

pkg/sharings/sharings.go
```diff
@@ -225,7 +226,8 @@ func ShareDoc(instance *instance.Instance, sharing *Sharing, recStatus *Recipien
 		docType := rule.Type
 		// Trigger the updates if the sharing is not one-shot
 		if sharing.SharingType != consts.OneShotSharing {
-			if err := AddTrigger(instance, rule, sharing.SharingID); err != nil {
+			err := AddTrigger(instance, rule, sharing.SharingID)
+			if err != nil {
```
nono: Why this change?

pkg/sharings/sharings.go
```diff
@@ -648,6 +654,74 @@ func deleteOAuthClient(ins *instance.Instance, id string) error {
...
+
+	// The doctype is not always set, at least in the tests, and is required in
+	// order to delete the document.
+	if doc.DocType() == "" {
+		doc.Type = doctype
+	}
+
+	loop := true
+	for loop {
+		loop = cursor.HasMore()
```
nono: You can use `break`:

```go
for {
  if !cursor.HasMore() {
      break
  }
  // ...
}
```

pkg/workers/sharings/share_data.go
```diff
@@ -98,12 +99,16 @@ func (opts *SendOptions) fillDetailsAndOpenFile(fs vfs.VFS, fileDoc *vfs.FileDoc
...
-	if err != nil {
-		return nil
+	var refs string
+	if opts.Selector == consts.SelectorReferencedBy {
+		sharedRefs := opts.getSharedReferences()
+		b, err := json.Marshal(sharedRefs)
+		if err != nil {
+			return err
+		}
+		refs = string(b[:])
```
nono: Why `b[:]` and not just `b`?

pkg/workers/sharings/share_data.go
```diff
@@ -132,6 +137,54 @@ func (opts *SendOptions) closeFile() error {
...
+			if ref.ID == sharedRef.ID {
+				match = true
+				break
+			}
+		}
+
+		if match {
+			res = append(res, refs[i])
+		}
+	}
```
nono: Why do you use `match`? Isn't this the same?

```go
for i, ref := range refs {
	for _, sharedRef := range sharedRefs {
		if ref.ID == sharedRef.ID {
			res = append(res, refs[i])
			break
		}
	}
}
```

web/files/referencedby.go
```diff
@@ -83,5 +83,5 @@ func RemoveReferencedHandler(c echo.Context) error {
 		return wrapVfsError(err)
 	}
 
-	return c.NoContent(http.StatusNoContent)
+	return c.NoContent(http.StatusOK)
```
nono: Why this change? 204 No Content is a good status code to indicate the success of a deletion (body is empty)

pkg/couchdb/couchdb.go
Ljinod: Yup totally.

pkg/workers/sharings/share_data.go
Gara64: @nono From what I understood, a byte array cannot be direclty converted to string, because it is actually a slice. So you need to first convert the byte array to slice using `[:]` , then convert to string.

pkg/consts/consts.go
Ljinod: I was not aware of that, I'll make the necessary changes.

pkg/sharings/sharings.go
Ljinod: Cosmetics reasons: I'm a bit maniac and the line exceeded 80 characters… I can revert it though.

web/files/referencedby.go
Ljinod: Returning 204 causes Go to produce the following error: "http: request method or response status code does not allow body". This is because in order to remove references we pass them in the body of the DELETE request.

I can revert back but I felt that having one error less would be better (or am I the only one that has it?).

pkg/workers/sharings/share_data.go
Ljinod: Yup it's the same, I think I didn't have my glasses one at the time…!

pkg/workers/sharings/share_data.go
nono: @Gara64 a byte array != a byte slice. But here, `json.Marshall` returns a byte slice, so I don't see why you try to force a slice

web/files/referencedby.go
nono: > Returning 204 causes Go to produce the following error: "http: request method or response status code does not allow body".

How have you got this error?

204 means that the response has no content, and it is the case, as we use `c.NoContent`. But it doesn't mean that the request must be empty.

Ljinod: I was able to reproduce but the culprit code was mine: I call `files.RemoveReferenceHandler` — which calls at the end `c.NoContent` — and then I returned `c.JSON(200, nil)`. Since the first call expects no body and the second provides one I get the error I mentionned. I corrected it.

pkg/couchdb/couchdb.go
```diff
@@ -192,8 +196,37 @@ func (j JSONDoc) Get(key string) interface{} {
...
+	if field == SelectorReferencedBy {
+		rawReferences := j.Get(field)
+		references, ok := rawReferences.([]interface{})
+		if !ok {
+			return false
+		}
+
+		values := strings.Split(value, "/")
+		valueType, valueID := values[0], values[1]
+
```
Gara64: It might be worth it to check the `values` length to avoid a nil pointer exception in case the permission was badly defined.

PR #662, opened 2017-05-30
export

AylinKocoglu:


imexport/export.go
```diff
@@ -0,0 +1,86 @@
...
+
+		AuthClient: authReq.ClientParams,
+		AuthScopes: authReq.Scopes,
+		Authorizer: opts.Authorizer,
+		UserAgent:  opts.UserAgent,
+	}
+
+	root := "/Documents"
+
+	if err := cClient.WalkByPath(root, func(path string, doc *client.DirOrFile, err error) error {
```
aenario: [nitpick] the combo err-in-if with a lambda function make this a bit hard to read : I'd recommend either naming the lambda  or having on separate lines `err := cClient.Walk....` & `if err != nil`

imexport/export.go
```diff
@@ -0,0 +1,86 @@
...
+		fmt.Printf("Visited: %s  type: %s\n", path, doc.Attrs.Type)
+
+		if doc.Attrs.Type == client.DirType {
+			fmt.Println("directory")
+		} else if doc.Attrs.Type == client.FileType {
+			readCloser, err := cClient.DownloadByPath(path)
+			if err != nil {
+				return err
+			}
```
jinroh: Do not forget to `Close()` the `readCloser` (which could be called `fileContent`). You can use a `defer` for that.

jinroh: Also, you could export this part in a separate method `func writeFile(cClient *client.Client, tw tar.Writer, doc *vfs.DirOrFileDoc)`, in order for the defered close to be called at the end of the copy instead of at the end of the copy of all files.

imexport/export.go
```diff
@@ -0,0 +1,86 @@
...
+
+	root := "/Documents"
+
+	if err := cClient.WalkByPath(root, func(path string, doc *client.DirOrFile, err error) error {
+		fmt.Printf("Visited: %s  type: %s\n", path, doc.Attrs.Type)
+
+		if doc.Attrs.Type == client.DirType {
+			fmt.Println("directory")
+		} else if doc.Attrs.Type == client.FileType {
+			readCloser, err := cClient.DownloadByPath(path)
```
jinroh: Prefer the use of `DownloadByID` to download the file using its ID.

imexport/job.go
```diff
@@ -0,0 +1,82 @@
...
+
+	mail, err := json.Marshal(b)
+
+	return mail, err
+
+}
+
+func sendMail(cClient *client.Client) error {
+
+	text := "Bonjour, vous pouvez des a presents recuperer vos documents"
```
nono: Il manque quelques accents : vous pouvez dès à présent récupérer vos documents

PR #669, opened 2017-05-31
feat: a document can be reshared

Ljinod:
With this PR it is possible to share several times the same document. We
detect that a document is reshared when we try to fetch it: we receive a
403 error — the document already exists but we can't access it.
When we receive such error we send the document as is if it was a
creation and the recipient will interpret it as an update.

### DIFF
--- a/pkg/couchdb/couchdb.go
+++ b/pkg/couchdb/couchdb.go
@@ -214,6 +214,9 @@ func (j JSONDoc) Valid(field, value string) bool {
 		}
 
 		values := strings.Split(value, "/")
+		if len(values) != 2 {
+			return false
+		}
 		valueType, valueID := values[0], values[1]
 
 		for _, ref := range references {
--- a/web/sharings/files.go
+++ b/web/sharings/files.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/cozy/cozy-stack/pkg/consts"
 	"github.com/cozy/cozy-stack/pkg/couchdb"
+	"github.com/cozy/cozy-stack/pkg/instance"
 	"github.com/cozy/cozy-stack/pkg/sharings"
 	"github.com/cozy/cozy-stack/pkg/vfs"
 	"github.com/cozy/cozy-stack/web/files"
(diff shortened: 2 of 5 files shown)


pkg/couchdb/couchdb.go
```diff
@@ -214,6 +214,9 @@ func (j JSONDoc) Valid(field, value string) bool {
 		}
 
 		values := strings.Split(value, "/")
+		if len(values) != 2 {
```
Ljinod: This doesn't directly concern the PR, it is to avoid errors due to a wrong formatting of the referenced_by values. I wanted to make a separate PR but I forgot…

web/sharings/files.go
```diff
@@ -345,3 +418,72 @@ func createDirForSharing(fs vfs.VFS, id, name string) error {
...
+			if reflect.DeepEqual(newReference, oldReference) {
+				exists = true
+				break
+			}
+		}
+
+		if !exists {
+			res = append(res, oldReference)
+		}
+	}
```
nono: You can simplify it to:

```go
for _, oldReference := range oldRefs {
	var exists bool
	for _, newReference := range newRefs {
		if reflect.DeepEqual(newReference, oldReference) {
			res = append(res, oldReference)
			break
		}
	}
}
```

web/sharings/files.go
```diff
@@ -345,3 +418,72 @@ func createDirForSharing(fs vfs.VFS, id, name string) error {
...
+			if newTag == oldTag {
+				exists = true
+				break
+			}
+		}
+
+		if !exists {
+			res = append(res, oldTag)
+		}
+	}
```
nono: Same here

web/sharings/files.go
Ljinod: Well actually here I don't want to append if there are equal, the aim is to merge without duplicating elements (it does look like a mistake I made in a previous PR).

web/sharings/files.go
Ljinod: (Same answer as above :))

web/sharings/files.go
nono: My mistake

PR #672, opened 2017-06-01
fix: prevent consecutive sendings

Ljinod:
When Alice shares a photo with Bob through a Master-Master sharing,
Bob's Cozy starts the sharing_update worker. It sees this event as a
creation and tries to send that very same photo to Alice. This led to
errors in the stack log.

This PR fixes this behavior: when we send a file we first try to get it
and unless we receive a 404 or 403 error we abort.

### DIFF
--- a/pkg/workers/sharings/share_data.go
+++ b/pkg/workers/sharings/share_data.go
@@ -901,8 +901,7 @@ func getDirOrFileMetadataAtRecipient(id string, recInfo *RecipientInfo) (*vfs.Di
 		},
 	})
 	if err != nil {
-		reqErr := err.(*request.Error)
-		return nil, nil, parseError(reqErr)
+		return nil, nil, parseError(err)
 	}
 
 	dirOrFileDoc, err := bindDirOrFile(res.Body)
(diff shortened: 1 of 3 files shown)


nono:
Ouch, it seems very expensive to redownload a photo we just got. At least, we should try to just make a `HEAD` request instead of `GET`. Better, to bundle several of those requests as a bulk. And the best, to make no request.

Ljinod:
@nono I'm not sure I'm following: the aim of this PR is to prevent a redownload (which is what is happening). The `GET` just fetches the metadata of the file, is it that expensive?
I have no problem doing a `HEAD` request instead (which one?) but in any case we do have to make at least a request to see if the file already exists in order to avoid a redownload.

pkg/workers/sharings/share_data.go
```diff
@@ -893,6 +916,41 @@ func getDirOrFileMetadataAtRecipient(id string, recInfo *RecipientInfo) (*vfs.Di
...
+		Method: http.MethodHead,
+		Path:   path,
+		Headers: request.Headers{
+			echo.HeaderContentType:   echo.MIMEApplicationJSON,
+			echo.HeaderAuthorization: "Bearer " + recInfo.Token,
+		},
+	})
+
+	if err != nil {
+		reqErr := err.(*request.Error)
```
jinroh: Are we sure this type assertion will not panic, ie: we do not have a request.Error ?
Maybe the `parseError` method could take an `error` type, and make a safe type assertion.

Gara64: I confirm it could, I just met this situation :)
I solved it with something like : 
`switch v := err.(type) {
  case *request.Error:
    //something
  default:
    //something
`
But the workaround proposed by @jinroh might be cleaner

PR #674, opened 2017-06-02
fix: IE11 does not handle HTML5 form attribute

enguerran:
Internet Explorer and Edge do not support HTML5 form.
You can check it on http://html5test.com/

So, either we update html markup to wrap all inputs into the form element (this very PR), either we add some crappy JavaScript code to trigger the form submission on button click.


nono:
@GoOz what do you think?

PR #675, opened 2017-06-02
Refresh token after auth error during sharing

Gara64:
When a sharing request fails with an authentication error (400 or 401), a token renewal is performed, before retrying the request.

Note this is based on #673


client/auth/auth.go
```diff
@@ -228,7 +228,7 @@ func (r *Request) GetAccessToken(c *Client, code string) (*AccessToken, error) {
 func (r *Request) RefreshToken(c *Client, t *AccessToken) (*AccessToken, error) {
 	q := url.Values{
 		"grant_type":    {"refresh_token"},
-		"code":          {t.RefreshToken},
+		"refresh_token": {t.RefreshToken},
```
nono: :+1:

PR #676, opened 2017-06-04
WIP: Initial Windows support

sebn:
**dislaimer: this is WIP**

The following few changes allow me to build the cozy-stack on Windows and make the cozy-desktop tests pass 😍 
It could make it possible to use the cozy-stack on AppVeyor (instead of depending on an instance hosted somewhere else).

Main issues encountered:
- logrus' syslog hook doesn't build on Windows (I moved setup to a dedicated package, conditionnally building a runtime-failing stub on unsupported OS)
- The default fs-url was mixing backslash-separated parts with slash-separated ones. Also `file://localhost/C:/...` [doesn't seem to work well on Windows](https://en.wikipedia.org/wiki/File_URI_scheme#Windows), and may be better replaced with `file:///C:/...`
- Using the instance's `domain:port` as a dir name can't work as colon is a reserved char on Windows. Replacing it with an underscore works, but I wonder whether underscores may be allowed in some cases, so we may better find a safer replacement char...

Except for those, it basically works, although I clearly didn't use the whole API.

Wdyt?
I'm aware it may not be the best timing to discuss this, so feel free to postpone it!

### DIFF
--- a/pkg/logger/hooks/syslog.go
+++ b/pkg/logger/hooks/syslog.go
@@ -1,0 +1,19 @@
+// +build !windows
+
+package hooks
+
+import (
+	"io/ioutil"
+	std_syslog "log/syslog"
+
+	logrus_syslog "github.com/Sirupsen/logrus/hooks/syslog"
+)
+
+func SetupSyslog() error {
+	hook, err := logrus_syslog.NewSyslogHook("", "", std_syslog.LOG_INFO, "cozy")
+	if err != nil {
+		return err
+	}
+	logrus.AddHook(hook)
+	logrus.SetOutput(ioutil.Discard)
+}
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -2,8 +2,9 @@ package cmd
 
 import (
 	"errors"
-	"fmt"
 	"os"
+	"net/url"
+	"path"
 	"path/filepath"
 	"runtime"
 	"strings"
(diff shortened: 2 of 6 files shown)


sebn:
(also, I assumed it was safe to use the new `PathSegment` method only in the affero vfs implementation, and keep using the domain in Swift)

pkg/logger/hooks/syslog.go
```diff
@@ -0,0 +1,19 @@
+// +build !windows
+
+package hooks
```
aenario: [nitpick] "hook" is the logrus term and is not very explicit, renaming this package to "syslog" would make thinks clearer IMHO

pkg/logger/hooks/syslog.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"io/ioutil"
+	std_syslog "log/syslog"
+
+	logrus_syslog "github.com/Sirupsen/logrus/hooks/syslog"
```
aenario: you lack a logrus import

sebn:
@aenario Thanks for the feedback. I couldn't run the tests yet because I was missing a redis instance. I'm almost sure tests will require more fixes to work on Windows though.

sebn:
Closing so you don't get a stale PR.
Code changed too much anyway.
I'll come up with anoter one with tests at some point.

PR #678, opened 2017-06-05
Add a first draft about /remote API documentation

nono:


aenario:
Some thoughts : 

- Having the doctypes defined in the repository needs more explanations : 
      - is the stack retrieving them from github at runtime or are they git submodule and build into the binary ?
      - Do we introduce a mechanism for defining these types locally for debugging 

- I am affraid we will have no choice but to provide a GET alternative, for tile services (osm, osmbuildings), the alternative is to ask our devs to change the code of leaflet & cie.

PR #685, opened 2017-06-07
fix: typo :pencil:

gregorylegarec:


nono:
It's not a typo. We had to enclose some markdown files with `{% raw %}` and `{% endraw %}` to avoid github sites to "eat" the `{{ variables }}`.

gregorylegarec:
Ok, my bad 👍

PR #687, opened 2017-06-07
Take the last update time into account to check app state

jinroh:
Also increase the couchdb timeout.

### DIFF
--- a/pkg/apps/installer.go
+++ b/pkg/apps/installer.go
@@ -4,6 +4,7 @@ import (
 	"io"
 	"net/url"
 	"regexp"
+	"time"
 
 	"github.com/cozy/cozy-stack/pkg/couchdb"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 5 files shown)


pkg/apps/installer.go
```diff
@@ -250,14 +249,28 @@ func (i *Installer) update() (Manifest, error) {
...
+// perform an update or deletion.
+func (i *Installer) checkState(man Manifest) error {
+	state := man.State()
+	if state == Ready ||
+		state == Installed ||
+		state == Errored {
+		return nil
+	}
+	timeAgo := time.Now().Add(-2 * time.Minute)
+	if man.LastUpdate().Before(timeAgo) {
```
nono: Isn't 2 minutes a bit short?
And you can shorten this to:

```go
if time.Since(man.LastUpdate()) > 15 * time.Minute
```

PR #693, opened 2017-06-08
feat: Add responsiveness + a little better permissions list

GoOz:
Permissions are now responsive and mobile compliant

Permissions list are not quite different for now, waiting for a bunch of icons but as it can't really wait, let's merge it like this, everything is just prepared but not used.

### DIFF
--- a/assets/locales/fr.po
+++ b/assets/locales/fr.po
@@ -63,7 +63,7 @@ msgid "Login Forgot password"
 msgstr "Mot de passe oublié ?"
 
 msgid "Authorize Title"
-msgstr "Autoriser %s à accéder à votre profil ?"
+msgstr "Autoriser %s à accéder à votre profil ?"
 
 msgid "Authorize App Title"
 msgstr ""
--- a/assets/templates/authorize.html
+++ b/assets/templates/authorize.html
@@ -40,9 +40,9 @@
                 <strong>{{.Domain}}</strong> :<br />
                 {{end}}
               </p>
-              <ul>
+              <ul class="perm-list">
                 {{range $index, $perm := .Permissions}}
-                <li>
+                <li class={{ $perm.Type }}>
                   {{- t $perm.TranslationKey -}}
                   {{- if $perm.Verbs.ReadOnly}}{{t "Permissions Read only"}}{{end -}}
                 </li>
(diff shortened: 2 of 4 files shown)


GoOz:
[(2) [GLOB-11] Permissions des apps oauth UI et permissions des apps tiers (consentement éclairés) UI](https://trello.com/c/XpE27X1N/268-2-glob-11-permissions-des-apps-oauth-ui-et-permissions-des-apps-tiers-consentement-%C3%A9clair%C3%A9s-ui)

assets/locales/fr.po
```diff
@@ -63,7 +63,7 @@ msgid "Login Forgot password"
 msgstr "Mot de passe oublié ?"
 
 msgid "Authorize Title"
-msgstr "Autoriser %s à accéder à votre profil ?"
+msgstr "Autoriser %s à accéder à votre profil ?"
```
GoOz: This is just a non-breakable space, so the question mark doesn't go to the next line by itself. It wasn't really pretty.

assets/templates/authorize.html
```diff
@@ -40,9 +40,9 @@
                 <strong>{{.Domain}}</strong> :<br />
                 {{end}}
               </p>
-              <ul>
+              <ul class="perm-list">
                 {{range $index, $perm := .Permissions}}
-                <li>
+                <li class={{ $perm.Type }}>
```
jinroh: Shouldn't this be in quotes: `class="{{ $perm.Type )}"` ?

assets/locales/fr.po
nono: Did you made this change on transifex?

assets/templates/authorize.html
GoOz: oh, right. I'm used to not put quotes in JS but yeah it think it should with Go. Browsers apparently add them automatically, but you're right, just to be sure i'll add them.

assets/locales/fr.po
GoOz: oh the stack uses Transifex too ? I thought not because the translations files are versioned unlike the apps we do, but I'll do it right away.

GoOz: Done.

PR #697, opened 2017-06-08
Swift: avoid file collision using index

jinroh:


pkg/vfs/couchdb_indexer.go
```diff
@@ -343,3 +343,15 @@ func (c *couchdbIndexer) DirLength(doc *DirDoc) (int, error) {
 	}
 	return int(f64), nil
 }
+
+func (c *couchdbIndexer) DirChildExists(dirID, name string) (bool, error) {
+	var res couchdb.ViewResponse
+
+	// consts.FilesByParentView keys are [parentID, type, name]
+	err := couchdb.ExecView(c.db, consts.FilesByParentView, &couchdb.ViewRequest{
+		Key:         []string{dirID, consts.FileType, name},
```
aenario: This wont match a child folder with the given name, is that on purpose ? You can use  a Keys to match both.

nono:
@sebn: as I understand this change, I think it can have an impact on cozy-desktop. A file is created as a document in couchdb at the start of its upload. It has no md5sum (and probably some other metadata are missing too) until the upload finishes. Then, if the upload is successful, we had the missing fields (else, the document is dropped).

@jinroh, @aenario: do you think it can impact the client side apps (drive, photos)? They may see files that are not yet ready to be used if I understand correctly.

PR #700, opened 2017-06-09
Add remote doctypes

nono:


pkg/remote/remote.go
```diff
@@ -0,0 +1,320 @@
...
+// Find finds the request defined for the given doctype
+func Find(ins *instance.Instance, doctype string) (*Remote, error) {
+	var raw string
+
+	if config.GetConfig().Doctypes == "" {
+		dt := Doctype{
+			DocID: consts.Doctypes + "/" + doctype,
+		}
+		err := couchdb.GetDoc(ins, consts.Doctypes, dt.DocID, &dt)
+		if err != nil {
```
jinroh: shouldn't we chose to go this branch only for "not found" errors ? is it wanted to always hit the github repo even for couchdb internal errors ?

nono: Yes, it was my intention to use github even for couchdb internal errors. Do you see some cons to that?

jinroh: Not really, although we could log internal errors to make them visible.

pkg/remote/remote.go
```diff
@@ -0,0 +1,320 @@
...
+	if err != nil {
+		return ErrInvalidRequest
+	}
+
+	req.Header.Set("User-Agent", "cozy-stack "+config.Version+" ("+runtime.Version()+")")
+	for k, v := range remote.Headers {
+		req.Header.Set(k, v)
+	}
+
+	res, err := http.DefaultClient.Do(req)
```
jinroh: I think the default client has a infinite timeout which could be problematic in the case of a proxy, with stall connection blocking others. The use of a specific `http.Client` instance with more sensible defaults may be useful here.

nono: Yup, good idea

pkg/remote/remote.go
```diff
@@ -0,0 +1,322 @@
...
+}
+
+var injectionRegexp = regexp.MustCompile(`{{\w+}}`)
+
+func injectVar(src string, vars map[string]string) string {
+	return injectionRegexp.ReplaceAllStringFunc(src, func(m string) string {
+		m = strings.TrimLeft(m, "{{")
+		m = strings.TrimRight(m, "}}")
+		m = strings.TrimSpace(m)
+		return vars[m]
```
aenario: We should probably fail on non-provided variable to ease debugging.

pkg/remote/remote.go
```diff
@@ -0,0 +1,322 @@
...
+
+	client := http.Client{Timeout: 20 * time.Second}
+	res, err := client.Do(req)
+	if err != nil {
+		log.Infof("Error on request %s: %s", remote.URL.String(), err)
+		return ErrRequestFailed
+	}
+	defer res.Body.Close()
+
+	ctype := strings.SplitN(res.Header.Get("Content-Type"), ";", 2)[0] // Drop the charset
```
aenario: You should lowercase the ctype, or use `mime.ParseMediaType`

pkg/remote/remote.go
```diff
@@ -0,0 +1,320 @@
...
+}
+
+var log = logger.WithNamespace("remote")
+
+// ParseRawRequest takes a string and parse it as a remote struct.
+// First line is verb and URL.
+// Then, we have the headers.
+// And for a POST, we have a blank line, and then the body.
+func ParseRawRequest(doctype, raw string) (*Remote, error) {
+	lines := strings.Split(raw, "\n")
```
nono: It's not the same format. Ours is:

```
GET https://www.example.org/foo
...
```

For `http.ReadRequest`, it should be:

```
GET /foo HTTP/1.0
Host: www.example.org
...
```

And I also think it doesn't play well with template markers.

PR #701, opened 2017-06-09
Make sure we do not leak user info in config urls

jinroh:


nono:
I don't see how these changes avoid leaking credentials in the errors returned in HTTP response. What am I missing?

jinroh:
This is a precaution to avoid what has happened with the couchdb URLs: a pass leak in a log that did not assume that a URL contains sensitive info. It makes sure that for all URLs we store in the config struct, the user@pass part is separated.

PR #715, opened 2017-06-13
Custom contexts

nono:


docs/cli/cozy-stack_settings.md
```diff
@@ -0,0 +1,44 @@
+## cozy-stack settings
```
nono: @aeris @nledez I think you can be interested by this command to update the `tracking` setting of some instances

PR #724, opened 2017-06-14
Avoid trying to generate thumbnails for an image still uploading

nono:
We have some uploaded photos that don't have thumbnails. From the logs, I think it's possible we try too soon, while the image is still uploading (since https://github.com/cozy/cozy-stack/pull/697, the document is created in couch at the beginning of the upload). What do you think?


aenario:
LGTM, Alternatively / optimimally, we might also NOT attempt to create the thumbnail if ByteSize == 0

PR #725, opened 2017-06-15
fix: english text

y-lohse:
Hi! There was a small english mistake in one of the default folder names.
I'm unsure about the complete process here; I think since the translation id changes, I'll need to update the other languages on transifex after the PR is merged. Is that correct?


nono:
I have made the changes on transifex

y-lohse:
Cool, thank you!

PR #726, opened 2017-06-15
Worker unzip

nono:


pkg/workers/unzip/unzip.go
```diff
@@ -0,0 +1,110 @@
...
+		if err != nil {
+			if couchdb.IsConflictError(err) {
+				doc.DocName = fmt.Sprintf("%s - conflict - %d", doc.DocName, time.Now().Unix())
+				file, err = fs.CreateFile(doc, nil)
+				if err != nil {
+					return err
+				}
+			}
+		}
+		defer file.Close()
```
jinroh: For file creation, the error handling of the close method may be important (indexing is done when closing). Also the closing could happened at the end of the `io.Copy` instead of at the end of all the file copies.

PR #727, opened 2017-06-15
URL sharing discovery

Gara64:
This implements a basic discovery process to share with a recipient without knowing his Cozy address. 
It is performed in few steps:

- The recipient receives a mail containing a URL pointing to a form hosted by the sharer
- The recipient type his Cozy address in the form
- The sharer completes the recipient info and starts the sharing protocol


assets/templates/sharing_discovery.html
```diff
@@ -0,0 +1,59 @@
+<!DOCTYPE html>
+<html lang="{{.Locale}}">
+  <head>
+    <meta charset="utf-8">
+    <title>Cozy</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <link rel="stylesheet" href="/assets/fonts/fonts.css">
+    <link rel="stylesheet" href="/assets/styles/stack.css">
+    <link rel="stylesheet" href="/assets/styles/login.css">
```
nono: I'm not sure it's a good idea to include both stack.css and login.css

pkg/sharings/errors.go
```diff
@@ -21,6 +21,8 @@ var (
 	ErrRecipientHasNoEmail = errors.New("Recipient has no email")
 	// ErrRecipientHasNoURL is used to signal that a recipient has no URL.
 	ErrRecipientHasNoURL = errors.New("Recipient has no URL")
+	// ErrRecipientBadParams is used at a recipient creation when the params are not well defined
+	ErrRecipientBadParams = errors.New("Recipient parameters are not corrects")
```
nono: s/are not corrects/are invalid/ ?

pkg/sharings/send_mails.go
```diff
@@ -18,20 +18,57 @@ type mailTemplateValues struct {
...
+	if err != nil {
+		return err
+	}
+	// Fill in the description.
+	var desc string
+	if s.Desc == "" {
+		desc = "[No description provided]"
+	} else {
+		desc = s.Desc
+	}
```
nono: Personally, I prefer:

```go
desc := s.Desc
if desc == "" {
    desc = "[No description provided]"
}
```

aenario:
:+1: LGTM but travis is not happy.

Gara64:
@aenario It complains on sharing_discovery.html, which cannot be loaded. It is fixed by building assets, by as @nono said, we prefer to avoid that in the PR. So what do I do? :)

nono:
OK, so don't listen to me :wink:

Gara64:
Ok, I'll make an exception then ;)

assets/templates/sharing_discovery.html
```diff
@@ -0,0 +1,58 @@
...
+              </div>
+              <footer>
+                <div class="controls">
+                  <button type="submit">{{t "Discovery Submit"}}</button>
+                </div>
+              </footer>
+            </form>
+          </div>
+        </div>
+    </section>
```
nono: Your `</section>` seems to be not correctly indented

assets/templates/sharing_discovery.html
```diff
@@ -0,0 +1,58 @@
...
+      </main>
+    </div>
+    <script src="/assets/scripts/async-submit.js"></script>
```
nono: Do you need to submit the form asynchronously?

nono: If it is the case, you need to answer JSON to that endpoint. See https://github.com/cozy/cozy-stack/blob/master/web/auth/auth.go#L127-L159

PR #730, opened 2017-06-16
Improve log message for konnector has failed definitively

nono:


pkg/workers/konnectors/konnector.go
```diff
@@ -319,19 +319,20 @@ func commit(ctx context.Context, m *jobs.Message, errjob error) error {
...
-	})
-	if err != nil {
-		return err
 	}
-	log.Infof("Konnector has failed definitively, should send mail.", msg)
+	// msg, err := jobs.NewMessage(jobs.JSONEncoding, &mail)
+	// if err != nil {
+	// 	return err
+	// }
+	log.Infof("Konnector has failed definitively, should send mail.", mail)
```
jinroh: This should be a `log.Info` instead of `log.Infof` or contain a `%#v`

PR #733, opened 2017-06-16
feat: Choose sharing destination directory

Ljinod:
This PR adds the possibility to choose the sharing directory for a given
app and doctype.

To enable that we made the following modifications:
* the application slug is now included in the sharing document;
* the sharingID is now transmitted in all requests to be able to fetch
the application slug at the recipient;
* a settings document was created for sharing;
* a route was added to set the destination directory given an
application slug and a doctype.

### DIFF
--- a/pkg/sharings/settings.go
+++ b/pkg/sharings/settings.go
@@ -1,0 +1,187 @@
+package sharings
+
+import (
+	"time"
+
+	"github.com/cozy/cozy-stack/pkg/consts"
+	"github.com/cozy/cozy-stack/pkg/couchdb"
+	"github.com/cozy/cozy-stack/pkg/instance"
+	"github.com/cozy/cozy-stack/pkg/vfs"
+)
+
+// SharingSettings is the list of destination directories set by the
+// different applications.
+type SharingSettings struct {
+	AppDestination      map[string]map[string]string `json:"app_destination"`
+	SharingsSettingsID  string                       `json:"_id,omitempty"`
+	SharingsSettingsRev string                       `json:"_rev,omitempty"`
+	SharedWithMeDirID   string                       `json:"shared_w_me_dir_id,omitempty"`
+}
+
+// ID returns the SharingSettings qualified identifier.
+func (s SharingSettings) ID() string { return s.SharingsSettingsID }
+
+// Rev returns the SharingSettings revision.
+func (s SharingSettings) Rev() string { return s.SharingsSettingsRev }
+
+// DocType returns the SharingSettings doctype.
+func (s SharingSettings) DocType() string { return consts.Settings }
+
+// Clone returns a new SharingSettings with the same values.
+func (s *SharingSettings) Clone() couchdb.Doc {
+	cloned := *s
+	cloned.AppDestination = make(map[string]map[string]string)
+	for app, dest := range s.AppDestination {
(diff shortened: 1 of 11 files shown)


pkg/sharings/settings.go
```diff
@@ -0,0 +1,187 @@
...
+	"github.com/cozy/cozy-stack/pkg/consts"
+	"github.com/cozy/cozy-stack/pkg/couchdb"
+	"github.com/cozy/cozy-stack/pkg/instance"
+	"github.com/cozy/cozy-stack/pkg/vfs"
+)
+
+// SharingSettings is the list of destination directories set by the
+// different applications.
+type SharingSettings struct {
+	AppDestination      map[string]map[string]string `json:"app_destination"`
```
nono: You should declare an intermediate type. Maybe something like:

```go
type DestinationDirectories map[string]string
```

pkg/sharings/settings.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+// SetID changes the SharingSettings qualified identifier.
+func (s *SharingSettings) SetID(id string) { s.SharingsSettingsID = id }
+
+// SetRev changes the SharingSettings revision.
+func (s *SharingSettings) SetRev(rev string) { s.SharingsSettingsRev = rev }
+
+// Check if the database `consts.Settings` exists, creates it if not, and
+// finally creates the `SharingSettings` document.
```
nono: I'm a bit lost here. Why not try to fetch the SharingSettings document, and if it does not exist, create it?

pkg/sharings/settings.go
nono: Maybe you can add a comment like:

```go
// slug -> doctype -> dirID
```

web/sharings/sharings.go
```diff
@@ -494,3 +538,18 @@ func wrapErrors(err error) error {
 	}
 	return err
 }
+
+func doesDoctypeExist(ins *instance.Instance, doctype string) (bool, error) {
+	doctypes, err := couchdb.AllDoctypes(ins)
```
nono: I think it would be better to use `couchdb.DbStatus` to check if a doctype exists or not, instead of `AllDoctypes` (which is not very efficient).

Ljinod: I didn't know that function existed which is, indeed, a better solution!

pkg/sharings/settings.go
Ljinod: That's a good point, I don't know why I did that actually.

Ljinod: I recall now, because the function `couchdb.CreateNamedDoc` doesn't create the database if it doesn't exist. But `couchdb.CreateNamedDocWithDB` does.

web/sharings/sharings.go
```diff
@@ -494,3 +529,8 @@ func wrapErrors(err error) error {
 	}
 	return err
 }
+
+func doctypeExists(ins *instance.Instance, doctype string) bool {
+	_, err := couchdb.DBStatus(ins, doctype)
+	return err != nil
```
aenario: Not sure why it works, but shouldn't it be err==nil ?

Ljinod: Yup totally!

Ljinod:
@jinroh It should be better now :)

PR #739, opened 2017-06-20
Fix triggers

aenario:


pkg/accounts/accounts.go
```diff
@@ -61,3 +65,36 @@ func (ac *Account) Clone() couchdb.Doc {
...
+	hooks.Add(consts.Accounts, hooks.EventDelete,
+		func(i *instance.Instance, doc couchdb.Doc, old couchdb.Doc) error {
+			trigs, err := stack.GetScheduler().GetAll(i.Domain)
+			if err != nil {
+				logger.WithDomain(i.Domain).Error(
+					"Failed to fetch triggers after account deletion: ", err)
+				return err
+			}
+			for _, t := range trigs {
+				var toDelete = false
```
nono: [nitpick] I prefer `toDelete := false`

pkg/accounts/accounts.go
```diff
@@ -61,3 +65,36 @@ func (ac *Account) Clone() couchdb.Doc {
...
+
+				var msg struct {
+					Account string `json:"account"`
+				}
+				err := t.Infos().Message.Unmarshal(&msg)
+				if err != nil {
+					toDelete = true
```
nono: Are you sure? Won't it delete all the triggers that are not tied to a konnector (sharings, thumbnails, etc.)?

pkg/hooks/hooks.go
```diff
@@ -0,0 +1,59 @@
...
+		hooks = make(map[string]map[string][]Hook)
+	}
+	m, ok := hooks[doctype]
+	if !ok {
+		m = make(map[string][]Hook)
+		hooks[doctype] = m
+	}
+	hs, ok := m[event]
+	if !ok {
+		hs = make([]Hook, 3)
```
nono: Why this hard-coded 3?

web/data/data.go
```diff
@@ -111,10 +113,19 @@ func createDoc(c echo.Context) error {
 		return err
 	}
 
+	if err := hooks.Run(instance, realtime.EventCreate, &doc, nil); err != nil {
```
nono: Why does this call happen before checking the permissions? And isn't redundant with nearly the same call at line 124?

web/data/data.go
```diff
@@ -111,10 +113,19 @@ func createDoc(c echo.Context) error {
...
+		return err
+	}
+
 	if err := permissions.Allow(c, permissions.POST, &doc); err != nil {
 		return err
 	}
 
+	errHoook := hooks.Run(instance, hooks.EventCreate, doc, nil)
```
nono: Why `errHoook` and not `err`? And why 3 'o' at `errHoook`?

web/data/data.go
```diff
@@ -162,7 +173,7 @@ func UpdateDoc(c echo.Context) error {
 
 	doc.Type = c.Param("doctype")
 
-	if err := CheckWritable(doc.Type); err != nil {
+	if err := hooks.Run(instance, realtime.EventUpdate, &doc, nil); err != nil {
```
nono: Why did you remove the call to `CheckWritable`?

nono: And why this call happen before checking that the doc is valid and the permissions are OK?

web/data/data.go
nono: Why do we call this from `web/data` and not `pkg/couchdb`? I would have thought that in `pkg/couchdb`, it would have been more solid if later we had an hook on a doctype that can be modified from another `web/something` API.

web/data/data.go
nono: Shouldn't we call it after the creation has been successful? If there is a conflict, we should not call this, right?

web/data/data.go
```diff
@@ -203,6 +214,11 @@ func UpdateDoc(c echo.Context) error {
 		}
 	}
 
+	errHoook := hooks.Run(instance, hooks.EventUpdate, doc, old)
```
nono: I'd have call this after the real update to avoid calling it if there the update fails (bad revision for example)

web/data/data.go
```diff
@@ -237,10 +253,6 @@ func DeleteDoc(c echo.Context) error {
 		return jsonapi.NewError(http.StatusBadRequest, "delete without revision")
 	}
 
-	if err := CheckWritable(doctype); err != nil {
```
nono: Why did you remove this call to `CheckWritable`?

web/jobs/jobs.go
```diff
@@ -205,6 +209,43 @@ func deleteTrigger(c echo.Context) error {
...
+				}
+				deleted++
+				continue
+			}
+
+			var a accounts.Account
+			err = couchdb.GetDoc(instance, consts.Accounts, msg.Account, &a)
+			if couchdb.IsNotFoundError(err) {
+				if err = sched.Delete(instance.Domain, t.ID()); err != nil {
+					logger.WithDomain(instance.Domain).Errorln("failled to delete orphan trigger", err)
```
nono: s/failled/failed/g

pkg/accounts/accounts.go
```diff
@@ -61,3 +65,36 @@ func (ac *Account) Clone() couchdb.Doc {
...
+				} else {
+					if msg.Account == doc.ID() {
+						toDelete = true
+					}
+				}
+				if toDelete {
+					if err := stack.GetScheduler().Delete(i.Domain, t.ID()); err != nil {
+						logger.WithDomain(i.Domain).Errorln("failled to delete orphan trigger", err)
```
nono: s/failled/failed/

web/data/data.go
```diff
@@ -180,16 +191,16 @@ func UpdateDoc(c echo.Context) error {
 		return createNamedDoc(c, doc)
 	}
 
+	var old couchdb.JSONDoc
+	errFetch := couchdb.GetDoc(instance, doc.DocType(), doc.ID(), &old)
+	if errFetch != nil {
+		return errFetch
+	}
+	old.Type = doc.DocType()
```
nono: By curiosity, why do we have to force the doctype here?

PR #743, opened 2017-06-20
Use a default context when an instance has no context

nono:


gregorylegarec:
Looks simple ! 👍 
Does it work when the `cozy.yaml` does not specify a `default` context ?

nono:
> Does it work when the cozy.yaml does not specify a default context?

It depends of your expectation of "work" in this case. It will return a 404 saying that the context has not been found.

gregorylegarec:
@nono Pretty much logic actually.

PR #753, opened 2017-06-22
Use contacts as sharing recipients

Gara64:
* Migrate the recipient doctype to `io.cozy.contacts`
* Check the recipient mail exist

### DIFF
--- a/pkg/consts/consts.go
+++ b/pkg/consts/consts.go
@@ -34,8 +34,8 @@ const (
 	Permissions = "io.cozy.permissions"
 	// Queues doc type for jobs queues
 	Queues = "io.cozy.queues"
-	// Recipients doc type for sharing recipients
-	Recipients = "io.cozy.contacts"
+	// Contacts doc type for sharing
+	Contacts = "io.cozy.contacts"
 	// RemoteRequests doc type for logging requests to remote websites
 	RemoteRequests = "io.cozy.remote.requests"
 	// Sessions doc type for sessions identifying a connection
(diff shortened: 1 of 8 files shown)


pkg/consts/consts.go
```diff
@@ -35,7 +35,7 @@ const (
 	// Queues doc type for jobs queues
 	Queues = "io.cozy.queues"
 	// Recipients doc type for sharing recipients
-	Recipients = "io.cozy.recipients"
+	Recipients = "io.cozy.contacts"
```
nono: We should probably call this constant `Contacts`

aenario:
:+1: LGTM

My only issue is that this PR defines part the structure of io.cozy.contacts (see `Recipient` definition [here](https://github.com/cozy/cozy-stack/blob/dd49ca72107e335eae0d73a255ecd92521a09a2e/pkg/sharings/recipients.go#L13-L19) ) with an `url` and  `email` field.

In VCARD, both these properties are multiple. A contact can have multiple email & multiple url.

How about : 
- Changing the `url` json field to `cozy_url`. We can have a rule that a contact only has one cozy.
- **Not sure what to do for email though**, should we decide now of the contacts properties structure.

I am not sure if i remember correctly, but the discovery route expects to receive an email with the contact-id ? This will allow to choose the email from the contact picker.

Gara64:
@aenario Yep I agree with your concerns. I'm not sure this is the right time to decide of the final contact structure as things are still under discussions from what I understood of @poupotte. 

So the idea here is to use the contact doctype for the sharing integration in the UX, even though we know the `email` and `url` won't remain like this, and redefine the structure / methods once the contact format is settled.

aenario:
Fine by me, but let's mention it on thursday and make sure everyone knows it should be changed **BEFORE GOING INTO PRODUCTION**.

PR #758, opened 2017-06-22
refactor: proper revocation

Ljinod:
The revocations of a sharing and a recipient were not properly handled,
and the permissions were not verified. This PR fixes those issues.


pkg/permissions/set.go
```diff
@@ -137,3 +138,25 @@ func (ps *Set) IsSubSetOf(parent Set) bool {
 
 	return true
 }
+
+// HasSameRules returns true if the two sets have exactly the same rules.
+func (ps Set) HasSameRules(other Set) bool {
```
aenario: This implementation actually checks that all `other` Rules exists in `ps`.
Was it the purpose?

nono: So, s/HasSameRules/IsASubsetOf/ ?

web/sharings/sharings.go
```diff
@@ -413,35 +418,73 @@ func deleteDocument(c echo.Context) error {
...
+
+	err = checkRevokeSharingPermissions(c, ins, sharing, sharerID)
+	if err != nil {
+		return c.JSON(http.StatusForbidden, err)
+	}
+
+	recursiveRaw := c.QueryParam(consts.QueryParamRecursive)
+	var recursive bool
+	if recursiveRaw == "" {
+		recursive = true
```
nono: [nitpick] I'd prefer to shorten the code a bit to make it explicit that `recursive` is `true` by default:

```go
recursive := true
if recursiveRaw != "" {
    // ...
}
```

web/sharings/sharings.go
```diff
@@ -413,35 +418,73 @@ func deleteDocument(c echo.Context) error {
...
 	}
 
-	// TODO Add permission check
-	err = sharings.RevokeRecipient(ins, sharing, c.Param("recipient-client-id"))
+	recipientClientID := c.Param("recipient-client-id")
+
+	recursiveRaw := c.QueryParam(consts.QueryParamRecursive)
+	var recursive bool
+	if recursiveRaw == "" {
+		recursive = true
```
nono: Idem

web/sharings/sharings.go
```diff
@@ -597,8 +640,9 @@ func Routes(router *echo.Group) {
 	router.GET("/discovery", discoveryForm)
 	router.POST("/discovery", discovery)
 
-	router.DELETE("/:id", revokeSharing)
-	router.DELETE("/:id/recipient/:recipient-client-id", revokeRecipient)
+	router.DELETE("/:sharing-id", revokeSharing)
+	router.DELETE("/:sharing-id/recipient/:recipient-client-id",
```
Gara64: We might use the `recipient-id` instead of the `client-id`

pkg/permissions/set.go
aenario: This would be confusing, we already have `IsSubSetOf` which also check rules content : `io.cozy.files:GET:some-id < io.cozy.files`

Ljinod: The aim is to check that they are identical, i.e. exactly the same rules. I'll add a check to see if there is the same number of rules in both, which should do what I want.

web/sharings/sharings.go
Ljinod: As we discussed, yes I also think it's preferable.

pkg/permissions/set.go
nono: In that case, you should also check that `otherRule.Verbs` and `rule.Verbs` are really equal below

PR #761, opened 2017-06-23
Use domain as public name if missing

Gara64:
This avoids to block the sharing if the public name is missing. Instead, we use the first part of the domain as a default public name. For instance, zoebellot.mycozy.cloud -> zoebellot

### DIFF
--- a/pkg/sharings/recipients.go
+++ b/pkg/sharings/recipients.go
@@ -3,7 +3,6 @@ package sharings
 import (
 	"net/http"
 	"net/url"
-	"strings"
 
 	"github.com/cozy/cozy-stack/client/auth"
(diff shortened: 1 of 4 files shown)


pkg/sharings/recipients.go
```diff
@@ -209,6 +206,20 @@ func (rs *RecipientStatus) Register(instance *instance.Instance) error {
 	return nil
 }
 
+func getPublicName(instance *instance.Instance) (string, error) {
```
nono: It should be a method of `Instance`, I think

Gara64: If it's ok to return a default public name in the general case, no problem to make it part of `Instance`

nono: Well, it should be OK

PR #764, opened 2017-06-25
fix: valid function with empty field/selector

Ljinod:
Defaults the "field" parameter to "_id" if it is empty. We had a problem
in the sharings because of that.

### DIFF
--- a/pkg/couchdb/couchdb.go
+++ b/pkg/couchdb/couchdb.go
@@ -225,6 +225,9 @@ func (j JSONDoc) Valid(field, value string) bool {
 		return false
 	}
 
+	if field == "" {
+		field = "_id"
+	}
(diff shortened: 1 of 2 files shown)


nono:
Do you know from where it is called (stack trace)?

Ljinod:
@nono Yes, first here: https://github.com/cozy/cozy-stack/blob/master/pkg/sharings/sharings.go#L748 then here: https://github.com/cozy/cozy-stack/blob/master/pkg/permissions/rule.go#L97 and finally where I made the change.

nono:
> first here: https://github.com/cozy/cozy-stack/blob/master/pkg/sharings/sharings.go#L748

Can you call `permissions.Allow` here?

Ljinod:
@nono Yes, see my latest commit :)

PR #771, opened 2017-06-26
Show if instance is onboarded in cmd instances ls

jinroh:


web/instances/instances.go
```diff
@@ -146,10 +146,6 @@ func listHandler(c echo.Context) error {
 
 	objs := make([]jsonapi.Object, len(is))
 	for i, in := range is {
-		in.OAuthSecret = nil
```
aenario: Leaving the OauthSecret & SessionSecret in the client lets any app do anything...

jinroh: Do not forget this API is only available through the admin socket. But I agree we can continue to filter the secrets until we have a valid usecase.

PR #775, opened 2017-06-27
Oauth redirect from state

aenario:
This PR should allows supporting "normal" oauth provider by registering each stack as a Oauth Client.

### DIFF
--- a/web/konnectorsauth/oauth.go
+++ b/web/konnectorsauth/oauth.go
@@ -14,6 +14,7 @@ import (
 	"github.com/cozy/cozy-stack/pkg/accounts"
 	"github.com/cozy/cozy-stack/pkg/consts"
 	"github.com/cozy/cozy-stack/pkg/couchdb"
+	"github.com/cozy/cozy-stack/pkg/instance"
 	"github.com/cozy/cozy-stack/pkg/permissions"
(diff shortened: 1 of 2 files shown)


web/konnectorsauth/oauth.go
```diff
@@ -107,16 +116,24 @@ func redirect(c echo.Context) error {
 
 	if state == nil ||
 		state.AccountType != accountTypeID ||
-		state.InstanceDomain != instance.Domain {
+		(i != nil && state.InstanceDomain != i.Domain) {
 		return errors.New("bad state")
 	}
 
+	// TODO should we check if the req.Host is a given "_oauth_callback" domain?
```
nono: `TODO` spotted. Do you want to fix it in this PR?

web/konnectorsauth/oauth.go
```diff
@@ -88,14 +91,20 @@ func redirect(c echo.Context) error {
 		return err
 	}
 
-	if accessToken != "" {
+	i, _ := instance.Get(c.Request().Host)
+
+	if i != nil && accessToken != "" {
+		if i == nil {
```
nono: I don't see how `i` can not nil (line 96) and nil (line 97) here

PR #776, opened 2017-06-27
[WIP] Add realtime websocket, even on multi-stack

nono:


web/realtime/realtime.go
```diff
@@ -0,0 +1,240 @@
...
+	if err = ws.SetReadDeadline(time.Now().Add(pongWait)); err != nil {
+		return nil
+	}
+	ws.SetPongHandler(func(string) error {
+		return ws.SetReadDeadline(time.Now().Add(pongWait))
+	})
+
+	ds := realtime.GetHub().Subscriber(instance.Domain)
+	defer ds.Close()
+	errc := make(chan *wsError)
```
jinroh: Adding a buffer-size of `1` would prevent a goroutine (`readPump`) leak in case the reading out of `errc` stops for some reason (client closing its connection causing a `WriteMessage` to fail for example)

nono: I'm going to fix the issue. But using a buffered channel is not a good solution. `errc` is used in a loop: if we receive several messages in a burst, the buffer can be full and we still have the `readPump` goroutine stuck.

nono: Fixed by using a context: https://github.com/cozy/cozy-stack/pull/776/commits/ed1d4f435cf623b4f35b3ecdb5ec707ca945d1f0

PR #777, opened 2017-06-27
Allow to use include_docs on the changes feed

nono:
@aenario do you see why we didn't allow `include_docs` on the changes feed before? Just because nobody asked us? Or is there a good reason?


aenario:
@nono Yep, I think it was just to avoid allowing a potentially heavy request  before it was needed.

aenario:
but gometalinter is bad...

PR #782, opened 2017-06-28
Handle sharing directory

Gara64:
This supports the following features:

- [x] Share a directory will all its file and sub-directories
- [x] Updates on files and directories
- [x] Creation of files and sub-directories
- [x] Deletion of directories
- [x] Change of path inside the shared directory for sub-directories
- [x] Change of path outside the shared directory for sub-directories 

It doesn't support yet:

- [ ] Deletion of files
- [ ] Change of path for files

### DIFF
--- a/pkg/sharings/sharings.go
+++ b/pkg/sharings/sharings.go
@@ -239,7 +239,7 @@ func AddTrigger(instance *instance.Instance, rule permissions.Rule, sharingID st
 		eventArgs = rule.Type + ":CREATED,UPDATED,DELETED:" +
 			strings.Join(rule.Values, ",") + ":" + rule.Selector
 	} else {
-		eventArgs = rule.Type + ":UPDATED,DELETED:" +
+		eventArgs = rule.Type + ":CREATED,UPDATED,DELETED:" +
 			strings.Join(rule.Values, ",")
 	}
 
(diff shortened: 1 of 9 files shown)


web/files/files.go
```diff
@@ -375,6 +375,28 @@ func ReadFileContentFromIDHandler(c echo.Context) error {
 	return nil
 }
 
+// HeadDirOrFile handles HEAD requests on directory or file to check their
+// existence
+func HeadDirOrFile(c echo.Context) error {
```
nono: You should check permissions in this function

pkg/sharings/sharings.go
```diff
@@ -288,104 +287,171 @@ func ShareDoc(instance *instance.Instance, sharing *Sharing, recStatus *Recipien
...
+
+// sharingByValues returns the ids to share based on the Rule values
+func sharingByValues(instance *instance.Instance, rule permissions.Rule) ([]string, error) {
+	var values []string
+
+	// Iterate on values to detect directory sharing
+	for _, val := range rule.Values {
+		if rule.Type == consts.Files {
+			fs := instance.VFS()
+			dirDoc, _, err := fs.DirOrFileByID(val)
```
jinroh: could be better to use `fs.DirByID` directly if you expect a directory ? am i missing something ?

Gara64: You're not, I am the one who missed this function :)

Gara64: However, at this point we truly don't know if it's a directory or a file; so it might be nicer to use `DirOrFileByID` to avoid dealing with an error from `DirByID` in case it's a file?

jinroh: Hmm, I see. In this case the `DirOrFile` is a good call.

PR #786, opened 2017-06-29
PRO-3 | Add strength indicator in reset password form

m4dz:


assets/scripts/passwd-strength-indicator.js
```diff
@@ -0,0 +1,63 @@
+(function (window, document) {
+  const indicator = document.getElementById('password-strength')
+
+  document.getElementById('password').addEventListener('input', function(event) {
+    const strength = getStrength(event.target.value)
+    indicator.value = parseInt(strength.percentage, 10)
+    indicator.setAttribute('class', `pw-${strength.label}`)
+  }, false)
+
+  // Return given password srength as an object {percentage, label}
```
lespacedunmatin: s/srength/strength/

assets/scripts/passwd-strength-indicator.js
```diff
@@ -0,0 +1,63 @@
...
+      { regexp: /[!@#$%^&*()_+\-=[\]{};':"\\|,.<>/?]/g, size: 30 }
+    ]
+
+    const possibleChars = charsets.reduce(function (possibleChars, charset) {
+      if (charset.regexp.test(password)) possibleChars += charset.size
+      return possibleChars
+    }, 0)
+
+    const passwordStrength =
+    (Math.log(Math.pow(possibleChars, password.length)) / (Math.log(2)))
```
lespacedunmatin: Mmmmm… So "passwordpassword" would be a strong enough ?

assets/scripts/passwd-strength-indicator.js
```diff
@@ -0,0 +1,63 @@
...
+
+    const charsets = [
+      // upper
+      { regexp: /[A-Z]/g, size: 26 },
+      // lower
+      { regexp: /[a-z]/g, size: 26 },
+      // digit
+      { regexp: /[0-9]/g, size: 10 },
+      // special
+      { regexp: /[!@#$%^&*()_+\-=[\]{};':"\\|,.<>/?]/g, size: 30 }
```
lespacedunmatin: I can type "«" or “»" on my keyboard, I would be denied to use it by possibleChars ?

assets/styles/login.css
```diff
@@ -267,6 +267,54 @@ select.error {
 input[type=number] {
     width: 100%
 }
+progress {
+  margin: .5em 0;
+  width: 100%;
+  height: .5em;
+  color: #297EF2;
```
lespacedunmatin: Aren’t we supposed to use variables from cozy-ui here ?

assets/scripts/passwd-strength-indicator.js
```diff
@@ -0,0 +1,63 @@
...
+    ]
+
+    const possibleChars = charsets.reduce(function (possibleChars, charset) {
```
lespacedunmatin: Missing a comment here to make things more clear : `// possibleChars measures the char diversity of the password`.

So it might be renamed `charDiversity` ?

Not sure of what I wrote…

assets/scripts/passwd-strength-indicator.js
```diff
@@ -0,0 +1,63 @@
...
+
+  document.getElementById('password').addEventListener('input', function(event) {
+    const strength = getStrength(event.target.value)
+    indicator.value = parseInt(strength.percentage, 10)
+    indicator.setAttribute('class', `pw-${strength.label}`)
+  }, false)
+
+  // Return given password srength as an object {percentage, label}
+  getStrength = function (password) {
+    if (!password && password !== '') {
```
lespacedunmatin: Wat ?

assets/scripts/passwd-strength-indicator.js
lespacedunmatin: Ok, in fact `possibleChars` measures the total char pool size of the password, i.e. "toto" returns size 26, "Toto" returns 52, and "t0to" returns 36, is that it ?

assets/scripts/passwd-strength-indicator.js
lespacedunmatin: Ok, I misunderstood the function. 

But special shouldn’t be that small, we have a whole unicode world outside, that can be copy-pasted from wikipedia.

lespacedunmatin: We could treat all chars that are not in [a-zA-Z0-9] as a special char (but space char ?).

assets/scripts/async-submit.js
```diff
@@ -57,7 +57,4 @@
 
   passphraseInput.focus()
   submitButton.removeAttribute('disabled')
-
-  // Preload font awesome
-  try { document.fonts.load('14px FontAwesome') } catch (e) {}
```
nono: Why did you remove the preloading of the fontawesome?

m4dz: Hmm, because FontAwesome isn't in assets anymore so it triggers an unwanted error. Is it a mistake?

m4dz:
@lespacedunmatin I'm not responsible of the strength algorythm, ping @aeris for more context :wink:

assets/styles/login.css
m4dz: We can't: we don't have any build toolchain in stack assets

assets/scripts/async-submit.js
nono: It is still used here: https://github.com/cozy/cozy-stack/blob/master/assets/scripts/async-submit.js#L44

assets/scripts/passwd-strength-indicator.js
gregorylegarec: To be effective, the strength indicator needs to know the size the characters interval. So "everything except" could not work.

Furthermore, the goal is here to have an idea of the password strength, it is not important if it is not really accurate.

So I agree with you, we could add more character. **But** as this piece of code is duplicated in _onboarding_ and _settings_, we may update both of those apps to be consistent.

assets/scripts/passwd-strength-indicator.js
gregorylegarec: @lespacedunmatin Exactly !

assets/scripts/passwd-strength-indicator.js
gregorylegarec:
```js
const password = ''
console.log(!password)
// true
```
But we want to keep the case when the password is just an empty string.

assets/scripts/async-submit.js
```diff
@@ -4,11 +4,11 @@
 
   const d = window.document
   let form = d.getElementById('login-form')
-  const url = form.getAttribute('action')
+  const url = form && form.getAttribute('action')
```
gregorylegarec: @m4dz For my culture : what are the case when the form is not available ?

assets/scripts/passwd-strength-indicator.js
lespacedunmatin: Password strength should be mutualised into cozy-client-js.

assets/scripts/passwd-strength-indicator.js
nono: For me, the surprising thing is to test `!password`, not the `password !== ''` exception

gregorylegarec:
@nono May we merge this ? Is the failing build a blocker ?

nono:
I've cherry-picked the first 2 commits and kept the preloading of font awesome.

gregorylegarec:
@nono Thanks ! I missed this point ! Hopefully you're here 👍

PR #790, opened 2017-06-30
Scheduler convert

aenario:
The conversion was breaking : the stack would not stack in integration with it. 

I added tests but cant identify the cause. @nledez 

I might need help for debuging when this goes to int. let's do it on monday.


aenario:
Let's merge this whenever you want @jinroh , We will check if the int fail.

PR #793, opened 2017-07-03
Fix import a tarball

nono:


pkg/imexport/import.go
```diff
@@ -74,9 +74,14 @@ func Untardir(fs vfs.VFS, r io.Reader, dst string) error {
 				}
 			}
 
-			if _, err := io.Copy(file, tr); err != nil {
+			_, err = io.Copy(file, tr)
+			cerr := file.Close()
+			if err != nil {
 				return err
```
aenario: Are we sure the file will be closed if Copy fail ? Otherwise we need to defer the file.Close().

nono: `file.Close` is called before checking the error from `io.Copy`, so yes, the file will be closed if the copy has failed

PR #794, opened 2017-07-03
fix: emails buttons were not fully clickable

GoOz:
The first two are just improvements. The whole area was not clickable on every email clients.
The last one is a fix, as forms are forbidden in most email clients.


GoOz:
[(0.5) [GLOB] Intégration emails de partage](https://trello.com/c/Joc9c04p/418-05-glob-int%C3%A9gration-emails-de-partage)

PR #797, opened 2017-07-04
PRO-3 | Disable submit on weak reset password

m4dz:


assets/scripts/passwd-strength-indicator.js
```diff
@@ -1,10 +1,12 @@
 (function (window, document) {
   const indicator = document.getElementById('password-strength')
+  const submit = document.getElementById('login-submit')
 
   document.getElementById('password').addEventListener('input', function(event) {
     const strength = getStrength(event.target.value)
     indicator.value = parseInt(strength.percentage, 10)
     indicator.setAttribute('class', `pw-${strength.label}`)
+    submit[strength.label == 'weak' ? 'setAttribute' : 'removeAttribute']('disabled', '')
```
gregorylegarec: As this script is only an _indicator_, it should not update other DOM elements IMHO.

gregorylegarec: Actually, I think we should keep the password helper as an external module, and keep it at the exact same format of the one in [cozy-settings](https://github.com/cozy/cozy-settings/blob/master/src/lib/passwordHelper.js). It could be used in `passwd-strength-indicator` *and* in `async-submit`, to get the password strength. This strength could be use in `async-submit` to validate the password and then enable/disable the submit button *and* disallow form submission if the password is not valid. Do you think that we could mix those two logics without too much explicit if/else ? Actually I am a little bit surprised that both login form and password reset form are using the exact same script.

m4dz: @gregorylegarec I'm surprised too, essentially because the stack does the redirect itself after a reset form submission…

assets/scripts/submit.js
```diff
@@ -55,11 +58,24 @@
...
   passphraseInput.focus()
-  if (!/passphrase_renew$/.test(document.location.pathname)) {
-    submitButton.removeAttribute('disabled')
-  }
+  loginForm && submitButton.removeAttribute('disabled')
 
   // Preload font awesome
-  try { document.fonts.load('14px FontAwesome') } catch (e) {}
-})(window)
+  // try { document.fonts.load('14px FontAwesome') } catch (e) {}
```
nono: Hey, I want to keep it!

m4dz: Yup yup, a too quick merge on rebase, but see #799 :wink:

nono:
@m4dz can you rebase master?

m4dz:
@nono done :+1:

nono:
@gregorylegarec ok for you?

PR #798, opened 2017-07-04
export album

AylinKocoglu:


jinroh:
If I understand correctly, you put all albums into a single file ? why not having one file per album ?

pkg/imexport/export.go
```diff
@@ -50,8 +54,93 @@ func createDir(name string, tw *tar.Writer, dir *vfs.DirDoc) error {
...
+	hdr := &tar.Header{
+		Name:       metaDir + "album.json",
+		Size:       int64(0),
+		Mode:       0644,
+		AccessTime: time.Now(),
+		ChangeTime: time.Now(),
+		Typeflag:   tar.TypeReg,
+	}
+
+	var content [][]byte
```
nono: Using a `bytes.Buffer` for `content` may simplify the code

pkg/imexport/export.go
```diff
@@ -50,8 +54,93 @@ func createDir(name string, tw *tar.Writer, dir *vfs.DirDoc) error {
...
+		Mode:     0755,
+		Typeflag: tar.TypeDir,
+	}
+	if err := tw.WriteHeader(hdrDir); err != nil {
+		return err
+	}
+
+	hdr := &tar.Header{
+		Name:       metaDir + "album.json",
+		Size:       int64(0),
```
nono: You can remove this line, 0 is the already the default, and you set the value below

pkg/imexport/export.go
```diff
@@ -73,11 +162,15 @@ func export(tw *tar.Writer, fs vfs.VFS) error {
 
 		return nil
 	})
+	if err != nil {
+		return err
+	}
+	err = metadata(tw, fs, domain)
```
nono: You can shorten the last 2 lines of this function in `return metadata(tw, fs, domain)`

pkg/imexport/export.go
```diff
@@ -50,8 +54,95 @@ func createDir(name string, tw *tar.Writer, dir *vfs.DirDoc) error {
...
 }
 
-func export(tw *tar.Writer, fs vfs.VFS) error {
-	root := "/Documents"
+func metadata(tw *tar.Writer, fs vfs.VFS, domain string) error {
+	db := couchdb.SimpleDatabasePrefix(domain)
+	doctype := "io.cozy.files"
+	req := &couchdb.AllDocsRequest{}
+	var results []map[string]interface{}
+	if err := couchdb.GetAllDocs(db, doctype, req, &results); err != nil {
```
aenario: You can use the [FilesReferencedByView](https://github.com/cozy/cozy-stack/blob/bc8e21249be824144a5feeedf1a3d0786bc817f1/pkg/consts/views.go#L47) to fetch only files with a referenced_by.

pkg/imexport/export.go
```diff
@@ -50,8 +54,95 @@ func createDir(name string, tw *tar.Writer, dir *vfs.DirDoc) error {
...
+				if err = couchdb.GetDoc(db, v.Type, v.ID, out); err != nil {
+					return err
+				}
+				m := out.ToMapWithType()
+
+				albumName := m["name"]
+				val["album_name"] = albumName
+			}
+
+			jsonString, err := json.Marshal(val)
```
aenario: `val` contains the entire file object, which we already have in the files folder of the tarball. It would be more efficient to only save here the fields from the `io.cozy.album`,  `m` in your code. 

Careful, do not assume all `referenced_by` will be to albums, for now you can discard other types, but the export should not break when we introduce other references.

PR #799, opened 2017-07-04
Remove FontAwesome

m4dz:


nono:
Can you also remove it here: https://github.com/cozy/cozy-stack/blob/master/assets/externals#L29-L32 ?

m4dz:
> Can you also remove it here: https://github.com/cozy/cozy-stack/blob/master/assets/externals#L29-L32 ?

OH! It was there, thanks! Removed, now :wink:

PR #801, opened 2017-07-05
Check assets command result in travis

jinroh:


aenario:
Look good to me, but we need to add tx to travis `./scripts/build.sh: line 182: tx: command not found`

PR #803, opened 2017-07-05
refactor: change source id and fuse views

Ljinod:


pkg/permissions/permissions.go
```diff
@@ -435,25 +434,24 @@ func GetPermissionsByType(db couchdb.Database, doctype string, cursor couchdb.Cu
...
-		consts.SharedWithOthersPermissionsView)
+	return getSharedWithPermissionsByDoctype(db, doctype, cursor, true)
 }
 
-func getSharedWithPermissionsByDoctype(db couchdb.Database, doctype string, cursor couchdb.Cursor, view *couchdb.View) ([]*Permission, error) {
-
+func getSharedWithPermissionsByDoctype(db couchdb.Database, doctype string, cursor couchdb.Cursor, owner bool) ([]*Permission, error) {
 	var req = &couchdb.ViewRequest{
 		StartKey:    []string{doctype},
 		EndKey:      []string{doctype, couchdb.MaxString},
```
aenario: If you invert owner & sharing_id in view definition, you could filter directly here instead of in the for loop below.

Ljinod: @aenario I tried to do that but I think I have a problem: `owner` is a boolean not a string, thus I cannot pas it in `StartKey` or `EndKey`. Or can I?

nono: @Ljinod did you try to pass it as a string (`"true"` or `"false"`)? I think it should work

enguerran:
[(2) [PHO] Onglet "Who has access" dans la modale de partage en lecture seule](https://trello.com/c/dpz2Hml1/366-2-pho-onglet-who-has-access-dans-la-modale-de-partage-en-lecture-seule)

PR #812, opened 2017-07-08
Graceful restart

jinroh:
This PR implement the graceful restart. It closes gracefully the opened HTTP servers, waiting for connections to close, and shuts down the job system (brokers, workers and schedulers) waiting for their current processes to finish.

### DIFF
--- a/web/server.go
+++ b/web/server.go
@@ -9,9 +9,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
-	"os/signal"
 	"path"
-	"time"
 
 	"github.com/cozy/cozy-stack/pkg/apps"
 	"github.com/cozy/cozy-stack/pkg/config"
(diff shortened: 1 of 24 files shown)


web/server.go
```diff
@@ -143,31 +144,60 @@ func checkExists(filepath string) error {
...
+// Servers contains the started HTTP servers and implement the Shutdowner
+// interface.
+type Servers struct {
+	major *echo.Echo
+	admin *echo.Echo
+	errs  chan error
+}
+
+// Start starts the servers.
+func (e *Servers) Start() {
```
aenario: I am not sure why server .Start method errors have to be handled in go routine ?

nono: These methods won't return until the http server is stopped. So, without launching them in goroutines, the major interface won't start until the admin interface is stopped.

nono:
-> #863

PR #815, opened 2017-07-10
Allow to create a download link for a file with permission on its dir

nono:


enguerran:
[(3) [FIL] Partage par lien d'un dossier](https://trello.com/c/lcIc7af7/377-3-fil-partage-par-lien-dun-dossier)

PR #818, opened 2017-07-11
fix: now use one doctype for all types of sinistres

doubleface:
I changed the french and english po files but I am not sur if it works with transifex for the other locales. What should I do?


nono:
We are using transifex for cozy-stack. I've pushed your changes on it. Now, it's the job of the translators to update other locales.

PR #823, opened 2017-07-12
Do not keep translation locales in git

nono:
2 reasons for that:

- We are doing like this in other repositories (drive, photos, settings, etc.)
- It's painful to push translations from git to transifex (the tx client ask to pull before pushing, and doing so, lose the translations, so we have to play with git to restore them before pushing).


aenario:
Do we have the transifex client when we build in jenkins @nledez ?

nono:
@aenario we don't care. When we build the debian package on jenkins, we are using the locales that are already packed in `statik.go`

PR #824, opened 2017-07-12
Extract ID3 metadata from MP3, ogg vorbis, and FLAC

nono:


pkg/vfs/metadata.go
```diff
@@ -181,3 +186,82 @@ func (e *ExifExtractor) Result() Metadata {
...
+	e.w.CloseWithError(err)
+	<-e.ch
+}
+
+// Result is called to get the extracted metadata
+func (e *AudioExtractor) Result() Metadata {
+	m := NewMetadata()
+	tags := <-e.ch
+	switch tags := tags.(type) {
+	case tag.Metadata:
```
aenario: [curiosity] why not use a checked casting `if tagsM, ok := tags.(tag.Metadata); ok` ?

nono: Well, it would work too. The switch case on type may be easier to extend if we had to use several libs for extracting audio metadata (e.g. one for id3 and another for flac).

PR #830, opened 2017-07-13
import album

AylinKocoglu:


web/imexport/imexport.go
```diff
@@ -125,7 +126,7 @@ func importer(c echo.Context) error {
 		}
 	}
 
-	err = imexport.Untardir(fs, r, dst.ID())
+	err = imexport.Untardir(fs, r, dst.ID(), domain)
```
aenario: why not pass instance instead?

PR #838, opened 2017-07-18
docs: sharing revocation and destination directory

Ljinod:


docs/sharing.md
```diff
@@ -401,10 +401,92 @@ Receive a sharing request.
 
 Answer a sharing request.
 
+
 ### DELETE /sharings/:id
```
Gara64: Could you use `sharing-id` instead of `id`? I think this could be confusing with the `_id`.

docs/sharing.md
```diff
@@ -401,10 +401,92 @@ Receive a sharing request.
...
+```
+
+#### Response
+
+```http
+HTTP/1.1 204 No Content
+Content-Type: application/json
+```
+
+### DELETE /sharings/:id/recipient/:recipient-id
```
Gara64: Same remark

PR #842, opened 2017-07-20
Services & Notifications

aenario:


doubleface:
What about localization?

nono:
It's the job of the app/konnector to push the notification with a localized title, content, etc.

PR #854, opened 2017-07-25
Apps HTTP fetcher

jinroh:
This PR adds the possibility to install application via a simple http URL. It is a first step toward the support for the `registry://` scheme.

It also adds the support for the `registries` section of the configuration.

### DIFF
--- a/pkg/apps/fetcher_http.go
+++ b/pkg/apps/fetcher_http.go
@@ -148,7 +148,7 @@ func (f *httpFetcher) Fetch(src *url.URL, fs Copier, man Manifest) (err error) {
 	for {
 		hdr, err := tarReader.Next()
 		if err == io.EOF {
-			return nil
+			break
 		}
 		if err != nil {
 			return err
--- a/pkg/apps/fetcher_http.go
+++ b/pkg/apps/fetcher_http.go
@@ -93,6 +93,19 @@ func (f *httpFetcher) FetchManifest(src *url.URL) (r io.ReadCloser, err error) {
 }
 
 func (f *httpFetcher) Fetch(src *url.URL, fs Copier, man Manifest) (err error) {
+	exists, err := fs.Start(man.Slug(), man.Version())
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if errc := fs.Close(); errc != nil && err == nil {
+			err = errc
+		}
+	}()
+	if exists {
+		return nil
+	}
+
 	req, err := http.NewRequest(http.MethodGet, src.String(), nil)
 	if err != nil {
 		return err
@@ -129,19 +142,6 @@ func (f *httpFetcher) Fetch(src *url.URL, fs Copier, man Manifest) (err error) {
 	}
 	if err != nil {
 		return err
-	}
-
-	exists, err := fs.Start(man.Slug(), man.Version())
-	if err != nil {
-		return err
-	}
-	defer func() {
-		if errc := fs.Close(); errc != nil && err == nil {
-			err = errc
-		}
-	}()
-	if exists {
-		return nil
 	}
 
 	tarReader := tar.NewReader(reader)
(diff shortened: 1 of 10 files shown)


pkg/apps/fetcher_http.go
```diff
@@ -0,0 +1,157 @@
...
+		reader = resp.Body
+	}
+	if err != nil {
+		return nil, ErrManifestNotReachable
+	}
+	tarReader := tar.NewReader(reader)
+	for {
+		hdr, err := tarReader.Next()
+		if err == io.EOF {
+			return nil, ErrManifestNotReachable
```
nono: Shouldn't we close `resp.Body` here?

pkg/apps/fetcher_http.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+		if err != nil {
+			return nil, ErrManifestNotReachable
```
nono: Idem

pkg/apps/fetcher_http.go
```diff
@@ -0,0 +1,157 @@
...
+		if baseName != hdr.Name {
+			f.prefix = path.Dir(hdr.Name) + "/"
+		}
+		return utils.ReadCloser(tarReader, func() error {
+			return resp.Body.Close()
+		}), nil
+	}
+}
+
+func (f *httpFetcher) Fetch(src *url.URL, fs Copier, man Manifest) (err error) {
```
nono: I have suggested in https://github.com/cozy/cozy-stack/blob/master/docs/apps.md#sources to pass the checksum in the fragment of the source URL for http/https scheme. So, the marketplace can use the `http/https` scheme for installing the specific version that it has shown the permissions to the user.

pkg/apps/fetcher_http.go
```diff
@@ -0,0 +1,157 @@
...
+		return err
+	}
+
+	exists, err := fs.Start(man.Slug(), man.Version())
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if errc := fs.Close(); errc != nil {
+			err = errc
```
nono: Shouldn't we preserve `err` if it was not nil?

pkg/apps/fetcher_http.go
```diff
@@ -0,0 +1,157 @@
...
+	tarReader := tar.NewReader(reader)
+	for {
+		hdr, err := tarReader.Next()
+		if err == io.EOF {
+			return nil
+		}
+		if err != nil {
+			return err
+		}
+		if hdr.Typeflag != tar.TypeReg {
```
nono: What about the directories?

pkg/apps/fetcher_http.go
jinroh: The defer line 42 should ensure that.

pkg/apps/fetcher_http.go
jinroh: The defer line 42 should ensure that.

pkg/apps/fetcher_http.go
jinroh: Directories do not need to be specified, the copier takes care to create necessary folders given the file names.

nono: And we don't care about the empty directories?

jinroh: Didn't think about this use case since application folders are "read-only". Not sure how an empty folder can be useful, but I can miss a usecase... ?

pkg/apps/fetcher_http.go
jinroh: Great. Juste added the feature. Adding some test.

pkg/apps/fetcher_http.go
nono: Well, I don't see a use case for empty dirs too.

PR #857, opened 2017-07-25
Add some escaping features for remote variables

jinroh:


client/apps.go
```diff
@@ -98,7 +96,7 @@ func (c *Client) InstallApp(opts *AppOptions) (*AppManifest, error) {
 func (c *Client) UpdateApp(opts *AppOptions) (*AppManifest, error) {
 	res, err := c.Req(&request.Options{
 		Method:  "PUT",
-		Path:    makeAppsPath(opts.AppType, url.QueryEscape(opts.Slug)),
+		Path:    makeAppsPath(opts.AppType, opts.Slug),
```
nono: No `url.PathEscape` here?

nono:
cc: @jacquarg - IIRC, it was one of your suggestions

jinroh:
@jacquarg: there is a breaking change with this modification since now, the variable injected in the URL as query parameter or path component are automatically (and always) escaped. This may break applications that were doing this escaping themselves.

jacquarg:
Hi @jinroh , @nono ,
Thank's for the notice. 
It effectively breaks the app, and also the way I were using remote URI.

Until this change, I sended whole querystrings. 

Example :
[cozy-doctypes/org.wikidata.wbsearchentities/request](https://github.com/cozy/cozy-doctypes/blob/master/org.wikidata.wbsearchentities/request) :

```GET https://www.wikidata.org/w/api.php?action=wbsearchentities&format=json&{{params}}```

Request : 
```
let params = {
    search: filmTitle,
    language: 'fr',
    type: 'item',
    limit: limit,
  };

  params = $.param(params);
  params = encodeURIComponent(params);
  return cozy.client.fetchJSON('GET', `/remote/org.wikidata.wbsearchentities?params=${params}`);
```

With the new feature, this doesn't work, without modifying the remote doctypes definitions (beacuse '=' and '&' are escaped).

Does it means that I have to provides each parameters separately in the remote doctypes definition ? 
As we can't put optional parts in the remote doctype, I will have to create more remote doctypes.

So, whould I move the previous example, to the following form ? : 

```GET https://www.wikidata.org/w/api.php?action=wbsearchentities&format=json&search={{query search}}&language={{query language}}&type={{query type}}&limit={{query limit}}```

Request : 
```
let params = {
    search: filmTitle,
    language: 'fr',
    type: 'item',
    limit: limit,
  };

  params = $.param(params);
  return cozy.client.fetchJSON('GET', `/remote/org.wikidata.wbsearchentities?${params}`);
```

**Comment by jinroh** [comment]
@jacquarg yes you have to provide each parameters individually. They are automatically escaped so you do not need to use the `query` helper (helpers are only available for the body part):

```
GET https://www.wikidata.org/w/api.php?action=wbsearchentities&format=json&search={{search}}&language={{language}}&type={{type}}&limit={{limit}}
```

PR #875, opened 2017-08-02
Sharings should check the claims issuer and expiration

jinroh:
The permissions handler for sharings was missing the check to ensure the issuer of the JWT corresponds to the instance's domain.

### DIFF
--- a/web/sharings/sharings.go
+++ b/web/sharings/sharings.go
@@ -728,6 +728,8 @@ func checkRevokePermissions(c echo.Context, ins *instance.Instance, sharing *sha
 		return err
 	}
 
+	// TODO: avoid having to parse the JWT token twice — already done by the
+	// perm.GetPermission method.
 	var claims permissions.Claims
 	err = crypto.ParseJWT(token, func(token *jwt.Token) (interface{}, error) {
 		return ins.PickKey(token.Claims.(*permissions.Claims).Audience)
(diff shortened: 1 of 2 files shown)


web/sharings/sharings.go
```diff
@@ -733,7 +733,10 @@ func checkRevokePermissions(c echo.Context, ins *instance.Instance, sharing *sha
 		return ins.PickKey(token.Claims.(*permissions.Claims).Audience)
 	}, &claims)
 	if err != nil {
-		return err
+		return sharings.ErrForbidden
```
Ljinod: Out of curiosity why do you return `ErrForbidden` here? Because it would give away information?

web/sharings/sharings.go
```diff
@@ -733,7 +733,10 @@ func checkRevokePermissions(c echo.Context, ins *instance.Instance, sharing *sha
...
+	}
+	if claims.Issuer != ins.Domain || claims.Expired() {
```
Ljinod: @jinroh These tests might be done already:
1. we call `perm.GetPermissions` [here](https://github.com/cozy/cozy-stack/blob/master/web/sharings/sharings.go#L726)
1. which calls `extract` [here](https://github.com/cozy/cozy-stack/blob/master/web/permissions/extract.go#L138)
1. which calls `ParseJWT` [here](https://github.com/cozy/cozy-stack/blob/master/web/permissions/extract.go#L128)
1. which finally does the same check [here](https://github.com/cozy/cozy-stack/blob/master/web/permissions/extract.go#L67-L74)

I still think the PR is okay as these tests are done indirectly :)

jinroh:
Thanks for the feedback @Ljinod. I replaced this PR by a simple TODO.

PR #876, opened 2017-08-02
Add notification creation route

jinroh:
This PR lays the foundations for the notifications service by adding a route to create a notification into database.

### DIFF
--- a/pkg/notifications/notifications.go
+++ b/pkg/notifications/notifications.go
@@ -57,7 +57,7 @@ type Action struct {
 
 // Create a new notification in database.
 func Create(db couchdb.Database, sourceID string, n *Notification) error {
-	if n.Content == "" || n.Title == "" || len(n.Actions) == 0 {
+	if n.Content == "" || n.Title == "" {
 		return ErrBadNotification
 	}
 	n.Source = sourceID
--- a/pkg/notifications/notifications.go
+++ b/pkg/notifications/notifications.go
@@ -1,0 +1,67 @@
+package notifications
+
+import (
+	"github.com/cozy/cozy-stack/pkg/apps"
+	"github.com/cozy/cozy-stack/pkg/consts"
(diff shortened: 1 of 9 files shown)


pkg/notifications/notifications.go
```diff
@@ -0,0 +1,67 @@
...
+
+// Valid implements permissions.Validable
+func (n *Notification) Valid(k, f string) bool { return false }
+
+// Action describes the actions associated to a notification.
+type Action struct {
+	Text   string `json:"text"`
+	Intent struct {
+		Action string `json:"action"`
+		Type   string `json:"type"`
```
nono: I think we may want the `data` field of intents too

pkg/notifications/notifications.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+}
+
+// Create a new notification in database.
+func Create(db couchdb.Database, slug string, n *Notification) error {
+	man, err := apps.GetBySlug(db, slug, apps.Webapp)
+	if err != nil {
+		return err
+	}
+	if n.Reference == "" || n.Content == "" || n.Title == "" || len(n.Actions) == 0 {
```
nono: In the doc, content and actions are optional

web/notifications/notifications.go
```diff
@@ -0,0 +1,68 @@
...
+	return json.Marshal(n.n)
+}
+
+func createHandler(c echo.Context) error {
+	inst := middlewares.GetInstance(c)
+	n := &notifications.Notification{}
+	if _, err := jsonapi.Bind(c.Request(), &n); err != nil {
+		return err
+	}
+	slug, err := permissions.AllowForWebapp(c, permissions.POST, n)
```
nono: Konnectors should be able to create notifications too. In fact, while we don't have background services on server for webapps, it will be mostly konnectors that will create them!

nono:
Maybe one or two tests on web/ could be useful to avoid regressions later

PR #880, opened 2017-08-03
Introduce a jobs fixer command to fix stalled job

jinroh:
This PR also adds a `couchdb.ForeachDocs` method to iterate over all the document of a couchdb database.

### DIFF
--- a/pkg/couchdb/couchdb.go
+++ b/pkg/couchdb/couchdb.go
@@ -755,7 +755,7 @@ func GetAllDocs(db Database, doctype string, req *AllDocsRequest, results interf
 
 // ForeachDocs traverse all the documents from the given database with the
 // specified doctype and calls a function for each document.
-func ForeachDocs(db Database, doctype string, doc interface{}, fn func() error) error {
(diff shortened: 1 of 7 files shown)


pkg/couchdb/couchdb.go
```diff
@@ -754,6 +753,56 @@ func GetAllDocs(db Database, doctype string, req *AllDocsRequest, results interf
...
+		err = makeRequest(db, "GET", url, nil, &res)
+		if err != nil {
+			return err
+		}
+
+		var count int
+		startKey = ""
+		for _, row := range res.Rows {
+			if !strings.HasPrefix(row.ID, "_design") {
+				if err = json.Unmarshal(row.Doc, doc); err != nil {
```
nono: Do you think it is safe to reuse the same document? For example, if one row has a value for on optional field, and the next row hasn't it, won't the doc will keep it?

nono: For example, `started_at` and `error` for jobs

PR #885, opened 2017-08-10
Remove google/gops graceful shutdown as it breaks our

jinroh:
see https://github.com/google/gops/blob/b0c5c02b30c5a0eab10869d14a3238be15a7f4a5/agent/agent.go#L127-L136

### DIFF
--- a/web/server.go
+++ b/web/server.go
@@ -160,7 +160,9 @@ func listenAndServe(appsHandler echo.HandlerFunc) (*Servers, error) {
 		}))
 	}
 
-	if err = agent.Listen(nil); err != nil {
+	// XXX: Removed gops graceful shutdown as it broke our own...
+	//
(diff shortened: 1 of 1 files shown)


web/server.go
```diff
@@ -160,7 +160,9 @@ func listenAndServe(appsHandler echo.HandlerFunc) (*Servers, error) {
 		}))
 	}
 
-	if err = agent.Listen(nil); err != nil {
+	// XXX: Removed gops graceful shutdown as it broke our own...
+	//
+	if err = agent.Listen(&agent.Options{NoShutdownCleanup: true}); err != nil {
```
nono: We should call `agent.Close()` from our graceful stop function, no?

jinroh: cf #890

PR #894, opened 2017-08-18
Add state and fixer to specify that onboarding in over

jinroh:


web/jobs/jobs.go
```diff
@@ -287,6 +287,9 @@ func getJob(c echo.Context) error {
 
 func cleanJobs(c echo.Context) error {
 	instance := middlewares.GetInstance(c)
+	if err := permissions.AllowWholeType(c, permissions.GET, consts.Jobs); err != nil {
+		return err
+	}
```
nono: Not sure why it is in this PR, but OK :+1:

web/settings/instance.go
```diff
@@ -43,6 +43,7 @@ func getInstance(c echo.Context) error {
 		return err
 	}
 	doc.M["locale"] = instance.Locale
+	doc.M["onboarding_finished"] = instance.OnboardingFinished
```
nono: We should also modify the `updateInstance` below

jinroh: I was hesitant to allow mutation from this endpoint and decided to set the flag on the `/context/onboarded` handler which provide the redirect at the end of the onboarding: see https://github.com/cozy/cozy-stack/pull/894/files#diff-0ff4658f63295d93dc43ccf67d16d07cR43

nono: OK. Let's fix the tests and merge as is

PR #905, opened 2017-08-25
Windows build

sebn:
This could be a first step to run the cozy-stack on Windows.

Benefits:
- Run all the cozy-desktop tests on AppVeyor
- Develop Cozy Apps on every Windows versions

Next steps could be:
- [x] Set up a Windows build (without tests) of the cozy-stack on AppVeyor
- [ ] Fix cozy-stack tests on Windows and run them on AppVeyor

I would prefer to address those points in future PR.

WDYT?

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,25 @@
+version: "{build}"
+
+platform: x64
+
+clone_folder: c:\gopath\src\github.com\cozy\cozy-stack
+
+environment:
+  GOPATH: c:\gopath
+  GOVERSION: 1.8
+
+install:
+  - rmdir c:\go /s /q
+  - appveyor DownloadFile https://storage.googleapis.com/golang/go%GOVERSION%.windows-amd64.msi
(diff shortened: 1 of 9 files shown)


sebn:
The instance names on Linux / macOS are the same as before, so this should have no impact on existing instances.

- [ ] I still need to manually test the instance creation with default FS URL on Linux / macOS.
- [ ] I'll also fix the build.

Also some issues remain, e.g. installing apps doesn't work (looks like it's not checking out the right commit...). But this could be fixed later.

We don't need to officially support Windows yet.
Building the Windows binary on AppVeyor should prevent most regressions, to allow some experiments.
And if we need to later introduce some feature that is hard to support on Windows, we could just drop Windows support, since the time spent on it was quite low.

sebn:
Build is green, everything seems to work as expected on the Unix side. I can `rebase --autosquash` anytime. :ship: ?

sebn:
Thanks! 😄

PR #914, opened 2017-08-30
Aggregate registries list endpoints

jinroh:


pkg/registry/registry.go
```diff
@@ -115,3 +221,61 @@ func resolveInRegistry(registry *url.URL, ref *url.URL) (rc io.ReadCloser, ok bo
...
+
+func sortedList(list []json.RawMessage, cursor int, sortBy string, limit int, reverse, starved bool) ([]json.RawMessage, bool) {
+	if cursor-len(list) < limit {
+		if !starved {
+			return list, false
+		}
+		if cursor > len(list) {
+			cursor = len(list)
+		}
+		list = list[cursor:]
```
nono: I don't understand how you can remove the beginning of the list before having sorted it. What is the trick?

pkg/registry/registry.go
```diff
@@ -115,3 +221,61 @@ func resolveInRegistry(registry *url.URL, ref *url.URL) (rc io.ReadCloser, ok bo
 	}
 	return resp.Body, true, nil
 }
+
+func sortedList(list []json.RawMessage, cursor int, sortBy string, limit int, reverse, starved bool) ([]json.RawMessage, bool) {
```
nono: A comment for this function and its parameters would be welcomed!

jinroh:
@nono good :). Adding them

PR #918, opened 2017-08-31
Omit folder_to_save field when empty

nono:


gregorylegarec:
[Erreur de dossier missing](https://trello.com/c/6eT9GYhh/584-erreur-de-dossier-missing)

PR #920, opened 2017-08-31
Debounce trigger

nono:


pkg/scheduler/redis_scheduler.go
```diff
@@ -154,7 +154,21 @@ func (s *RedisScheduler) eventLoop(eventsCh <-chan *realtime.Event) {
 					event.Domain, triggerID, err.Error())
 				continue
 			}
-			_, err = s.broker.PushJob(t.(*EventTrigger).Trigger(event))
+			et := t.(*EventTrigger)
+			if et.infos.Debounce != "" {
+				var d time.Duration
+				if d, err = time.ParseDuration(et.infos.Debounce); err == nil {
```
jinroh: Shouldn't we validate at the trigger creation that the `Debounce` field is correct ?

PR #923, opened 2017-09-01
Fix a crash when commiting from a konnector not ready

nono:


pkg/workers/exec/konnector.go
```diff
@@ -213,6 +213,9 @@ func (w *konnectorWorker) Error(i *instance.Instance, err error) error {
 }
 
 func (w *konnectorWorker) Commit(ctx context.Context, msg *jobs.Message, errjob error) error {
+	if w.opts == nil {
+		return nil
```
jinroh: `return errjob` maybe ?

nono: Done

PR #924, opened 2017-09-02
WIP: Disable CSP in development

sebn:
This would allow one to use the *cozy-stack* while developping a client app to serve its development build dir including multiple dynamically-loaded js files (e.g. Google Closure with no optimizations, which is used by some language to compile to js) without having to figure out how to make this work with a third-party web server.

To do:
- [ ] Fix *401 You must be authenticated* errors when loading sourcemaps
- [ ] Test it with *cozy-app-dev*

Does it make sense? WDYT?


sebn:
Oh my, I just discovered the `--disable-csp` and `--appdir` options. Not so useful in the end 😄

PR #928, opened 2017-09-05
Disable CSP from config file

sebn:


jinroh:
I wonder with this PR if it does not break the precedence of the CLI flag `--disable-csp` over the configuration. In the case where `disable_csp: true` and --disable-csp is set on CLI.

jinroh:
Using `viper.BindPFlag` may work: https://github.com/cozy/cozy-stack/blob/master/cmd/serve.go#L199

sebn:
Thanks for the hint @jinroh, I'll give it a try as soon as possible and make sure it works in every case.

nono:
@sebn can we close this PR and let you open a new one when you will have fixed it?

sebn:
@nono No problem! Sorry for the delay.

PR #948, opened 2017-09-11
Auto updates

jinroh:
This PR adds the auto updates process. The instance struct has a new `AutoUpdate` field to activate the auto update.

The admin API has new entries to execute manually the auto-updates. The configuration also has a new block to configure the auto-updates scheduler.

~~Still missing: /settings route to set an instance `AutoUpdate`. We also coud decide to make it the default value.~~

### DIFF
--- a/cmd/instances.go
+++ b/cmd/instances.go
@@ -440,8 +440,8 @@ var oauthClientInstanceCmd = &cobra.Command{
 
 var updateCmd = &cobra.Command{
 	Use:   "update [domain] [slugs...]",
-	Short: "Starts the updates for the specified domain instance.",
-	Long: `Starts the updates for the specified domain instance. Use whether the --domain
+	Short: "Start the updates for the specified domain instance.",
+	Long: `Start the updates for the specified domain instance. Use whether the --domain
 flag to specify the instance or the --all-domains flags to updates all domains.
 The slugs arguments can be used to select which applications should be
 updated.`,
--- a/cmd/instances.go
+++ b/cmd/instances.go
@@ -20,6 +20,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var flagDomain string
 var flagLocale string
 var flagTimezone string
 var flagEmail string
@@ -437,32 +438,27 @@ var oauthClientInstanceCmd = &cobra.Command{
 	},
 }
 
-var updateAllCmd = &cobra.Command{
-	Use:   "update-all [slugs...]",
-	Short: "Starts the updates for all instances.",
-	Long: `Starts the updates for all instances. The slugs arguments can be used
-to select which applications should be updated.`,
-	RunE: func(cmd *cobra.Command, args []string) error {
-		c := newAdminClient()
-		return c.Updates(&client.UpdatesOptions{
-			Slugs: args,
-		})
-	},
-}
-
 var updateCmd = &cobra.Command{
 	Use:   "update [domain] [slugs...]",
 	Short: "Starts the updates for the specified domain instance.",
-	Long: `Starts the updates for the specified domain instance. The slugs
-arguments can be used to select which applications should be updated.`,
+	Long: `Starts the updates for the specified domain instance. Use whether the --domain
+flag to specify the instance or the --all-domains flags to updates all domains.
+The slugs arguments can be used to select which applications should be
+updated.`,
+	Aliases: []string{"updates"},
 	RunE: func(cmd *cobra.Command, args []string) error {
-		if len(args) < 1 {
-			return cmd.Help()
-		}
 		c := newAdminClient()
+		if flagAllDomains {
+			return c.Updates(&client.UpdatesOptions{
+				Slugs: args,
+			})
+		}
+		if flagDomain == "" {
+			return errAppsMissingDomain
+		}
 		return c.Updates(&client.UpdatesOptions{
-			Domain: args[0],
-			Slugs:  args[1:],
+			Domain: flagDomain,
+			Slugs:  args,
 		})
 	},
 }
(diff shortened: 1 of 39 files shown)


web/jsonapi/data.go
```diff
@@ -94,13 +94,6 @@ func MarshalObject(o Object) (json.RawMessage, error) {
 	links := o.Links()
 	rels := o.Relationships()
 
-	o.SetID("")
-	o.SetRev("")
-	defer func() {
-		o.SetID(id)
-		o.SetRev(rev)
-	}()
-
```
jinroh: @nono

I removed this hack because it provoked some race-conditions that would break installation APIs that rely on the fact that the manifest should not be mutated by consumer of the API. Their may be other race-conditions of the same nature in the code.

The side effect is that the `_id` and `_rev` in the `data` field.

cmd/instances.go
```diff
@@ -437,6 +437,36 @@ var oauthClientInstanceCmd = &cobra.Command{
 	},
 }
 
+var updateAllCmd = &cobra.Command{
+	Use:   "update-all [slugs...]",
```
nono: Maybe we can use an `--all-domains` flag like many other commands

cmd/instances.go
```diff
@@ -437,6 +437,36 @@ var oauthClientInstanceCmd = &cobra.Command{
...
+	RunE: func(cmd *cobra.Command, args []string) error {
+		c := newAdminClient()
+		return c.Updates(&client.UpdatesOptions{
+			Slugs: args,
+		})
+	},
+}
+
+var updateCmd = &cobra.Command{
+	Use:   "update [domain] [slugs...]",
```
nono: What about the `cozy-stack apps update` command? Does this command replace it? Or are they are complementary?

pkg/apps/webapp.go
```diff
@@ -398,7 +401,7 @@ func GetWebappBySlug(db couchdb.Database, slug string) (*WebappManifest, error)
 // TODO: pagination
 func ListWebapps(db couchdb.Database) ([]*WebappManifest, error) {
 	var docs []*WebappManifest
-	req := &couchdb.AllDocsRequest{Limit: 100}
+	req := &couchdb.AllDocsRequest{Limit: 200}
```
nono: Why this change? Are we near the limit of 100?

pkg/instance/instance.go
```diff
@@ -76,11 +76,12 @@ var (
...
-	DocRev string `json:"_rev,omitempty"` // couchdb _rev
-	Domain string `json:"domain"`         // The main DNS domain, like example.cozycloud.cc
-	Locale string `json:"locale"`         // The locale used on the server
-	Dev    bool   `json:"dev"`            // Whether or not the instance is for development
+	DocID        string `json:"_id,omitempty"`  // couchdb _id
+	DocRev       string `json:"_rev,omitempty"` // couchdb _rev
+	Domain       string `json:"domain"`         // The main DNS domain, like example.cozycloud.cc
+	Locale       string `json:"locale"`         // The locale used on the server
+	NoAutoUpdate bool   `json:"no_auto_update"` // Whether or not the instance has auto updates for its applications
+	Dev          bool   `json:"dev"`            // Whether or not the instance is for development
```
nono: Maybe we should use `omitempty` for `NoAutoUpdate` and `Dev`

pkg/instance/updates.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+
+	u := &updateCron{
+		stopped:  make(chan struct{}),
+		finished: make(chan struct{}),
+	}
+
+	scheduleValue := autoUpdates.Schedule
+	if scheduleValue == "" {
+		scheduleValue = "@midnight"
```
nono: I thought `""` means that the auto update is disabled, https://github.com/cozy/cozy-stack/pull/948/files#diff-091ace02b49a46f55f2586cfec89a160R306

cozy.example.yaml
```diff
@@ -76,6 +76,10 @@ mail:
 # directory with the hooks scripts - flags: --hooks
 hooks: ./scripts/hooks
 
+# Auto updates scheduler
+auto_updates:
+  schedule: "@cron 0 0 0 * * *"
```
nono: For a cluster, only one node should have the `auto_updates.schedule`? Or should all node have it?

pkg/instance/updates.go
```diff
@@ -0,0 +1,248 @@
...
+}
+
+// UpdateInstance starts the auto-update process on the given instance. The
+// slugs parameters can be used to filter (whitelist) the applications' slug
+func UpdateInstance(inst *Instance, slugs ...string) error {
+	insc := make(chan *apps.Installer)
+	errc := make(chan error)
+
+	var g sync.WaitGroup
+	g.Add(4)
```
nono: What is this `4`?

web/jsonapi/data.go
nono: I don't like that `_id` and `_rev` are present in the `data.attributes` field.

web/settings/instance.go
```diff
@@ -68,9 +69,20 @@ func updateInstance(c echo.Context) error {
 		return err
 	}
 
+	var needUpdate bool
```
nono: [nitpick] I'd prefer `needUpdate := false`

cmd/instances.go
```diff
@@ -437,6 +437,36 @@ var oauthClientInstanceCmd = &cobra.Command{
...
+	Short: "Starts the updates for the specified domain instance.",
+	Long: `Starts the updates for the specified domain instance. The slugs
```
nono: s/Starts/Start/ to be coherent with other commands

cozy.example.yaml
jinroh: For a cluster, I think it is better to control the auto updates via the administration API to have better supervision of when starting the autoupdates.

PR #964, opened 2017-09-20
Registry publish

jinroh:
Documentation and script example to automate the publishing of an application on a registry.

### DIFF
--- a/cozy.example.yaml
+++ b/cozy.example.yaml
@@ -80,6 +80,10 @@ hooks: ./scripts/hooks
 auto_updates:
   schedule: "@cron 0 0 0 * * *"
 
+# Registries used for applications and konnectors
+registries:
+  - https://registry.cozy.io/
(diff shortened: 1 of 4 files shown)


cozy.example.yaml
```diff
@@ -80,6 +80,10 @@ hooks: ./scripts/hooks
 auto_updates:
   schedule: "@cron 0 0 0 * * *"
 
+# Registries used for applications and konnectors
+registries:
+  - https://registry.cozy.io/
```
nono: Do you think we should use this registry by default when no config file is used (for the cozy-app-dev docker image for example)?

CPatchane: That would be more useful I think.

jinroh: done in 60de01bf78d3f50de7ada58d5a2915ce37e1ac35

PR #969, opened 2017-09-26
Registry: add routes to /icon and /screenshots

jinroh:


web/registry/registry.go
```diff
@@ -51,7 +54,9 @@ func proxyListReq(c echo.Context) error {
 func Routes(router *echo.Group) {
 	router.GET("", proxyListReq)
 	router.GET("/", proxyListReq)
-	router.GET("/:app", proxyReq)
-	router.GET("/:app/:version", proxyReq)
-	router.GET("/:app/:channel/latest", proxyReq)
+	router.GET("/:app", proxyReq(registry.WithCache))
+	router.GET("/:app/icon", proxyReq(registry.NoCache))
+	router.GET("/:app/screenshots/:filename", proxyReq(registry.NoCache))
```
nono: Icon and screenshots are images and will probably be loaded via an `` tag. Thus, the `Authorization` header won't be sent for those requests but `proxyReq` expects this header to check if the permissions are tied to a web app. Maybe we should pass another argument to `proxyReq` to say to just check for the authentication, not the permissions.

PR #980, opened 2017-09-29
Make jobs permissions more strict: specific to a worker 

jinroh:
Followup of #950.

Improve the `GET /jobs/queue/:worker-type` route to include the queued and running jobs via a restricted permission on the worker.

This PR also includes some simplifications of the `pkg/jobs` package, mainly removing some unused abstraction for persistence layer.


web/data/permissions.go
```diff
@@ -26,7 +26,6 @@ var blackList = map[string]bool{
 	consts.Files:            readable,
 	consts.Jobs:             readable,
 	consts.Notifications:    readable,
-	consts.Queues:           readable,
```
nono: Why this change?

web/data/permissions.go
```diff
@@ -26,7 +26,6 @@ var blackList = map[string]bool{
 	consts.Files:            readable,
 	consts.Jobs:             readable,
```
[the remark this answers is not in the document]
jinroh: Indeed. Done in a335d0e119c1f5d5a9d6e14a900fe8cbd1b13077

web/data/permissions.go
jinroh: I removed completely the `io.cozy.queues` doctype. It wasn't really required looking back.

jinroh:
With these last to commits, the jobs permissions is not *actually* reduced by this PR to ease the update of applications that still depend on "whole-type" permissions on io.cozy.jobs. As soon as these applications are updated, we will revert these last two commits.

PR #982, opened 2017-10-02
Apply more strict jobs permissions

jinroh:
Waiting for update in some applications (mainly collect) to apply this patch.
cc @CPatchane @gregorylegarec

This reverts commit 616f9fa0764a50a6fe2bcf297da938a67b2e5a87.

### DIFF
--- a/web/data/data_test.go
+++ b/web/data/data_test.go
@@ -115,10 +115,8 @@ func TestAllDoctypes(t *testing.T) {
 		"io.cozy.contacts",
 		"io.cozy.events",
 		"io.cozy.files",
-		"io.cozy.jobs",
 		"io.cozy.konnectors",
 		"io.cozy.settings",
-		"io.cozy.triggers",
 	}
 	assert.Equal(t, expected, dbs)
 }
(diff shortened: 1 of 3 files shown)


nono:
Ping @CPatchane @gregorylegarec -> when will we be able to apply this change?

CPatchane:
Is this about have restrictions according to a specific working?

nono:
@CPatchane yes, the goal is the forbid the permission for all the workers and only let apps have permissions on specific worker(s).

gregorylegarec:
What the status of this PR ? I totally forgot what it was all about.

jinroh:
@gregorylegarec depends on the applications using permissions on `io.cozy.jobs`, and especially collect.
If applications still have permissions on the whole `io.cozy.jobs` doctype, this PR cannot be merged. 

However, if applications have their permissions limited to the `worker_type` they actually care about, we should be ready to merge this.

gregorylegarec:
@jinroh Thanks for those precisions ! 👌

nono:
This PR is opened for a long time and has several files in conflict. Some applications like cozy-home still have permissions on the whole `io.cozy.jobs` and `io.cozy.triggers` doctypes (I think it is required for `io.cozy.jobs` for the realtime in the current state of the stack). Let's close the PR, we will see later what to do about that.

PR #986, opened 2017-10-02
Log every login into a session history

jinroh:


jinroh:
@nono: https://github.com/cozy/cozy-doctypes/commit/e27b3a6d790becee66a0015d00a768cbc27bcb8f#diff-d16a7c4aeb0e608e61aa6e5a4253e472

PR #987, opened 2017-10-02
Add a command to run a konnector

jinroh:


cmd/apps.go
```diff
@@ -135,6 +138,40 @@ var lsKonnectorsCmd = &cobra.Command{
...
+			return cmd.Usage()
+		}
+		slug := args[0]
+		c := newClient(flagAppsDomain, consts.Jobs+":POST:konnector:worker")
+		j, err := c.JobPush(&client.JobOptions{
+			Worker: "konnector",
+			Arguments: map[string]interface{}{
+				"konnector":      slug,
+				"account":        flagKonnectorAccount,
+				"folder_to_save": flagKonnectorFolder,
```
nono: I don't think it is easy/intuitive to know what you should pass in options. Is it the name or the id of the account? Is it a path or an ID for the folder?

PR #996, opened 2017-10-05
Avoid autocomplete on clear password login field

CPatchane:
`current-password` works only on password type input, so use `autocomplete="off"`


CPatchane:
[[forum-4991] Firefox dévoile le mot de passe](https://trello.com/c/TqnpVNAO/627-forum-4991-firefox-d%C3%A9voile-le-mot-de-passe)

PR #1004, opened 2017-10-11
Add _bulk_docs route for replications

jinroh:


tests/pouchdb-integration/tests.js
```diff
@@ -92,7 +92,7 @@ function destroyObjects (done) {
   }, done)
 }
 
-describe('Replication stack -> pouchdb', function () {
+xdescribe('Replication stack -> pouchdb', function () {
```
nono: Was it just for tests and forgotten? Or is there a good reason to disable those tests?

web/data/replication.go
```diff
@@ -79,14 +78,57 @@ func bulkGet(c echo.Context) error {
...
+	}
+
+	return proxy(c, "/")
+}
+
 func fullCommit(c echo.Context) error {
 	doctype := c.Get("doctype").(string)
 
-	if err := permissions.AllowWholeType(c, permissions.GET, doctype); err != nil {
+	if err := permissions.AllowWholeType(c, permissions.POST, doctype); err != nil {
```
nono: `_ensure_full_commit` is not used for replication from couchdb to pouchdb (ie when we are just reading from couchdb)?

PR #1007, opened 2017-10-12
Trigger run

jinroh:
This PR adds a `POST /jobs/triggers/:trigger-id/launch` route to allow launching manually the trigger.

It also removes the specific message transformation made by the the `@event` trigger to include the event (`*realtime.Event`) data inside the message field of the job. (`"message": { "message": ..., "event": {} }`). The event is now forwarded to the worker in the worker execution context in a specific key. This was used in the `sharing_updates` worker.

This way, jobs always have an `"message"` field matching the original message of the trigger, independently of the trigger type.

### DIFF
--- a/web/jobs/jobs.go
+++ b/web/jobs/jobs.go
@@ -198,6 +198,22 @@ func getTrigger(c echo.Context) error {
 	return jsonapi.Data(c, http.StatusOK, &apiTrigger{t}, nil)
 }
 
+func launchTrigger(c echo.Context) error {
+	instance := middlewares.GetInstance(c)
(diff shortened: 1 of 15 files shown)


web/jobs/jobs.go
```diff
@@ -225,6 +241,7 @@ func getAllTriggers(c echo.Context) error {
 	if err != nil {
 		return wrapJobsError(err)
 	}
+	// TODO: we could potentially benefit from an index on 'worker_type' field.
```
nono: In theory, yes. In practice, if we have so many triggers that this loop is slow, we will have so more issues that we don't really care about this particular performance issue.

PR #1008, opened 2017-10-12
Create konnector folder in the stack

jinroh:


pkg/workers/exec/konnector.go
```diff
@@ -119,6 +121,27 @@ func (w *konnectorWorker) PrepareWorkDir(i *instance.Instance, m jobs.Message) (
 		return
 	}
 
+	// Create the folder in which the konnector has the right to write.
+	{
+		folderToSave, _ := msg["folder_to_save"].(string)
+		if folderToSave != "" {
+			defaultFolderPath, _ := msg["default_folder_path"].(string)
+			if defaultFolderPath == "" {
+				defaultFolderPath = fmt.Sprintf("/???/%s", slug)
```
jinroh: @nono @poupotte what name should we default to if no `default_folder_path` is provided ?

nono: @jinroh let's say `/Administrative/%{name}` in english and `/Administatif/%{name}` in french, where `%{name}` is the name of konnector

PR #1014, opened 2017-10-18
link to source schemes documentation

enguerran:
As source schemes are documented here https://github.com/cozy/cozy-stack/blob/master/docs/apps.md#sources I found it was a good idea to link both documentation.

_Note: I created #1013 with github interface and I created a branch in the main project instead of a fork of mine. Here is the correction._


nono:
Thanks! It's a good idea.

`docs/cli` is generated from the code. I've changed the code and regenerated the docs in https://github.com/cozy/cozy-stack/commit/d13f126656b77e355b493c59be968e682b691aa6.

PR #1020, opened 2017-10-25
 Remove konnectors result and add new routes to fetch triggers' jobs 

jinroh:
This PR get rids of `io.cozy.konnectors.results` and only rely on `io.cozy.jobs`. The `io.cozy.jobs` now contain the `trigger_id` field when launched from a trigger, which allows to fetch the jobs associated with a specific trigger.

New routes:

  - `/triggers/jobs?Worker=???`: fetch all the last jobs (from a specific worker)
  - `/triggers/:trigger-id/jobs`: fetch all completed the jobs launched by a trigger
  - ~~`/triggers/:trigger-id/jobs/last`: fetch the last completed job launched by a trigger~~

/cc @CPatchane 

~~⚠️ This is not a retro-compatible change~~

### DIFF
--- a/docs/jobs.md
+++ b/docs/jobs.md
@@ -503,10 +503,14 @@ To use this endpoint, an application needs a permission on the type
 
 Get the jobs launched by the trigger with the specified ID.
 
+Query parameters:
+
+  - `Limit`: to specify the number of jobs to get out
+
 #### Request
 
 ```http
-GET /jobs/triggers/123123/jobs HTTP/1.1
+GET /jobs/triggers/123123/jobs?Limit=1 HTTP/1.1
 Accept: application/vnd.api+json
 ```
 
--- a/docs/jobs.md
+++ b/docs/jobs.md
@@ -635,7 +635,7 @@ Query parameters:
 #### Request
 
 ```http
-GET /jobs/triggers?Worker=konnector HTTP/1.1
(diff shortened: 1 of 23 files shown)


docs/jobs.md
```diff
@@ -569,24 +621,46 @@ Accept: application/vnd.api+json
...
 Accept: application/vnd.api+json
 ```
 
+#### Response
+
+```json
+{
+  "data": [
+    {
+      "type": "io.cozy.jobs",
```
nono: s/jobs/triggers/ ?

pkg/workers/exec/konnector.go
```diff
@@ -218,7 +236,59 @@ func (w *konnectorWorker) Error(i *instance.Instance, err error) error {
 }
 
 func (w *konnectorWorker) Commit(ctx *jobs.WorkerContext, errjob error) error {
-	return nil
+	// TODO: remove this retro-compatibility block
+	// <<<<<<<<<<<<<
```
jinroh: This commit puts back the `konnectorResult` logic in order to keep compatibility with current `cozy-collect` and ease its update (cc @CPatchane).

CPatchane: Thanks 👍 (poke @gregorylegarec )

PR #1024, opened 2017-10-30
Add two factor authentication

jinroh:
This PR adds two factor authentication using the following scheme:

  - when user asks for two factor authentication, this parameter is saved on the instance document directly
  - on each connection, when the TFA is activated:
    * the user is asked for its passphrase first
    * when entering correct passphrase, the user is then asked for:
      - a TOTP (Timebased One-Time password, RFC 6238) derived from the `SessionSecret` with a period of 30secondes and a skew validity of 4 periods (for ~2 minutes validity) composed of six digits, named `two-factor-passcode`
      - a short term timestamped MAC with the same validity time-range and also derived from the `SessionSecret` named `two-factor-token`
    * when sending a correct and still-valid pair `(two-factor-passcode, two-factor-token)`, the user is granted with authentication cookie

The `two-factor-passcode` is sent to the instance's owner email (more transport shall be added later) and the `two-factor-token` is sent with the response to a successful passphrased login directly in the response (in a `two_factor_token` field for `application/json` or in a hidden field for html response).

The library used for TOTP generation is: https://github.com/pquerna/otp/totp

### DIFF
--- a/assets/scripts/submit.js
+++ b/assets/scripts/submit.js
@@ -10,23 +10,36 @@
   const passphraseInput = document.getElementById('password')
   const redirectInput = document.getElementById('redirect')
   const submitButton = document.getElementById('login-submit')
+  const twoFactorPasscodeInput = document.getElementById('two-factor-passcode')
+  const twoFactorTokenInput = document.getElementById('two-factor-token')
+  const twoFactorForm = document.getElementById("two-factor-form")
+  const passwordForm = document.getElementById("password-form")
 
   let errorPanel = loginForm && loginForm.querySelector('.errors')
 
   const showError = function (error) {
-    error = error || 'The Cozy server is unavailable. Do you have network?'
+    if (error) {
+      error = '' + error
+    } else {
+      error = 'The Cozy server is unavailable. Do you have network?'
+    }
 
     if (!errorPanel) {
       errorPanel = document.createElement('div')
       errorPanel.classList.add('errors')
+      errorPanel.appendChild(document.createElement('p'))
       loginForm.appendChild(errorPanel)
     }
 
-    errorPanel.innerHTML = '<p>' + error + '</p>'
+    if (!errorPanel.firstChild) {
+      errorPanel.appendChild(document.createElement('p'))
+    }
+
+    errorPanel.firstChild.textContent = error;
     submitButton.removeAttribute('disabled')
   }
 
-  loginForm && loginForm.addEventListener('submit', (event) => {
+  const onSubmitPassphrase = function(event) {
     event.preventDefault()
     submitButton.setAttribute('disabled', true)
 
@@ -44,6 +57,10 @@
       const loginSuccess = response.status < 400
       response.json().then((body) => {
         if (loginSuccess) {
+          if (body.two_factor_token) {
+            renderTwoFactorForm(body.two_factor_token)
+            return
+          }
           submitButton.innerHTML = '<svg width="16" height="16"><use xlink:href="#fa-check"/></svg>'
           submitButton.classList.add('btn-success')
           if (body.redirect) {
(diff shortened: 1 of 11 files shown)


assets/scripts/submit.js
```diff
@@ -10,23 +10,36 @@
...
+    }
 
     if (!errorPanel) {
       errorPanel = document.createElement('div')
       errorPanel.classList.add('errors')
+      errorPanel.appendChild(document.createElement('p'))
       loginForm.appendChild(errorPanel)
     }
 
-    errorPanel.innerHTML = '<p>' + error + '</p>'
```
jinroh: XSS warning here: even though the current possible errors are all independent from user input, this may not always be the case.

assets/scripts/submit.js
```diff
@@ -10,23 +10,36 @@
   const passphraseInput = document.getElementById('password')
   const redirectInput = document.getElementById('redirect')
   const submitButton = document.getElementById('login-submit')
+  const twoFactorPasscodeInput = document.getElementById('two-factor-passcode')
+  const twoFactorTokenInput = document.getElementById('two-factor-token')
+  const twoFactorForm = document.getElementById("two-factor-form")
+  const passwordForm = document.getElementById("password-form")
```
nono: nitpick, s/"/'/ to be consistent with the rest of this script

assets/scripts/submit.js
```diff
@@ -10,23 +10,36 @@
...
+    if (error) {
+      error = '' + error
+    } else {
+      error = 'The Cozy server is unavailable. Do you have network?'
+    }
 
     if (!errorPanel) {
       errorPanel = document.createElement('div')
       errorPanel.classList.add('errors')
+      errorPanel.appendChild(document.createElement('p'))
```
nono: Why this line? It looks like a `<p>` children is created below if it doesn't exist

assets/templates/login.html
```diff
@@ -38,14 +38,21 @@
               <div role="region">
                   <input id="redirect" type="hidden" name="redirect" value="{{.Redirect}}" />
                   <p class="help" id="login-password-tip">{{.Help}}</p>
-                  <p class="line">
+                    {{if not .TwoFactorForm}}
```
nono: Indentation looks weird here

assets/templates/login.html
```diff
@@ -38,14 +38,21 @@
...
                     <button id="password-visibility-button" class="icon password-visibility-icon masked"
                         type="button"
                         title="{{t "Login Password show"}}"
                         name="password-visibility"></button>
                     <input id="password" name="passphrase" placeholder="{{t "Login Password field"}}" type="password" autofocus="true" autocomplete="current-password" />
-                  </p>
+                    </p>
+                    {{end}}
+                    <p id="two-factor-form" class="line{{if not .TwoFactorForm}} display-none{{end}}">
+                    <label for="two-factor-passcode" aria-describedby="login-two-factor-passcode-tip">{{t "Login Two factor field"}}</label>
```
nono: I don't see a node with id `login-two-factor-passcode-tip`

docs/settings.md
```diff
@@ -125,7 +125,8 @@ Cookie: sessionid=xxxx
       "locale":"fr",
       "auto_update": true,
       "email": "alice@example.com",
-      "public_name":"Alice Martin"
+      "public_name":"Alice Martin",
+      "auth_mode": "basic"
```
nono: Can you also update the auth documentation? https://cozy.github.io/cozy-stack/auth.html#faq and https://cozy.github.io/cozy-stack/auth.html#get-authlogin

pkg/instance/instance.go
```diff
@@ -47,6 +49,13 @@ const (
 // reset token.
 var passwordResetValidityDuration = 15 * time.Minute
 
+var twoFactorTOTPOptions = totp.ValidateOpts{
+	Period:    30, // 30s
+	Skew:      4,  // 30s +- 4*30s = [-2min; 2,5min]
```
nono: I think we should have a larger window. Mail can be slow, particularly with greylisting. Maybe something 5 minutes

docs/settings.md
jinroh: done https://github.com/cozy/cozy-stack/pull/1024/commits/7e1d6d4461789c6799cd4bf379fd93c96d8c0095

Yamakaky:
When will this be deployed on cozy.io instances? I can only find the email OTP options.

nono:
@Yamakaky it's already deployed. This pull request was for adding 2FA, with email as a second factor. We want to support specialized apps for providing the second factor, but it's not on our short-term top priority list.

Yamakaky:
Hum, in Settings > Profile, there is a 2FA section, but if I disable then enable it, it says that the email 2FA is enabled. The doc also only mentions email OTP.
This PR is for TOTP, right?

Yamakaky:
Oh, my bad, I missed the email-transport-only part in the original message.
What's preventing from adding full TOTP support?

PR #1025, opened 2017-10-30
 Two factor authentification: allow to skip it on a "trusted" machine 

jinroh:
⚠️ Based on #1024


pkg/instance/two_factor_auth.go
```diff
@@ -99,3 +121,12 @@ func (i *Instance) totpMACConfig() *crypto.MACConfig {
 		MaxLen: 256,
 	}
 }
+
+func (i *Instance) trustedUserMACConfig() *crypto.MACConfig {
+	return &crypto.MACConfig{
+		Name:   "trusted-user",
```
nono: Maybe s/trusted-user/trusted-device/ ?

PR #1026, opened 2017-10-30
Make io.cozy.sessions readable and link sessions history to sessions

jinroh:


web/data/permissions.go
```diff
@@ -31,6 +30,7 @@ var blackList = map[string]bool{
 	consts.Files:          readable,
 	consts.Notifications:  readable,
 	consts.RemoteRequests: readable,
+	consts.Sessions:       readable,
```
nono: I'd prefer to add a routes for this instead of making sessions readable. It's not for security but just to keep the logic about knowing when a session has expired in the stack. I think we delete a session in couchdb only when the user explicitly log out, not when he/she doesn't use his/her session for a long time.

PR #1041, opened 2017-11-08
Use couchdb 2.1.1

nono:


pkg/vfs/iter.go
```diff
@@ -77,6 +81,8 @@ func (i *iter) fetch() error {
 	}
 	err := couchdb.FindDocs(i.db, consts.Files, req, &i.list)
 	if err != nil {
+		fmt.Printf("iter err = %v\n", err)
+		fmt.Printf("sel = %#v\n", i.sel)
```
jinroh: missing printfs

nono: Good catch

PR #1044, opened 2017-11-09
Add fsck --prune flag to try and repare a broken filesystem

jinroh:


pkg/vfs/fsck.go
```diff
@@ -0,0 +1,133 @@
...
+					entry.PruneError = err
+					continue
+				}
+				fileDoc.DirID = orphanDir.ID()
+			}
+			if err := indexer.CreateFileDoc(fileDoc); err != nil {
+				entry.PruneError = err
+			}
+		case TypeMismatch:
+			// TODO: determine how to handle type mismatch... This probably require a
```
nono: What are the cases that lead here?

- If we have a directory in couchdb and a file in swift, we should keep the directory and recreate the file under a new name (like for a conflict).
- If we have a file in couchdb and a directory, we should delete the file entry and create a directory entry in couch.
- Others?

PR #1045, opened 2017-11-09
Add a content_html field to notifications

jinroh:


pkg/notifications/notifications.go
```diff
@@ -75,12 +76,21 @@ func Create(db couchdb.Database, sourceID string, n *Notification) error {
 }
 
 func sendMail(db couchdb.Database, n *Notification) error {
+	var parts []*mails.Part
+	if n.ContentHTML {
```
nono: It should be `n.ContentHTML == ""`, no?

PR #1046, opened 2017-11-13
Fix building cozy-stack on FreeBSD

nono:


aeris:
After merging, I will ask @TheKinrar for a BSD test

PR #1055, opened 2017-11-17
Distinguish notifications for client registrations and normal logins

jinroh:


pkg/sessions/login_history.go
```diff
@@ -28,9 +29,11 @@ type LoginEntry struct {
...
 	Country   string `json:"country,omitempty"`
 	// XXX No omitempty on os and browser, because they are indexed in couchdb
-	UA        string    `json:"user_agent"`
-	OS        string    `json:"os"`
-	Browser   string    `json:"browser"`
+	UA      string `json:"user_agent"`
+	OS      string `json:"os"`
+	Browser string `json:"browser"`
+	// Used when the login entry is linked to an OAuth client registration
+	ClientID  string    `json:"client_id,omitempty"`
```
nono: I'd prefer that we don't keep the `client_id` in this couchdb collection, as it is available to applications IIRC. If it is not too complicated, can we pass the client_id in redis? I think it is just used for building a link in the mail

PR #1075, opened 2017-11-27
Make the coverage script skip faster the packages with no tests

nono:


jinroh:
Looks like it skipped them all in travis.

nono:
@jinroh it should be better now

PR #1082, opened 2017-11-29
Refactor sharings

nono:


pkg/permissions/claims.go
```diff
@@ -88,3 +89,16 @@ func (claims *Claims) Expired() bool {
 	validUntil := claims.IssuedAtUTC().Add(validityDuration)
 	return validUntil.Before(time.Now().UTC())
 }
+
+// CreateCode returns a new sharecode to put the codes field of a permissions document
+func CreateCode(oauthSecret []byte, domain, name string) (string, error) {
```
jinroh: nit: why not using the `instance.MakeJWT()` ?

pkg/sharings/send_mails.go
```diff
@@ -5,206 +5,106 @@ import (
...
-func SendSharingMails(instance *instance.Instance, s *Sharing) error {
-	sharerPublicName, err := instance.PublicName()
-	if err != nil {
-		return err
-	}
-	// Fill in the description.
-	desc := s.Description
-	if desc == "" {
-		desc = "[No description provided]"
+		desc = "Surprise!"
```
jinroh: nit: you could use `i18n.Translate` to internationalize this pseudo-description.

pkg/permissions/claims.go
nono: I've just extracted some code to use at another place, but I should use `instance.MakeJWT()`

PR #1092, opened 2017-12-05
Send HTML mail for new connections

jinroh:


web/settings/passphrase.go
```diff
@@ -48,6 +50,13 @@ func registerPassphrase(c echo.Context) error {
 func updatePassphrase(c echo.Context) error {
 	instance := middlewares.GetInstance(c)
 
+	// Event if the current passphrase is sent in this request to work, we only
+	// allow enforce a valid permission to avoid having an unauthorized enpoint
+	// for bruteforce.
```
nono: This sentence seems to have some typos and is hard to read. Maybe something like:

> Even if the current passphrase is needed for this request to work, we enforce a valid permission to avoid having an unauthorized enpoint that can be bruteforced.

jinroh: Sorry about that... rewrote this comment without reading over...

PR #1100, opened 2017-12-07
Add skeleton for a notification worker with iOS and Android

jinroh:


pkg/config/config.go
```diff
@@ -338,6 +352,16 @@ func UseViper(v *viper.Viper) error {
...
+		Notifications: Notifications{
+			Development: v.GetBool("notifications.development"),
+
+			AndroidAPIKey: v.GetString("notifications.android_api_key"),
+
+			IOSCertificateKeyPath:  v.GetString("notifications.ios_certificate_key_path"),
+			IOSCertificatePassword: v.GetString("notifications.ios_certificate_password"),
+			IOSKeyID:               v.GetString("notifications.ios_key_id"),
+			IOSTeamID:              v.GetString("notifications.ios_team_id"),
+		},
```
nono: Can you update the config example file?

pkg/workers/notifications/notification.go
```diff
@@ -0,0 +1,211 @@
...
+	IOS = "ios"
+)
+
+var (
+	fcmClient *fcm.Client
+	iosClient *apns.Client
+)
+
+func init() {
+	jobs.AddWorker("notification", &jobs.WorkerConfig{
```
nono: Can you add documentation about this worker in https://github.com/cozy/cozy-stack/blob/master/docs/workers.md?

nono: I'd have named this package and this woker `push`, not `notification`. Why this choice?

pkg/workers/notifications/notification.go
```diff
@@ -0,0 +1,211 @@
...
+	if err := ctx.UnmarshalMessage(&msg); err != nil {
+		return err
+	}
+	switch msg.Platform {
+	case Android:
+		return pushToAndroid(ctx, &msg)
+	case IOS:
+		return pushToIOS(ctx, &msg)
+	default:
+		return fmt.Errorf("notifications: unknown platform %q", msg.Platform)
```
nono: `%q` -> is it a typo?

pkg/workers/notifications/notification.go
```diff
@@ -0,0 +1,211 @@
...
+			}
+		}
+	}
+
+	return errm
+}
+
+func pushToIOS(ctx *jobs.WorkerContext, msg *Message) error {
+	if iosClient == nil {
+		ctx.Logger().Warn("Could not send android notification: not configured")
```
nono: s/android/ios/

pkg/workers/notifications/notification.go
jinroh: Sure, will do.

FYI I'm still not really confident in the message type that is given to this worker. I built It as a kind of intersection between what I found for FCM (Android) and APNS (iOS) notifications, but it may not cover all use-cases, or worse, not contain enough information.

jinroh: Indeed, I like `push` better too. It also avoid confusions with our `pkg/notifications` package.

pkg/workers/notifications/notification.go
jinroh: No, it formats as a quoted string: `fmt.Sprintf("my name is %q", "pierre") -> "my name is \"pierre\""`

pkg/config/config.go
jinroh: done

docs/workers.md
```diff
@@ -3,18 +3,61 @@
...
+The `push` worker can be used to send push-notifications to a user's device.
+The options are:
+
+* `client_id`: the ID of the oauth client to push a notification to.
+* `title`: the title of the notification
+* `message`: the content of the notification
+* `priority`: the notification priority: `high` or `normal` (optional)
+* `topic`: the topic identifier of the notification (optional)
+* `sound`: a sound associated with the notification (optional)
+
```
nono: There is a `Data` thing in the code that is not documented here. Is it normal?

pkg/oauth/client.go
```diff
@@ -156,7 +170,14 @@ func (c *Client) checkMandatoryFields(i *instance.Instance) *ClientRegistrationE
 			Description: "software_id is mandatory",
 		}
 	}
-
+	switch c.NotificationPlatform {
```
nono: Maybe we can transform the platform to lowercase, before doing this check.

PR #1107, opened 2017-12-12
Change the passphrase reset confirmation page

jinroh:


assets/templates/error.html
```diff
@@ -17,7 +17,7 @@
         </header>
         <div class="container">
           <h1>{{if .ErrorTitle}}{{t .ErrorTitle}}{{else}}{{t "Error Title"}}{{end}}</h1>
-          <footer>
+          <content>
```
nono: Is `<content>` an official html tag? I don't see it on https://w3c.github.io/html-reference/elements.html#elements

assets/locales/en.po
```diff
@@ -90,6 +90,17 @@ msgstr "Are you sure you want to reset your password?"
 msgid "Passphrase reset Submit"
 msgstr "Reset password"
 
+msgid "Passphrase resetted Title"
```
nono: I think it is `reset` and not `resetted`

assets/templates/error.html
GoOz: Oh right ! O_O
When I saw it I assumed it was some templating keyword…
You're absolutely right, `<div>` is good enough.

PR #1112, opened 2017-12-13
Add created_at for notifications

### LINKED ISSUE #1108: No date for io.cozy.notifications
I just found out that notifications have no creation_date, I think it would be useful, I am debugging a very strange bug and I think this information would be useful.

nono:
Fix #1108

### DIFF
--- a/docs/notifications.md
+++ b/docs/notifications.md
@@ -17,6 +17,7 @@ content: text
 content_html: html
 icon?: image
 actions?: [{text, intent}]
+created_at: timestamp
 ```
 
 All applications, services and konnectors can create notifications if they have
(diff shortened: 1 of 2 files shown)


ptbrowne:
Thanks @nono ! Mini question, why do you use snake_case for JSON attributes (created_at and content_html) ? I thought in JS world camelCase was the default and it seems that [Google's JSON Style guide](https://google.github.io/styleguide/jsoncstyleguide.xml?showone=Property_Name_Format#Property_Name_Format) recommends camelCase.

nono:
In JS, camelCase is more used than snake_case. But for JSON, it's the opposite. There are some people advocating for camelCase, but most popular Rest API use snake_case (twitter, facebook, and even google in some places like https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages).

PR #1113, opened 2017-12-14
Add support for redis clustering or failover

jinroh:


pkg/config/config.go
```diff
@@ -154,34 +156,63 @@ type Notifications struct {
 
 // RedisConfig contains the configuration values for a redis system
 type RedisConfig struct {
-	Auth *url.Userinfo
-	URL  *url.URL
-
-	opt *redis.Options
-	cli *redis.Client
+	cli redis.UniversalClient
```
nono: Just to be sure: with this change, we won't leak the redis credentials in error messages?

jinroh: Not really, it is more because these fields were dead codes. They weren't used anywhere and have no real justification.

PR #1117, opened 2017-12-15
Immutable caching for embedded assets

jinroh:
This PR introduces immutable HTTP caching control for assets.

All assets are now served via a unique URI, containing a hash of the associated resource. The `(resource-uri, resource-content)` pair is now immutable, using a long `max-age` or the [`immutable`](https://hacks.mozilla.org/2017/01/using-immutable-caching-to-speed-up-the-web/) header value.

`Etag`s are always used for assets fetched without their unique identifier associated. In this case the `cache-control` is `no-cache`.

For application assets, we also detect the presence of a unique identifier (long hexadecimal subpart in the name of the asset), meaning that assets built with tools like webpack which add this identifier in the filename will be considered immutable. The homepage of applications are now `private, no-store, must-revalidate`, because it contains user tokens data.

The `Accept-Encoding: gzip` is also supported for the embedded assets, reducing drastically the output data of assets.


nono:
Are the old URL still available? I think we have a use case where a CSS file is loaded from outside the webapps/stack, but I don't remember which one...

jinroh:
Yes, they are. The ID is only used to set a special cache-control when used.

For instance:

```
GET /css/style.css HTTP/1.1

Cache-Control: no-cache
Etag: abcdef
```

```
GET /css/style.abcdef.css HTTP/1.1

Cache-Control: max-age=31536000, immutable
```

web/apps/serve.go
```diff
@@ -141,14 +142,27 @@ func ServeAppFile(c echo.Context, i *instance.Instance, fs apps.FileServer, app
...
 	version := app.Version()
+
 	if file != route.Index {
+		// If file is not the index, it is considered an asset of the application
+		// (JS, image, ...). For theses assets we check if it contains an unique
+		// identifier to help caching. In such case, a long cache (1 year) is set.
+		//
+		// A unique identifier is matched when the file base contains a "long"
+		// hexadecimal subpart between '.', of at least 10 characters: for instance
+		// "app.badf00dbadf00d.js".
```
nono: Can you add that to the documentation (`docs/apps.md`)?

web/middlewares/cors.go
```diff
@@ -9,59 +9,82 @@ import (
...
+
+	return func(next echo.HandlerFunc) echo.HandlerFunc {
+		return func(c echo.Context) error {
+			req := c.Request()
+			res := c.Response()
 
-		path := c.Path()
-		for _, route := range corsBlackList {
-			if strings.HasPrefix(path, route) {
+			// @TODO validate origin against oauth clients ?
```
nono: I think you can remove this TODO. Most of our OAuth clients have no origin (cordova, electron, etc.), but they don't need CORS. CORS is currently used by the web apps (with flat subdomains).

PR #1130, opened 2017-12-27
Fix a lot of things about sharing

nono:


pkg/permissions/permissions.go
```diff
@@ -13,13 +13,15 @@ import (
...
-	ExpiresAt   int               `json:"expires_at,omitempty"`
-	Codes       map[string]string `json:"codes,omitempty"`
+	PID         string `json:"_id,omitempty"`
+	PRev        string `json:"_rev,omitempty"`
+	Type        string `json:"type,omitempty"`
+	SourceID    string `json:"source_id,omitempty"`
+	Permissions Set    `json:"permissions,omitempty"`
+
+	// XXX omitempty does not work for time.Time, thus the interface{} type
+	ExpiresAt interface{}       `json:"expires_at,omitempty"`
```
jinroh: Why not using a `*time.Time` ?

nono: Good question. I have copied what was done here: https://github.com/cozy/cozy-stack/blob/master/pkg/oauth/client.go#L63

PR #1141, opened 2018-01-02
Configurable delay between password reset

aeris:


pkg/config/config.go
```diff
@@ -253,6 +255,10 @@ func GetConfig() *Config {
 	return config
 }
 
+func PasswordResetInterval() time.Duration {
```
nono: Travis build failed on the lint step. Exported functions (those with a name that starts by an upper case letter) must have a comment. Can you fix it?

jinroh:
Also you should pass a `go fmt ./...`  command to format the code.

pkg/config/config.go
```diff
@@ -253,6 +255,11 @@ func GetConfig() *Config {
 	return config
 }
 
+// PasswordResetInterval returns the minimal delay between two password reset
+func PasswordResetInterval() time.Duration {
+	return config.PasswordResetInterval
```
jinroh: You could make this method return a default like that, for instance:

```go
func PasswordResetInterval() time.Duration {
	const defaultPasswordResetInterval = 15 * time.Minute
	if config.PasswordResetInterval == 0 {
		return defaultPasswordResetInterval
	}
	return config.PasswordResetInterval
}
```

pkg/config/config.go
```diff
@@ -253,6 +255,11 @@ func GetConfig() *Config {
...
+}
+
 // Setup Viper to read the environment and the optional config file
 func Setup(cfgFile string) (err error) {
 	viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
```
jinroh: You can add a 

```go
viper.SetDefault("password_reset_interval", 15 * time.Minute)
```

jinroh:
Did a followup of this PR in #1160. It was failing due to malfunctions in our configuration defaults not applied for tests.

PR #1149, opened 2018-01-04
Update accept sharing template

y-lohse:
⚠️ ⚠️ ⚠️ ⚠️ ⚠️  **I wrote some Go code** ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ 

I'll explain what I wanted to do below so you can point me to better ways to do things. This is what I'm after:

capture d ecran 2018-01-04 12 53 31

### DIFF
--- a/web/auth/auth.go
+++ b/web/auth/auth.go
@@ -541,21 +541,37 @@ func authorizeForm(c echo.Context) error {
 			"Error":  "Error Invalid scope",
 		})
 	}
+	readOnly := true
+	for _, p := range permissions {
+		if !p.Verbs.ReadOnly() {
+			readOnly = false
+		}
+	}
 	params.client.ClientID = params.client.CouchID
+
+	clientURL, err := url.Parse(params.client.ClientURI)
+	var clientDomain string
+	if err != nil {
+		clientDomain = params.client.ClientURI
+	} else {
+		clientDomain = clientURL.Hostname()
+	}
 
 	tmpl := "authorize.html"
 	if params.resType == consts.SharingResponseType {
 		tmpl = "authorize_sharing.html"
 	}
 	return c.Render(http.StatusOK, tmpl, echo.Map{
-		"Domain":      instance.Domain,
-		"Locale":      instance.Locale,
-		"Client":      params.client,
-		"State":       params.state,
-		"RedirectURI": params.redirectURI,
-		"Scope":       params.scope,
-		"Permissions": permissions,
-		"CSRF":        c.Get("csrf"),
+		"Domain":       instance.Domain,
+		"ClientDomain": clientDomain,
+		"Locale":       instance.Locale,
+		"Client":       params.client,
+		"State":        params.state,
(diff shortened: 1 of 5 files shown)


web/auth/auth.go
```diff
@@ -541,21 +541,32 @@ func authorizeForm(c echo.Context) error {
 			"Error":  "Error Invalid scope",
 		})
 	}
+  readOnly := permissions[0].Verbs.ReadOnly()
```
y-lohse: We just want to show the permission type *once*, and not once per permission as was the case until now.
I guess it would be better to check all the permissions, but I didn't find a straightforward way to do it? Also, can `permissions` be empty, in which case this line would break things?

web/auth/auth.go
```diff
@@ -541,21 +541,32 @@ func authorizeForm(c echo.Context) error {
...
 	params.client.ClientID = params.client.CouchID
+  
+  clientURL, err := url.Parse(params.client.ClientURI)
+  var clientDomain string
+  if err != nil {
+    clientDomain = params.client.ClientURI
+  } else {
+    clientDomain = clientURL.Hostname()
+  }
```
y-lohse: We want to show the owner's domain, and the client's domain, but we only have the client's URL. Is it ok to sort of ignore the error, since I don't think it's worth showing an error when this happens (can it even happen ???) ?

y-lohse:
[(5) ETQ utilisateur je peux partager à un ou plusieurs contacts un dossier ou un fichier ou un album photos](https://trello.com/c/zmmbUM50/686-5-etq-utilisateur-je-peux-partager-%C3%A0-un-ou-plusieurs-contacts-un-dossier-ou-un-fichier-ou-un-album-photos)

web/auth/auth.go
nono: You can do something like this:

```go
readOnly := true
for _, p := range permissions {
  if !p.Verbs.ReadOnly() {
    readOnly = false
  }
}
```

nono:
@y-lohse can you rebase master, now that #1148 has been merged? It should fix travis and appveyor builds

nono:
@y-lohse the build on travis failed because of a test that check that a string is in the page you have modified. See https://travis-ci.org/cozy/cozy-stack/jobs/325136073#L1011. Do you want to fix the test yourself? Or do you prefer that we do it?

web/auth/auth.go
```diff
@@ -541,21 +541,37 @@ func authorizeForm(c echo.Context) error {
 			"Error":  "Error Invalid scope",
 		})
 	}
+  readOnly := true
+  for _, p := range permissions {
+    if !p.Verbs.ReadOnly() {
+      readOnly = false
+    }
+  }
```
nono: Oh, indentation looks weird. Can you use [gofmt](https://blog.golang.org/go-fmt-your-code) on this file?

y-lohse:
Fixed it, thanks for the pointer!

jinroh:
Thanks for the contribution @y-lohse.

PR #1150, opened 2018-01-04
Fix accept sharing

nono:


pkg/sharings/workflow.go
```diff
@@ -248,8 +253,12 @@ func SharingAccepted(i *instance.Instance, shareCode, clientID, accessCode strin
 		}
 	}
 
-	// TODO is it too soon (the recipient may not be ready)?
 	// Share all the documents with the recipient
-	err = ShareDoc(i, s, m)
-	return res, err
+	go func() {
+		time.Sleep(1 * time.Second)
```
jinroh: Maybe you could add a little comment to explain out the reason of this sleep and goroutine.

nono: It's probably something just temporary. But I'm not sure, I don't really understand how the initial replication works.

PR #1153, opened 2018-01-05
doc: add a precision about internet access for services

doubleface:


doubleface:
Closed because already fixed by another commit

PR #1167, opened 2018-01-10
Less verbose thumbnail worker

jinroh:


pkg/workers/thumbnail/thumbnail.go
```diff
@@ -66,7 +62,7 @@ func Worker(ctx *jobs.WorkerContext) error {
 		return generateThumbnails(ctx, i, &img.Doc)
 	case "UPDATED":
 		if err = removeThumbnails(i, &img.Doc); err != nil {
-			log.Infof("[jobs] failed to remove thumbnails for %s", img.Doc.ID())
+			log.Warnf("failed to remove thumbnails for %s", img.Doc.ID())
```
nono: I think we can use `debug` for this log. The files are now created with trashed: true, and when the upload finishes, updated with trashed false. It means that we have this log every time a file is uploaded.

PR #1174, opened 2018-01-11
Compress application assets

jinroh:
Adds compression before storing application assets and serve them as is when the user accepts gzip encoding with `Accept-Encoding: gzip` header.

### DIFF
--- a/pkg/apps/copier.go
+++ b/pkg/apps/copier.go
@@ -2,11 +2,14 @@ package apps
 
 import (
 	"archive/tar"
+	"compress/gzip"
 	"io"
 	"os"
 	"path"
+	"strconv"
 	"time"
 
+	"github.com/cozy/cozy-stack/pkg/magic"
 	"github.com/cozy/swift"
 	"github.com/spf13/afero"
 )
@@ -73,7 +76,19 @@ func (f *swiftCopier) Copy(stat os.FileInfo, src io.Reader) (err error) {
 		}
 	}()
 	objName := path.Join(f.rootObj, stat.Name())
-	file, err := f.c.ObjectCreate(f.container, objName, false, "", "", nil)
+	objMeta := swift.Metadata{
+		"content-encoding":        "gzip",
+		"original-content-length": strconv.FormatInt(stat.Size(), 10),
+	}
+
+	var contentType string
+	contentType, src = magic.MIMETypeFromReader(src)
+	if contentType == "" {
+		contentType = magic.MIMETypeByExtension(path.Ext(stat.Name()))
+	}
+
+	file, err := f.c.ObjectCreate(f.container, objName, false, "",
+		contentType, objMeta.ObjectHeaders())
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


pkg/apps/copier.go
```diff
@@ -82,7 +97,18 @@ func (f *swiftCopier) Copy(stat os.FileInfo, src io.Reader) (err error) {
...
 	}()
-	_, err = io.Copy(file, src)
+
+	gw, err := gzip.NewWriterLevel(file, gzip.BestCompression)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if err == nil {
+			err = gw.Close()
```
nono: We should close `gw` even if err is not nil, no?

pkg/apps/copier.go
```diff
@@ -133,7 +161,18 @@ func (f *aferoCopier) Copy(stat os.FileInfo, src io.Reader) (err error) {
...
 	}()
-	_, err = io.Copy(dst, src)
+
+	gw, err := gzip.NewWriterLevel(dst, gzip.BestCompression)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if err == nil {
+			err = gw.Close()
```
nono: Idem

pkg/apps/copier.go
```diff
@@ -114,7 +140,8 @@ func (f *aferoCopier) Copy(stat os.FileInfo, src io.Reader) (err error) {
 	if !f.started {
 		panic("copier should call Start() before Copy()")
 	}
-	fullpath := path.Join(f.appDir, stat.Name())
+
+	fullpath := path.Join(f.appDir, stat.Name()) + ".gz"
```
nono: Does travis build fail because we are now storing a file `app.tar.gz` for konnectors but still looking for `app.tar` (just a guess)?

nono: Or is it just the konnectors _tests_ are still looking for `app.tar`?

pkg/apps/server.go
```diff
@@ -47,26 +49,55 @@ func NewSwiftFileServer(conn *swift.Connection, appsType AppType) FileServer {
...
+			r, err = gzip.NewReader(f)
+			if err != nil {
+				return err
+			}
+		}
 	}
 
-	http.ServeContent(w, req, objName, unixZeroEpoch, f)
+	if contentType == "" {
+		contentType = magic.MIMETypeByExtension(path.Ext(file))
```
nono: It looks like we are already trying to guess the content-type before storing the file. Why should we try again when serving a file? Is it for old files (retro-compatibility)?

pkg/apps/server.go
```diff
@@ -111,24 +153,52 @@ func (s *aferoServer) ServeFileContent(w http.ResponseWriter, req *http.Request,
...
+		if err != nil {
+			return err
+		}
+		content = rc
 	}
 
-	if req.Header.Get("Cache-Control") == "" && req.Header.Get("Range") == "" {
-		w.Header().Set("Etag", fmt.Sprintf(`"%s"`, hex.EncodeToString(h.Sum(nil))))
+	if isGzipped {
+		w.Header().Set("Content-Encoding", "gzip")
```
nono: I don't see where do you check `Accept-Encoding` (ie the client supports gzip)?

pkg/apps/server.go
jinroh: Yes, this is only for retro-compatibility reasons

pkg/apps/copier.go
jinroh: The problem was in the tests that I had to adapt.

PR #1178, opened 2018-01-17
Add encryption to sensible fields of io.cozy.accounts

jinroh:


cmd/config.go
```diff
@@ -119,8 +120,61 @@ example: cozy-stack config passwd ~/.cozy/
 	},
 }
 
+var genRSAKeysCmd = &cobra.Command{
+	Use:   "gen-rsa-keys [filepath]",
+	Short: "Generate an RSA key pair",
+	Long: `
+cozy-stack config gen-rsa-keys generate a RSA key-pair and save them in the
+specified path. The public key filename is given a ".pub" extension suffix.
+The files permissions are 0400 for private key and 0444 for the public key.
```
nono: Can we use 0440 for the public key? I don't see a reason to allow the whole world to read this key.

web/data/data.go
```diff
@@ -466,6 +403,22 @@ func Routes(router *echo.Group) {
 	router.GET("/", dataAPIWelcome)
 	router.GET("/_all_doctypes", allDoctypes)
 
+	// Accounts are handled specifically to remove the auth fields
+	{
+		accountGroup := router.Group("/"+consts.Accounts, accountDoctype)
```
nono: Can you add some tests? I had some surprises with echo router while working on the sharing and I'd prefer to be sure that all the routes work as intented

pkg/accounts/credentials.go
```diff
@@ -0,0 +1,94 @@
...
+	// make a buffer containing the length of the login in bigendian over 4
+	// bytes, followed by the login and password contatenated.
+	creds := make([]byte, plainPrefixLen+loginLen+len(password))
+
+	// put the length of login in the first 4 bytes
+	binary.BigEndian.PutUint32(creds[0:], uint32(loginLen))
+
+	// copy the concatenation of login + password in the end
+	copy(creds[plainPrefixLen:], login)
+	copy(creds[plainPrefixLen+loginLen:], password)
```
nono: Can you add a test where you encrypt and then decrypt a login+password where the login has some not-ascii characters?

It looks like `loginLen` is the number of runes of login, but `creds` is a slice of bytes and it should be the number of bytes of login that should be used to compute the offet. Maybe I'm missing something, so a test would be a nice way to be sure that we are doing this correctly.

web/data/accounts.go
```diff
@@ -0,0 +1,218 @@
...
+		return err
+	}
+
+	if encryptAccount(out) {
+		if err = couchdb.UpdateDoc(instance, out); err != nil {
+			return err
+		}
+	}
+
+	perm, err := permissions.GetPermission(c)
```
nono: I'd prefer if we check the permissions before doing the `encryptAccount`. No good reason, it's just that is seems more logical.

web/data/accounts.go
```diff
@@ -0,0 +1,218 @@
...
+		return createNamedDoc(c, doc)
+	}
+
+	errWhole := permissions.AllowWholeType(c, permissions.PUT, doc.DocType())
+	if errWhole != nil {
+		// we cant apply to whole type, let's fetch old doc and see if it applies there
+		var old couchdb.JSONDoc
+		errFetch := couchdb.GetDoc(instance, doc.DocType(), doc.ID(), &old)
+		if errFetch != nil {
+			return errFetch
```
nono: Is it expected that we forbid the creation of a new account for permissions that are not on the whole type?

web/data/accounts.go
```diff
@@ -0,0 +1,218 @@
...
+				login, _ := m["login"].(string)
+				passwordEncrypted, err = accounts.EncryptCredentials(login, password)
+				encrypted = err == nil
+			}
+		} else if mm, ok := v.(map[string]interface{}); ok && encryptMap(mm) {
+			encrypted = true
+		}
+	}
+	if len(passwordEncrypted) > 0 {
+		delete(m, "password")
```
nono: "login" is not deleted here: does it mean that we have login in clear text + in the encrypted field?

web/data/accounts.go
jinroh: Not really. I'll simplify this handler to only handle the whole type.

web/data/accounts.go
jinroh: The permissions are already checked before, line 52. This is a specific check for the token type.

pkg/accounts/credentials.go
jinroh: `loginLen` is actually the [number of bytes in the string](https://golang.org/pkg/builtin/#len). The number of runes would be returned by the method [RuneCountInString](https://golang.org/pkg/unicode/utf8/#RuneCountInString).

However, you're right and tests are on their way :)

cmd/config.go
```diff
@@ -119,8 +120,61 @@ example: cozy-stack config passwd ~/.cozy/
 	},
 }
 
+var genKeysCmd = &cobra.Command{
+	Use:   "gen-keys [filepath]",
+	Short: "Generate an key pair for encryption and decryption of credentials",
+	Long: `
+cozy-stack config gen-keys generate a key-pair and save them in the
+specified path. The decryptor key filename is given a ".dec" extension suffix
+and the encryptor filename a ".dec". The files permissions are 0400.
```
aenario: typo ;-)

web/data/accounts.go
jinroh: Actually, it may be used by some konnectors who need to update their account during their execution. They have a specific permission on the account with its ID.

PR #1188, opened 2018-01-23
Add RandomStringFast method

jinroh:
This PR adds a `RandomStringFast` method allowing to pass a specific RNG (not thread-safe). It maximises the entropy for each roundtrip with the RNG, wasting now 4bits of entropy per 64bits , instead of 56bits wasted by the `RandomString` method. It encodes 6 bits of entropy per character with a 64 character set `[0-9a-zA-Z_-]` (the same as `base64.URLEncoding`).

In addition, the statestore used to store OAuth credentials now uses a rand/crypto for its security-sensitive work.

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -33,7 +33,7 @@ func RandomString(n int) string {
 // RandomStringFast returns a random string containing printable ascii
 // characters: [0-9a-zA-Z_-]{n}. Each character encodes 6bits of entropy. To
 // avoid wasting entropy, it is better to create a string whose length is a
-// multiple of 10. For instance a 20 bytes string will encode 120 bytes of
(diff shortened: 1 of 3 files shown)


pkg/utils/utils.go
```diff
@@ -30,6 +30,26 @@ func RandomString(n int) string {
 	return string(b)
 }
 
+// RandomStringFast returns a random string containing printable ascii
+// characters: [0-9a-zA-Z_-]{n}. Each character encodes 6bits of entropy. To
+// avoid wasting entropy, it is better to create a string whose length is a
+// multiple of 10. For instance a 20 bytes string will encode 120 bytes of
```
nono: s/120 bytes/120 bits/

PR #1203, opened 2018-01-29
Add global read/write timeouts

jinroh:
This PR adds global timeouts on the created `http.Server`. To support longer timeouts for upload and download handler, a special read/writer is used using a moving window as timeout, shifted on each read/write operation.

The `ErrorHandler` is also modified to better support the `Accept` header negotiation.


jinroh:
Closing. Too long in WIP.

PR #1211, opened 2018-01-31
fix: account delete

ptbrowne:
Seems we were matching on the wrong type ?

I added a default case and I was always passing in it :/

I printed the type and it was couchdb.JSONDoc. I admit that I don't really know what I'm doing.


jinroh:
Closing in favor of: https://github.com/cozy/cozy-stack/pull/1212.

PR #1212, opened 2018-01-31
Account deletion: add couchdb.JSONDoc value case

jinroh:


pkg/accounts/accounts.go
```diff
@@ -122,17 +122,29 @@ func init() {
...
 				}
 				if konnector == "" {
+					log.Info("No associated konnector for account: " +
+						"cannot create on_delete job")
 					return nil
 				}
 
+				logger.
+					WithField("konnector", konnector).
+					Info("Pusing job for konnector on_delete")
```
ptbrowne: s/Pusing/Pushing/

PR #1224, opened 2018-02-06
Add a no-retry flag on jobs to avoid a retry when needed

jinroh:
This PR adds a flag to avoid a retry for jobs. A konnector can send in its `critical` JSON message log a `no_retry` field as `true`, the job won't be retried, regardless the worker configuration.

### DIFF
--- a/pkg/jobs/worker.go
+++ b/pkg/jobs/worker.go
@@ -121,13 +121,13 @@ func (c *WorkerContext) WithCookie(cookie interface{}) *WorkerContext {
 }
 
 // SetNoRetry set the no-retry flag to prevent a retry on the next execution.
-func (w *WorkerContext) SetNoRetry() {
-	w.noRetry = true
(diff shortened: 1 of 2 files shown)


pkg/jobs/worker.go
```diff
@@ -367,7 +382,7 @@ func (t *task) exec(ctx *WorkerContext) (err error) {
 			if !ok {
 				err = fmt.Errorf("%v", r)
 			}
-			joblog.Errorf("%s: %s", r, debug.Stack())
+			ctx.Logger().Errorf("[panic] %s: %s", r, debug.Stack())
```
ptbrowne: c'est quoi la diff ?

jinroh: le logger est plus précis en ajoute des informatiosn sur le domaine / jobID / workerID etc. ce qui permet d'avoir plus de contexte sur le panic en question.

PR #1229, opened 2018-02-08
fix: add Promise polyfill for IE11 🚑

gregorylegarec:


gregorylegarec:
[Promise is undefined sous IE11](https://trello.com/c/dxF7kAP7/855-promise-is-undefined-sous-ie11)

assets/templates/login.html
```diff
@@ -81,6 +81,7 @@ <h1>{{.Title}}</h1>
         </div>
       </main>
     </div>
+    <script src="{{asset .Domain "/scripts/polyfills/promise.js"}}"></script>
```
nono: Promise and Fetch needs polyfills only for IE? Maybe we can use a conditional comment to load the polyfills only on IE, like this:

```html
<!--[if IE]>
  <script src="{{asset .Domain "/scripts/polyfills/promise.js"}}"></script>
  <script src="{{asset .Domain "/scripts/polyfills/fetch.js"}}"></script>
<![endif]-->
```

nono: Forget that, it looks like IE11 doesn't support conditional comments...

gregorylegarec: Also @clochix has reported some user experimenting the same kind of problem on Android.

gregorylegarec:
[Erreur de login sur IE](https://trello.com/c/AR2hp2n4/854-erreur-de-login-sur-ie)

PR #1237, opened 2018-02-14
fix: fix trust device layout

Narkfr:


Narkfr:
[(2) Améliorer la page de double authentification](https://trello.com/c/yRv5bwKC/861-2-am%C3%A9liorer-la-page-de-double-authentification)

assets/styles/login.css
```diff
@@ -194,6 +194,15 @@ label {
     font-size: .9em;
     padding: .55em 0
 }
+/* Specific styles for trust-device checkbox */
+label.two-factor-trust-device-label {
```
gregorylegarec: Does this selector need to be so specific ? Do you think we can only keep `.two-factor-trust-device-label` ?

gregorylegarec:
Thanks @Narkfr ⚽️

PR #1255, opened 2018-02-27
Remove sharing stuff

nono:


nono:
@goldoraf with this PR, the endpoints `/permissions/doctype/:doctype/sharedWithMe` and `/permissions/doctype/:doctype/sharedByMe` will respond a 404. Is it an issue for Drive/Photos?

PR #1263, opened 2018-03-02
Do not copy security HTTP headers for /remote

nono:


pkg/remote/remote.go
```diff
@@ -442,13 +442,31 @@ func ProxyRemoteAsset(name string, w http.ResponseWriter) error {
...
+}
+
 func copyHeader(dst, src http.Header) {
 	for k, vv := range src {
-		if k == "Set-Cookie" {
-			continue
+		copy := true
+		for _, h := range doNotCopyHeaders {
+			if k == h {
+				copy = false
```
jinroh: You could break here.

nono: Done

PR #1265, opened 2018-03-02
Add the route to create a sharing

nono:


web/routing.go
```diff
@@ -141,6 +142,7 @@ func SetupRoutes(router *echo.Echo) error {
 	if config.IsDevRelease() {
 		router.GET("/dev/mails/:name", devMailsHandler)
 		router.GET("/dev/templates/:name", devTemplatesHandler)
+		sharings.Routes(router.Group("/sharings"))
```
nono: Since we don't have feature flags yet and I don't want the sharing stuff to go in production, I've put the routing for it only for dev release. It is obviously only temporary.

PR #1268, opened 2018-03-06
Always ask for a new code to activate two-factor authentication

jinroh:
This PR changes the API to activate the two-factor authentication.

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -162,30 +162,6 @@ HTTP/1.1 204 No Content
 Set-Cookie: cozysessid=AAAAShoo3uo1Maic4VibuGohlik2eKUyMmZiN2Q0YTYzNDAxN2Y5NjCmp2Ja56hPgHwufpJCBBGJC2mLeJ5LCRrFFkHwaVVa; Path=/; Domain=alice.example.com; Max-Age=604800; HttpOnly; Secure
 ```
 
-### PUT /settings/confirm_mail
-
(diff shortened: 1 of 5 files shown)


docs/settings.md
```diff
@@ -299,7 +275,31 @@ Content-type: application/json
 To use this endpoint, an application needs a permission on the type
 `io.cozy.settings` for the verb `PUT`.
 
-### PUT /settings/instance/confirm_mail_tfa
+### PUT /settings/instance/activate_tfa
```
nono: Maybe we can use a more resourceful URLs like `PUT /settings/instance/tfa` and `POST /settings/instance/tfa/code` (I don't like verbs in paths, it's too RPC in my taste)

docs/settings.md
```diff
@@ -299,7 +275,31 @@ Content-type: application/json
...
+
+The user can activate of two-factor authentication. The code used as
+confirmation should have been sent via email.
+
+Status codes:
+  * `204 No Content`: when the mail has been confirmed and two-factor authentication is activated
+  * `422 Unprocessable Entity`: when the confirmation code given is not good.
```
nono: s/the confirmation code given/the given confirmation code/

docs/settings.md
nono: And later, we may have `DELETE /settings/instance/tfa` to disable two-factor authentication

PR #1271, opened 2018-03-07
Move the "trash" logic in the indexer

jinroh:
This PR moves the "move to/from trash" logic in the indexer. It allows a more robust implementation of this process.

### DIFF
--- a/pkg/vfs/couchdb_indexer.go
+++ b/pkg/vfs/couchdb_indexer.go
@@ -173,6 +173,32 @@ func (c *couchdbIndexer) DeleteDirDoc(doc *DirDoc) error {
 	return couchdb.DeleteDoc(c.db, doc)
 }
 
+func (c *couchdbIndexer) DeleteDirDocAndContent(doc *DirDoc, onlyContent bool) (ids []string, err error) {
+	var files []interface{}
+	if !onlyContent {
+		files = append(files, doc)
+	}
+	err = walk(c, doc.Name(), doc, nil, func(name string, dir *DirDoc, file *FileDoc, err error) error {
+		if err != nil {
+			return err
+		}
+		if dir != nil {
+			if dir.ID() == doc.ID() {
+				return nil
+			}
+			files = append(files, dir)
+		} else {
+			files = append(files, file)
+			ids = append(ids, file.ID())
+		}
+		return err
+	}, 0)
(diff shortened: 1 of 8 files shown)


pkg/vfs/couchdb_indexer.go
```diff
@@ -136,36 +137,82 @@ func (c *couchdbIndexer) CreateNamedDirDoc(doc *DirDoc) error {
...
-		return err
-	}
 
-	couchdocs := make([]interface{}, len(children))
-	for i, child := range children {
-		child.Fullpath = path.Join(newpath, child.Fullpath[len(oldpath)+1:])
-		couchdocs[i] = child
+	skip := 0
+	for {
+		limit := 256
```
nono: [nitpick] I'd have put `limit` outside of the `for` loop (with `skip`)

pkg/vfs/couchdb_indexer.go
```diff
@@ -136,36 +137,82 @@ func (c *couchdbIndexer) CreateNamedDirDoc(doc *DirDoc) error {
...
+			break
+		}
+		if cap(docs) < len(children) {
+			docs = make([]interface{}, 0, len(children))
+		}
+		for _, child := range children {
+			child.Fullpath = path.Join(newpath, child.Fullpath[len(oldpath)+1:])
+			docs = append(docs, child)
+		}
+		if err = couchdb.BulkUpdateDocs(c.db, consts.Files, docs); err != nil {
```
nono: I'm not sure I follow you here: why do you copy documents from `children` to `docs`, and not use `children` directly?

pkg/vfs/couchdb_indexer.go
```diff
@@ -136,36 +137,82 @@ func (c *couchdbIndexer) CreateNamedDirDoc(doc *DirDoc) error {
...
+	skip := 0
+	for {
+		limit := 256
+		sel := mango.StartWith("path", oldpath+"/")
+		req := &couchdb.FindRequest{
+			UseIndex: "dir-by-path",
+			Selector: sel,
+			Skip:     skip,
+			Limit:    limit,
+		}
```
nono: Humm, we should keep `skip` to 0. The files we want to skip are those we have updated the path and will no longer match the index. Or, am I missing something?

PR #1272, opened 2018-03-07
feat: event delete is correctly broadcasted

ptbrowne:
Deleting via `_bulk_docs` would not send realtime events as
deletions but as updates. Here we check if `_deleted` is set
to true to correctly broadcast the Delete event.

### DIFF
--- a/pkg/couchdb/proxy.go
+++ b/pkg/couchdb/proxy.go
@@ -114,6 +114,8 @@ func ProxyBulkDocs(db Database, doctype string, req *http.Request) (*httputil.Re
 					var event string
 					if doc.Rev() == "" {
 						event = realtime.EventCreate
+					} else if doc.Get("_deleted") == true {
+						event = realtime.EventDelete
 					} else {
 						event = realtime.EventUpdate
 					}


pkg/couchdb/proxy.go
```diff
@@ -114,6 +114,8 @@ func ProxyBulkDocs(db Database, doctype string, req *http.Request) (*httputil.Re
 					var event string
 					if doc.Rev() == "" {
 						event = realtime.EventCreate
+					} else if doc.Get("_deleted") == true {
```
jinroh: You probably need a type conversion here:

```go
} else if deleted, ok := doc.Get("_deleted").(bool); ok && deleted {
```

ptbrowne: It worked as is.

nono: It works, but I'm surprised that gometalinter didn't say anything about that

ptbrowne: It is working since the result of doc.Get is an interface and interface are comparable.

>  Two interface values are equal if they have identical dynamic types and equal dynamic values or if both have value nil.

ptbrowne:
Ca veut dire quoi "codecov/patch — 0% of diff hit (target 52.94%)" ? Je peux merger ?

pkg/couchdb/proxy.go
ptbrowne: @nono Why  could the problem be here ?

nono: Not a problem. It's just that gometalinter can be very nitpicking sometimes.

nono:
@ptbrowne Codecov/patch just says that your changes is not covered by unit test

PR #1282, opened 2018-03-14
Declarative notifications and notifications center

jinroh:
A declarative way to specify the notifications that a certain application is allowed to send.
For more detail, see the updated [notifications documentation page](https://github.com/jinroh/cozy-stack/blob/192ba76988dfdd486d1b80724b7e8c2a056adc86/docs/notifications.md).

### DIFF
--- a/pkg/notifications/notifications.go
+++ b/pkg/notifications/notifications.go
@@ -1,33 +1,51 @@
 package notifications
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/cozy/cozy-stack/pkg/consts"
 	"github.com/cozy/cozy-stack/pkg/couchdb"
-	"github.com/cozy/cozy-stack/pkg/globals"
-	"github.com/cozy/cozy-stack/pkg/instance"
-	"github.com/cozy/cozy-stack/pkg/jobs"
-	"github.com/cozy/cozy-stack/pkg/workers/mails"
-	"github.com/cozy/cozy-stack/pkg/workers/push"
 )
+
+// Properties is a notification type parameters, describing how a specific
+// notification group should behave.
+type Properties struct {
+	Description     string            `json:"description,omitempty"`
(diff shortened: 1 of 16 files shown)


pkg/notifications/notificationscenter/notification_center.go
```diff
@@ -0,0 +1,164 @@
+package notificationscenter
```
nono: I think `pkg/notifications/center/notification_center.go` and `packer center` should be enough. Or do you think it can cause confusion?

pkg/notifications/notifications.go
```diff
@@ -1,33 +1,51 @@
 package notifications
```
nono: For sharings, I've used a singular for the package in `pkg` and a plural for the package in `web`: `pkg/sharing` and `web/sharings` to avoid the confusion between the two packages when they are used both in a third package. Let's say it's an experiment, but we find it OK, I really hope to apply it to `permissions` (the `pkg/permissions` and `web/permissions` are often used together). Maybe, you can consider it for `notifications` too.

pkg/notifications/notificationscenter/notification_center.go
```diff
@@ -0,0 +1,164 @@
...
+	if n.ContentHTML == "" {
+		parts = []*mails.Part{
+			{Body: n.Content, Type: "text/plain"},
+		}
+	} else {
+		parts = []*mails.Part{
+			{Body: n.ContentHTML, Type: "text/html"},
+			{Body: n.Content, Type: "text/plain"},
+		}
+	}
```
nono: [nitpick] I'd have written this like that:

```go
parts := []*mails.Part{
    {Body: n.Content, Type: "text/plain"},
}
if n.ContentHTML == "" {
    parts = append(parts, &mails.Part{Body: n.ContentHTML, Type: "text/html"})
}
```

PR #1286, opened 2018-03-15
Compare two shared databases to known what to send

nono:


pkg/sharing/replicator.go
```diff
@@ -0,0 +1,190 @@
...
+// ComputeRevsDiff takes a map of id->[revisions] and returns the missing
+// revisions for those documents on the current instance.
+func (s *Sharing) ComputeRevsDiff(inst *instance.Instance, changes Changes) (*Missings, error) {
+	ids := make([]string, 0, len(changes))
+	for id := range changes {
+		ids = append(ids, id)
+	}
+	results := make([]SharedDoc, 0, len(changes))
+	req := couchdb.AllDocsRequest{Keys: ids}
+	err := couchdb.GetAllDocs(inst, consts.Shared, &req, &results)
```
jinroh: `GetAllDocs` has a default `include_docs` set to `1`: could be useful to add a negation of this default property in the `AllDocsRequest` struct, like `NoIncludeDocs`, to prevent the inclusion of the document content.

jinroh: see: https://github.com/cozy/cozy-stack/blob/master/pkg/couchdb/couchdb.go#L841

pkg/sharing/replicator.go
```diff
@@ -0,0 +1,190 @@
...
+		return ErrInvalidURL
+	}
+
+	// TODO get the last sequence number
+
+	changes, err := s.callChangesFeed(inst)
+	if err != nil {
+		return err
+	}
+	fmt.Printf("changes = %#v\n", changes)
```
jinroh: `fmt.Printf` to remove

jinroh: Or maybe you're doing because its a WIP, and dont want the compiler to yell ?

nono: Yes, this function is a WIP

PR #1287, opened 2018-03-15
Remove the globals package and merge pkg/scheduler into pkg/jobs

jinroh:
Re-factorization of the jobs packages:
  - merges the `pkg/scheduler` into `pkg/jobs`: no real reason to separate these two packages, and it is useful to have them in the same module to share primitives
  - removes the `pkg/globals` package which is no longer needed
  - new `JobSystem` struct gathering a `Broker` and `Scheduler` interfaces, exported as a global via the `jobs.System()` method


pkg/jobs/scheduler.go
```diff
@@ -174,7 +173,7 @@ func GetJobs(t Trigger, limit int) ([]*jobs.Job, error) {
 }
 
 // GetTriggerState returns the state of the trigger, calculated from the last
-// launched jobs.
+// launched
```
nono: Your `sed` was a bit too violent :laughing:

jinroh: Indeed ><

PR #1291, opened 2018-03-19
Error case to auto-delete triggers

jinroh:
Add error type to convey from the broker to the scheduler that a trigger can be removed. This is used for trigger executing konnector or service that is not installed anymore, or does not match the one defined in the service manifest document.

### DIFF
--- a/pkg/jobs/worker.go
+++ b/pkg/jobs/worker.go
@@ -262,6 +262,17 @@ func (w *Worker) work(workerID string, closed chan<- struct{}) {
 			parentCtx.Logger().Errorf("[job] error while acking job done: %s",
 				err.Error())
 		}
+
+		// Delete the trigger associated with the job (if any) when we receive a
+		// ErrBadTrigger.
+		//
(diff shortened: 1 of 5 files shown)


pkg/jobs/worker.go
```diff
@@ -391,9 +402,14 @@ func (t *task) exec(ctx *WorkerContext) (err error) {
...
-	case ErrAbort, ErrMessageUnmarshal, ErrMessageNil:
-		return false, 0, 0
+	{
+		if _, ok := prevError.(ErrBadTrigger); ok {
+			return false, 0, 0
+		}
+		switch prevError {
+		case ErrAbort, ErrMessageUnmarshal, ErrMessageNil:
+			return false, 0, 0
+		}
```
nono: Why not just this?

```go
switch prevError {
case ErrAbort, ErrMessageUnmarshal, ErrMessageNil, ErrBadTrigger:
        return false, 0, 0
}
```

jinroh: `ErrBadTrigger` is an error *type*, not a value.

PR #1293, opened 2018-03-20
Notifications: field "notId" has to be an integer

jinroh:
The `notId` field for notification push needs to be an integer. This PR changes then encoding of this value.

### DIFF
--- a/pkg/workers/push/push.go
+++ b/pkg/workers/push/push.go
@@ -4,6 +4,7 @@ import (
 	"crypto/ecdsa"
 	"crypto/md5"
 	"crypto/tls"
+	"encoding/binary"
 	"encoding/hex"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


pkg/workers/push/push.go
```diff
@@ -148,7 +150,20 @@ func pushToAndroid(ctx *jobs.WorkerContext, msg *Message) error {
...
+	var hashedSource []byte
+	if msg.Collapsible {
+		hashedSource = hashSource(msg.Source)
+	} else {
+		hashedSource = hashSource(msg.Source + msg.NotificationID)
+	}
+
+	// notID should be an integer, we take the first 32bits of the hashed source
+	// value.
+	notID := int32(binary.BigEndian.Uint32(hashedSource[:8]))
```
nono: Why do yo take 8 bytes (64 bits) from `hashedSource` if you just want 32 bits?

jinroh: 4 bytes is enough yes.

PR #1294, opened 2018-03-20
Encrypt "auth" field from io.cozy.accounts

jinroh:


jinroh:
Closing for now. The auth field may contain data used by collect.

PR #1295, opened 2018-03-21
Send documents in bulk in the sharing replicator

nono:


pkg/sharing/replicator.go
```diff
@@ -102,31 +101,35 @@ func (s *Sharing) callChangesFeed(inst *instance.Instance) (*Changes, error) {
...
-// callRevsDiff asks the other cozy to compute the revs_diff
+// callRevsDiff asks the other cozy to compute the _revs_diff
 // http://docs.couchdb.org/en/2.1.1/api/database/misc.html#db-revs-diff
 func (s *Sharing) callRevsDiff(m *Member, changes *Changes) (*Missings, error) {
 	u, err := url.Parse(m.Instance)
 	if err != nil {
 		return nil, err
 	}
-	body, err := json.Marshal(changes)
+	leafRevs := make(map[string][]string) // "doctype-docid" -> [leaf revisions]
```
jinroh: nit: looks like you know in advance the size of the map: `make(map[string][]string, len(*changes))` ?

pkg/sharing/replicator.go
```diff
@@ -150,14 +153,49 @@ func (s *Sharing) callRevsDiff(m *Member, changes *Changes) (*Missings, error) {
 	return &missings, nil
 }
 
+// computePossibleAncestors find the revisions in haves that have a generation
+// number just inferior to a generation number of a revision in wants.
+func computePossibleAncestors(wants []string, haves []string) []string {
+	// Build a sorted array of unique generation number for revisions of wants
+	var wgs []int
+	seen := make(map[int]bool)
```
jinroh: I think a `map[int]struct{}` is more compact for a set.

PR #1298, opened 2018-03-26
Notifications one job per push

jinroh:
With one job, send all the necessary push to the clients. The worker handle finding the necessary clients instead on creating one job per client.

Also, this way we do not need to store the device token into the persisted job message value.

### DIFF
--- a/pkg/workers/push/push.go
+++ b/pkg/workers/push/push.go
@@ -47,7 +47,6 @@ func init() {
 type Message struct {
 	NotificationID string `json:"notification_id"`
 	Source         string `json:"source"`
-	ClientID       string `json:"client_id,omitempty"`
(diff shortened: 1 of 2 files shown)


pkg/workers/push/push.go
```diff
@@ -48,8 +48,6 @@ type Message struct {
 	NotificationID string `json:"notification_id"`
 	Source         string `json:"source"`
 	ClientID       string `json:"client_id,omitempty"`
```
nono: Why do you still have a `ClientID` here? Is it just for retro-compatibility? If yes, you should add a comment to say so

jinroh: No we could get rid of it actually, and it is not used for the moment. I'll remove it as you suggest since I don't think we'll ever need it.

PR #1300, opened 2018-03-27
User action required errors

aenario:
For the last part, I am not sure if we should use the notification API instead.

### DIFF
--- a/docs/user-action-required.md
+++ b/docs/user-action-required.md
@@ -1,0 +1,50 @@
+[Table of contents](README.md#table-of-contents)
+
+# HTTP 402 Error
+
+In some cases the stack will return a 402 error to API requests. 
+402 will be used to denote error that require an user's action.
+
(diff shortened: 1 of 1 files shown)


docs/user-action-required.md
```diff
@@ -0,0 +1,50 @@
...
+be provided within the body of the response (see examples below)
+
+```http
+HTTP/1.1 402 Payment Required
+Content-Length: ...
+Content-Type: application/json
+
+{
+  "status": 402,
+  "error": "cgu_updated",
```
nono: s/cgu/tos/

aenario:
Change done. What do you thing about the warning phase ?

nono:
> For the last part, I am not sure if we should use the notification API instead.

The stack may use the notifications API to send a mail to the user but I don't think we should use it to display the warning in the desktop or mobile clients. Let's go for what you suggest, and we can add a variable in the HTML to help the webapps avoiding another HTTP call when booting.

PR #1303, opened 2018-03-27
Add the initial copy to the io.cozy.shared database

nono:


pkg/sharing/setup.go
```diff
@@ -0,0 +1,140 @@
...
+
+// Setup is used when a member accept a sharing to prepare the io.cozy.shared
+// database and start an initial replication. It is meant to be used in a new
+// goroutine and, as such, does not return errors but log them.
+func (s *Sharing) Setup(inst *instance.Instance, m *Member) {
+	// TODO lock
+	// TODO add triggers to update io.cozy.shared if not yet configured
+	for i, rule := range s.Rules {
+		if err := s.InitialCopy(inst, rule, i); err != nil {
+			logger.WithDomain(inst.Domain).Warnf("[sharing] Error on initial copy for %s: %s", rule.Title, err)
```
jinroh: `inst.Logger()` is handy in this case.

nono: Thanks for the tip, I will use it now

PR #1317, opened 2018-04-05
First round of debug for sharings

nono:


pkg/sharing/setup.go
```diff
@@ -114,6 +141,7 @@ func (s *Sharing) AddReplicateTrigger(inst *instance.Instance) error {
 		WorkerType: "share-replicate",
 		Message:    msg,
 		Arguments:  args,
+		Debounce:   "5s",
```
Gara64: What is the purpose of the `debounce` here?

nono: The replicator can send several documents at once, so it's more efficient to wait a bit after a change before starting the replicator, because we can group several changes in one replicator execution. For example, a connector can create several bills in the short time period.

Gara64: Ok, smart :)

PR #1325, opened 2018-04-10
Do not add share-replicate trigger on receiver side if one-way sharing

Gara64:
This prevents a nil exception on the receiver side for `push` sharings

### DIFF
--- a/pkg/sharing/rule.go
+++ b/pkg/sharing/rule.go
@@ -155,3 +155,14 @@ func (r Rule) TriggerArgs(owner bool) string {
 	}
 	return args
 }
+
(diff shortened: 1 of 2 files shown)


pkg/sharing/rule.go
```diff
@@ -155,3 +155,14 @@ func (r Rule) TriggerArgs(owner bool) string {
 	}
 	return args
 }
+
+// RuleHasSync returns true if at least one rule has an ActionRuleSync defined
+func (s *Sharing) RuleHasSync() bool {
```
nono: [nitpick] I'd have called it `TwoWays`

PR #1332, opened 2018-04-12
Konnector icon

jinroh:
Store konnectors in a flat hierarchy, like applications, to be able to extract independent files, like its icon.


nono:
Travis is red

PR #1333, opened 2018-04-12
Include shared documents for GET sharings/sharing-id

Gara64:

PR #1336, opened 2018-04-13
Sharing get doctypes

Gara64:
This PR is based on #1333.

This is a naive implementation to get the the sharing info from a doctype. Actually, this will returns as many sharing info as there are a shared documents, i.e. a sharing involving 5 docs will be returned 5 times. We might use a smarter map function and/or a reduce.

I am not confident on the json-api format either.

### DIFF
--- a/docs/sharing.md
+++ b/docs/sharing.md
@@ -310,6 +310,141 @@ Content-Type: application/vnd.api+json
 }
 ```
 
+### GET /sharings/doctype/:doctype
+
+Get information about all the sharings that have a rule for the given doctype. This includes the content of the rules, the members, as well as the already shared documents for this sharing.
+
+#### Request
+
+```http
+GET /sharings/doctype/io.cozy.files HTTP/1.1
+Host: alice.example.net
+Accept: application/vnd.api+json
+```
+
+#### Response
+
+```http
+HTTP/1.1 200 OK
+Content-Type: application/vnd.api+json
+```
+
+```json
+{
+  "data": null,
+  "meta": {
+    "count": 0
+  },
+  "included": [
+    {
+      "type": "io.cozy.sharings",
+      "id": "ce8835a061d0ef68947afe69a0046722",
+      "attributes": {
+        "description": "sharing test",
+        "preview_path": "/preview-sharing",
+        "app_slug": "drive",
+        "owner": true,
+        "created_at": "2018-01-04T12:35:08Z",
+        "updated_at": "2018-01-04T13:45:43Z",
+        "members": [
+          {
+            "status": "owner",
+            "name": "Alice",
+            "email": "alice@example.net",
+            "instance": "alice.example.net"
+          },
+          {
+            "status": "ready",
+            "name": "Bob",
+            "email": "bob@example.net"
+          }
+        ],
+        "rules": [
+          {
+            "title": "Hawaii",
+            "doctype": "io.cozy.files",
+            "values": ["612acf1c-1d72-11e8-b043-ef239d3074dd"],
+            "add": "sync",
+            "update": "sync",
+            "remove": "sync"
+          }
+        ],
+      },
+      "meta": {
+        "rev": "1-4859c6c755143adf0838d225c5e97882"
+      },
+      "links": {
+        "self": "/sharings/ce8835a061d0ef68947afe69a0046722"
+      },
+      "relationships": {
+        "shared_docs": {
+          "data": [
+            {
+              "id": "612acf1c-1d72-11e8-b043-ef239d3074dd",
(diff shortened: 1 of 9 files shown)


pkg/consts/views.go
```diff
@@ -165,6 +165,26 @@ function(doc) {
 }`,
 }
 
+// SharedDocsByDocType is the view for fetching a list of sharingid
+// associated with a docType. The reduce function is to avoid duplicates
+var SharedDocsByDocType = &couchdb.View{
```
nono: I don't think you need a view. You can query _all_docs with `start_key: "io.cozy.files/", end_key: "io.cozy.files/\uFFFF", include_docs: true`

web/sharings/sharings.go
```diff
@@ -111,6 +111,37 @@ func GetSharing(c echo.Context) error {
...
+		if err != nil {
+			return wrapErrors(err)
+		}
+		as := &sharing.APISharing{
+			Sharing:     s,
+			Credentials: nil,
+			SharedDocs:  docs,
+		}
+		sharings[i] = as
+	}
```
nono: We should avoid to put the same sharings several times in the response, and to query the sharings one by one. In go, we should use a map to deduplicate the sharing ids (`var seen map[string]interface{}`) and load all the sharings at once with a query on _all_docs with `keys: sharingsIDs`.

pkg/couchdb/encoding.go
```diff
@@ -49,3 +49,29 @@ func (vr *ViewRequest) Values() (url.Values, error) {
 
 	return v, nil
 }
+
+// Values transforms a AllDocsRequest into a query string suitable for couchdb
+// ie, where non-scalar fields have been JSON+URL encoded.
+func (adr *AllDocsRequest) Values() (url.Values, error) {
```
nono: I don't understand why you have to introduce this method. We already use requests with startkey and endkey at several places in the stack, and I don't see what is different in your case.

pkg/sharing/shared.go
```diff
@@ -175,4 +175,31 @@ func GetSharedDocsBySharingID(inst *instance.Instance, sharingID string) ([]couc
...
+	var docs []*SharedRef
+	err := couchdb.GetAllDocs(inst, consts.Shared, req, &docs)
+	if err != nil {
+		return nil, err
+	}
+	// Use a map to easily avoid duplicates
+	keys := make(map[string]bool)
+	var sharingIDs []string
+	for _, doc := range docs {
+		for id := range doc.Infos {
```
nono: You don't look at the `removed` flag of Infos, so the files/documents that have been removed/deleted/taken out of the sharing will be included. Is it what you want to do? The description of the endpoint is not really clear on that.

I'd have not done like that, but both points of view can be defended.

docs/sharing.md
```diff
@@ -310,6 +310,141 @@ Content-Type: application/vnd.api+json
...
+#### Response
+
+```http
+HTTP/1.1 200 OK
+Content-Type: application/vnd.api+json
+```
+
+```json
+{
+  "data": null,
```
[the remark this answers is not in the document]
Gara64: It seemed to me that the data had to refer to a single document content, but apparently not. I'll put the sharings into data as an array then.

pkg/couchdb/encoding.go
Gara64: From what I saw, the startkey and endkey are used for regular views but never for _all_docs, which use a specific `AllDocsRequest` struct https://github.com/cozy/cozy-stack/blob/master/pkg/couchdb/bulk.go#L15
Therefore I needed to redefine the Values() method for this kind of request.

pkg/sharing/shared.go
Gara64: From my perspective, even a deleted document could have the legitimity to be included as it has been shared at some point. 
That being said, this could be confusing for the querier, so we might filter out the removed documents here and add an option `include_removed=true` in the future, if needed.

web/sharings/sharings.go
```diff
@@ -111,6 +111,39 @@ func GetSharing(c echo.Context) error {
...
+
+	sharings, err := sharing.FindSharings(inst, sharingIDs)
+	if err != nil {
+		return wrapErrors(err)
+	}
+	for i, s := range sharings {
+		if err = checkGetPermissions(c, s); err != nil {
+			return wrapErrors(err)
+		}
+		docs, err := sharing.GetSharedDocsBySharingID(inst, s.ID())
```
nono: It doesn't look optimal. Maybe you can modify `GetSharingsByDocType` to build a map of `sharingID -> []SharedDoc`, and just use the map here to fill the SharedDocs field of the APISharing

PR #1338, opened 2018-04-16
Allowing definition of default-src CSP in configuration file

jinroh:


cmd/serve.go
```diff
@@ -209,8 +210,13 @@ func init() {
 	flags.String("password-reset-interval", "15m", "minimal duration between two password reset")
 	checkNoErr(viper.BindPFlag("password_reset_interval", flags.Lookup("password-reset-interval")))
 
+	flags.BoolVar(&flagAllowRoot, "allow-root", false, "Allow to start as root (disabled by default)")
```
nono: Does it work with `viper.BindPFlag("...", flags.Lookup("..."))`?

PR #1349, opened 2018-04-24
Upload new files for a sharing

nono:


pkg/sharing/replicator.go
```diff
@@ -237,6 +237,9 @@ func (s *Sharing) callChangesFeed(inst *instance.Instance, since string) (*Chang
 		if !ok {
 			continue
 		}
+		if _, ok = info["binary"]; ok {
+			continue
```
Gara64: Shouldn't it be `!ok` ?

pkg/workers/share/share.go
```diff
@@ -23,20 +32,42 @@ func init() {
...
-		MaxExecCount: 2,
-		Timeout:      30 * time.Second,
-		WorkerFunc:   WorkerTrack,
+		WorkerType:  "share-upload",
+		Concurrency: runtime.NumCPU(),
+		// TODO the worker is not idempotent: if it fails, it adds a new job to
+		// retry, but with MaxExecCount > 1, it can amplifies a lot the number
+		// of retries
+		MaxExecCount: 1,
+		Timeout:      1 * time.Hour,
```
Gara64: A large video might fail with a slow connection, but do we support this kind of case ?

pkg/sharing/setup.go
```diff
@@ -84,7 +84,13 @@ func (s *Sharing) Setup(inst *instance.Instance, m *Member) {
 	if err := s.ReplicateTo(inst, m, true); err != nil {
 		inst.Logger().WithField("nspace", "sharing").
 			Warnf("Error on initial replication (%s): %s", s.SID, err)
-		s.retryReplicate(inst, 0)
+		s.retryWorker(inst, "share-replicate", 0)
+	} else {
+		if err := s.InitialUpload(inst, m); err != nil {
```
Gara64: We should check the sharing involves files here

pkg/sharing/upload.go
```diff
@@ -95,11 +98,63 @@ func (s *Sharing) UploadTo(inst *instance.Instance, m *Member) (bool, error) {
...
 	}
 
 	return true, s.UpdateLastSequenceNumber(inst, m, "upload", seq)
 }
+
+// findNextFileToUpload uses the changes feed to find the next file that needs
+// to be uploaded. It returns a file document if there is one file to upload,
+// and the sequence number where it is in the changes feed.
+func (s *Sharing) findNextFileToUpload(inst *instance.Instance, since string) (*vfs.FileDoc, string, error) {
+	for {
```
Gara64: Isn't a bit dangerous? Maybe a MAX_CHANGES condition could be added

docs/sharing.md
```diff
@@ -561,4 +561,61 @@ Content-Type: application/json
...
+PUT /sharings/ce8835a061d0ef68947afe69a0046722/io.cozy.files/0c1116b028c6ae6f5cdafb949c088265/metadata HTTP/1.1
+Host: bob.example.net
+Accept: application/json
+Content-Type: application/json
+Authorization: Bearer ...
+```
+
+#### Response
+
+If only the metadata has changed (not the content), the response will be a
```
Gara64: If only -> Only if :)

pkg/sharing/upload.go
```diff
@@ -305,3 +309,32 @@ func (s *Sharing) uploadFile(inst *instance.Instance, m *Member, file *couchdb.J
...
+// HandleFileUpload is used to receive a file upload when synchronizing just
+// the metadata was not enough.
+func (s *Sharing) HandleFileUpload(inst *instance.Instance, key string, body io.ReadCloser) error {
+	newdoc := getCache().Get(inst.Domain, key)
+	if newdoc == nil {
+		return ErrMissingFileMetadata
+	}
+
+	indexer := NewSharingIndexer(inst, nil) // TODO
+	fs := inst.VFS().UseSharingIndexer(indexer)
```
Gara64: Why do we need a special indexer for the sharings?

pkg/sharing/replicator.go
nono: No, I really meant `ok`. We are here in the replicator (the first phase of synchronization of a sharing), and binaries are managed in the second phase ("the upload"), so we want to skip them here.

pkg/workers/share/share.go
nono: We will support this kind of case, but latter. For the moment, the limit of 1 hour seems to be a good tradeoff.

pkg/sharing/upload.go
nono: Well, maybe we can

docs/sharing.md
nono: What I wanted to say is that if the metadata has changed, the response will be a 204, but if the metadata and the content (or just the content) have changed, the response will be a 200. In short: if "only the metadata has changed", then 204.

pkg/sharing/upload.go
```diff
@@ -166,6 +167,43 @@ func (s *Sharing) findNextFileToUpload(inst *instance.Instance, since string) (*
...
+type KeyToUpload struct {
+	Key string `json:"key"`
+}
+
+func (s *Sharing) createUploadKey(inst *instance.Instance, target *vfs.FileDoc) (*KeyToUpload, error) {
+	key := getCache().Save(inst.Domain, target)
+	return &KeyToUpload{Key: key}, nil
+}
+
+// SyncFile tries to synchroniza a file with just the metadata. If it can't,
```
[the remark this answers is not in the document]
nono: Good catch!

pkg/sharing/upload.go
```diff
@@ -245,12 +283,11 @@ func (s *Sharing) uploadFile(inst *instance.Instance, m *Member, file *couchdb.J
...
 	defer content.Close()
 
-	// TODO send resBody["token"]
 	res2, err := request.Req(&request.Options{
 		Method: http.MethodPut,
 		Scheme: u.Scheme,
 		Domain: u.Host,
-		Path:   "/sharings/" + s.SID + "/io.cozy.files/" + file.ID(),
+		Path:   "/sharings/" + s.SID + "/io.cozy.files/" + resBody.Key,
 		Headers: request.Headers{
```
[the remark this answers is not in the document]
nono: I use the status code to know if we need to upload the binary or not.

PR #1351, opened 2018-04-25
Move and rename files/folders

nono:


pkg/sharing/files.go
```diff
@@ -349,15 +350,44 @@ func (s *Sharing) UpdateDir(inst *instance.Instance, target map[string]interface
...
+			parent, err := fs.DirByID(dirID)
+			// TODO better handling of this conflict
+			if err != nil {
+				inst.Logger().WithField("nspace", "replicator").
+					Debugf("Conflict for parent on updating dir: %s", err)
+				return err
+			}
+			dir.DirID = parent.DocID
+		}
+	} else {
```
Gara64: In which case the `dir_id` wouldn't exist?

nono: When the parent directory is not the same on both instances. For example, it can be files shared by references (the files are shared, but not the directory in which they are).

PR #1354, opened 2018-04-26
Allow to overwrite a file in a sharing

nono:


tests/sharing/lib/rule.rb
```diff
@@ -25,7 +25,8 @@ def self.create_from_obj(obj, what)
              values: values,
              add: what,
              update: what,
-             remove: what
+             remove: what,
+             object: obj
```
Gara64: What is the purpose of `object`?

nono: When playing in `console.rb`, it's sometimes practical to inspect the folder or album associated to a rule.

PR #1357, opened 2018-04-27
Settings/warnings: dont panic if empty

aenario:


web/jsonapi/jsonapi.go
```diff
@@ -152,9 +152,13 @@ func DataErrorList(c echo.Context, errs ...*Error) error {
 	doc := Document{
 		Errors: errs,
 	}
+	status := http.StatusNotFound
+	if len(errs) > 0 {
+		status = errs[0].Status
+	}
```
jinroh: Could we move this logic into the `settings/warning` handler ? I think this method is good as it is, and should not expect an empty error list.

PR #1359, opened 2018-04-27
Improve VFS Afero when uploading a new version of a file

nono:


pkg/vfs/vfsafero/impl.go
```diff
@@ -748,22 +738,21 @@ func (f *aferoFileCreation) Close() (err error) {
 	defer func() {
 		if err == nil {
 			if f.olddoc != nil {
-				// remove the backup if no error occured
-				f.afs.fs.Remove(f.bakpath) // #nosec
+				// move the temporary file to its final location
+				f.afs.fs.Rename(f.tmppath, f.newpath) // #nosec
```
jinroh: Looks to me like this `Rename` should happened whether or not the `f.olddoc` is nil ?

nono: If `f.olddoc` is nil, `tmppath` and `newpath` are the same (ie for a new file, we don't bother to use a tempory file that we rename at the end).

PR #1363, opened 2018-04-30
Detect when a file/folder is trashed or removed from a sharing

nono:

PR #1368, opened 2018-05-02
Fix missing rev for a received dir 

Gara64:


nono:
:+1: Good job on this debug session!

I'm suggesting to fix it in the indexer, like this: https://github.com/cozy/cozy-stack/pull/1363/commits/dd1ce70404c14082cd115be9569947bff21d2f62. Currently, it does the same thing but I feel this code should be in the indexer, and it may be easier to handle conflicts later with this.

Is it OK for you?

Gara64:
Sure, I actually did like this at first, but it felt more logical to put it where the other fields are set. No problem to use your commit, I'll close this PR then!

PR #1369, opened 2018-05-03
Export worker for exporting the content of a cozy

jinroh:
This PR adds a worker responsible for making a snapshot of the metadata content of the user into a tar file, with text-format export of the user's content.

Access control is specific and requires and MAC sent to the user via e-mail to access to the data. 

Archives have an expiry date.

The binary data of the users (io.cozy.files) are handled specifically. The index is exported in a tree-like datatype, but the actual files are not stored into the archive and are multi-part streamed to the user as requested.

WIP: documentation


pkg/consts/views.go
```diff
@@ -13,6 +13,8 @@ const IndexViewsVersion int = 16
 // properly.
 var GlobalIndexes = []*mango.Index{
 	mango.IndexOnFields(Instances, "by-domain", []string{"domain"}),
+
+	mango.IndexOnFields(Exports, "by-domain", []string{"domain", "created_at"}),
```
nono: Don't forget to increase `IndexViewsVersion`

pkg/vfs/directory.go
```diff
@@ -29,10 +29,10 @@ type DirDoc struct {
 
 	CreatedAt time.Time `json:"created_at"`
 	UpdatedAt time.Time `json:"updated_at"`
-	Tags      []string  `json:"tags"`
+	Tags      []string  `json:"tags,omitempty"`
 
 	// Directory path on VFS
-	Fullpath string `json:"path"`
+	Fullpath string `json:"path,omitempty"`
```
nono: In which case we can have the Fullpath empty?

pkg/workers/move/export.go
```diff
@@ -0,0 +1,393 @@
...
+		err = errc
+	}
+	size += n
+
+	return
+}
+
+// BucketSize is the default size of a file bucket, to split the index into
+// equal-sized parts.
+const BucketSize = 35 * 1024 * 1024 // 50 MB
```
nono: Is it 35 or 50 Mo?

pkg/workers/move/export.go
```diff
@@ -0,0 +1,393 @@
...
+		if len(withDoctypes) > 0 && !utils.IsInArray(doctype, withDoctypes) {
+			continue
+		}
+		switch doctype {
+		case consts.Jobs, consts.KonnectorLogs,
+			consts.Archives,
+			consts.OAuthClients, consts.OAuthAccessCodes:
+			// ignore these doctypes
+		case consts.Sharings, consts.SharingsAnswer:
+			// ignore sharings ? TBD
```
nono: Yes, we should ignore sharings for the moment. And we should add `consts.Shared` to this list too.

pkg/workers/move/export.go
```diff
@@ -0,0 +1,393 @@
...
+	doctypes, err := couchdb.AllDoctypes(in)
+	if err != nil {
+		return
+	}
+	for _, doctype := range doctypes {
+		if len(withDoctypes) > 0 && !utils.IsInArray(doctype, withDoctypes) {
+			continue
+		}
+		switch doctype {
+		case consts.Jobs, consts.KonnectorLogs,
```
nono: Why do we ignore jobs? And what about sessions?

pkg/workers/updates/updates.go
```diff
@@ -1,110 +1,70 @@
...
+	"github.com/cozy/cozy-stack/pkg/instance"
+	"github.com/cozy/cozy-stack/pkg/jobs"
 	"github.com/cozy/cozy-stack/pkg/registry"
-	"github.com/cozy/cozy-stack/pkg/utils"
 	multierror "github.com/hashicorp/go-multierror"
-	"github.com/robfig/cron"
 )
 
-const numUpdaters = 50
+const numUpdaters = 25
```
nono: Not for this PR, but shouldn't the number of updaters be configurable?

pkg/workers/move/export.go
jinroh: Thanks for the note, i'll fix it. I'll make it a configurable value I think though.

pkg/workers/updates/updates.go
jinroh: Yes, and I'll remove this diff.

pkg/vfs/directory.go
jinroh: `DirDoc` is used as the base of `DirOrFileDoc`, this is the reason why I added an omitempty. I'll add a comment to make it clear.

pkg/consts/views.go
jinroh: Done 62a7d65bc013d0b8c9d068830b5ae8ed66c6057f.

PR #1371, opened 2018-05-03
Put removed files of a sharing in the trash

nono:


pkg/sharing/replicator.go
```diff
@@ -208,9 +209,17 @@ func (s *Sharing) replicationID(m *Member) (string, error) {
...
 }
 
-// Changes is a map of "doctype/docid" -> [revisions]
-// It's the format for the request body of our _revs_diff
-type Changes map[string][]string
+// Changed is a map of "doctype/docid" -> [revisions]
+type Changed map[string][]string
+
+// Removed is a set of "doctype/docid"
+type Removed map[string]struct{}
```
Gara64: Why a struct and not a bool?

pkg/sharing/files.go
```diff
@@ -257,15 +257,30 @@ func (s *Sharing) ApplyBulkFiles(inst *instance.Instance, docs DocsList) error {
...
-		if ref == nil && doc == nil {
+		if _, ok := target["_deleted"]; ok {
+			if ref == nil || (dir == nil && file == nil) {
+				continue
+			}
+			if dir != nil {
+				err = s.TrashDir(inst, dir)
+			} else {
+				err = s.TrashFile(inst, file)
+			}
```
Gara64: When a directory containing files is deleted, the dir + files are sent in bulk. If a file arrives before the dir here, it's going to be trashed with a ` dir_id: io.cozy.files.trash-dir`, causing unconsistency in the fs.

pkg/sharing/replicator.go
nono: It's an optimization. A bool takes one byte, but an empty struct is empty and takes 0 bytes.

pkg/sharing/files.go
nono: When dir+files are sent in the same bulk, they are sorted so that a file will come after its parent directory. If they are sent in two bulks, the file can trashed before its parent: both will appear in the trash: it looks the other user has first trashed the file and then trashed the directory. It's a tradeoff: we consider this situation as OK (not optimal).

PR #1376, opened 2018-05-09
Allow usage of HTTPs for admin client

jinroh:
Choosing the root CA and client certificate as an option.

New env variables are:
  - `COZY_ADMIN_HTTPS_CLIENT` (set to `1` or `true` to activate https)
  - `COZY_ADMIN_HTTPS_CLIENT_ROOTCA_FILE`
  - `COZY_ADMIN_HTTPS_CLIENT_CERT_FILE`
  - `COZY_ADMIN_HTTPS_CLIENT_KEY_FILE`
  - `COZY_ADMIN_HTTPS_CLIENT_PINNED_KEY`

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -1,9 +1,15 @@
 package cmd
 
 import (
+	"crypto/tls"
+	"crypto/x509"
 	"errors"
 	"fmt"
+	"io/ioutil"
+	"net/http"
 	"os"
+	"strconv"
+	"time"
 
 	"github.com/cozy/cozy-stack/client"
 	"github.com/cozy/cozy-stack/client/request"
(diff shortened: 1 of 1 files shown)


cmd/root.go
```diff
@@ -76,23 +86,112 @@ func newClient(domain string, scopes ...string) *client.Client {
...
+		InsecureSkipVerify:    false, // should be false, we *need* rootca verification
+	}
+	return &http.Client{
+		Timeout:   15 * time.Second,
+		Transport: &http.Transport{TLSClientConfig: tlsConfig},
+	}
+}
+
+func sslVerifyPinnedKey(pinnedFingerPrint []byte) func(rawCerts [][]byte, _ [][]*x509.Certificate) error {
+	if len(pinnedFingerPrint) != sha256.Size {
```
aeris: Already verified L153

jinroh: This is just me being extra-careful if we use this method in the future without this verification...

PR #1379, opened 2018-05-11
Revocation of a sharing

Gara64:
First PR for the revocation ; the target route /sharings/:sharing-id will be added for a next PR and include some tests

### DIFF
--- a/pkg/sharing/member.go
+++ b/pkg/sharing/member.go
@@ -1,9 +1,11 @@
 package sharing
 
 import (
+	"net/http"
 	"net/url"
 
 	"github.com/cozy/cozy-stack/client/auth"
+	"github.com/cozy/cozy-stack/client/request"
 	"github.com/cozy/cozy-stack/pkg/contacts"
 	"github.com/cozy/cozy-stack/pkg/couchdb"
 	"github.com/cozy/cozy-stack/pkg/crypto"
@@ -22,6 +24,8 @@ const (
 	MemberStatusPendingInvitation = "pending"
 	// MemberStatusReady is for recipient that have accepted the sharing
 	MemberStatusReady = "ready"
+	// MemberStatusRevoked is for a revoked member
+	MemberStatusRevoked = "revoked"
 )
 
 // Member contains the information about a recipient (or the sharer) for a sharing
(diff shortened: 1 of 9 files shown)


pkg/sharing/member.go
```diff
@@ -148,3 +155,47 @@ func (c *Credentials) Refresh(inst *instance.Instance, s *Sharing, m *Member) er
...
+	}
+
+	m.Status = MemberStatusRevoked
+
+	// Do not remove the credential to preserve the members / credentials order
+	c.State = ""
+	c.XorKey = nil
+	c.Client = nil
+	c.AccessToken = nil
+	c.LocalClientID = ""
```
nono: `*c = Credentials{}` does not work?

pkg/sharing/shared.go
```diff
@@ -239,6 +244,39 @@ func UpdateShared(inst *instance.Instance, msg TrackMessage, evt TrackEvent) err
 	return couchdb.UpdateDoc(inst, &ref)
 }
 
+// RemoveReferences deletes the references containing the sharingid
```
nono: s/RemoveReferences/RemoveSharedRefs/

[nitpick] The "references"  are the referenced_by of the files in my mind, I'd prefer to avoid this term here

pkg/sharing/sharing.go
```diff
@@ -184,6 +202,77 @@ func (s *Sharing) CreateRequest(inst *instance.Instance) error {
...
 
+// Revoke remove the credentials for all members, contact them, removes the
+// triggers and set the active flag to false.
+func (s *Sharing) Revoke(inst *instance.Instance) error {
+	var errm error
+
+	if !s.Owner {
+		return ErrInvalidSharing
+	}
+	for i := range s.Members {
```
nono: I'd have done this (no `if`):

```go
for i := range s.Credentials {
    if err := s.RevokeMember(inst, &s.Members[i+1], &s.Credentials[i]); err != nil {
        multierror.Append(errm, err)
    }
}
```

pkg/sharing/sharing.go
```diff
@@ -184,6 +202,77 @@ func (s *Sharing) CreateRequest(inst *instance.Instance) error {
...
+		return err
+	}
+	return errm
+}
+
+// RemoveTriggers remove all the triggers associated to this sharing
+func (s *Sharing) RemoveTriggers(inst *instance.Instance) error {
+	var err error
+	sched := jobs.System()
+	// TODO: a view on WorkerType would be more efficient
```
nono: A third way would be to put the trigger IDs on the `Triggers` struct of `Sharing`. Currently, we have bool but it should be easy to modify it to have the IDs there.

pkg/sharing/sharing.go
```diff
@@ -6,14 +6,21 @@ import (
...
 	"github.com/cozy/cozy-stack/pkg/permissions"
+	multierror "github.com/hashicorp/go-multierror"
 )
 
 const (
 	// StateLen is the number of bytes for the OAuth state parameter
 	StateLen = 16
 )
 
+// Msg is an interface for the sharing worker messages
```
nono: This interface looks weird (too generic?)

nono: I was thinking of the methods: with just one `ID` method, you can have a lots of structs that implement this interface even if they don't have anything to do with a message for a sharing worker

Gara64: Ok I see. The last commit removes this interface anyways, as it is no longer necessary by using the trigger ids in the sharing struct

PR #1380, opened 2018-05-14
Manage files/folders conflicts that can be reconcialiated

nono:


tests/sharing/tests/multiple_sharings.rb
```diff
@@ -2,7 +2,7 @@
 
 require_relative '../boot'
 require 'minitest/autorun'
-require 'pry-rescue/minitest'
+require 'pry-rescue/minitest' unless ENV['CI']
```
Gara64: What is it for?

nono: `pry-rescue/minitest` starts a pry session (the console) when an assertion fails. It's handy for debugging. But it blocks travis when the assertion fails on travis, so we require it only when we are not on a CI environment (aka travis)

pkg/sharing/indexer.go
```diff
@@ -21,17 +23,35 @@ type sharingIndexer struct {
...
 }
 
+func (s *sharingIndexer) IncrementRevision() {
+	var start int64
+	switch s := s.bulkRevs.Revisions["start"].(type) {
+	case float64:
+		start = int64(s)
+	case int64:
+		start = s
+	default:
```
Gara64: Why could it be float?

docs/sharing-design.md
```diff
@@ -200,11 +200,10 @@ are:
   `application/octet-steam`)
 * `tags` and `referenced_by`: a union of the two versions is made
 
-For 3., we rename the file or folder with the later `updated_at`.
+For 3., we rename the file or folder with the smaller `id`.
```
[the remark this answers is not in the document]
nono: I think that in a case of a sharing with more than 2 people, we can have a case with several conflicts on the same file or folder can be lead to a situation where we don't converge (no eventual consistency). It's hard to describe here, but it looks safer to use the `id` to resolve the conflict in the same way on all the members.

pkg/sharing/files.go
```diff
@@ -396,6 +400,93 @@ func resolveConflictSamePath(inst *instance.Instance, id, pth string) (string, e
...
+		}
+	}
+	return nil, ErrFolderNotFound
+}
+
+// recreateParent is used when a file or folder is added by a cozy, and sent to
+// this instance, but its parent directory was trashed and deleted on this
+// cozy. To resolve the conflict, this instance will fetch informations from
+// the other instance about the parent directory and will recreate it. It can
+// be necessery to recurse if there were several levels of directories deleted.
```
Gara64: necessery -> necessary

PR #1381, opened 2018-05-15
Add route to inform a recipient about a revocation

Gara64:


pkg/sharing/oauth.go
```diff
@@ -326,3 +328,23 @@ func (s *Sharing) ProcessAnswer(inst *instance.Instance, creds *Credentials) (*A
...
 	return nil, ErrMemberNotFound
 }
+
+// RefreshToken is used after a failed request with a 4xx error code.
+// It renews the access token and retries the request
+func RefreshToken(inst *instance.Instance, s *Sharing, m *Member, creds *Credentials, opts *request.Options) (*http.Response, error) {
+	if err := creds.Refresh(inst, s, m); err != nil {
+		return nil, err
+	}
+	res, err := request.Req(opts)
```
nono: You have to update the `Authorization` header of `opts` with the new token

pkg/sharing/replicator.go
```diff
@@ -326,38 +326,19 @@ func (s *Sharing) callRevsDiff(inst *instance.Instance, m *Member, creds *Creden
...
-			Domain: u.Host,
-			Path:   "/sharings/" + s.SID + "/_revs_diff",
-			Headers: request.Headers{
-				"Accept":        "application/json",
-				"Content-Type":  "application/json",
-				"Authorization": "Bearer " + creds.AccessToken.AccessToken,
-			},
-			Body: bytes.NewReader(body),
-		})
+		res, err = RefreshToken(inst, s, m, creds, opts)
```
nono: It won't work: the body is a stream, and it is consumed on the first request. When it is retried in `RefreshToken`, the body will be empty.

pkg/sharing/oauth.go
```diff
@@ -326,3 +328,23 @@ func (s *Sharing) ProcessAnswer(inst *instance.Instance, creds *Credentials) (*A
...
+	if err != nil {
+		return nil, err
+	}
+	defer res.Body.Close()
```
nono: By closing `res.Body` in `RefreshToken`, it won't be possible to parse the response from the calling function.

pkg/sharing/member.go
```diff
@@ -165,22 +165,29 @@ func (s *Sharing) RevokeMember(inst *instance.Instance, m *Member, c *Credential
...
-		})
+		}
+		res, err := request.Req(opts)
 		if err != nil {
 			return err
 		}
 		if res.StatusCode/100 != 2 {
 			return ErrRequestFailed
 		}
+		if res.StatusCode/100 == 4 {
```
nono: You have to put this block before the `if res.StatusCode/100 != 2 {`

pkg/sharing/replicator.go
```diff
@@ -326,38 +326,19 @@ func (s *Sharing) callRevsDiff(inst *instance.Instance, m *Member, creds *Creden
...
 		if err != nil {
 			return nil, err
 		}
 	}
-	defer res.Body.Close()
-	if res.StatusCode/100 == 5 {
```
nono: We should keep this block: if the first request returns a 5xx, we won't try to refresh token and we should still return this error

pkg/sharing/upload.go
```diff
@@ -210,38 +209,19 @@ func (s *Sharing) uploadFile(inst *instance.Instance, m *Member, file map[string
...
-			Method: http.MethodPut,
-			Scheme: u.Scheme,
-			Domain: u.Host,
-			Path:   "/sharings/" + s.SID + "/io.cozy.files/" + xoredFileID + "/metadata",
-			Headers: request.Headers{
-				"Accept":        "application/json",
-				"Content-Type":  "application/json",
-				"Authorization": "Bearer " + creds.AccessToken.AccessToken,
-			},
-			Body: bytes.NewReader(body),
```
nono: Idem about body

pkg/sharing/upload.go
```diff
@@ -210,38 +209,19 @@ func (s *Sharing) uploadFile(inst *instance.Instance, m *Member, file map[string
...
-		})
+		res, err = RefreshToken(inst, s, m, creds, opts)
 		if err != nil {
 			return err
 		}
 	}
-	defer res.Body.Close()
-	if res.StatusCode/100 == 5 {
```
nono: Idem

pkg/sharing/sharing.go
```diff
@@ -252,6 +251,31 @@ func removeSharingTrigger(inst *instance.Instance, triggerID string) error {
...
+// RevokeByNotification is called on the recipient side, after a revocation
+// performed by the sharer
+func (s *Sharing) RevokeByNotification(inst *instance.Instance) error {
+	if s.Owner {
+		return ErrInvalidSharing
+	}
+	if err := DeleteOAuthClient(inst, &s.Members[0], &s.Credentials[0]); err != nil {
+		return err
+	}
+	if !s.ReadOnly() {
```
nono: I think you can skip the `if !s.ReadOnly()` condition, and just rely on the sharing trigger IDs to know if there are some triggers to remove (or not).

nono: 3 checks that are doing no I/O or memory allocation are really fast, probably even faster than calling `ReadOnly()`. But we don't care, it's a micro-optimization on a cold path, readability and solidity are way more important than speed here.

pkg/sharing/member.go
```diff
@@ -165,21 +165,28 @@ func (s *Sharing) RevokeMember(inst *instance.Instance, m *Member, c *Credential
...
+				"Authorization": "Bearer " + c.AccessToken.AccessToken,
 			},
-		})
+		}
+		res, err := request.Req(opts)
 		if err != nil {
 			return err
 		}
-		if res.StatusCode/100 != 2 {
-			return ErrRequestFailed
```
nono: You should close `res.Body`

pkg/sharing/member.go
```diff
@@ -165,21 +165,28 @@ func (s *Sharing) RevokeMember(inst *instance.Instance, m *Member, c *Credential
...
-		res, err := request.Req(&request.Options{
+		opts := &request.Options{
 			Method: http.MethodDelete,
 			Scheme: u.Scheme,
 			Domain: u.Host,
 			Path:   "/sharings/" + s.SID,
 			Headers: request.Headers{
-				"Accept":       "application/vnd.api+json",
-				"Content-Type": "application/vnd.api+json",
+				"Accept":        "application/vnd.api+json",
```
nono: It looks like you are not using the response, so it's probably useless to put an `Accept` header

pkg/sharing/member.go
```diff
@@ -165,21 +165,28 @@ func (s *Sharing) RevokeMember(inst *instance.Instance, m *Member, c *Credential
...
+				"Content-Type":  "application/vnd.api+json",
```
nono: You have no body for the request, so you should remove this `Content-Type`

pkg/sharing/member.go
```diff
@@ -165,21 +165,28 @@ func (s *Sharing) RevokeMember(inst *instance.Instance, m *Member, c *Credential
...
 		if err != nil {
 			return err
 		}
-		if res.StatusCode/100 != 2 {
-			return ErrRequestFailed
+		if res.StatusCode/100 == 5 {
+			return ErrInternalServerError
+		}
+		if res.StatusCode/100 == 4 {
+			if _, err := RefreshToken(inst, s, m, c, opts, nil); err != nil {
```
nono: You should assign the response from `RefreshToken` and close its `Body`

pkg/sharing/replicator.go
```diff
@@ -326,38 +326,24 @@ func (s *Sharing) callRevsDiff(inst *instance.Instance, m *Member, creds *Creden
...
 		},
 		Body: bytes.NewReader(body),
-	})
+	}
+	var res *http.Response
+	res, err = request.Req(opts)
 	if err != nil {
 		return nil, err
 	}
+	if res.StatusCode/100 == 5 {
```
nono: You have to close `res.Body` in this block

pkg/sharing/replicator.go
```diff
@@ -326,38 +326,24 @@ func (s *Sharing) callRevsDiff(inst *instance.Instance, m *Member, creds *Creden
...
-			Domain: u.Host,
-			Path:   "/sharings/" + s.SID + "/_revs_diff",
-			Headers: request.Headers{
-				"Accept":        "application/json",
-				"Content-Type":  "application/json",
-				"Authorization": "Bearer " + creds.AccessToken.AccessToken,
-			},
-			Body: bytes.NewReader(body),
-		})
+		res, err = RefreshToken(inst, s, m, creds, opts, body)
```
nono: I think that we should close `res.Body` if err is `ErrInternalServerError` or `ErrClientError`

pkg/sharing/replicator.go
```diff
@@ -553,38 +539,22 @@ func (s *Sharing) sendBulkDocs(inst *instance.Instance, m *Member, creds *Creden
...
 		},
 		Body: bytes.NewReader(body),
-	})
+	}
+	res, err := request.Req(opts)
 	if err != nil {
 		return err
 	}
+	if res.StatusCode/100 == 5 {
+		return ErrInternalServerError
```
nono: We have to close `res.Body` in this block

pkg/sharing/replicator.go
```diff
@@ -553,38 +539,22 @@ func (s *Sharing) sendBulkDocs(inst *instance.Instance, m *Member, creds *Creden
...
-			Path:   "/sharings/" + s.SID + "/_bulk_docs",
-			Headers: request.Headers{
-				"Accept":        "application/json",
-				"Content-Type":  "application/json",
-				"Authorization": "Bearer " + creds.AccessToken.AccessToken,
-			},
-			Body: bytes.NewReader(body),
-		})
-		if err != nil {
+		if res, err = RefreshToken(inst, s, m, creds, opts, body); err != nil {
```
nono: I think that we should close `res.Body` if err is `ErrInternalServerError` or `ErrClientError`

pkg/sharing/upload.go
```diff
@@ -210,42 +209,27 @@ func (s *Sharing) uploadFile(inst *instance.Instance, m *Member, file map[string
...
 		},
 		Body: bytes.NewReader(body),
-	})
+	}
+	var res *http.Response
+	res, err = request.Req(opts)
 	if err != nil {
 		return err
 	}
+	if res.StatusCode/100 == 5 {
```
nono: We have to close `res.Body` in this block

pkg/sharing/upload.go
```diff
@@ -210,42 +209,27 @@ func (s *Sharing) uploadFile(inst *instance.Instance, m *Member, file map[string
...
-			Domain: u.Host,
-			Path:   "/sharings/" + s.SID + "/io.cozy.files/" + xoredFileID + "/metadata",
-			Headers: request.Headers{
-				"Accept":        "application/json",
-				"Content-Type":  "application/json",
-				"Authorization": "Bearer " + creds.AccessToken.AccessToken,
-			},
-			Body: bytes.NewReader(body),
-		})
+		res, err = RefreshToken(inst, s, m, creds, opts, body)
```
nono: I think that we should close `res.Body` if err is `ErrInternalServerError` or `ErrClientError`

web/sharings/sharings_test.go
```diff
@@ -603,19 +603,23 @@ func TestCheckPermissions(t *testing.T) {
 func TestCheckSharingInfoByDocType(t *testing.T) {
 	sharedDocs1 := []string{"fakeid1", "fakeid2", "fakeid3"}
 	sharedDocs2 := []string{"fakeid4", "fakeid5"}
-	s1 := createSharing(aliceInstance, sharedDocs1)
+	s1, err := createSharing(aliceInstance, sharedDocs1)
```
nono: Another way is to pass `t` in the parameter of `createSharing`, so that the `assert.NoError(t, err)` can be done in `createSharing`

pkg/sharing/member.go
```diff
@@ -165,22 +165,30 @@ func (s *Sharing) RevokeMember(inst *instance.Instance, m *Member, c *Credential
...
-		})
+		}
+		var res *http.Response
+		res, err := request.Req(opts)
 		if err != nil {
 			return err
 		}
-		if res.StatusCode/100 != 2 {
-			return ErrRequestFailed
+		if res.StatusCode/100 == 5 {
```
nono: `res.Body.Close()`

pkg/sharing/member.go
```diff
@@ -165,22 +165,30 @@ func (s *Sharing) RevokeMember(inst *instance.Instance, m *Member, c *Credential
...
+			return ErrInternalServerError
 		}
+		if res.StatusCode/100 == 4 {
```
nono: `res.Body.Close()`

PR #1384, opened 2018-05-18
Add a route to revoke to recipient from a sharing

nono:


web/sharings/sharings.go
```diff
@@ -186,6 +187,28 @@ func RevokeSharing(c echo.Context) error {
...
+	s, err := sharing.FindSharing(inst, sharingID)
+	if err != nil {
+		return wrapErrors(err)
+	}
+	_, err = checkCreatePermissions(c, s)
+	if err != nil {
+		return echo.NewHTTPError(http.StatusForbidden)
+	}
+	index, err := strconv.Atoi(c.Param("index"))
+	if err != nil || index == 0 || index >= len(s.Members) {
```
Gara64: If the index starts at 1, it might be useful to say it in the documentation

nono: `index == 0` is the sharer him-self

Gara64: Yes, but it might be not obvious for a dev not familiar with the sharing, that would have to implement the revocation and would just read the doc of this route :)

nono: Yep -> https://github.com/cozy/cozy-stack/pull/1384/commits/9436ba11b37b823d996b03406146c50432cf47fa

PR #1387, opened 2018-05-21
Revoke sharing from recipient

Gara64:


pkg/sharing/member.go
```diff
@@ -177,33 +177,53 @@ func (s *Sharing) RevokeMember(inst *instance.Instance, m *Member, c *Credential
...
+
+	if err := s.NotifyMemberRevocation(inst, m, c); err != nil {
+		inst.Logger().WithField("nspace", "sharing").
+			Warnf("Error on revocation notification: %s", err)
+	}
+	if err := DeleteOAuthClient(inst, m, c); err != nil {
+		return err
+	}
+	m.Status = MemberStatusRevoked
+	s.Credentials = nil
```
nono: I would also set `s.Active` to false. What do you think?

pkg/sharing/oauth.go
```diff
@@ -305,19 +301,28 @@ func (s *Sharing) ProcessAnswer(inst *instance.Instance, creds *Credentials) (*A
...
-				ac.Credentials.AccessToken = token
+			// Create the credentials for the recipient
+			cli, err := CreateOAuthClient(inst, &s.Members[i+1])
+			if err != nil {
+				return &ac, nil
 			}
+			s.Credentials[i].InboundClientID = cli.ClientID
+			ac.Credentials.Client = ConvertOAuthClient(cli)
+			var verb permissions.VerbSet
+			// In case of read-only, The recipient only needs read access on the
```
nono: [nitpick] s/The/the/

pkg/sharing/member.go
Gara64: I actually set it after in `RevokeRecipientBySelf` It follows the same logic than the other revocation functions, eg. revoke the member-related things (notify it, member revoked, crendials reset) and then the sharing-related things (active to false, remove triggers and shared references)

PR #1389, opened 2018-05-22
Manage referenced_by for shared files/folders

nono:


docs/README.md
```diff
@@ -35,6 +35,7 @@
 * `/jobs` - [Jobs](jobs.md)
   * [Konnectors](konnectors.md)
   * [Workers](workers.md)
+- `/move` - [Move, export and import an instance](move.md)
```
Gara64: Should it be part of the PR? :)

docs/toc.yml
```diff
@@ -27,6 +27,7 @@
   - "/jobs Jobs": ./jobs.md
   - "Konnectors": ./konnectors.md
   - "Workers": ./workers.md
+  - "/move - Move, export and import an instance": ./move.md
```
Gara64: Same question

pkg/sharing/rule.go
```diff
@@ -200,6 +201,39 @@ func (s *Sharing) FirstFilesRule() *Rule {
...
+		allFound := true
+		for _, ref := range file.ReferencedBy {
+			v := ref.Type + "/" + ref.ID
+			found := false
+			for _, val := range rule.Values {
+				if val == v {
+					found = true
+					break
+				}
+			}
```
Gara64: You could use `rule.hasReferencedBy(ref)`

docs/README.md
nono: No, it's not related. It was just that I was working on fixing a typo on the sharings docs, and I saw this entry was missing. I added too in a commit about documentation.

pkg/sharing/rule.go
```diff
@@ -200,6 +201,39 @@ func (s *Sharing) FirstFilesRule() *Rule {
...
+			if !found {
+				allFound = false
+				break
+			}
+		}
+		if allFound {
+			return &s.Rules[i]
+		}
```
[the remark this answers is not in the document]
nono: When a cozy send a new file to another cozy, it is expected that it sends only the relevant references (ie the references tied to a sharing rule). In practice, having several relevant references on the same file will be an exception. Here, it's more a sanity check that the received file has no references that the sender has forgot to remove (ie a reference to a not shared document).

PR #1392, opened 2018-05-23
Replace revisions slice by a tree in SharedRef 

nono:


Gara64:
Not an easy one, but looks good ;)

PR #1397, opened 2018-05-24
Add tests with referenced_by

Gara64:


tests/sharing/tests/referenced_by.rb
```diff
@@ -0,0 +1,128 @@
...
+require 'pry-rescue/minitest' unless ENV['CI']
+
+def file_has_album_reference(file, album_id)
+  return false if file.referenced_by.nil?
+  file.referenced_by.each do |ref|
+    if ref["type"] == Album.doctype && ref["id"] == album_id
+      return true
+    end
+  end
+  false
```
nono: This method can be shortened by using `any?` like this:

```ruby
   file.referenced_by.any? do |ref|
     ref["type"] == Album.doctype && ref["id"] == album_id
   end
```

PR #1398, opened 2018-05-25
Clone olddoc after file close

Gara64:


pkg/vfs/vfsafero/impl.go
```diff
@@ -802,7 +802,7 @@ func (f *aferoFileCreation) Close() (err error) {
 		newdoc.Trashed = false
 	}
 	if olddoc == nil {
-		olddoc = newdoc
+		olddoc = newdoc.Clone().(*vfs.FileDoc)
```
nono: I think we should also fix `pkg/vfs/vfsswift`:

- https://github.com/cozy/cozy-stack/blob/master/pkg/vfs/vfsswift/impl_v2.go#L774
- https://github.com/cozy/cozy-stack/blob/master/pkg/vfs/vfsswift/impl_v1.go#L871

PR #1399, opened 2018-05-25
Increase timeout for CLI

nono:


cmd/root.go
```diff
@@ -8,20 +8,21 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
-	"github.com/cozy/cozy-stack/client"
-	"github.com/cozy/cozy-stack/client/request"
-	"github.com/cozy/cozy-stack/pkg/config"
-	"github.com/cozy/cozy-stack/pkg/permissions"
-	"github.com/howeyc/gopass"
-	"github.com/spf13/cobra"
-	"github.com/spf13/viper"
```
nono: @aeris you don't use [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports) ?

PR #1403, opened 2018-05-28
Add conflict resolution for files

nono:


pkg/sharing/files.go
```diff
@@ -648,7 +651,10 @@ func (s *Sharing) CreateDir(inst *instance.Instance, target map[string]interface
 		return err
 	}
 	dir.SetID(target["_id"].(string))
+	ref.SID = consts.Files + "/" + dir.DocID
 	copySafeFieldsToDir(target, dir)
+	_, ruleIndex := s.findRuleForNewDirectory(dir)
```
Gara64: We might want to check if the rule is `ǹil` to make sure the ruleIndex is legit

pkg/sharing/upload.go
```diff
@@ -317,6 +318,10 @@ func (s *Sharing) SyncFile(inst *instance.Instance, target *FileDocWithRevisions
 	if _, ok := ref.Infos[s.SID]; !ok {
 		return nil, ErrSafety
 	}
+	if ref.Revisions.Find(target.DocRev) != nil {
+		// It's just the echo, there is nothing to do
```
Gara64: :)

Gara64:
Well done with this PR! It's sometimes quite complex, so tests are definitely needed :)

pkg/sharing/files.go
nono: Good catch!

PR #1405, opened 2018-05-28
Add generic prefix for database namespaces

jinroh:
Instead of using the domain name of instances as the prefix for our databases (couchdb, redis, vfs, ...), a generic prefix is generated at the instance creation and used as an identifier.

### DIFF
--- a/pkg/couchdb/couchdb.go
+++ b/pkg/couchdb/couchdb.go
@@ -70,10 +70,8 @@ func RTEvent(db Database, verb string, doc, oldDoc Doc) {
 			Verb:   verb,
 			Doc:    doc.Clone(),
 			OldDoc: oldDoc,
-			Domain: db.DomainName(),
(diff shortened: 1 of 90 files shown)


pkg/couchdb/couchdb.go
```diff
@@ -252,14 +252,35 @@ var couchdbClient = &http.Client{
...
 
 func escapeCouchdbName(name string) string {
-	name = strings.Replace(name, ".", "-", -1)
-	name = strings.Replace(name, ":", "-", -1)
-	return strings.ToLower(name)
+	name = strings.ToLower(name)
+	if config.GetConfig().DomainAnyAllowed {
+		return dbUnderscoreEscaper.Replace(name)
+	}
+	return dbDashEscaper.Replace(name)
```
nono: Won't it break compatibility for self-hosted?

PR #1409, opened 2018-05-31
Add scalability tests

Gara64:


tests/sharing/scalability/base_tests.rb
```diff
@@ -0,0 +1,84 @@
+require_relative '../boot'
+
+$stdout.sync = true
```
nono: By curiosity, why do you use this?

tests/sharing/scalability/base_tests.rb
```diff
@@ -0,0 +1,84 @@
...
+  insts
+end
+
+def create_sharing(insts, obj)
+  inst_sharer = insts[0]
+  sharing = Sharing.new
+  sharing.rules << Rule.sync(obj)
+  sharing.members << inst_sharer
+  insts.each_with_index do |inst, i|
+    if i != 0
```
nono: You can write:

```ruby
recipients = insts.drop 1
recipients.each do |inst|
  # ...
end
# ...
recipients.each { |inst| inst.accept sharing }
```

tests/sharing/scalability/base_tests.rb
```diff
@@ -0,0 +1,84 @@
...
+end
+
+def create_files(inst, n_files, dir_id)
+  print "Create #{n_files} files... "
+  files = Array.new(n_files)
+  n_files.times do |i|
+    file_name = "#{Faker::Internet.unique.slug}.txt"
+    files[i] = CozyFile.create inst, dir_id: dir_id, name: file_name
+  end
+  print "Done.\n"
```
nono: In Ruby, it's more common to us `puts` than `print`. Cf https://www.garethrees.co.uk/2013/05/04/p-vs-puts-vs-print-in-ruby/

tests/sharing/lib/folder.rb
```diff
@@ -29,6 +29,49 @@ def self.load_from_url(inst, path)
...
+            file_class: child["class"],
+            metadata: child["metadata"]
+          )
+        end
+        f.couch_id = id
+        f.couch_rev = rev
+        children << f
+      end
+    end
+    children
```
nono: You can simplify a bit the code:

```ruby
def self.children(inst, path)
  # ...
  j = JSON.parse(res.body)["included"]
  (j || []).map do |child|
    id = child["id"]
    rev = child["rev"]
    # ...
    f.couch_id = id
    f.couch_rev = rev
    f
  end
end
```

tests/sharing/scalability/base_tests.rb
Gara64: This is needed if we wanted to progressively show the progression of the tasks, as I originally wanted to. So for now, it is not useful

PR #1415, opened 2018-06-04
Release scripts

aeris:
Automate binaries generation and release publication

### DIFF
--- a/scripts/release.rb
+++ b/scripts/release.rb
@@ -1,0 +1,31 @@
+#!/usr/bin/env ruby
+require 'git'
+require 'octokit'
+require 'filemagic'
+
+REPO = 'cozy/cozy-stack'
+filemagic = FileMagic.new FileMagic::MAGIC_MIME
+root_dir = File.dirname File.dirname File.expand_path __FILE__
+
+git = Git.open root_dir
+tag = git.describe '--tags'
+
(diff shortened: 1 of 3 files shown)


scripts/release.rb
```diff
@@ -0,0 +1,31 @@
...
+git = Git.open root_dir
+tag = git.describe '--tags'
+
+token = ENV.fetch 'GITHUB_RELEASE_TOKEN'
+github = Octokit::Client.new access_token: token
+user = github.user
+user.login
+
+# releases = github.releases REPO
+# releases.each { |r| github.delete_release r.url }
```
nono: Is this commented code for when an upload failed and you want to recreate the release?

scripts/release.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+set -xe
+
+RELEASE="$(git describe --tags)"
+
+go get -u -v ./... || true
```
nono: Why the `|| true`?

scripts/release.rb
aeris: For testing purpose.
Very tedious to have to manually destroy release to be able to rerun the script without failure…
Can be removed

scripts/release.sh
aeris: "go get" sometime returns error return code even when no real error, and because of the "set -xe", fails all the script :cry:

PR #1418, opened 2018-06-05
Add a member to a sharing

nono:


web/sharings/sharings.go
```diff
@@ -175,6 +181,45 @@ func AnswerSharing(c echo.Context) error {
...
 
+// AddRecipient is used to add a member to a sharing
+func AddRecipient(c echo.Context) error {
+	inst := middlewares.GetInstance(c)
+	sharingID := c.Param("sharing-id")
+	s, err := sharing.FindSharing(inst, sharingID)
+	if err != nil {
+		return wrapErrors(err)
+	}
+	if err = checkGetPermissions(c, s); err != nil {
```
Gara64: Shouldn't be `checkCreatePermissions`?

nono: Yes, we don't have explicit permissions for the `io.cozy.sharings` doctype, so it's not very clear what are the good way to check permissions in `web/sharings`, but it looks like `checkCreatePermissions` would be a better fit than `checkGetPermissions` here.

PR #1420, opened 2018-06-06
fix: mail rendering on Gmail client

GoOz:
Gmail was failling its style parsing because of this extra character and therefore disregarding everything else in the `<style>` tag.


GoOz:
[Bouton cassé sur mail d'ajout de device](https://trello.com/c/GXyW6i0b/170-bouton-cass%C3%A9-sur-mail-dajout-de-device)

PR #1423, opened 2018-06-07
Update the members list for the recipients of a sharing

nono:


pkg/sharing/member.go
```diff
@@ -86,6 +88,9 @@ func (s *Sharing) AddContact(inst *instance.Instance, contactID string) error {
 // UpdateRecipients updates the list of recipients
 func (s *Sharing) UpdateRecipients(inst *instance.Instance, members []Member) error {
 	for i, m := range members {
+		if i >= len(s.Members) {
+			s.Members = append(s.Members, Member{})
```
Gara64: Don't you want to append `m` instead of `Member{}`?

pkg/sharing/sharing.go
```diff
@@ -196,6 +197,12 @@ func (s *Sharing) CreateRequest(inst *instance.Instance) error {
 	s.UpdatedAt = time.Now()
 	s.Credentials = make([]Credentials, 1)
 
+	for i, m := range s.Members {
+		if contact, err := contacts.FindByEmail(inst, m.Email); err == nil {
```
Gara64: `err != nil` ?

pkg/sharing/member.go
```diff
@@ -81,6 +81,22 @@ func (s *Sharing) AddContact(inst *instance.Instance, contactID string) error {
 	return nil
 }
 
+// UpdateRecipients updates the list of recipients
+func (s *Sharing) UpdateRecipients(inst *instance.Instance, members []Member) error {
+	for i, m := range members {
+		if m.Email != s.Members[i].Email {
+			if contact, err := contacts.FindByEmail(inst, m.Email); err == nil {
```
Gara64: `err != nil` ?

pkg/sharing/oauth.go
```diff
@@ -146,7 +146,7 @@ func CreateOAuthClient(inst *instance.Instance, m *Member) (*oauth.Client, error
 	}
 	cli := oauth.Client{
 		RedirectURIs: []string{m.Instance + "/sharings/answer"},
-		ClientName:   "Sharing " + m.Name,
+		ClientName:   "Sharing " + m.PublicName,
```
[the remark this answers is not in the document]
nono: In the oauth package, we add a suffix when trying a register a client with an already used client name: https://github.com/cozy/cozy-stack/blob/master/pkg/oauth/client.go#L224-L252

pkg/sharing/sharing.go
nono: No, it's if `err` is nil that the contact is found and valid and that we can use its name. We can safely ignore errors here.

Gara64: My mistake, I should have reviewed this review

PR #1440, opened 2018-06-21
Fix push jobs on a server where workers are disabled

nono:
It fixes a regression introduced by https://github.com/cozy/cozy-stack/pull/1435/commits/b9531f2ef40ca988812cdb3ecb07723e3cb572bd#diff-656325a475171641eede0030442145d6

On web servers, the workers are disabled and when we push a job, we try to find the worker from `b.workers` to execute the PreHook. So, we need to initialize `b.workers` even if the workers are disabled.


gregorylegarec:
[Pb de mise à jour du mot de passe](https://trello.com/c/4LjnlxiG/1073-pb-de-mise-%C3%A0-jour-du-mot-de-passe)

PR #1443, opened 2018-06-25
doc: document some pouchdb-find quirks

y-lohse:


y-lohse:
[(5) [forum-5712] Le tri est lent](https://trello.com/c/UxXUjiar/965-5-forum-5712-le-tri-est-lent)

PR #1453, opened 2018-07-03
Add again a revoked recipient to a sharing

nono:


pkg/sharing/files.go
```diff
@@ -281,6 +281,23 @@ func (s *Sharing) GetSharingDir(inst *instance.Instance) (*vfs.DirDoc, error) {
 	return inst.VFS().DirByID(res.Rows[0].ID)
 }
 
+// RemoveSharingDir removes the reference on the sharing directory.
+// It should be called when a sharing is revoked, on the recipient Cozy.
+func (s *Sharing) RemoveSharingDir(inst *instance.Instance) error {
```
Gara64: The name might be a bit confusing, maybe use something like `RemoveSharingDirRef` or `RemoveFromSharingDir` ?

PR #1458, opened 2018-07-11
Add expire flag for generating tokens from CLI

nono:


CPatchane:
[(2) [COL] Améliorer le style de la modale de configuration](https://trello.com/c/8JGaB46v/1110-2-col-am%C3%A9liorer-le-style-de-la-modale-de-configuration)

PR #1462, opened 2018-07-18
Improve deletion of VFS relying on Swift layout V1

jinroh:
Use the same logic as layout v2 by first marking containers as "to-be-deleted" before deleting on the containers' data.

### DIFF
--- a/pkg/vfs/vfsswift/impl_v1.go
+++ b/pkg/vfs/vfsswift/impl_v1.go
@@ -22,6 +22,7 @@ import (
 )
 
 const swiftV1ContainerPrefix = "cozy-"
+const swiftV1DataContainerPrefix = "data-"
 
 const versionSuffix = "-version"
 const maxFileSize = 5 << (3 * 10) // 5 GiB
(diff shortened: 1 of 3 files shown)


pkg/vfs/vfsswift/impl_v1.go
```diff
@@ -102,20 +105,38 @@ func (sfs *swiftVFS) InitFs() error {
...
+	containerMeta := swift.Metadata{"to-be-deleted": "1"}.ContainerHeaders()
+	sfs.log.Infof("Marking containers %q, %q and %q as to-be-deleted",
+		sfs.container, sfs.version, sfs.dataContainer)
+	err1 := sfs.c.ContainerUpdate(sfs.container, containerMeta)
+	err2 := sfs.c.ContainerUpdate(sfs.dataContainer, containerMeta)
+	err3 := sfs.c.ContainerUpdate(sfs.version, containerMeta)
+	if err1 != nil {
+		sfs.log.Errorf("Could not mark container %q as to-be-deleted: %s",
+			sfs.container, err1)
+		return err1
```
sblaisot: Is there a reason why you removed the exact same `return` in impl_v2 and keep it here ?

jinroh: Mistake. Thanks for spotting it.

PR #1485, opened 2018-08-08
Specific OAuth client for authorization of manager

jinroh:


web/auth/auth.go
```diff
@@ -529,6 +535,33 @@ func authorizeForm(c echo.Context) error {
...
+		u, err := url.ParseRequestURI(params.redirectURI)
+		if err != nil {
+			return c.Render(http.StatusBadRequest, "error.html", echo.Map{
+				"Domain": instance.ContextualDomain(),
+				"Error":  "Error Invalid redirect_uri",
+			})
+		}
+
+		q := u.Query()
+		q.Set("access_code", access.Code)
```
nono: It should be `code` instead of `access_code` (my mistake, see #1488)

web/auth/auth.go
```diff
@@ -529,6 +535,33 @@ func authorizeForm(c echo.Context) error {
...
+		q.Set("state", params.state)
+		q.Set("client_id", params.clientID)
```
nono: `client_id` should not be sent in the redirection. But I can see that I'm doing that also below.

Cf https://tools.ietf.org/html/rfc6749#page-26

PR #1498, opened 2018-08-24
feat: Login errors more accessible

GoOz:
Login errors are now more accessible for screen readers as they are inserted in the DOM at the top of the login form but put back to its end with CSS.

You shouldn't see any difference with before in the UI.

### DIFF
--- a/assets/scripts/submit.js
+++ b/assets/scripts/submit.js
@@ -36,7 +36,7 @@
     if (!errorPanel) {
       errorPanel = document.createElement('div')
       errorPanel.classList.add('errors')
-      loginForm.appendChild(errorPanel)
+      loginForm.insertBefore(errorPanel, loginForm.firstChild);
     }
 
     if (!errorPanel.firstChild) {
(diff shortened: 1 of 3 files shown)


GoOz:
[Améliorer l’accessibilité de la page de connexion](https://trello.com/c/qgMRznqb/1158-am%C3%A9liorer-laccessibilit%C3%A9-de-la-page-de-connexion)

assets/scripts/submit.js
```diff
@@ -36,7 +36,8 @@
     if (!errorPanel) {
       errorPanel = document.createElement('div')
       errorPanel.classList.add('errors')
-      loginForm.appendChild(errorPanel)
+      // loginForm.prepend(errorPanel)
```
nono: Can you remove this commented line?

GoOz: Sorry, I removed it in the end and forgot to save before commiting and pushing… (¬_¬)

edas:
ça mériterait probablement un ou deux rôles ARIA, type role=alert

PR #1517, opened 2018-09-07
Fix emails

aeris:


cmd/fixer.go
```diff
@@ -311,6 +316,132 @@ var thumbnailsFixer = &cobra.Command{
...
+						address = semicolonAtEnd.ReplaceAllString(address, "")
+						address = dotAtStart.ReplaceAllString(address, "")
+						address = dotAtEnd.ReplaceAllString(address, "")
+						address = lessThanStart.ReplaceAllString(address, "")
+						address = greaterThanEnd.ReplaceAllString(address, "")
+						address = strings.TrimSpace(address)
+						_, err := mail.ParseAddress(address)
+						if err == nil {
+							fmt.Printf("    Email fixed: \"%s\" → \"%s\"\n", old, address)
+							changed = true
```
nono: I'd have expected something like `changed = address != old`

aeris: There is a ParseAddress error before and not here, so address was inevitably changed.

PR #1522, opened 2018-09-13
Build dev version

aeris:
Release cozy-app-dev with the others artifacts


scripts/build.sh
```diff
@@ -241,8 +241,8 @@ case "${1}" in
 		do_assets
 		;;
 
-	docker-dev)
-		do_docker_dev_image
```
nono: We should avoid dead code and remove the `do_docker_dev_image` function

aeris: Fixed

jinroh:
You have included the cozy-stack binary from the release script: https://github.com/cozy/cozy-stack/pull/1522/files#diff-2ad17b8c8e916bda029051ab8b93009d

PR #1531, opened 2018-09-26
Add cozy-stack CLI commands 

maximemenager:
Add 3 more commands to cozy-stack CLI :
- encrypt-data
- encrypt-credentials
- decrypt-data

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -133,9 +133,9 @@ specified path.
 The decryptor key filename is given the ".dec" extension suffix.
 The encryptor key filename is given the ".enc" extension suffix.
 
-The files permissions are 0400.
+The files permissions are 0400.`,
 
-example: cozy-stack config gen-keys ~/credentials-key
+	Example: `$ cozy-stack config gen-keys ~/credentials-key
 keyfiles written in:
 	~/credentials-key.enc
 	~/credentials-key.dec
@@ -150,9 +150,11 @@ keyfiles written in:
 		decryptorFilename := filename + ".dec"
 
 		marshaledEncryptorKey, marshaledDecryptorKey, err := keymgmt.GenerateEncodedNACLKeyPair()
+
 		if err != nil {
 			return nil
 		}
+
 		if err = writeFile(encryptorFilename, marshaledEncryptorKey, 0400); err != nil {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


cmd/config.go
```diff
@@ -164,6 +166,96 @@ keyfiles written in:
 	},
 }
 
+var encryptCredentialsDataCmd = &cobra.Command{
+	Use:     "encrypt-data <encoding keyfile> <text>",
+	Aliases: []string{"encrypt-data"},
```
nono: I'm not sure what you want to do here. `Aliases` is for giving other names to a command, but you use the same name, so it probably does nothing.

cmd/config.go
```diff
@@ -164,6 +166,96 @@ keyfiles written in:
...
+	Short:   "Encrypt data with the specified encryption keyfile.",
+	Long: `
+cozy-stack config encrypt-data encrypts any valid JSON data
+
+example :
```
nono: [nitpick] There are no spaces before `:` in english

cmd/config.go
```diff
@@ -164,6 +166,96 @@ keyfiles written in:
...
+./cozy-stack config encrypt-data ~/.cozy/key.enc "{\"foo\": \"bar\"}"
```
nono: It looks like `cobra.Command` now has an [Example](https://godoc.org/github.com/spf13/cobra#Command) field. We should try to use it

cmd/config.go
```diff
@@ -164,6 +166,96 @@ keyfiles written in:
...
+		data := base64.StdEncoding.EncodeToString(dataEncrypted)
+		fmt.Printf("%s\n", data)
+
+		return nil
+	},
+}
+
+var decryptCredentialsDataCmd = &cobra.Command{
+	Use:     "decrypt-data <decoding keyfile> <ciphertext>",
+	Aliases: []string{"decrypt-data"},
```
nono: Idem

cmd/config.go
```diff
@@ -215,10 +303,26 @@ func writeFile(filename string, data []byte, perm os.FileMode) error {
 	return err
 }
 
+func readKeyFromFile(filepath string) (*keymgmt.NACLKey, error) {
+	keyBytes, err := ioutil.ReadFile(filepath)
+	if err != nil {
+		panic(err)
```
nono: Why do you use `panic(err)` instead of `return nil, err` here?

cmd/config.go
```diff
@@ -215,10 +303,26 @@ func writeFile(filename string, data []byte, perm os.FileMode) error {
...
+	}
+
+	keyStruct, err := keymgmt.UnmarshalNACLKey(keyBytes)
```
nono: You can shorten this function by replacing the last 5 lines with `return keymgmt.UnmarshalNACLKey(keyBytes)`

PR #1544, opened 2018-10-01
Block konnectors in maintenance

jinroh:


pkg/workers/exec/konnector.go
```diff
@@ -79,12 +80,36 @@ type konnectorResult struct {
...
+
+	if err := json.Unmarshal(job.Message, &konnectorMessage); err == nil {
+		inst, err := instance.Get(job.DomainName())
+		if err != nil {
+			return false, err
+		}
+		registries, _ := inst.Registries()
+		app, err := registry.GetApplication(konnectorMessage.Konnector, registries)
+		if err != nil {
+			job.Logger().Errorf("konnector %q could not get application to fetch maintenance status", konnectorMessage.Konnector)
```
nono: As I understand, it does not block developers to run konnectors that are not yet on the apps registry to test them. Am I right?

pkg/workers/exec/konnector.go
```diff
@@ -79,12 +80,36 @@ type konnectorResult struct {
...
+		}
+	}
+
+	if job.Manual || job.TriggerID != "" {
 		return true, nil
 	}
-	trigger := req.Trigger
-	state, err := jobs.GetTriggerState(trigger)
+
+	state, err := jobs.GetTriggerState(job, job.TriggerID)
```
nono: Something looks wrong.

If `jobs.TriggerID` was not the empty string (`""`), the function would have already returned (`if job.Manual || job.TriggerID != "" { return true, nil }`). So, here , `jobs.TriggerID` must be the empty string, and `GetTriggerState` would fail.

pkg/workers/exec/konnector.go
jinroh: Yes it was the motivation. Also, we prefer running konnectors (even if in maintenance) if the registries has issues to respond to the stack.

pkg/workers/exec/konnector.go
jinroh: Fixed.

PR #1557, opened 2018-10-08
Add CLI/API to retrieve an OAuth client

maximemenager:


cmd/instances.go
```diff
@@ -653,6 +654,49 @@ var oauthClientInstanceCmd = &cobra.Command{
 	},
 }
 
+var findOauthClientCmd = &cobra.Command{
+	Use:   "find-oauth-client <domain> <software_id>",
+	Short: "Finds a OAuth client",
```
nono: s/Finds a/Find an/

cmd/instances.go
```diff
@@ -653,6 +654,49 @@ var oauthClientInstanceCmd = &cobra.Command{
...
+		}{
+			Domain:     args[0],
+			SoftwareID: args[1],
+		}
+		// Make request
+		b, err := json.Marshal(body)
+		req := &request.Options{
+			Method: "GET",
+			Path:   "instances/oauth_client",
+			Body:   bytes.NewReader(b),
```
nono: HTTP specs say that GET with a body has no semantics, ie it shouldn't be used. We can use a query-string for these 2 parameters instead.

pkg/couchdb/couchdb.go
```diff
@@ -259,6 +259,10 @@ func dbNameHasPrefix(dbname, dbprefix string) (bool, string) {
 	return true, strings.Replace(dbname, dbprefix, "", 1)
 }
 
+func FindWithoutIndex(db Database, doctype, method, path string, reqBody interface{}, resBody interface{}) error {
```
nono: I don't where this function is used. And if you start a function by an upper-case letter, it means the function will be exported and can be used in other modules. For these functions, a comment is expected to explain what the function does (and the linter will insist on that).

pkg/oauth/client.go
```diff
@@ -157,6 +157,27 @@ func FindClient(i *instance.Instance, id string) (*Client, error) {
...
+	// We should have very few requests. Only on instance creation.
+	err := couchdb.FindDocsUnoptimized(i, consts.OAuthClients, &req, &results)
+	if err != nil {
+		return nil, err
+	}
+	if len(results) == 1 {
+		c = results[0]
+		return c, nil
+	}
+	return nil, err
```
nono: `err` is nil here. You should return a new error instead (or explicitly returns `nil, nil`, but I prefer to avoid this).

pkg/oauth/client.go
```diff
@@ -157,6 +157,27 @@ func FindClient(i *instance.Instance, id string) (*Client, error) {
...
+		Limit:    1,
+	}
+	// We should have very few requests. Only on instance creation.
+	err := couchdb.FindDocsUnoptimized(i, consts.OAuthClients, &req, &results)
+	if err != nil {
+		return nil, err
+	}
+	if len(results) == 1 {
+		c = results[0]
+		return c, nil
```
nono: [nitpick] You can avoid the declaration of `var c *Client` and just `return results[0], nil`

PR #1568, opened 2018-10-12
Fix the Content-Disposition header

nono:
We need to quote the quoted string of a Content-Disposition header
for filenames with a separator like , or ;

https://tools.ietf.org/html/rfc2616#section-3.6

### DIFF
--- a/pkg/vfs/archive.go
+++ b/pkg/vfs/archive.go
@@ -53,7 +53,7 @@ func ContentDisposition(disposition, filename string) string {
 		escaped = "download"
 	}
 	if filename == escaped {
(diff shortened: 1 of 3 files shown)


nono:
[[3.10.0-beta.1] quentin : net::ERR_RESPONSE_HEADERS_MULTIPLE_CONTENT_DISPOSITION](https://trello.com/c/DiI5m1BQ/329-3100-beta1-quentin-neterrresponseheadersmultiplecontentdisposition)

jinroh:
What if there is a `"` is the filename, shouldn't we escape it like that with a `%q`:
```
fmt.Sprintf(`%s; filename=%q`, disposition, escaped)
```

PR #1572, opened 2018-10-12
Improve swift-prefix command formatting

maximemenager:


nono:
[(𝐒) Connaitre les bases/buckets à partir d'une instance](https://trello.com/c/bJcSAWR1/156-%F0%9D%90%92-connaitre-les-bases-buckets-%C3%A0-partir-dune-instance)

PR #1596, opened 2018-10-23
Add option to install specific version of konnector

maximemenager:


nono:
[(?) ETD je peux installer une version particulière d'un connecteur/ d'une application](https://trello.com/c/D24B3yEr/379-etd-je-peux-installer-une-version-particuli%C3%A8re-dun-connecteur-dune-application)

CPatchane:
Is it specific to konnectors? We are still not able to install a specific version of a web app?

maximemenager:
> Is it specific to konnectors? We are still not able to install a specific version of a web app?

It should also work for apps

PR #1602, opened 2018-10-30
Checks the presence of the directory in batch actions

maximemenager:


maximemenager:
Done in https://github.com/cozy/cozy-stack/pull/1603

PR #1604, opened 2018-10-30
Force page refresh every hour for CSRF token

maximemenager:


CPatchane:
TIL this trick to refresh page using html :P, thanks!

PR #1611, opened 2018-11-05
Add CLI to interact with instance apps versions 

maximemenager:
- `show-app-version` to list instances with a precise app version
- `apps-versions` to count apps versions of all instances


cmd/instances.go
```diff
@@ -838,6 +839,79 @@ var showSwiftPrefixInstanceCmd = &cobra.Command{
 	},
 }
 
+var appsVersionsCmd = &cobra.Command{
+	Use:     "apps-versions",
```
nono: I'd prefer to have this command in `cozy-stack apps versions` instead of `cozy-stack instances apps-versions`

PR #1613, opened 2018-11-06
Add a middleware to deprecate IE11

maximemenager:


web/middlewares/ie_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+	ins = &instance.Instance{Domain: "cozy.local", Locale: "en"}
+
+	rec := httptest.NewRecorder()
+	c := e.NewContext(req, rec)
+	c.Set("instance", ins)
+
+	h := CheckIE(echo.NotFoundHandler)
+	err := h(c)
+	assert.NotEqual(t, err, nil)
```
nono: [nitpick] You can use `assert.Error` and `assert.NoError` to make your tests clearer

PR #1617, opened 2018-11-07
Move OAuth parameters outside public context

aeris:


pkg/config/config.go
```diff
@@ -641,6 +648,7 @@ func UseViper(v *viper.Viper) error {
 		},
 		Contexts:   v.GetStringMap("contexts"),
 		Registries: regs,
+		Clouderies: clouderies,
```
nono: Why you don't use `v.GetStringMap("clouderies")` like for the contexts?

pkg/config/config.go
```diff
@@ -439,6 +440,12 @@ func UseViper(v *viper.Viper) error {
 		return err
 	}
 
+	tmp := v.Get("clouderies")
```
nono: Can you also update the example config file? https://github.com/cozy/cozy-stack/blob/master/cozy.example.yaml

pkg/instance/manager.go
```diff
@@ -8,7 +8,11 @@ import (
 	"strings"
 	"time"
 
-	"github.com/cozy/cozy-stack/pkg/utils"
+	"github.com/cozy/cozy-stack/pkg/config"
```
nono: Do you use `gofmt` or `goimports`? It is weird to have a blank line here

aeris: I use goimports

nono: It's weird, goimports remove the blank lines on my machine.

pkg/instance/manager.go
```diff
@@ -26,35 +30,89 @@ const (
...
-// ManagerURL returns an external string for the given ManagerURL kind.
-func (i *Instance) ManagerURL(k ManagerURLKind) (s string, ok bool) {
-	defer func() {
-		if !ok {
-			s = i.PageURL("/manager_url_is_not_specified", nil)
-		}
-	}()
+type managerConfig struct {
+	Api *struct {
+		Url   string
```
nono: Can you use `API` and `URL` like it's the convention in go? Cf https://github.com/golang/go/wiki/CodeReviewComments#initialisms

pkg/instance/manager.go
```diff
@@ -26,35 +30,89 @@ const (
...
+	if err != nil {
+		return "", nil
+	}
+
+	base, ok := config["manager_url"]
+	if !ok {
+		return "", nil
+	}
+
+	baseUrl, err := url.Parse(base.(string))
```
nono: Same here, with `s/baseUrl/baseURL/`

pkg/instance/manager.go
aeris: Done

pkg/config/config.go
aeris: Done

pkg/instance/manager.go
aeris: Done

PR #1620, opened 2018-11-08
Fix install app for instances with no context

nono:


nono:
[🐛🔥 Retirer les tokens OAuth du contexte](https://trello.com/c/1NKvqoen/464-%F0%9F%90%9B%F0%9F%94%A5-retirer-les-tokens-oauth-du-contexte)

PR #1624, opened 2018-11-13
Add OAuth checking before rendering login page

maximemenager:
This PR also use introduces a new Cozy-UI CSS asset and use it in the login page.

capture d ecran 2018-11-28 a 12 30 22


gregorylegarec:
I think we're ready on this one 👍

gregorylegarec:
Thanks for the test fix, we should have seen them, next time let us handle them 👍

PR #1629, opened 2018-11-20
Fix overwritten sharings on new sharing creation

maximemenager:
Already existing tokens where overwritten when a new token was generated
for a share. This fix should keep the old tokens.

### DIFF
--- a/pkg/sharing/sharing.go
+++ b/pkg/sharing/sharing.go
@@ -152,7 +152,7 @@ func (s *Sharing) BeOwner(inst *instance.Instance, slug string) error {
 // CreatePreviewPermissions creates the permissions doc for previewing this sharing,
 // or updates it with the new codes if the document already exists
 func (s *Sharing) CreatePreviewPermissions(inst *instance.Instance) (map[string]string, error) {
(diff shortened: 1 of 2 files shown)


maximemenager:
[PARTAGE - successifs, 1 seul lien fonctionnel](https://trello.com/c/AHrHivYt/500-partage-successifs-1-seul-lien-fonctionnel)

pkg/sharing/sharing.go
```diff
@@ -152,16 +152,32 @@ func (s *Sharing) BeOwner(inst *instance.Instance, slug string) error {
 // CreatePreviewPermissions creates the permissions doc for previewing this sharing,
 // or updates it with the new codes if the document already exists
 func (s *Sharing) CreatePreviewPermissions(inst *instance.Instance) (map[string]string, error) {
+	doc, err := permissions.GetForSharePreview(inst, s.SID)
```
nono: [nitpick] You are doing the comparison two times below to know if there was an existing permissions document, and it is not done the same:

```go
if doc != nil {
```

```go
if err == nil {
```

You should stick to one way, let's say `if doc != nil {` and drop `err` declaration here.

PR #1632, opened 2018-11-26
Use only the first image to generate a gif thumbnail

maximemenager:


nono:
Image Magick can use `[0]` after the input filename to extract the first image from a GIF. In our case, we are using `-` to tell Image Magick to take its input from stdin. I think we can try `-[0]` instead.

maximemenager:
> Image Magick can use `[0]` after the input filename to extract the first image from a GIF. In our case, we are using `-` to tell Image Magick to take its input from stdin. I think we can try `-[0]` instead.

Yep, it seems to work smoothly (tested on gifs >= 40mo and jpegs)

PR #1634, opened 2018-11-27
Fix set-disk-quota 0 command to disable the quota

nono:


nono:
[Correction de la commande set-disk-quota à 0 pour désactiver le quota d'une instance](https://trello.com/c/VXTKBPqj/545-correction-de-la-commande-set-disk-quota-%C3%A0-0-pour-d%C3%A9sactiver-le-quota-dune-instance)

PR #1635, opened 2018-11-28
Onboarding redirection on mobile

maximemenager:


web/auth/auth.go
```diff
@@ -1122,4 +1149,5 @@ func Routes(router *echo.Group) {
 	authorizeGroup.POST("/app", authorizeApp)
 
 	router.POST("/access_token", accessToken)
+	router.POST("/secret_exchange", secretExchange)
```
nono: Note for later: this route should be documented in `docs/auth.md`

pkg/instance/instance.go
```diff
@@ -98,9 +98,14 @@ type Instance struct {
...
 	Dev           bool     `json:"dev,omitempty"`            // Whether or not the instance is for development
 
-	OnboardingFinished bool  `json:"onboarding_finished,omitempty"` // Whether or not the onboarding is complete.
-	BytesDiskQuota     int64 `json:"disk_quota,string,omitempty"`   // The total size in bytes allowed to the user
-	IndexViewsVersion  int   `json:"indexes_version"`
+	OnboardingFinished bool `json:"onboarding_finished,omitempty"` // Whether or not the onboarding is complete.
+
+	OnboardingSecret      string `json:"onboarding_secret,omitempty"`
+	OnboardingApp         string `json:"onboarding_app,omitempty"`
+	OnboardingPermissions string `json:"onboarding_permissions,omitempty"`
```
nono: I'd have put these 3 fields in the oauth client, not in the instance.

web/auth/auth.go
```diff
@@ -1086,6 +1087,32 @@ func passphraseRenew(c echo.Context) error {
...
+
+	doc, err := oauth.FindClient(instance, instance.OnboardingClientID)
+
+	if err == nil {
+		if instance.OnboardingSecret != "" && instance.OnboardingSecret == e.Secret {
+			return c.JSON(http.StatusOK, doc)
+		}
+		r = "Bad secret"
+	}
+	return jsonapi.Errorf(http.StatusBadRequest, "%s", r)
```
nono: I prefer to handle the error first, and to use the correct HTTP codes:

```go
doc, err := oauth.FindClient(instance, instance.OnboardingClientID)
if err != nil {
    return jsonapi.NotFound(err)
}
if instance.OnboardingSecret == "" || instance.OnboardingSecret != e.Secret {
    return jsonapi.InvalidAttribute("secret", errors.New("Invalid secret"))
}
return c.JSON(http.StatusOK, doc)
}

File: web/auth/auth.go [hunk 4]
```diff
@@ -1086,6 +1087,32 @@ func passphraseRenew(c echo.Context) error {
 	return c.Redirect(http.StatusSeeOther, inst.PageURL("/auth/login", nil))
 }
 
+// Used to trade a secret for OAuth client informations
+func secretExchange(c echo.Context) error {
+	type exchange struct {
+		Secret string `json:"secret,omitempty"`
```
nono: Why `omitempty` here?

web/auth/auth.go
```diff
@@ -1086,6 +1087,32 @@ func passphraseRenew(c echo.Context) error {
 	return c.Redirect(http.StatusSeeOther, inst.PageURL("/auth/login", nil))
 }
 
+// Used to trade a secret for OAuth client informations
+func secretExchange(c echo.Context) error {
```
nono: Unit tests are expected for this function

web/settings/context.go
```diff
@@ -47,6 +54,40 @@ func onboarded(c echo.Context) error {
...
+		deeplink := fmt.Sprintf("cozy%s://%s", i.OnboardingApp, i.Domain)
+
+		// Generate the OAuth client
+		OAuthClient := oauth.Client{
+			RedirectURIs: []string{deeplink},
+			ClientName:   i.OnboardingApp,
+			SoftwareID:   "github.com/cozy/cozy-stack",
+		}
+		OAuthClient.Create(i)
+		i.OnboardingClientID = OAuthClient.ClientID
```
nono: In our sequence diagram, it's the cloudery that calls the stack to create the OAuth client

web/settings/context.go
```diff
@@ -47,6 +54,40 @@ func onboarded(c echo.Context) error {
...
+
+		// Redirection
+		// Generate state
+		b := make([]byte, 32)
```
nono: I've updated the sequence diagram: it should be the webview that creates this state. And the state should be checked by the webview after the redirection (ie the state the webview has generated == the state in the query string in the redirected URL)

web/auth/auth.go
```diff
@@ -1128,6 +1130,36 @@ func passphraseRenew(c echo.Context) error {
...
+
+	if err != nil {
+		return jsonapi.NotFound(err)
+	}
+
+	if doc.OnboardingSecret == "" || doc.OnboardingSecret != e.Secret {
+		return jsonapi.InvalidAttribute("secret", errors.New("Invalid secret"))
+	}
+
+	return c.JSON(http.StatusOK, doc)
```
nono: You should use the same format for the response that the `GET /auth/register/:client_id`: https://docs.cozy.io/en/cozy-stack/auth/#get-authregisterclient-id

web/settings/context.go
```diff
@@ -47,6 +51,34 @@ func onboarded(c echo.Context) error {
 		}
 	}
 	redirect := i.OnboardedRedirection().String()
+
+	// Retreiving client
+	clientID := c.QueryParam("client_id")
```
nono: I don't think we have a `client_id` in the query-string here!

web/settings/context.go
```diff
@@ -47,6 +51,34 @@ func onboarded(c echo.Context) error {
...
+	client, err := oauth.FindClient(i, clientID)
+	if err != nil {
+		return err
```
nono: No, on the web onboarding, we have no client, but we should not `return err` here!

nono: And this shows the need of unit tests ;-)

pkg/oauth/client.go
```diff
@@ -70,6 +70,11 @@ type Client struct {
 
 	// XXX omitempty does not work for time.Time, thus the interface{} type
 	SynchronizedAt interface{} `json:"synchronized_at,omitempty"` // Date of the last synchronization, updated by /settings/synchronized
+
+	OnboardingSecret      string `json:"onboarding_secret,omitempty"`
+	OnboardingApp         string `json:"onboarding_app,omitempty"`
+	OnboardingPermissions string `json:"onboarding_permissions,omitempty"`
+	OnboardingState       string `json:"onboarding_state,omitempty"`
```
nono: How these fields will be filled? We need a CLI command to do that, no?

nono: Note: it can be done in another PR if you don't want to work too long on this one

web/settings/context.go
```diff
@@ -47,6 +50,27 @@ func onboarded(c echo.Context) error {
...
+		queryParams := url.Values{
+			"client_id":     {client.ClientID},
+			"redirect_uri":  {deeplink},
+			"state":         {client.OnboardingState},
+			"response_type": {"code"},
+			"scope":         {client.OnboardingPermissions},
+		}
+		redirect = i.PageURL("/auth/authorize", queryParams)
+
+	}
```
nono: I would also like a unit test for this

web/auth/auth.go
edas: done

PR #1638, opened 2018-11-30
Remove swift versionning from swift vfs impl v1 & v2

sblaisot:
Remove swift versionning in swift VFS (implementation v1 and v2).

Swift versionning causes a lot of problems with zero gain:
- There is a bug in file removal in swift implementation v2 which doesn't remove all file versions unlike it is done in implementation v1 (compare https://github.com/cozy/cozy-stack/blob/master/pkg/vfs/vfsswift/impl_v1.go#L383-L412 and https://github.com/cozy/cozy-stack/blob/master/pkg/vfs/vfsswift/impl_v2.go#L380-L394)
- There is a bug in openstack swift versions pior to 2.19.1 that prevent using unicode filenames in versionned containers: https://github.com/openstack/swift/commit/0e3e7b9b0953baaf7a0686881fd2348fde7d1e59

Simply removing container versionning is the quickest way to fix this.

### DIFF
--- a/pkg/vfs/vfsswift/impl_v1.go
+++ b/pkg/vfs/vfsswift/impl_v1.go
@@ -24,7 +24,6 @@ import (
 const swiftV1ContainerPrefix = "cozy-"
 const swiftV1DataContainerPrefix = "data-"
 
-const versionSuffix = "-version"
 const maxFileSize = 5 << (3 * 10) // 5 GiB
 const dirContentType = "directory"
 
@@ -35,7 +34,6 @@ type swiftVFS struct {
 	domain        string
 	prefix        string
 	container     string
-	version       string
 	dataContainer string
 	mu            lock.ErrorRWLocker
 	log           *logrus.Entry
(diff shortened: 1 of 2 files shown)


nono:
[Swift - Suppression d'un fichier avec des versions](https://trello.com/c/0qZqiiQT/498-swift-suppression-dun-fichier-avec-des-versions)

nono:
[Ne plus utiliser le bucket "versions" pour swift](https://trello.com/c/ieyuAzIL/323-ne-plus-utiliser-le-bucket-versions-pour-swift)

pkg/vfs/vfsswift/impl_v1.go
```diff
@@ -710,8 +668,9 @@ func (f *swiftFileCreation) Close() (err error) {
 				vfs.PushDiskQuotaAlert(f.fs, true)
 			}
 		} else {
-			// Deleting the object should be secure since we use X-Versions-Location
-			// on the container and the old object should be restored.
+			// TODO(xxx): We previously used X-Versions-Location on the container and
+			// the old object should be restored so deleting the object was secure.
+			// This not the case anymore so maybe we should add some error handling here.
```
nono: We really need to fix this (layout v1 and v2) before merging this PR, or we may lose files in production

nono:
I close this pull request. We still want to remove Swift versioning, but the way to do that is moving to a new Swift layout.

PR #1639, opened 2018-12-03
Make it easier to work with the stack assets

nono:


nono:
[[DOC] Expliquer comment utiliser --assets pour des assets externals](https://trello.com/c/x7c11Poo/233-doc-expliquer-comment-utiliser-assets-pour-des-assets-externals)

PR #1640, opened 2018-12-03
Add a shortcode to sharings

maximemenager:


pkg/consts/views.go
```diff
@@ -147,6 +147,21 @@ function(doc) {
...
 }
 
+// PermissionsShareByShortcodeView is the view for fetching the permissions associated
+// to a document via a token code.
+var PermissionsShareByShortcodeView = &couchdb.View{
+	Name:    "by-short-code",
+	Doctype: Permissions,
+	Map: `
+function(doc) {
+	if(doc.shortcodes) {
```
nono: [nitpick] Can you indent this view like the others?

pkg/crypto/utils.go
```diff
@@ -39,3 +39,15 @@ func Base64Decode(value []byte) ([]byte, error) {
 	}
 	return dec[:b], nil
 }
+
+// GenerateRandomString generates a secure random string of length N
+func GenerateRandomString(n int) string {
+	const letters = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
```
nono: Can you add the digits to this list to increase the entropy?

pkg/sharing/sharing.go
```diff
@@ -155,27 +158,32 @@ func (s *Sharing) CreatePreviewPermissions(inst *instance.Instance) (map[string]
...
 		var err error
 
 		var previousVal string
+		var previousShortCode string
 		var ok bool
 
 		// Checks that we don't already have a sharing code
 		if doc != nil {
 			previousVal, ok = doc.Codes[m.Email]
+			previousShortCode, ok = doc.ShortCodes[m.Email]
```
nono: You overwrite the `ok` from the previous line. Are you sure you want to do that?

web/permissions/permissions.go
```diff
@@ -39,6 +40,8 @@ const ContextPermissionSet = "permissions_set"
 // #nosec
 const ContextClaims = "token_claims"
 
+const ShortCodeLen = 12
```
nono: Why do you declare two const for the short code length? One in this file and the other in pkg/sharing/sharing.go

pkg/permissions/permissions.go
```diff
@@ -276,6 +281,42 @@ func GetForShareCode(db prefixer.Prefixer, tokenCode string) (*Permission, error
...
+	if len(res.Rows) == 0 {
+		return "", fmt.Errorf("no permission doc for shortcode %v", shortcode)
+	}
+
+	if len(res.Rows) > 1 {
+		return "", fmt.Errorf("Bad state: several permission docs for shortcode %v", shortcode)
+	}
+
+	perm := &Permission{}
+	err = json.Unmarshal(res.Rows[0].Doc, &perm)
```
nono: Are you sure about `&perm`? `perm` is already a pointer, it looks weird to use `&` with a pointer. Oh, and you should add a unit test for this function, I don't want to guess if it works!

web/permissions/permissions.go
```diff
@@ -92,10 +93,17 @@ func createPermission(c echo.Context) error {
 	}
 
 	var codes map[string]string
+	var shortcodes map[string]string
```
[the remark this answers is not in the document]
nono: https://github.com/cozy/cozy-stack/blob/master/docs/permissions.md#post-permissions

PR #1641, opened 2018-12-03
Remove useless params for front

maximemenager:


web/auth/auth.go
```diff
@@ -150,9 +150,6 @@ func renderLoginForm(c echo.Context, i *instance.Instance, code int, credsErrors
 	} else if strings.Contains(redirectStr, "reconnect") {
 		title = i.Translate("Login Reconnect title")
 		help = i.Translate("Login Reconnect help")
-	} else if oauth {
-		title = i.Translate("Login Connect from oauth title")
-		help = i.Translate("Login Connect from oauth help")
```
nono: @maximemenager I think these translations are not used anywhere else, and can be removed in `assets/locale/en.po`

PR #1645, opened 2018-12-04
Fix Swift encoding

nono:
Disable Accept-Encoding for requests to Swift. For files with a gzipped
size inferior to the real size, a reverse-proxy between the stack and
Swift can gzip the response, and when it is the case, the stack may
return a 500 because it can no longer infer the file size from the
Content-Length, as http.ServeContent does.


nono:
[Erreur 500 au téléchargement d'un fichier](https://trello.com/c/Ky1Rfby5/568-erreur-500-au-t%C3%A9l%C3%A9chargement-dun-fichier)

PR #1650, opened 2018-12-07
Add konnector version in job execution message & improve Service logs

maximemenager:


pkg/workers/exec/konnector.go
```diff
@@ -55,6 +55,7 @@ type KonnectorMessage struct {
 	FolderToSave   string `json:"folder_to_save"` // FolderToSave is the identifier of the folder
 	AccountDeleted bool   `json:"account_deleted,omitempty"`
 
+	Version string `json:"version"`
```
nono: I think it should be removed (just a remaining line of code from a reverted commit), no?

pkg/workers/exec/service.go
```diff
@@ -50,6 +50,9 @@ func (w *serviceWorker) PrepareWorkDir(ctx *jobs.WorkerContext, i *instance.Inst
 		}
 		return
 	}
+
+	w.slug = slug
```
nono: By curiosity, why did you move this assignation?

maximemenager: Only for the logs. If we get the manifest but have any kind of error after (app not ready, service not found, ....), we will still have the slug name instead of an empty one

pkg/workers/exec/konnector.go
maximemenager: Yep, nice find ;)

PR #1657, opened 2018-12-12
Add endpoint and CLI to remove assets

maximemenager:


cmd/config.go
```diff
@@ -357,6 +357,45 @@ var insertAssetCmd = &cobra.Command{
...
+
+		if len(args) != 2 {
+			return cmd.Usage()
+		}
+		assetOption := fs.AssetOption{
+			Context: args[0],
+			Name:    args[1],
+		}
+
+		customAssets = append(customAssets, assetOption)
```
nono: You can simplify this part:

```
if len(args) != 2 {
    return cmd.Usage()
}
customAssets := []fs.AssetOption{
    { Context: args[0], Name: args[1] },
}
```

web/instances/instances.go
```diff
@@ -485,6 +498,7 @@ func Routes(router *echo.Group) {
 	router.POST("/redis", rebuildRedis)
 	router.GET("/assets", assetsInfos)
 	router.POST("/assets", addAssets)
+	router.DELETE("/assets", deleteAssets)
```
nono: I'd prefer a route `DELETE /assets/:context/:name` to avoid having a JSON body for a DELETE request

web/instances/instances.go
```diff
@@ -260,6 +260,19 @@ func addAssets(c echo.Context) error {
...
+
+	if err := json.NewDecoder(c.Request().Body).Decode(&unmarshaledAssets); err != nil {
+		return err
+	}
+	err := config_dyn.RemoveAssets(unmarshaledAssets)
+	if err != nil {
+		wrapError(err)
+	}
+	return c.NoContent(http.StatusNoContent)
+}
```
nono: Do you think you can take some time to add unit tests for dynamic assets?

PR #1665, opened 2018-12-17
Handles agregator accounts

gregorylegarec:
Expose aggregator data from manifest, used in https://github.com/cozy/cozy-home/pull/1041

### DIFF
--- a/pkg/apps/konnector.go
+++ b/pkg/apps/konnector.go
@@ -22,6 +22,8 @@ type KonnManifest struct {
 	Editor     string `json:"editor"`
 	Icon       string `json:"icon"`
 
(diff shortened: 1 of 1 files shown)


pkg/apps/konnector.go
```diff
@@ -22,6 +22,8 @@ type KonnManifest struct {
 	Editor     string `json:"editor"`
 	Icon       string `json:"icon"`
 
+	Agregator   *json.RawMessage `json:"agregator,omitempty"`
```
nono: In english, it is spelled with two `g`: aggregator ;-)

Can you add `cloned.Agregator = cloneRawMessage(m.Agregator)` in the `Clone` method below?

PR #1667, opened 2018-12-17
Update konnectors workflow with aggregator informations

gregorylegarec:
Following https://github.com/cozy/cozy-stack/pull/1665

konnector-workflow

[...]

#### Aggregator accounts

Some konnectors are based on a aggregator service. An aggregator is declared
in the konnector manifest and specify an `accountId` property. When Cozy-Home
detects this property, it check if an account with this given id exists. If not,
it creates it. Every account created for a konnector based on an aggregator is
then related to this aggregator account. This relationship is called `master`
account.

```json
{
    "auth": {
        "login": "000000000",
        "password": "**********"
    },
    "folderPath": "/Administratif/Bankbank",
    "label": "Bankbank",
    "namePath": "Bankbank",
    "relationships": {
      "master": {
        "_id": "service-aggregator-account",
        "_type": "io.cozy.accounts"
      }
    }
}
```
[...]


docs/konnectors-workflow.md
```diff
@@ -29,6 +29,33 @@ saved in CouchDB, but we will show them as the konnectors will see them):
 
 Accounts are manipulated through the `/data/` API.
 
+#### Aggregator accounts
+
+Some konnectors are based on a aggregator service. An aggregator is declared
```
nono: [nitpick] s/a aggregator/an aggregator/

docs/konnectors-workflow.md
```diff
@@ -29,6 +29,33 @@ saved in CouchDB, but we will show them as the konnectors will see them):
...
+in the konnector manifest and specify an `accountId` property. When Cozy-Home
+detects this property, it check if an account with this given id exists. If not,
```
nono: s/check/checks/

docs/konnectors-workflow.md
```diff
@@ -29,6 +29,33 @@ saved in CouchDB, but we will show them as the konnectors will see them):
...
+it creates it. Every account created for a konnector based on an aggregator is
+then related to this aggregator account. This relationship is called `master`
```
nono: Why do you call this relationship `master` and not `aggregator` or `aggregator-account`?

docs/diagrams/konnector-workflow.mmdc
```diff
@@ -17,6 +19,7 @@ sequenceDiagram
     K ->> S: login
     K -->> C: [log] login successful
     C -->> H: [ws] login successful
+    K -->> C: update aggregator account
     K ->> S: fetch data
     K ->> C: save documents and files
     C -->>H: [ws] success / error
```
nono: Can you also update the png in the repository?

docs/konnectors-workflow.md
gregorylegarec: As seen with @edas, this relation should also be used for account sharing the same authentification information.

For example, in the future an user should be able to use its Google io.cozy.accounts, already defined for another konnector, and have different options for this konnector.

So the documents could be linked one with another in this way:

Google account

```json
{
  "_id": "7225fa4cb2fa4ed09ea9211abcb41294",
  "account_type": "google",
  "auth": {
     "email": "random@gmail.com"
   },
   label: "My Google Account !"
}
```

Linked account:

```json
  "_id": "4aadd2da4e46456a83c35ca94e4683a2",
  "account_type": "another service"
  "label": "My service using Google account !"
  "relationships": {
    "master": {
      "data": [
        "_id": "7225fa4cb2fa4ed09ea9211abcb41294",
        "_doctype": "io.cozy.accounts"
      ]
    }
  }
}
  
```

docs/diagrams/konnector-workflow.mmdc
gregorylegarec: Oops, sorry.

docs/konnectors-workflow.md
nono: OK, in that case, we can use `linked`, `delegated`, or `authentication` instead of `master`. `master` seems really too generic and will confuse people later.

gregorylegarec: I have pretty much no opinion, I like `linked`. What @edas think about it ?

edas: `master` was pretty generic, so you're right to search something better.

`delegated` seems in the wrong direction for the link. 

`linked` is ok but you loose the direction and which one is the master between the two.

None seems perfect IMHO, all 3 are fine with me unless you find something better

gregorylegarec: What about `extends`, `main`, `base`, `merge-with`, `aggregator`, `father`, `ancestor`, `lead`, `leader`, `root` ?

ptbrowne: parent ?

gregorylegarec: Yes `parent`, that the one I was looking for to complete the list.

nono: `extends` and `parent` both looks good to me

edas: GFM

gregorylegarec: PR updated 👍

PR #1673, opened 2018-12-19
Add terms & partnership fields to app/konnectors manifests

maximemenager:


pkg/apps/konnector.go
```diff
@@ -33,6 +33,7 @@ type KonnManifest struct {
 	Developer   *json.RawMessage `json:"developer,omitempty"`
 	Screenshots *json.RawMessage `json:"screenshots,omitempty"`
 	Tags        *json.RawMessage `json:"tags,omitempty"`
+	Partnership *json.RawMessage `json:"partnership,omitempty"`
```
nono: Can you update the `Clone` method?

pkg/apps/installer.go
```diff
@@ -334,6 +334,17 @@ func (i *Installer) update() error {
 		}
 	}
 
+	oldTermsVersion := oldManifest.Terms().Version
+	newTermsVersion := newManifest.Terms().Version
+
+	termsAdded := oldTermsVersion == "" && newTermsVersion != ""
+	termsUpdated := oldTermsVersion != newTermsVersion
```
nono: Just to be sure, we agree that `termsAdded` implies `termsUpdated`? Ie, when the `oldTermsVersion` is empty, and `newTermsVersion` is not, they can't be equal, and thus `termsUpdated` is true.

maximemenager: Yes, this is exactly that

PR #1677, opened 2018-12-21
Fix konnectors install after new terms

nono:


nono:
[Versions incohérentes sur le registry](https://trello.com/c/OCHnQuTU/683-versions-incoh%C3%A9rentes-sur-le-registry)

PR #1678, opened 2018-12-21
Fix cozies export

maximemenager:
With these unexported JSON fields, there were not written to the `files-index.json` metadata file during the export, thus not exported to the archive.


maximemenager:
The the commit which introduced the regression is https://github.com/cozy/cozy-stack/commit/2c834cd22926a0883ee60adafcd25bdb752e682d

Maybe there is some good reason for JSON field naming removing

PR #1691, opened 2019-01-04
Override default config file with .local config file

aeris:
Manage the config file "the-debian-way".
Allow config split between system config and user config for easier packaging


nono:
@aeris travis is red:
- the linter is not happy
- the stack doesn't start for sharing tests, with the message `Error: Unable to open and parse configuration file template : open : no such file or directory`

aeris:
Yep, I see it. Going to fix that.

PR #1700, opened 2019-01-11
Fix for same app/version but not code from 2 registries

nono:
When an app is available on two registry spaces, at the same version
number, but not with the same code (customized app for example), the
stack was serving the code from the first version installed. With
this fix, we can differentiate these cases and serve the right code
for the webapp/konnector.

### DIFF
--- a/pkg/apps/apps.go
+++ b/pkg/apps/apps.go
@@ -73,7 +73,7 @@ type Manifest interface {
 	Permissions() permissions.Set
 	Source() string
 	Version() string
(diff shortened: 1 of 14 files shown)


nono:
[La stack se mélange les pinceaux quand un connecteur a du code différent sur 2 espaces de registry](https://trello.com/c/KMikXckC/770-la-stack-se-m%C3%A9lange-les-pinceaux-quand-un-connecteur-a-du-code-diff%C3%A9rent-sur-2-espaces-de-registry)

PR #1706, opened 2019-01-16
Avoid crashing on out-of-memory with several scrypt calls

nono:


nono:
[Éviter que cozy-stack ne crash quand on utilise beaucoup le port d'admin](https://trello.com/c/gmkh1eaQ/801-%C3%A9viter-que-cozy-stack-ne-crash-quand-on-utilise-beaucoup-le-port-dadmin)

PR #1710, opened 2019-01-17
Prevent webapp uninstallation if there is a linked-app oauth client

maximemenager:


web/instances/instances.go
```diff
@@ -518,6 +518,7 @@ func Routes(router *echo.Group) {
 	router.POST("/updates", updatesHandler)
 	router.POST("/token", createToken)
 	router.GET("/oauth_client", findClientBySoftwareID)
+	router.GET("/linked_client", findLinkedClient)
```
nono: Why do you add this route? I don't see any CLI tools that use it

web/apps/apps_test.go
```diff
@@ -408,6 +411,106 @@ func TestThemeWithContext(t *testing.T) {
...
+			Slug:       "drive",
+			SourceURL:  "registry://drive",
+			Registries: inst.Registries(),
+		},
+	)
+	assert.NoError(t, err)
+	_, err = installer.RunSync()
+	assert.NoError(t, err)
+
+	// Link an OAuthClient to drive
```
nono: Humm, the OAuth client just below is NOT linked to drive, right?

web/apps/apps_test.go
```diff
@@ -408,6 +411,106 @@ func TestThemeWithContext(t *testing.T) {
...
+
+	req.Header.Add("Authorization", "Bearer "+token)
+	res, err := client.Do(req)
+	assert.NoError(t, err)
+	assert.Equal(t, 400, res.StatusCode)
+	body, err := ioutil.ReadAll(res.Body)
+	assert.NoError(t, err)
+	assert.Contains(t, string(body), "linked OAuth client exists")
+
+	// Removing the client and uninstall the app to cleanly finish the test
```
nono: I don't see where you remove the client

web/apps/apps_test.go
```diff
@@ -408,6 +411,106 @@ func TestThemeWithContext(t *testing.T) {
...
+	uninstaller.RunSync()
+
+}
+
+func TestUninstallAppWithoutLinkedClient(t *testing.T) {
+	instance.Destroy("testinstancewithoutlinked")
+	inst, err := instance.Create(&instance.Options{
+		Domain: "testinstancewithoutlinked",
+		Locale: "en",
+	})
```
nono: In general, we try to not create/destroy instance in a test, but to do that in the main. At least, we should try to destroy the instance at the end of the test if it is successful (and it's better if we can also delete the instance when the test fails) to avoid polluting the CouchDB for the developers.

PR #1713, opened 2019-01-18
feat: add session encrypted attribute to accounts

doubleface:
The connectors will soon be able to save their cookie in the associated io.cozy.account. This data
is as sensitive as the password and then should also be encrypted.

This PR is naive, I don't know if there are other files to update accordingly


nono:
[Chiffrement de la session pour les io.cozy.accounts](https://trello.com/c/R3eDlOAH/818-chiffrement-de-la-session-pour-les-iocozyaccounts)

PR #1727, opened 2019-01-24
Prevent login page access for blocked instances

maximemenager:


web/routing.go
```diff
@@ -57,6 +57,10 @@ func SetupAppsHandler(appsHandler echo.HandlerFunc) echo.HandlerFunc {
 	mws := []echo.MiddlewareFunc{
 		middlewares.LoadAppSession,
 		middlewares.CheckIE,
+		middlewares.Accept(middlewares.AcceptOptions{
```
nono: Wouldn't it be better to put this middleware before `CheckIE`, like in `SetupRoutes`?

web/routing.go
```diff
@@ -57,6 +57,10 @@ func SetupAppsHandler(appsHandler echo.HandlerFunc) echo.HandlerFunc {
...
+			DefaultContentTypeOffer: echo.MIMETextHTML,
+		}),
+		middlewares.CheckInstanceBlocked,
```
nono: How does it play with https://github.com/cozy/cozy-stack/blob/master/web/apps/serve.go#L50-L58?

web/routing.go
```diff
@@ -136,6 +140,7 @@ func SetupRoutes(router *echo.Echo) error {
 				DefaultContentTypeOffer: echo.MIMETextHTML,
 			}),
 			middlewares.CheckIE,
+			middlewares.CheckInstanceBlocked,
```
nono: What will be shown to the user when their instance is blocked?

maximemenager: The user will see the default HTTP error page with a `StatusPaymentRequired` HTTP code sent by the middleware, which is like this :

blocked

A better-looking page is planned for the future

web/routing.go
maximemenager: Not so good, because the user will never be redirected to the `ManagerURL` error page (instead, the default stack error page will be used).

Another implementation option is to add a condition in the `CheckInstanceBlocked()` to render a `StatusPaymentRequired` error page (like the existing middleware) in case of a non-defined `ManagerURL` for the instance.

With this option, we are using a middleware to check if an instance is blocked in some code sections and using a simple condition in other places, which is not the best we can do.

The last option would be the integration of [this test](https://github.com/cozy/cozy-stack/blob/master/web/apps/serve.go#L50-L58) in the middleware

nono: Well, we should have a message that don't say payment required when the instance is blocked because of too many login attempts. And i'd like to keep the redirection to the manager when a manager URL is defined and the stack doesn't have the reason of the blocking.

Can you do that?

PR #1730, opened 2019-01-24
Allow to use swift CLI comands with swift+https

nono:


nono:
[(𝐒) Lister le nombre de layout v1 vs v2](https://trello.com/c/HGvliCnb/765-%F0%9D%90%92-lister-le-nombre-de-layout-v1-vs-v2)

PR #1731, opened 2019-01-24
Add cli to set an instance passphrase

maximemenager:


cmd/instances.go
```diff
@@ -272,6 +274,52 @@ settings for a specified domain.
 	},
 }
 
+var updateInstancePassphraseCmd = &cobra.Command{
+	Use:     "set-passphrase <domain> <current-passphrase> <new-passphrase>",
```
nono: Hmmm, I wouldn't have asked for the current passphrase on this command, just the new one. If you have access to the CLI, you have special access rights and you are expected to know what you are doing.

cmd/instances.go
```diff
@@ -272,6 +274,52 @@ settings for a specified domain.
...
+			Headers: request.Headers{
+				"Content-Type": "application/json",
+			},
+		})
+
+		switch res.StatusCode {
+		case http.StatusNoContent:
+			fmt.Println("Passphrase has been changed for instance ", domain)
+		case http.StatusOK:
+			return fmt.Errorf("Instance %s has 2FA, you cannot change its password", domain)
```
nono: Why?

cmd/instances.go
```diff
@@ -272,6 +274,52 @@ settings for a specified domain.
...
+			return err
+		}
+		res, err := c.Req(&request.Options{
+			Method: "PUT",
+			Path:   "/settings/passphrase",
+			Body:   bytes.NewReader(reqBody),
+			Headers: request.Headers{
+				"Content-Type": "application/json",
+			},
+		})
```
nono: We should check that `err` is not nil (for example, when we can't connect to the stack).

PR #1732, opened 2019-01-25
Add logs for uploading errors on copy

nono:


nono:
[`Error on uploading file (copy)` bloque la synchro d’un fichier](https://trello.com/c/HBYVMvkW/808-error-on-uploading-file-copy-bloque-la-synchro-dun-fichier)

PR #1737, opened 2019-01-28
Fix the cozy-app-dev image for docker

nono:


CPatchane:
LGTM as far as I understand 👍

PR #1739, opened 2019-01-28
cozy-stack must listen not only on localhost to be accessible from host

aeris:


scripts/cozy-app-dev.sh
```diff
@@ -104,6 +104,7 @@ do_start() {
 
 	${COZY_STACK_PATH} serve --allow-root \
 		--appdir "${appdir}" \
+		--host "::" \
```
nono: I'd have put `0.0.0.0` instead of `::`, but it should work too

aeris: IPv6 soon ? :joy:

nono:
[Corriger l'image docker cozy-app-dev](https://trello.com/c/SpDuDS7t/890-corriger-limage-docker-cozy-app-dev)

PR #1740, opened 2019-01-28
List stack contexts

maximemenager:


sblaisot:
I'm surprised to see it in `/instances` rather than `/config`

sblaisot:
oh, ok. thanks.

PR #1742, opened 2019-01-29
Add instance context name to error page

maximemenager:


web/errors/errors.go
```diff
@@ -152,6 +152,7 @@ func HTMLErrorHandler(err error, c echo.Context) {
 			"Error":       value,
 			"ActionTitle": actionTitle,
 			"ActionURL":   actionURL,
+			"ContextName": i.ContextName,
```
nono: I think `i` may be nil here, and we should do like for `domain`

web/auth/auth_test.go
```diff
@@ -71,6 +71,29 @@ func getSessionID(cookies []*http.Cookie) string {
...
+
+	res, err := client.Do(req)
+	assert.NoError(t, err)
+	assert.Equal(t, http.StatusPaymentRequired, res.StatusCode)
+
+	// Trying with a Accept: text/html header to simulate a browser
+	req.Header.Add("Accept", "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8")
+	res2, err := client.Do(req)
+	assert.NoError(t, err)
+	assert.Equal(t, http.StatusPaymentRequired, res2.StatusCode)
```
nono: Can you also check that we don't have a blank page? Maybe just just look at a title in the body

PR #1748, opened 2019-01-31
Add expire time for redis keys

maximemenager:


cozy.example.yaml
```diff
@@ -205,6 +205,7 @@ redis:
     konnectors: 5
     realtime: 6
     log: 7
+    rate_limiting: 8
```
nono: @sblaisot we will need to update the config file for this pull-request

sblaisot: no problem in our infra.
ping @aeris for self-hosted

web/auth/rate_limiting.go
```diff
@@ -98,7 +98,14 @@ func NewRedisCounter(client redis.UniversalClient) Counter {
 }
 
 func (r *redisCounter) Increment(key string, timeLimit time.Duration) (int64, error) {
-	return r.Client.Incr(key).Result()
+	count, err := r.Client.Incr(key).Result()
+	if err != nil {
+		return 0, err
+	}
+	if count == 1 {
```
sblaisot: Why do you count time from first failed password and not from last failed password ?

nono:
For the self-hosted, there is nothing to do: we use an in-memory version in the stack, not redis.

web/auth/rate_limiting.go
```diff
@@ -98,7 +98,14 @@ func NewRedisCounter(client redis.UniversalClient) Counter {
...
+		r.Client.Expire(key, timeLimit)
```
edas: Non blocking : You probably should have used Redis transactions to deal with concurrency. You may have 1000 requests done by another client between L101 and L106 here

nono: I don't see the concurrency issue here. Can you explain why it matters if 1000 requests have been made between L101 and L106?

edas: It's not an issue (so non blocking) but probably a bad pattern to not use a transaction. No real consequence here

nono: And, even if we have a race condition, I don't think we can do what we want with a redis transaction (neither multi/exec nor watch/discard), but there is still the lua script way.

nono: > probably a bad pattern to not use a transaction

I disagree: it's true for most databases, but not redis. Redis is single-threaded and its transactions are not comparable to transactions in SQL world for example.

PR #1749, opened 2019-01-31
Deprecate the dev flag on instances

nono:
A stack can be built in two release modes: dev and prod. We also has a flag per instance, but it was confusing to have both and, in practice, using a prod release with a dev instance didn't work.

Note that a prod release can be used as a dev release by starting the server like this:

	$ cozy-stack serve --dev


nono:
[L'option --dev ne marche pas quand la stack est buildée en release de prod](https://trello.com/c/HkMGrmIM/250-loption-dev-ne-marche-pas-quand-la-stack-est-build%C3%A9e-en-release-de-prod)

PR #1758, opened 2019-02-04
Improve instance blocked checking

maximemenager:
- Split the notion of "blocked instance" in really "blocked" state and "ToS not signed and deadline exceeded". It implies that a user can access his login page and connect even if his ToS deadline is in error. He will be redirected to his `ManagerURL` later

- Add a field `blocking_reason` in Instance to store the information about the blocking reason

### DIFF
--- a/web/middlewares/instance.go
+++ b/web/middlewares/instance.go
@@ -56,20 +56,15 @@ func CheckInstanceBlocked(next echo.HandlerFunc) echo.HandlerFunc {
 			}
 
 			// Fallback by trying to determine the blocking reason
-			var returnCode int
 			reason := i.BlockingReason
+			returnCode := http.StatusPaymentRequired
 
-			switch i.BlockingReason {
-			case instance.BlockedPaymentFailed.Code:
-				returnCode = http.StatusPaymentRequired
+			if reason == "" {
+				reason = http.StatusText(http.StatusPaymentRequired)
+			} else if reason == instance.BlockedPaymentFailed.Code {
 				reason = instance.BlockedPaymentFailed.Message
-			default:
-				// Used for retro-compatibility reasons. Will be deprecated.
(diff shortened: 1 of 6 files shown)


web/middlewares/instance.go
```diff
@@ -46,13 +46,55 @@ func CheckInstanceBlocked(next echo.HandlerFunc) echo.HandlerFunc {
 			return next(c)
 		}
 		if i.CheckInstanceBlocked() {
+			if url, _ := i.ManagerURL(instance.ManagerBlockedURL); url != "" {
+				return c.Redirect(http.StatusFound, url)
```
nono: I think it is creating a loop of redirections: when the instance is blocked, the user is redirected from the stack to the manager, but as the user is not logged in on the manager, he is redirected to the login page on the stack, but the instance is blocked, and the user is redirected again to the manager, etc.

web/middlewares/instance.go
```diff
@@ -46,18 +46,64 @@ func CheckInstanceBlocked(next echo.HandlerFunc) echo.HandlerFunc {
...
+			case instance.BlockedPaymentFailed.Code:
+				returnCode = http.StatusPaymentRequired
+				reason = instance.BlockedPaymentFailed.Message
+			default:
+				// Used for retro-compatibility reasons. Will be deprecated.
+				returnCode = http.StatusPaymentRequired
+				if reason == "" {
+					reason = http.StatusText(http.StatusPaymentRequired)
+				}
+			}
```
nono: It looks like we can simplify this code:

```go
returnCode := http.StatusPaymentRequired
reason := i.BlockingReason
if reason == "" {
    reason = http.StatusText(http.StatusPaymentRequired)
} else if reason == instance.BlockedPaymentFailed.Code {
    reason = instance.BlockedPaymentFailed.Message
}
```

PR #1762, opened 2019-02-06
Improve instance blocked message

maximemenager:
capture du 2019-02-06 16-55-34


assets/templates/instance_blocked.html
```diff
@@ -0,0 +1,31 @@
+<!DOCTYPE html>
+<html lang="en">
```
nono:
```suggestion
<html lang="{{.Locale}}">

File: assets/templates/instance_blocked.html [hunk 2]
```diff
@@ -0,0 +1,31 @@
...
+
+<head>
+    <meta charset="utf-8">
+    <title>Cozy</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <link rel="stylesheet" href="{{asset .Domain "/fonts/fonts.css"}}">
+    <link rel="stylesheet" href="{{asset .Domain "/styles/stack.css"}}">
+
```
nono:
```suggestion
{{.ThemeCSS}}
```

pkg/instance/instance.go
```diff
@@ -87,7 +87,7 @@ type BlockingReason struct {
 
 var (
 	// BlockedLoginFailed is used when a security issue has been detected on the instance
-	BlockedLoginFailed = BlockingReason{Code: "LOGIN_FAILED", Message: "The instance was block because of too many login failed attempts"}
+	BlockedLoginFailed = BlockingReason{Code: "LOGIN_FAILED", Message: "The instance was blocked because of too many login failed attempts"}
```
nono: The message is weird: it should be too many login failed or too many login attempts, but not both failed and attempts, no?

web/middlewares/instance.go
```diff
@@ -48,7 +48,11 @@ func CheckInstanceBlocked(next echo.HandlerFunc) echo.HandlerFunc {
 		if i.CheckInstanceBlocked() {
 			// Standard checks
 			if i.BlockingReason == instance.BlockedLoginFailed.Code {
-				return echo.NewHTTPError(http.StatusServiceUnavailable, instance.BlockedLoginFailed.Message)
+				return c.Render(http.StatusServiceUnavailable, "instance_blocked.html", echo.Map{
+					"Domain":      i.ContextualDomain(),
+					"ContextName": i.ContextName,
+					"Reason":      instance.BlockedLoginFailed.Message,
```
nono: You need to pass the Locale and ThemeCSS here

web/middlewares/instance.go
```diff
@@ -70,7 +74,11 @@ func CheckInstanceBlocked(next echo.HandlerFunc) echo.HandlerFunc {
 			case jsonapi.ContentType, echo.MIMEApplicationJSON:
 				return c.JSON(returnCode, i.Warnings())
 			default:
-				return echo.NewHTTPError(returnCode, reason)
+				return c.Render(returnCode, "instance_blocked.html", echo.Map{
+					"Domain":      i.ContextualDomain(),
+					"ContextName": i.ContextName,
+					"Reason":      reason,
```
nono: Same here

web/middlewares/instance.go
nono: OK, it's harder that what it looks like: we have some circular import issues

PR #1763, opened 2019-02-06
Gzip the responses for /registry

nono:


nono:
[Gzipper les réponses sur /registry](https://trello.com/c/yzYDt06c/938-gzipper-les-r%C3%A9ponses-sur-registry)

PR #1769, opened 2019-02-11
Add favicon context for apps

maximemenager:


web/apps/serve.go
```diff
@@ -302,6 +304,25 @@ func BuildTemplates() {
...
+func favicon(i *instance.Instance) template.HTML {
+	buf := new(bytes.Buffer)
+	err := faviconTemplate.Execute(buf, echo.Map{
+		"Domain":      i.ContextualDomain(),
+		"ContextName": i.ContextName,
+	})
+	if err != nil {
+		panic(err)
+	}
+	return template.HTML(buf.String()) // #nosec
```
nono: Do you think it would be complicated to move `favicon` and `themecss` to `web/middlewares`, so that we can have the same helpers for both the stack assets and the webapps pages?

maximemenager: Yep, nice pick ;)

PR #1774, opened 2019-02-13
Fix a panic when there was an empty old settings

nono:


nono:
[Panic quand les settings sont vides](https://trello.com/c/GrnSLrVO/966-panic-quand-les-settings-sont-vides)

PR #1775, opened 2019-02-13
Fix a panic on fsck for swift layout v1

nono:


nono:
[Panic sur le fsck en layout swift v1](https://trello.com/c/t2OikmYc/967-panic-sur-le-fsck-en-layout-swift-v1)

PR #1779, opened 2019-02-18
Deeplink desktop

maximemenager:


maximemenager:
Firefox browser always tries to reach a URL with an unknown protocol (cozy://...), even if it cannot handle it. For this reason, we have to check with a background iframe if the URL is reachable before trying to open it.

assets/scripts/check-deeplink.js
```diff
@@ -0,0 +1,145 @@
...
+  function onBlur() {
+    clearTimeout(timeout);
+    handler.remove();
+    successCb();
+  }
+
+  window.location = uri;
+}
+
+function openUriUsingFirefox(uri, failCb, successCb) {
```
Crash--: We can remove this function. We use the same on Safari and Firefox now

assets/scripts/check-deeplink.js
```diff
@@ -0,0 +1,145 @@
...
+    successCb();
+  } catch (e) {
+    if (e.name == "NS_ERROR_UNKNOWN_PROTOCOL") {
+      failCb();
+    }
+  }
+}
+
+function openUriWithMsLaunchUri(uri, failCb, successCb) {
+  navigator.msLaunchUri(uri, successCb, failCb);
```
Crash--: I didn't test the code on MS Edge, needs to be done during the validation of this PR.

assets/scripts/check-deeplink.js
```diff
@@ -0,0 +1,132 @@
+function _registerEvent(target, eventType, cb) {
+  if (target.addEventListener) {
+    target.addEventListener(eventType, cb);
+    return {
+      remove: function() {
+        target.removeEventListener(eventType, cb);
+      }
+    };
+  } else {
+    target.attachEvent(eventType, cb);
```
nono: `attachEvent`, really? It is not only for IE6/7/8?

assets/scripts/check-deeplink.js
```diff
@@ -0,0 +1,132 @@
...
+  return iframe;
+}
+
+function openUriWithHiddenFrame(uri, failCb, successCb) {
+  var timeout = setTimeout(function() {
+    failCb();
+    handler.remove();
+  }, 1000);
+
+  var iframe = document.querySelector("#hiddenIframe");
```
nono:
```suggestion
  var iframe = document.getElementById("hiddenIframe");
```

assets/scripts/check-deeplink.js
```diff
@@ -0,0 +1,132 @@
...
+  var isOpera = !!window.opera || navigator.userAgent.indexOf(" OPR/") >= 0;
+  var ua = navigator.userAgent.toLowerCase();
+  return {
+    isOpera: isOpera,
+    isFirefox: typeof InstallTrigger !== "undefined",
+    isSafari:
+      (~ua.indexOf("safari") && !~ua.indexOf("chrome")) ||
+      Object.prototype.toString
+        .call(window.HTMLElement)
+        .indexOf("Constructor") > 0,
```
nono: Maybe we can extract this code to a variable, like for `isOpera`.

assets/scripts/check-deeplink.js
```diff
@@ -0,0 +1,132 @@
...
+    openUriWithMsLaunchUri(uri, failCb, successCb);
+  } else {
+    var browser = checkBrowser();
+
+    if (browser.isChrome || browser.isIOS) {
+      openUriWithTimeoutHack(uri, failCallback, successCallback);
+    } else if (browser.isSafari || browser.isFirefox) {
+      openUriWithHiddenFrame(uri, failCallback, successCallback);
+    } else {
+      unsupportedCb();
```
nono: For `failCb` and `successCb`, we are checking that we have a function before calling it (`failCb && failCb();`), but we are not doing it for `unsupportedCb`. In practice, in the call to `check` below, `unsupportedCb` is not function.

assets/templates/authorize.html
```diff
@@ -19,13 +19,14 @@
           <a href="https://cozy.io" target="_blank" title="Cozy Website" class="shield"></a>
         </header>
         <div class="container">
-          <form method="POST" action="/auth/authorize" class="login auth">
+          <form method="POST" action="/auth/authorize" class="login auth" id="loginform">
```
nono: Please use another id that `loginform`, as this form isn't for _login_.

assets/templates/authorize.html
```diff
@@ -92,5 +93,6 @@ <h1>{{t "Authorize Title" .Client.ClientName}}</h1>
       </section>
     </main>
     <script src="{{asset .Domain "/scripts/cancel-button.js"}}"></script>
+    <script src="{{asset .Domain "/scripts/check-deeplink.js"}}"></script>
```
nono: Can we insert this script only if the redirect URI is a deeplink?

assets/scripts/check-deeplink.js
```diff
@@ -0,0 +1,133 @@
...
+
+(function(window) {
+  let url = window.location.href;
+  let queryString = url.split("?")[1];
+  let params = new URLSearchParams(queryString);
+  const fallbackUri = params.get("fallback_uri");
+  const form = document.getElementById("authorizeform");
+  form.addEventListener("submit", async e => {
+    /*
+      We want to call manullaly the /authorize route
```
nono: [nit] *manually

assets/scripts/check-deeplink.js
```diff
@@ -0,0 +1,133 @@
...
+        encodeURIComponent(el.name) + "=" + encodeURIComponent(el.value)
+      );
+    }
+    const bodyString = arr.join("&");
+    const action = form.action;
+    const response = await fetch(action, {
+      method: "POST",
+      body: bodyString,
+      headers: {
+        "Content-Type": "application/x-www-form-urlencoded"
```
nono: Can you also add a `Accept: application/json` header? It would prefer if the server decides to response with JSON or a redirection based on this HTTP header.

assets/scripts/check-deeplink.js
```diff
@@ -0,0 +1,133 @@
...
+      }
+    });
+    const deeplink = await response.json();
```
nono: I'd prefer if the JSON response could be an object with a `deeplink` field, and not just a string. It would allow in he future to send more things in the response without having to change the format.

assets/scripts/check-deeplink.js
```diff
@@ -0,0 +1,133 @@
...
+    isSafari,
+    isIOS: /iPad|iPhone|iPod/.test(navigator.userAgent) && !window.MSStream,
+    isChrome: !!window.chrome && !isOpera
+  };
+}
+
+function check(uri, failCb) {
+  function failCallback() {
+    failCb && failCb();
+  }
```
nono: What is the purpose of `failCallback`, as `check` is always called with a `failCb` function?

assets/scripts/check-deeplink.js
```diff
@@ -0,0 +1,133 @@
...
+    failCb();
+    handler.remove();
+  }, 500);
+
+  let iframe = document.querySelector("hiddenIframe");
+  if (!iframe) {
+    iframe = _createHiddenIframe(document.body, "about:blank");
+  }
+
+  const handler = window.addEventListener("blur", onBlur);
```
nono: Sam here

assets/scripts/check-deeplink.js
```diff
@@ -0,0 +1,133 @@
...
+    handler.remove();
+  }, 500);
+
+  //handle page running in an iframe (blur must be registered with top level window)
+  let target = window;
+  while (target != target.parent) {
+    target = target.parent;
+  }
+
+  const handler = target.addEventListener("blur", onBlur);
```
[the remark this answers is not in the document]
gregorylegarec: You made me doubt, but as `handler` is encapsulated in a function, [it seems to work](https://jsfiddle.net/og3q0t7f/).

assets/scripts/check-deeplink.js
```diff
@@ -0,0 +1,133 @@
...
+    isFirefox: typeof InstallTrigger !== "undefined",
+    isSafari,
+    isIOS: /iPad|iPhone|iPod/.test(navigator.userAgent) && !window.MSStream,
+    isChrome: !!window.chrome && !isOpera
+  };
+}
+
+function check(uri, failCb) {
+  function failCallback() {
+    failCb && failCb();
```
gregorylegarec: nit: `typeof failCb === 'function' && failCb()`

assets/scripts/check-deeplink.js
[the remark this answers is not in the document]
gregorylegarec: (Btw the code should be clearer with the timeout affected after `handler` declaration.)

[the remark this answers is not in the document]
nono: My bad, the actual code works. See http://2ality.com/2015/02/es6-scoping.html#the-temporal-dead-zone for an explanation.

assets/scripts/check-deeplink.js
maximemenager: Fixed in 456bada

assets/scripts/check-deeplink.js
```diff
@@ -0,0 +1,134 @@
...
+    }
+  }
+}
+
+(function(window) {
+  let params = new URLSearchParams(document.location.search.substring(1));
+
+  const fallbackUri = params.get("fallback_uri");
+  const form = document.getElementById("authorizeform");
+  form.addEventListener("submit", async e => {
```
nono: [nit] s/async e =>/function (e)/

PR #1793, opened 2019-02-27
Improve the MIME-Type handling

nono:
The magic package was not maintained and some types like webp images were missing. So, I've replaced it with an external library, filetype.

And, I think we were using a Content-Type with a charset at some places where a type and subtype, but no charset, would be better. I've changed that too.


nono:
[Refacto sur les content-type](https://trello.com/c/ndL5th4b/1053-refacto-sur-les-content-type)

PR #1794, opened 2019-02-28
Fix the error handling in the konnector worker

nono:
It a konnector fails on the first, and then succeeds on a next retry, we should reset the error from the previous run to not see it as a whole failure.


nono:
[Le retry sur un connecteur garde le resultat de la 1ere execution](https://trello.com/c/MV4uDbSp/1062-le-retry-sur-un-connecteur-garde-le-resultat-de-la-1ere-execution)

PR #1795, opened 2019-03-04
Fix a panic when updating the settings

nono:
We were comparing the old and new values for the settings to know if we really need to persist the changed to CouchDB. But the comparison can panic if a value is a slice or a map. Using reflect.DeepEqual fix this.


nono:
[Erreur sur le changement de langue](https://trello.com/c/435iyb8m/1069-erreur-sur-le-changement-de-langue)

PR #1801, opened 2019-03-05
Include the check deeplink JS only when useful

nono:
If we don't have a fallback URI, it is useless to try to detect if the device can handle a deeplink or not, as we can't do anything if it does not. So, let's include this script only if needed.


nono:
[Bug sur l'oauth pour les apps mobiles](https://trello.com/c/4nCnEWuC/1077-bug-sur-loauth-pour-les-apps-mobiles)

PR #1806, opened 2019-03-07
Use a map for contacts

nono:
We were using structs for the contacts doctype, but it forces to have exactly the same mapping as the front applications. It can be dangerous and lead to data loss. For example, if a contact was added to a group, and she accepts a sharing, the contact will be removed from the group.


nono:
[Gérer trashed et les groupes pour les contacts](https://trello.com/c/L1MZJGqZ/702-g%C3%A9rer-trashed-et-les-groupes-pour-les-contacts)

PR #1807, opened 2019-03-07
Fix the error message for konnector failures

nono:
If we had some logs from the konnector after the error message, this message was lost.


nono:
[Le retry sur un connecteur garde le resultat de la 1ere execution](https://trello.com/c/MV4uDbSp/1062-le-retry-sur-un-connecteur-garde-le-resultat-de-la-1ere-execution)

PR #1809, opened 2019-03-11
Use /_up for checking CouchDB status

nono:


web/status/status.go
```diff
@@ -15,7 +15,7 @@ func Status(c echo.Context) error {
 	checker := checkup.HTTPChecker{
 		Name:     "CouchDB",
 		Client:   config.GetConfig().CouchDB.Client,
-		URL:      config.CouchURL().String(),
+		URL:      config.CouchURL().String() + "_up",
```
sblaisot: you suppose here that config.CouchURL().String() ends with `/`
better use go equivalent of 'path.join(config.CouchURL().String(), '_up')` to add the `/` if it is missing.

nono: It doesn't work: `http://` is transformed to `http:/` (with a single slash)

sblaisot: ok so maybe you have to normalize at config read to add a `/` at the end if it is missing.

PR #1810, opened 2019-03-11
Quick wins for the dynamic assets

nono:


nono:
[Quick wins sur les assets dynamiques](https://trello.com/c/QDkSNgN9/1106-quick-wins-sur-les-assets-dynamiques)

PR #1813, opened 2019-03-11
Fix refresh the TTL for the assets cache on read

nono:


nono:
[Quick wins sur les assets dynamiques](https://trello.com/c/QDkSNgN9/1106-quick-wins-sur-les-assets-dynamiques)

PR #1818, opened 2019-03-12
Do not block emails when the TOS are not signed

nono:
Most workers are inactive when an instance is blocked for not signed terms of services. But the user may need to receive an email to be able to login and accept the terms of services.


nono:
[Les Cozy avec la 2FA activé et les TOS non signés sont bloqués](https://trello.com/c/JUaNOJxG/1114-les-cozy-avec-la-2fa-activ%C3%A9-et-les-tos-non-sign%C3%A9s-sont-bloqu%C3%A9s)

PR #1822, opened 2019-03-21
Make CSP more permissive for deeplinks

nono:


web/routing.go
```diff
@@ -47,6 +47,12 @@ const (
 	cspImgSrcWhitelist = "https://piwik.cozycloud.cc https://matomo.cozycloud.cc " +
 		"https://*.tile.openstreetmap.org https://*.tile.osm.org " +
 		"https://*.tiles.mapbox.com https://api.mapbox.com"
+
+	// cspFrameSrcWhiteList is a whitelist of custom procols that are allow in
```
maximemenager:
```suggestion
	// cspFrameSrcWhiteList is a whitelist of custom protocols that are allowed in
```

Crash--:
It seems that the generated CSP are: `cozydrive:///`& `cozybanks:///` with triple `///` 

```
Content Security Policy: Couldn’t parse invalid host cozydrive:///
Content Security Policy: Couldn’t parse invalid host cozybanks:///
```

PR #1826, opened 2019-04-01
Add a fixer script for 64k

maximemenager:


cmd/fixer.go
```diff
@@ -489,10 +492,95 @@ var linkedAppFixer = &cobra.Command{
...
+			if content["type"] != "content_mismatch" {
+				continue
+			}
+			cm := content["content_mismatch"].(map[string]interface{})
+			sizeIndex := cm["size_index"].(float64)
+			sizeFile := cm["size_file"].(float64)
+
+			// SizeFile should be 64k shorter than SizeIndex
+			size := int64(64 * 1024)
+			if math.Abs(sizeIndex-sizeFile) != float64(size) {
```
nono: No, we don't compare with `==` or `!=` two floats. You should use some structs to parse the JSON response to force an integer type.

cmd/fixer.go
```diff
@@ -489,10 +492,95 @@ var linkedAppFixer = &cobra.Command{
...
+
+			// Fixing :
+			// - Appending a corrupted suffix to the file
+			// - Force the file index size to the real file size
+			doc.DocName = doc.DocName + corruptedSuffix
+			doc.ByteSize = int64(cm["size_file"].(float64))
+
+			fmt.Printf("Updating index document for file %s\n", fileDoc["path"].(string))
+			if noDryRunFlag {
+				return couchdb.UpdateDoc(inst, doc)
```
nono: For Swift layout v1, we need to also rename the file in Swift

cmd/fixer.go
```diff
@@ -489,10 +492,95 @@ var linkedAppFixer = &cobra.Command{
...
+
+			// Checks if the file is trashed
+			if fileDoc["restore_path"] != nil {
+				// This is a trashed file, just delete it
+				instanceVFS := inst.VFS()
+				fmt.Printf("Removing file %s from instance %s\n", fileDoc["path"].(string), domain)
+				if noDryRunFlag {
+					return instanceVFS.DestroyFile(doc)
+				}
+				return nil
```
nono: If you are doing a `return` here, how do you work on the next file?

PR #1827, opened 2019-04-02
Add a context option to skip perms verification

maximemenager:


pkg/apps/installer.go
```diff
@@ -300,6 +301,31 @@ func (i *Installer) install() error {
 	})
 }
 
+// checkSkipPermissions checks if the instance contexts is configured to skip
+// permissions
+func (i *Installer) checkSkipPermissions() (bool, error) {
+	domain := i.Domain()
+	if i.Domain() == prefixer.UnknownDomainName {
```
nono:
```suggestion
	if domain == prefixer.UnknownDomainName {
```

pkg/apps/installer.go
```diff
@@ -339,7 +365,14 @@ func (i *Installer) update() error {
 		newPermissions := newManifest.Permissions()
 		samePermissions := newPermissions != nil && oldPermissions != nil &&
 			newPermissions.HasSameRules(oldPermissions)
-		if !samePermissions && !i.permissionsAcked {
+
+		// Check if we are going to skip the permissions
+		skip, err := i.checkSkipPermissions()
+		if err != nil {
+			return err
+		}
```
nono: It would be better to move that inside the `if !samePermissions && !i.permissionsAcked` block. There is no need to load the instance and check if we skip the permissions on this instance context in most cases (when permissions haven't changed).

pkg/prefixer/prefixer.go
```diff
@@ -12,11 +12,13 @@ type prefixer struct {
 	prefix string
 }
 
+const UnknownDomainName string = "<unknown>"
```
nono: [nit] Can you add a comment for this constant?

docs/apps.md
```diff
@@ -326,37 +326,44 @@ Content-Type: application/vnd.api+json
...
-  - Tells that the user accepted the permissions/ToS. It is useful if there are
+-   `PermissionsAcked`: (defaults to `true`)
+      - Tells that the user accepted the permissions/ToS. It is useful if there are
     newer permissions or Terms Of Service and you want to be sure they were read
     or accepted. If set to `false`, the update will be blocked and the user will
-    be told that a new app version is available
-
-- `Source` (defaults to `SourceURL` installation parameter):
-  - Use a different source to update this app (e.g. to install a `beta` or `dev`
+    be told that a new app version is available.\
```
nono: [nit] Why do we have a `\` at the end of the line?

maximemenager: I was not sure if the MKdocs markdown interpreter would add the carriage return, so I forced it. In the past, I noticed some differences between my markdown previews and the generated documentation.

PR #1833, opened 2019-04-05
Avoid losing custom permissions during a webapp/konnector update 

maximemenager:
This PR allows to keep the extra-permissions added to a webapp or konnector during updates. 

Here is the workflow:
1 - A diff is made between the app/konnector original manifest and the actual permissions doc of the app. This diff contains the added `verbs` or `values` of each `Rule` of the `Set`
2 - The webapp/konnector is updated, this diff is given to the update function
3 - Once the app is updated, the diff is applied to the permissions of the app

### DIFF
--- a/pkg/permissions/permissions.go
+++ b/pkg/permissions/permissions.go
@@ -353,6 +353,21 @@ func createAppSet(db prefixer.Prefixer, typ, docType, slug string, set Set) (*Pe
 func MergeExtraPermissions(perms, extraPermissions Set) (Set, error) {
 	var permissions Set
 
+	// Appending the extraPermissions which are not in the target permissions
+	for _, ep := range extraPermissions {
+		found := false
+		for _, p := range perms {
+			if ep.Title == p.Title {
+				found = true
+				break
+			}
+		}
+		if !found {
+			permissions = append(permissions, ep)
+		}
+	}
+
+	// Merging the rules already existing
 	for _, rule := range perms {
 		for _, newRule := range extraPermissions {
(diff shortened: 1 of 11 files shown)


pkg/permissions/permissions.go
```diff
@@ -348,6 +348,30 @@ func createAppSet(db prefixer.Prefixer, typ, docType, slug string, set Set) (*Pe
...
+// in the "perms" one
+func MergeExtraPermissions(perms, extraPermissions Set) (Set, error) {
+	var err error
+	var mergedRule *Rule
+	var permissions Set
+
+	for _, rule := range perms {
+		for _, newRule := range extraPermissions {
+			if rule.Title == newRule.Title {
+				mergedRule, err = rule.Merge(newRule)
```
nono: [nit] I think it would be better to declare the two variables here (`mergedRule, err := rule.Merge(newRule)`)

pkg/permissions/set.go
```diff
@@ -211,3 +211,52 @@ func (ps Set) HasSameRules(other Set) bool {
...
 	return true
 }
+
+// Diff returns a the differences between two sets.
+// Useful to see what rules had been added between a original manifest
+// permissions and now.
+//
+// TODO: We are ignoring removed values/verbs between rule 1 and rule 2.
+// - At the moment, it onlys show the added values & verbs
+// - It does not handle new rules added
```
nono: I think we need to handle new rules added: this is how intents work, they add a rule, and it is also how the home adds permissions for on a folder for a konnector

PR #1847, opened 2019-04-12
fix: Put back fonts on pwd reset & renew templates

GoOz:


assets/templates/passphrase_renew.html
```diff
@@ -4,6 +4,7 @@
     <meta charset="utf-8">
     <title>{{.Title}}</title>
     <meta name="viewport" content="width=device-width, initial-scale=1">
+    <link rel="stylesheet" href="{{asset .Domain "/fonts/fonts.css"}}">
```
nono: Can you add `.ContextName`?

assets/templates/passphrase_reset.html
```diff
@@ -4,6 +4,7 @@
     <meta charset="utf-8">
     <title>{{.Title}}</title>
     <meta name="viewport" content="width=device-width, initial-scale=1">
+    <link rel="stylesheet" href="{{asset .Domain "/fonts/fonts.css"}}">
```
nono: Idem

assets/templates/passphrase_renew.html
GoOz: oh for this too? I wonder what it is for because some other templates like `login.html` don't have it on `fonts.css` so I thought it wasn't needed.

nono: It's better to put it on every call to `asset` in templates (but it's not a big deal if it is missing). For the moment, we don't override `fonts.css` via the dynamic asset, but if we want to do that one day, it will be nice to have it ready.

GoOz: Copy that 👍

GoOz:
[🔶 (𝐒) ETUO ma page de reset de mot de passe est personnalisée](https://trello.com/c/ONlJtSbF/1239-%F0%9F%94%B6-%F0%9D%90%92-etuo-ma-page-de-reset-de-mot-de-passe-est-personnalis%C3%A9e)

PR #1848, opened 2019-04-15
feat: Add custom apps URL schemes

ptbrowne:
Add schemes for custom applications. IMHO it would be much better to do that via `cozy.yml`.


nono:
> IMHO it would be much better to do that via cozy.yml.

Yes, let's do that in the config.

PR #1859, opened 2019-04-19
fix: Do not trash files inside a no longer shared dir

nono:
When a directory is inside a shared folder and is moved outside of it, the files of the directory must not be put in the trash by retroaction. This commit fixes the removed flag of the io.cozy.shared documents.


nono:
[Partage - Alice déplace un sous dossier, le contenu va dans la corbeille d'Alice (et Bob)](https://trello.com/c/1eMVlqSU/1285-partage-alice-d%C3%A9place-un-sous-dossier-le-contenu-va-dans-la-corbeille-dalice-et-bob)

PR #1864, opened 2019-04-24
Improve the lint

nono:


nono:
[Améliorer le lint de la stack](https://trello.com/c/xF9zwP6u/1334-am%C3%A9liorer-le-lint-de-la-stack)

PR #1865, opened 2019-04-24
Add an API endpoint/CLI to purge old jobs

maximemenager:


nono:
@sblaisot does this PR do what you want?

docs/jobs.md
```diff
@@ -664,6 +664,34 @@ Accept: application/vnd.api+json
 To use this endpoint, an application needs a permission on the type
 `io.cozy.triggers` for the verb `POST`.
 
+### GET /jobs/purge
+
+This endpoint allows to purge old jobs of an instance.
+Some parameters can be given to this route:
+
+* `months` is the number of months to keep. Jobs queued at a date prior to these
+  months will be removed.
```
nono: [nit] I'd have used something where it's possible to keep only 2 weeks of jobs

pkg/consts/views.go
```diff
@@ -253,6 +254,18 @@ function(doc) {
...
+// state filters also can be used for granular selection
+var JobByQueuedAt = &couchdb.View{
+	Name:    "jobs-by-queued-at",
+	Doctype: Jobs,
+	Map: `
+function (doc) {
+	emit(doc.queued_at, doc._id);
+}
+`,
+}
```
nono: Why do you need both a mango index on `queued_at` and a view? The mango index should be enough, and I prefer to not add views when we can use mango index.

pkg/jobs/broker.go
```diff
@@ -304,6 +313,66 @@ func GetQueuedJobs(db prefixer.Prefixer, workerType string) ([]*Job, error) {
...
+
+	err := couchdb.ExecView(db, consts.JobByQueuedAt, &couchdb.ViewRequest{
+		IncludeDocs: true,
+		EndKey:      date.Format(time.RFC3339Nano),
+	}, &res)
+	if err != nil {
+		return nil, err
+	}
+	if len(res.Rows) == 0 {
+		return nil, ErrNotFoundJob
```
nono: Why do you return an error if there are no rows?

pkg/jobs/broker.go
```diff
@@ -304,6 +313,66 @@ func GetQueuedJobs(db prefixer.Prefixer, workerType string) ([]*Job, error) {
...
+	}
+
+	// Returning docs
+	doc := &Job{}
```
nono: I'd prefer to put this declaration inside the loop. IIRC, `json.Unmarshal` does not reset the document, and we may keep one field from the previous iteration, like `manual_execution` which has `omitempty`.

pkg/jobs/broker.go
```diff
@@ -304,6 +313,66 @@ func GetQueuedJobs(db prefixer.Prefixer, workerType string) ([]*Job, error) {
...
+// GetLastsJobs returns the N lasts job of each state for an instance/worker
+// type pair
+func GetLastsJobs(db prefixer.Prefixer, workerType string, limits map[State]int) ([]*Job, error) {
+	var result []*Job
+	var limit int
+	var ok bool
+
+	for _, state := range []State{Queued, Running, Done, Errored} {
+		jobs := []*Job{}
+		limit, ok = limits[state]
```
nono: `limit, ok := limits[state]`

pkg/jobs/workers_list.go
```diff
@@ -59,6 +59,17 @@ func GetWorkersList() ([]*WorkerConfig, error) {
...
 
+// GetWorkersNamesList returns the names of the configured workers
+func GetWorkersNamesList() []string {
+	workers, _ := GetWorkersList()
+	defaultWorkers := make([]string, len(workers))
+
+	for _, w := range workers {
+		defaultWorkers = append(defaultWorkers, w.WorkerType)
+	}
+	return defaultWorkers
```
nono: [nit] s/defaultWorkers/workerNames/

pkg/jobs/workers_list.go
```diff
@@ -59,6 +59,17 @@ func GetWorkersList() ([]*WorkerConfig, error) {
...
 }
 
+// GetWorkersNamesList returns the names of the configured workers
+func GetWorkersNamesList() []string {
+	workers, _ := GetWorkersList()
+	defaultWorkers := make([]string, len(workers))
+
+	for _, w := range workers {
+		defaultWorkers = append(defaultWorkers, w.WorkerType)
+	}
```
nono: Don't you get a slice with twice the number of expected strings, and the first half with empty strings?

I'd have written the loop as:

```go
	for i, w := range workers {
		defaultWorkers[i] = w.WorkerType
	}
```

cmd/jobs.go
```diff
@@ -80,6 +135,10 @@ func init() {
 	jobsRunCmd.Flags().BoolVar(&flagJobPrintLogs, "logs", false, "print jobs log in stdout")
 	jobsRunCmd.Flags().BoolVar(&flagJobPrintLogsVerbose, "logs-verbose", false, "verbose logging (with --logs flag)")
 
+	jobsPurgeCmd.Flags().StringSliceVar(&flagJobWorkers, "workers", nil, "worker types to iterate over (all workers by default)")
+	jobsPurgeCmd.Flags().IntVar(&flagJobsPurgeMonth, "months", 2, "number of months to look for")
```
nono: I'd be better to have this default of `2` in a single place: maybe something like `jobs.DefaultNumberOfMonthsToKeep`.

web/jobs/jobs.go
```diff
@@ -466,6 +467,94 @@ func cleanJobs(c echo.Context) error {
...
+		}
+	}
+	workers := jobs.GetWorkersNamesList()
+	if workersParam != "" {
+		workers = strings.Split(workersParam, ",")
+	}
+
+	// Step 1: We want to get all the jobs prior to the date parameter.
+	// Jobs returned are the ones we want to remove
+	d := time.Now().AddDate(0, -months, 0)
```
nono: [nit] It feels wrong. I'm not sure why exactly, maybe the fact that we are using a negative number of months instead of a duration.

pkg/jobs/broker.go
```diff
@@ -304,6 +313,66 @@ func GetQueuedJobs(db prefixer.Prefixer, workerType string) ([]*Job, error) {
...
+		if err = json.Unmarshal(row.Doc, &doc); err == nil {
+			jobs = append(jobs, doc)
+		}
+	}
+	return jobs, err
+}
+
+// GetLastsJobs returns the N lasts job of each state for an instance/worker
+// type pair
+func GetLastsJobs(db prefixer.Prefixer, workerType string, limits map[State]int) ([]*Job, error) {
```
nono: I'd prefer to avoid the last parameter (`limits`) if we don't use it anywhere

web/jobs/jobs.go
```diff
@@ -466,6 +467,94 @@ func cleanJobs(c echo.Context) error {
...
+	}
+
+	// No dry run, removing the jobs
+	for _, j := range finalJobs {
+		err := couchdb.DeleteDoc(instance, j)
+		if err != nil {
+			return err
+		}
+	}
+	return c.JSON(200, map[string]int{"deleted": len(finalJobs)})
```
nono: s/200/http.StatusOK/

web/jobs/jobs.go
```diff
@@ -466,6 +467,94 @@ func cleanJobs(c echo.Context) error {
...
+
+			if !found {
+				finalJobs = append(finalJobs, job)
+			}
+		}
+	}
+
+	// No dry run, removing the jobs
+	for _, j := range finalJobs {
+		err := couchdb.DeleteDoc(instance, j)
```
nono: Using `couchdb.BulkDeleteDocs` should be more efficient

web/jobs/jobs.go
```diff
@@ -466,6 +467,94 @@ func cleanJobs(c echo.Context) error {
...
+	// Step 2: We also want to keep a minimum number of jobs for each state.
+	// Jobs returned will be kept.
+	limits := map[jobs.State]int{
+		jobs.Queued:  0, // "0" is equivalent of the default limit
+		jobs.Running: 0,
+		jobs.Done:    0,
+		jobs.Errored: 0,
+	}
+
+	var lastsJobs []*jobs.Job
```
nono: I'd have use a map of id for `lastJobs` (`map[string]struct{}{}`) to make it easier to look if a job id is one of the last jobs or not.

cmd/jobs.go
```diff
@@ -68,6 +78,51 @@ var jobsRunCmd = &cobra.Command{
 	},
 }
 
+var jobsPurgeCmd = &cobra.Command{
+	Use:     "purge-old-jobs <domain>",
```
nono: Can you a `Short` description of this command?

pkg/consts/views.go
maximemenager: Views offers the `endkey` param, which is very useful to get all the jobs before a specific date with `queued_at` sorted view.

nono: And mango has a `$lt` keyword in `selector` to do that.

cozy.example.yaml
```diff
@@ -168,6 +168,9 @@ mail:
   # skip the certificate validation (may be useful on localhost)
   skip_certificate_validation: false
 
+# Sets the default duration of jobs database documents to keep
+defaultDurationToKeep: "2W" # Keep 2 weeks
```
nono: Can we move this parameter to the `jobs` section of the config file?

pkg/jobs/broker.go
```diff
@@ -304,6 +313,60 @@ func GetQueuedJobs(db prefixer.Prefixer, workerType string) ([]*Job, error) {
...
+		return nil, err
+	}
+
+	// Returning docs
+	for _, row := range res.Rows {
+		doc := &Job{}
+		if err = json.Unmarshal(row.Doc, &doc); err == nil {
+			jobs = append(jobs, doc)
+		}
+	}
```
nono: Do we need this loop?

web/jobs/jobs.go
```diff
@@ -466,6 +470,89 @@ func cleanJobs(c echo.Context) error {
...
+	var dur time.Duration
+	var err error
+
+	instance := middlewares.GetInstance(c)
+	if err := middlewares.AllowWholeType(c, webpermissions.DELETE, consts.Jobs); err != nil {
+		return err
+	}
+
+	workersParam := c.QueryParam("workers")
+	durationParam := c.QueryParam("duration")
```
nono: I don't see where we use this duration param

web/jobs/jobs.go
```diff
@@ -466,6 +470,92 @@ func cleanJobs(c echo.Context) error {
...
+
+	conf := config.GetConfig().Jobs
+	dur, err = bigduration.ParseDuration(conf.DefaultDurationToKeep)
```
nono: [nit] You can now do `dur, err := ...`

PR #1868, opened 2019-04-25
Add the fallback URI for authorize with linked apps

nono:
When a mobile app is linked to a web app, the stack will add a fallback parameter on the redirection after the authorize step of OAuth. This fallback parameter is the URL of the web app on this instance. It will allow to send the user to this URL if the mobile app can't be opened.


nono:
[(𝐒) Adapter notre onboarding aux universal link](https://trello.com/c/o36OlYuc/1332-%F0%9D%90%92-adapter-notre-onboarding-aux-universal-link)

PR #1873, opened 2019-04-26
Small improvements for universal links

maximemenager:
* Do not generate a `fallbackURI` if the `redirectURI` is not a deeplink. This allows to handle both the universal links and the deeplinks.
* Use `instance.SubDomain()` instead of `instance.PageURL()` for app url generation

### DIFF
--- a/web/settings/context.go
+++ b/web/settings/context.go
@@ -76,9 +76,15 @@ func finishOnboarding(c echo.Context) error {
 		if err != nil {
 			return err
 		}
-		appSlug := strings.TrimLeft(r.Scheme, "cozy")
-		fallbackURI := i.SubDomain(appSlug).String()
-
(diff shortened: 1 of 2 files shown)


web/settings/context.go
```diff
@@ -76,9 +76,15 @@ func finishOnboarding(c echo.Context) error {
 		if err != nil {
 			return err
 		}
-		appSlug := strings.TrimLeft(r.Scheme, "cozy")
-		fallbackURI := i.SubDomain(appSlug).String()
-
+		// If the redirectURI scheme is not starting with cozy<app>://, it means
+		// that we probably are on a recent mobile, handling universal/android
+		// links. We won't provide a fallbackURI because the redirectURI should
+		// be enough to handle the redirection on the mobile-side
```
Crash--: 👍

PR #1874, opened 2019-04-29
Add sharing by Cozy URL feature

nono:
The goal of this pull request is to make possible to add recipients to a sharing from contacts with a cozy URL (but no email addresses).


nono:
[(𝐋) [REN-PAR-1] ETU je peux faire un partage via une url de Cozy (sans email du destinataire)](https://trello.com/c/jSmVKLJi/258-%F0%9D%90%8B-ren-par-1-etu-je-peux-faire-un-partage-via-une-url-de-cozy-sans-email-du-destinataire)

PR #1875, opened 2019-04-29
Add Swift connection initialization for CLI

maximemenager:


cmd/jobs.go
```diff
@@ -85,6 +85,15 @@ var jobsPurgeCmd = &cobra.Command{
...
 		}
+		if err := config.Setup(cfgFile); err != nil {
+			return err
+		}
+		if config.FsURL().Scheme == config.SchemeSwift ||
+			config.FsURL().Scheme == config.SchemeSwiftSecure {
+			if err := config.InitSwiftConnection(config.GetConfig().Fs); err != nil {
+				return err
+			}
+		}
```
nono: Why do we need a connection to swift for purging old jobs?

cmd/fixer.go
```diff
@@ -497,6 +497,15 @@ var contentMismatch64Kfixer = &cobra.Command{
 		if len(args) == 0 {
 			return cmd.Usage()
 		}
+		if err := config.Setup(cfgFile); err != nil {
```
nono: Why do we need to call `config.Setup` here? The call in `cmd/root.go` is not enough?

cmd/fixer.go
```diff
@@ -497,6 +497,15 @@ var contentMismatch64Kfixer = &cobra.Command{
...
+			return err
+		}
+		if config.FsURL().Scheme == config.SchemeSwift ||
+			config.FsURL().Scheme == config.SchemeSwiftSecure {
+			if err := config.InitSwiftConnection(config.GetConfig().Fs); err != nil {
+				return err
+			}
+		}
```
nono: Well, maybe it would be better to have a `PersistentPreRunE` on the whole cmd/fixer.go to avoid duplicating this for each fixer command.

cmd/jobs.go
maximemenager: The reason is that `instance.GetFromCouch` method calls the `instance.MakeVFS()`. If swift connection is not initialized, the stack panics (https://github.com/cozy/cozy-stack/blob/4c1765674e2c09988bd728927577e5fc4f158c76/pkg/instance/instance.go#L541)

cmd/fixer.go
maximemenager: I based the swift initialization on the linked-app function: https://github.com/cozy/cozy-stack/commit/d33e0af59bc7f1ed71454f3f22a8601e49329357

You are right, a `PersistentPreRunE` on cmd/fixer would certainly be better

cmd/fixer.go
```diff
@@ -41,6 +41,16 @@ var softwareIDs = map[string]string{
 var fixerCmdGroup = &cobra.Command{
 	Use:   "fixer <command>",
 	Short: "A set of tools to fix issues or migrate content for retro-compatibility.",
+	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+		if err := config.Setup(cfgFile); err != nil {
+			return err
+		}
+		if config.FsURL().Scheme != config.SchemeSwift &&
+			config.FsURL().Scheme != config.SchemeSwiftSecure {
+			return fmt.Errorf("swift: the configured filesystem does not rely on OpenStack Swift")
```
nono: Hey, it looks like it make it impossible to run the fixer command for the self-hosted and dev docker images!

PR #1882, opened 2019-04-30
Fix for installing apps with ALL+GET permissions

### LINKED ISSUE #1831: Malformed permissions verbset make installation fails
If an app serve a manifest containing malformed permissions verbset (for example `["ALL", "GET"]`, it does not throw any error from front tests to registry, until app installation.

Imho this should not be a reason for making an app installation fail.

nono:
If an application manifest contains a permission where the verbs contain ALL and other things, we should still allow to install the app and consider the permissions for ALL.

Resolves #1831


nono:
[Permettre d'installer une app dont le manifeste a une permission pour ALL et GET](https://trello.com/c/KSHlYppw/1358-permettre-dinstaller-une-app-dont-le-manifeste-a-une-permission-pour-all-et-get)

PR #1890, opened 2019-05-07
Make apps install/update more reliable

nono:
When installing or updating an application in Swift, the stack puts its file in a temporary directory, and if everything is fine, copies them to their final destination. To know the list of files to copy, the stack calls ObjectNamesAll of the swift library. Using the X-Newest flag should improve reliability by ensuring a consistent response from Swift.


nono:
[Pb de mise à jour rencontré](https://trello.com/c/aiEmkSGd/1373-pb-de-mise-%C3%A0-jour-rencontr%C3%A9)

PR #1892, opened 2019-05-07
Reorganize packages

nono:
After #1868 and #1888, I continue to reorganize the packages. My goal is to have packages organized in those directories:

- `main` and `cmd` are the top level packages
- `web` is where we have the routers and handlers for web requests
- `worker` is where we define the workers for our job system
- `model` is for high-level internal packages (in general one package is used for one doctype)
- `client` is a small number of packages used for writing clients for the stack
- `pkg` is the low-level packages (most of those packages are just a couple of structs and functions).


nono:
[Refacto de la stack pour mieux organiser les modules](https://trello.com/c/RKOUBuar/1395-refacto-de-la-stack-pour-mieux-organiser-les-modules)

PR #1893, opened 2019-05-09
Fix an email template

nono:
The `<mj-image>` tag doesn't accept an inner-padding attribute. See https://www.npmjs.com/package/mjml-image


nono:
[Petites corrections sur les envois de mail](https://trello.com/c/Gsv5QES1/1396-petites-corrections-sur-les-envois-de-mail)

PR #1894, opened 2019-05-09
WIP: fix: Break instance url if too long

GoOz:
Too long instance url wouldn't break and caused the page to enlarge itself, especially problematic on mobile.
- [x] Add break-word property
- [x] Split the instance url in two parts (Slug + instance hostname) so we can add a `<wbr />` tag between them in order to control the breaking and get a more elegant one in every case possible.

Fix: #1884


GoOz:
[Problème de marge pour la page d'oauth de permissions avec une url longue](https://trello.com/c/wpC9Zvnl/1363-probl%C3%A8me-de-marge-pour-la-page-doauth-de-permissions-avec-une-url-longue)

GoOz:
So I checked, it should be good like this, no change needed.

assets/templates/authorize.html
```diff
@@ -31,16 +31,16 @@
...
                   </svg>
-                  <strong>{{.Domain}}</strong>
+                  <strong>{{.InstanceSlugName}}<wbr/>.{{.InstanceDomain}}</strong>
```
nono: By curiosity, why do you use a manuel `<wbr/>` instead of a pure CSS solution (like https://css-tricks.com/snippets/css/prevent-long-urls-from-breaking-out-of-container/) ?

GoOz: Because I can't control where it breaks only with CSS. The browser kinda decide by itself with special characters, space, etc.
Since I want it to break on the domain first then break the slug if still too long there's no other way than adding a `wbr`tag to achieve that.

PR #1896, opened 2019-05-10
Set oldServices field to manifest in case of no updates

maximemenager:
When an app update is being performed, a diff is done on `services` declared in the both old and new manifests. This diff result is applied at the end the update.
The comparison is done by setting an `oldServices` field in the `newManifest`, containing the `oldManifest` services.

In the event of a blocking update (new permissions, terms added, ...) or an unneeded update (same version), the manifest `Update` function is still called, but with the current manifest. 
The manifest field `oldServices` is therefore not set, causing a diff with an empty `oldServices` field. This results in the stack believing that a service has been added, and creates a new trigger.

### DIFF
--- a/model/app/webapp.go
+++ b/model/app/webapp.go
@@ -109,7 +109,7 @@ type WebappManifest struct {
 	FromAppsDir bool        `json:"-"` // Used in development
 	Instance    SubDomainer `json:"-"` // Used for JSON-API links
 
-	OldServices Services `json:"-"` // Used to diff against when updating the app
+	oldServices Services // Used to diff against when updating the app
 
 	Err string `json:"error,omitempty"`
 	err error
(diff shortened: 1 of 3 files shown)


model/app/webapp.go
```diff
@@ -109,7 +109,7 @@ type WebappManifest struct {
 	FromAppsDir bool        `json:"-"` // Used in development
 	Instance    SubDomainer `json:"-"` // Used for JSON-API links
 
-	oldServices Services // Used to diff against when updating the app
+	OldServices Services // Used to diff against when updating the app
```
nono: Why do you export this field? At least, if you need to, add `` `json:"-"` `` to this line.

maximemenager: We need this field to cast the manifest field in `model/apps/installer.go`

nono: But why? It's in the same package, I don't see why it should be exported

maximemenager: My bad, my linter was oddly triggered by this

PR #1899, opened 2019-05-15
Handle CozyMetaData in the stack

maximemenager:


model/metadata/metadata.go
```diff
@@ -0,0 +1,157 @@
...
+		tmp := *cm.CreatedAt
+		cloned.CreatedAt = &tmp
+	}
+	if cm.ImportedAt != nil {
+		tmp := *cm.ImportedAt
+		cloned.ImportedAt = &tmp
+	}
+	cloned.UpdatedByApps = make([]*UpdatedByAppEntry, len(cm.UpdatedByApps))
+	for idx, app := range cm.UpdatedByApps {
+		cloned.UpdatedByApps[idx] = app
```
nono: You are sure about this line? I feel like something is wrong even if I can't say exactly what

model/metadata/metadata.go
```diff
@@ -0,0 +1,157 @@
...
+	if cm.MetadataVersion == 0 {
+		cm.MetadataVersion = defaultMetadata.MetadataVersion
+	}
+	if cm.UpdatedByApps == nil {
+		cm.UpdatedByApps = defaultMetadata.UpdatedByApps
+	}
+}
+
+// Update updates the UpdatedAt timestamp
+func (cm *CozyMetaData) Update() {
```
nono: I'd prefer to call it something like `ChangeUpdatedAt`. `Update` makes me think that we are persisting the document in CouchDB, not just changing a field in memory.

model/metadata/metadata.go
```diff
@@ -0,0 +1,157 @@
...
+		if entry.Slug == slug {
+			cm.UpdatedByApps[idx] = updated
+			return nil
+		}
+	}
+
+	// The entry has not been found, adding it
+	cm.UpdatedByApps = append(cm.UpdatedByApps, updated)
+	cm.UsedByApps = append(cm.UsedByApps, slug)
+	cm.UpdatedAt = &date
```
nono: I think this line should be sooner, as we should set the `UpdatedAt` even if we take the return on line 148

web/intents/intents_test.go
```diff
@@ -134,7 +135,7 @@ func TestMain(m *testing.M) {
 		fmt.Println(err)
 		os.Exit(1)
 	}
-	appPerms, err = permission.CreateWebappSet(ins, webapp.Slug(), webapp.Permissions())
+	appPerms, err = permission.CreateWebappSet(ins, webapp.Slug(), webapp.Permissions(), metadata.New())
```
nono: We have a `slug` here. We should use it for the metadata of the permission doc, no?

web/permissions/permissions.go
```diff
@@ -126,7 +127,23 @@ func createPermission(c echo.Context) error {
...
+	err = md.UpdatedByApp(sourceID, "") // We do not have a version here
+	if err != nil {
+		return err
+	}
+
+	// Adding metadata if it does not exist
+	if subdoc.Metadata == nil {
+		subdoc.Metadata = md
+	} else { // Otherwise, ensure we have all the needed fields
+		subdoc.Metadata.EnsureCreatedFields(md)
```
nono: It looks like you are changing what this route accepts in parameter. You should update documentation if it is the case.

web/permissions/permissions.go
```diff
@@ -278,6 +295,17 @@ func patchPermission(getPerms getPermsFunc, paramName string) echo.HandlerFunc {
 			}
 		}
 
+		// Handle metadata
+		// If the metadata has been given in the body request, just apply it to
+		// the patch
+		if patch.Metadata != nil {
+			toPatch.Metadata = patch.Metadata
+		} else if toPatch.Metadata != nil { // No metadata given in the request, but it does exist in the database: update it
+			updatedMD := toPatch.Metadata.Clone()
```
nono: It really feels weird that we have to call `Clone` here

web/permissions/permissions.go
```diff
@@ -278,6 +295,17 @@ func patchPermission(getPerms getPermsFunc, paramName string) echo.HandlerFunc {
...
+			updatedMD.Update()
+			toPatch.Metadata = &updatedMD
+		}
+
```
nono: We should also look at the token to know which app has made the update to fill `UpdatedByApps`

model/app/konnector.go
```diff
@@ -217,7 +222,11 @@ func (m *KonnManifest) Create(db prefixer.Prefixer) error {
 	if err := couchdb.CreateNamedDocWithDB(db, m); err != nil {
 		return err
 	}
-	_, err := permission.CreateKonnectorSet(db, m.Slug(), m.Permissions())
+	// Add metadata
+	md := metadata.NewWithApp(m.DocID, m.Version())
+	md.DocTypeVersion = KonnectorDocTypeVersion
```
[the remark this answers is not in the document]
edas: Agreed

model/app/konnector.go
```diff
@@ -217,7 +222,11 @@ func (m *KonnManifest) Create(db prefixer.Prefixer) error {
 	if err := couchdb.CreateNamedDocWithDB(db, m); err != nil {
 		return err
 	}
-	_, err := permission.CreateKonnectorSet(db, m.Slug(), m.Permissions())
+	// Add metadata
+	md := metadata.NewWithApp(m.DocID, m.Version())
```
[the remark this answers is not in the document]
edas: Agreed

docs/konnectors-workflow.md
```diff
@@ -304,7 +304,7 @@ instance is in debug mode. It would be too verbose to do otherwise.
 
 ### Doctypes
 
-`io.cozy.konnectors` gives their desiderata for an account
+`io.cozy.konnectors` gives their desired data for an account
```
nono: FYI, desiderata is a word that exists in both french and english: https://en.wiktionary.org/wiki/desiderata

docs/permissions.md
```diff
@@ -372,6 +372,20 @@ Accept: application/vnd.api+json
...
+            "createdAt": "2019-05-14T11:01:41.687737935+02:00",
+            "createdByApp": "drive",
+            "updatedAt": "2019-05-14T12:00:37.372193145+02:00",
+            "updatedByApps": [
+                {
+                    "slug": "drive",
+                    "date": "2019-05-14T12:00:37.37219352+02:00",
+                    "version": "1.1.0"
+                }
+            ]
```
nono: I think it would be safer to not let an application sets the `updatedByApps`. It should be the stack that keep a trace of which apps have made changes to a permission document.

model/app/konnector.go
```diff
@@ -217,7 +218,14 @@ func (m *KonnManifest) Create(db prefixer.Prefixer) error {
 	if err := couchdb.CreateNamedDocWithDB(db, m); err != nil {
 		return err
 	}
-	_, err := permission.CreateKonnectorSet(db, m.Slug(), m.Permissions())
+	// Add metadata
+	md, err := metadata.NewWithApp(m.DocSlug, m.Version())
+	if err != nil {
+		return err
+	}
+	md.DocTypeVersion = permission.DocTypeVersion
```
nono: [nit] I think this line should be moved to `permission.CreateKonnectorSet`

model/metadata/metadata.go
```diff
@@ -0,0 +1,130 @@
...
+	// Version of the cozyMetadata
+	MetadataVersion int `json:"metadataVersion"`
+	// Creation date of the cozy document
+	CreatedAt time.Time `json:"createdAt"`
+	// Slug of the app or konnector which created the document
+	CreatedByApp string `json:"createdByApp,omitempty"`
+	// Version identifier of the app
+	CreatedByAppVersion string `json:"createdByAppVersion,omitempty"`
+	// Last modification date of the cozy document
+	UpdatedAt *time.Time `json:"updatedAt"`
```
nono: Why `CreatedAt` is a `time.Time` and `UpdatedAt` a `*time.Time` (pointer)?

model/permission/permissions.go
```diff
@@ -455,12 +470,13 @@ func CreateShareSet(db prefixer.Prefixer, parent *Permission, sourceID string, c
 }
 
 // CreateSharePreviewSet creates a Permission doc for previewing a sharing
-func CreateSharePreviewSet(db prefixer.Prefixer, sharingID string, codes map[string]string, set Set) (*Permission, error) {
+func CreateSharePreviewSet(db prefixer.Prefixer, sharingID string, codes map[string]string, set Set, metadata *metadata.CozyMetaData) (*Permission, error) {
```
nono: Why didn't you choose the same solution for giving metadata for both `CreateShareSet` and `CreateSharePreviewSet`? In one case, there is a single parameter `subdoc` (a struct with a field for the set and another field for the metadata), and in the other case, there are 2 parameters.

web/middlewares/permissions.go
```diff
@@ -388,3 +388,18 @@ func HasWebAppToken(c echo.Context) bool {
...
 	return pdoc.Type == permission.TypeWebapp
 }
+
+// GetTokenSub returns the subject of a token from the context
+func GetTokenSub(c echo.Context) (string, error) {
+	var claims permission.Claims
+	inst := GetInstance(c)
+	header := c.Request().Header.Get(echo.HeaderAuthorization)
+	token := header[len("Bearer "):]
+	err := crypto.ParseJWT(token, func(token *jwt.Token) (interface{}, error) {
```
nono: I'd prefer to avoid parsing the JWT and checking its validity twice. We are already doing this for checking permissions, can't we reuse that?

web/permissions/permissions.go
```diff
@@ -126,7 +128,30 @@ func createPermission(c echo.Context) error {
...
+			return err
+		}
+	}
+
+	// Handles the metadata part
+	md, err := metadata.NewWithApp(slug, "")
+	if err != nil {
+		return err
+	}
+	md.DocTypeVersion = permission.DocTypeVersion
```
nono: I'd prefer if this line was in `metadata.NewWithApp`

docs/permissions.md
```diff
@@ -372,6 +372,13 @@ Accept: application/vnd.api+json
             "codes": {
                 "jane": "Yohyoo8BHahh1lie"
             }
+        },
+        "cozyMetadata": {
+            "doctypeVersion": 1,
+            "metadataVersion": 1,
+            "createdAt": "2019-05-14T11:01:41.687737935+02:00",
+            "createdByApp": "drive",
```
nono: I think it should not be possible to change the `createdAt` and `createdByApp` once the permissions was created.

pkg/metadata/metadata.go
```diff
@@ -0,0 +1,126 @@
...
+// slug/version already exists or by appending a new entry to the list
+func (cm *CozyMetaData) UpdatedByApp(slug, version string) error {
+	if slug == "" {
+		return ErrSlugEmpty
+	}
+
+	now := time.Now()
+	cm.UpdatedAt = now
+	updated := &UpdatedByAppEntry{Slug: slug, Date: now, Version: version}
+	for idx, entry := range cm.updatedByApps {
```
nono: s/idx/i/ : it is more idiomatic in go to use `i` for the index on loop

web/permissions/permissions.go
nono: Did you find why we need to call `Clone` here?

model/app/konnector.go
```diff
@@ -217,7 +218,13 @@ func (m *KonnManifest) Create(db prefixer.Prefixer) error {
...
 		return err
 	}
-	_, err := permission.CreateKonnectorSet(db, m.Slug(), m.Permissions())
+	// Add metadata
+	md, err := metadata.NewWithApp(m.DocSlug, m.Version(), permission.DocTypeVersion)
+	if err != nil {
+		return err
+	}
+
+	_, err = permission.CreateKonnectorSet(db, m.Slug(), m.Permissions(), md)
```
nono: I think it may be cleaner to pass `m.Version()` to `permission.CreateKonnectorSet`, and let `permission.CreateKonnectorSet` creates the `metadata.NewWithApp`.

docs/permissions.md
edas: I do not mind if it is technically possible (there is no need to add a protection if it's not planned), but no one should have to modify it in a normal situation

maximemenager: I kept the option to modify it, but removed the documentation lines.

web/permissions/permissions.go
```diff
@@ -278,6 +300,20 @@ func patchPermission(getPerms getPermsFunc, paramName string) echo.HandlerFunc {
 			}
 		}
 
+		// Handle metadata
+		// If the metadata has been given in the body request, just apply it to
+		// the patch
+		if patch.Metadata != nil {
```
nono: I'd have added those lines to keep the createdAt from the old metadata if the new metadata doesn't say anything about that (and for some other fields):

```go
if toPatch.Metadata != nil {
    patch.Metadata.EnsureCreatedFields(toPatch.Metadata)
}
```

PR #1901, opened 2019-05-16
Update bar to v6.25.7

cedricmessiant:
We need this for new Drive and Contacts versions


cedricmessiant:
[Publication Drive/Photos 1.18.3](https://trello.com/c/I7mzoNNn/1937-publication-drive-photos-1183)

PR #1908, opened 2019-05-22
Fix updating app/konnector that had no metadata

nono:


nono:
[(𝐒) [DOCTYPE] Ajouter CozyMetadata au io.cozy.permissions](https://trello.com/c/8R2Ril2V/1391-%F0%9D%90%92-doctype-ajouter-cozymetadata-au-iocozypermissions)

PR #1911, opened 2019-05-24
Add CozyMetadata to io.cozy.sessions doctype

maximemenager:


maximemenager:
> I don't see any interest to do that: we already have a `created_at` field on session, and we don't use any other metadata.

You are right, I'm closing the PR.

PR #1912, opened 2019-05-24
Add CozyMetadata for io.cozy.oauth.clients doctype

maximemenager:


model/oauth/client.go
```diff
@@ -357,6 +368,14 @@ func (c *Client) Create(i *instance.Instance) *ClientRegistrationError {
 	c.GrantTypes = []string{"authorization_code", "refresh_token"}
 	c.ResponseTypes = []string{"code"}
 
+	// Adding Metadata
+	appSlug := c.SoftwareID
+	if strings.HasPrefix(c.SoftwareID, "registry://") {
+		appSlug = strings.TrimPrefix(c.SoftwareID, "registry://")
+	}
+	md, _ := metadata.NewWithApp(appSlug, c.SoftwareVersion, DocTypeVersion)
```
nono: > createdByApp: Slug of the app or connector which created the document

As it is described, I don't think we should put the `SoftwareID` (when it is not linked to a webapp) in the `createdByApp`. For example, we are using OAuth clients for sharing, and the SoftwareID in this case is `github.com/cozy/cozy-stack`.

cc: @edas

model/oauth/client.go
```diff
@@ -426,6 +445,20 @@ func (c *Client) Update(i *instance.Instance, old *Client) *ClientRegistrationEr
 		c.NotificationDeviceToken = old.NotificationDeviceToken
 	}
 
+	// Updating metadata
+	appSlug := c.SoftwareID
+	if strings.HasPrefix(c.SoftwareID, "registry://") {
+		appSlug = strings.TrimPrefix(c.SoftwareID, "registry://")
+	}
+
+	md, _ := metadata.NewWithApp(appSlug, c.SoftwareVersion, DocTypeVersion)
```
nono: [nit] You can move this line inside the `if` as it is neither used in the `else` branch or after the if/else block

model/oauth/client.go
edas: The purpose is to know where the change or creation comes from. I don't mind having a "cozy-stack" slug, or even a "cozy/*" namespace for the stack. 
"github.com/cozy/cozy-stack" is somewhat out of scope IMHO.

If it's simplier for you, I agree to discard anything which is not on registry.

Also, be carefull not to include any channel in the slug (could a "registry://" string contain one ?)

model/oauth/client.go
```diff
@@ -357,6 +368,15 @@ func (c *Client) Create(i *instance.Instance) *ClientRegistrationError {
 	c.GrantTypes = []string{"authorization_code", "refresh_token"}
 	c.ResponseTypes = []string{"code"}
 
+	// Adding Metadata
+	md := metadata.New()
+	if strings.HasPrefix(c.SoftwareID, "registry://") {
+		md.CreatedByApp = strings.TrimPrefix(c.SoftwareID, "registry://")
+	}
+	md.CreatedByAppVersion = c.SoftwareVersion
```
nono: We should move this line inside the if block: the `CreatedByAppVersion` is relevant only if `CreatedByApp` is set.

model/oauth/client.go
```diff
@@ -426,6 +446,20 @@ func (c *Client) Update(i *instance.Instance, old *Client) *ClientRegistrationEr
...
 	}
 
+	// Updating metadata
+	appSlug := c.SoftwareID
+	if strings.HasPrefix(c.SoftwareID, "registry://") {
+		appSlug = strings.TrimPrefix(c.SoftwareID, "registry://")
+	}
+
+	if old.Metadata == nil {
+		md, _ := metadata.NewWithApp(appSlug, c.SoftwareVersion, DocTypeVersion)
```
nono: To be consistent, we should not set the `CreatedByApp` and `CreatedByAppVersion` fields if the SoftwareID is not linked to a registry webapp.

model/oauth/client.go
```diff
@@ -104,6 +111,10 @@ func (c *Client) Clone() couchdb.Doc {
 		props := (&v).Clone()
 		cloned.Notifications[k] = *props
 	}
+	if cloned.Metadata != nil {
+		tmp := *c.Metadata
```
nono: Oh, and we should call `Clone` on the `Metadata`!

PR #1913, opened 2019-05-24
Add CozyMetadata for io.cozy.triggers doctype

maximemenager:


model/instance/lifecycle/create.go
```diff
@@ -267,6 +268,10 @@ func buildSettings(opts *Options) *couchdb.JSONDoc {
 
 // Triggers returns the list of the triggers to add when an instance is created
 func Triggers(db prefixer.Prefixer) []job.TriggerInfos {
+	// Init metadata
+	md := metadata.New()
+	md.DocTypeVersion = job.DocTypeVersionTrigger
```
nono: I don't see why we are initializing the metadata here, as we are also doing the same thing in `NewTrigger`

model/sharing/replicator.go
```diff
@@ -112,10 +113,14 @@ func (s *Sharing) retryWorker(inst *instance.Instance, worker string, errors int
...
 		return
 	}
+	// Adding metadata
+	md := metadata.New()
+	md.DocTypeVersion = job.DocTypeVersionTrigger
 	t, err := job.NewTrigger(inst, job.TriggerInfos{
 		Type:       "@in",
 		WorkerType: worker,
 		Arguments:  backoff.String(),
+		Metadata:   md,
```
nono: Idem

model/sharing/setup.go
```diff
@@ -148,10 +149,14 @@ func (s *Sharing) AddTrackTriggers(inst *instance.Instance) error {
...
 			DocType:   rule.DocType,
 		}
+		// Adding metadata
+		md := metadata.New()
+		md.DocTypeVersion = job.DocTypeVersionTrigger
 		t, err := job.NewTrigger(inst, job.TriggerInfos{
 			Type:       "@event",
 			WorkerType: "share-track",
 			Arguments:  args,
+			Metadata:   md,
```
nono: Idem

model/sharing/setup.go
```diff
@@ -175,13 +180,17 @@ func (s *Sharing) AddReplicateTrigger(inst *instance.Instance) error {
...
+	md := metadata.New()
+	md.DocTypeVersion = job.DocTypeVersionTrigger
 	args := consts.Shared + ":CREATED,UPDATED:" + s.SID + ":sharing"
 	t, err := job.NewTrigger(inst, job.TriggerInfos{
 		Domain:     inst.ContextualDomain(),
 		Type:       "@event",
 		WorkerType: "share-replicate",
 		Arguments:  args,
 		Debounce:   "5s",
+		Metadata:   md,
```
nono: Idem

model/sharing/setup.go
```diff
@@ -352,13 +361,17 @@ func (s *Sharing) AddUploadTrigger(inst *instance.Instance) error {
...
+	md := metadata.New()
+	md.DocTypeVersion = job.DocTypeVersionTrigger
 	args := consts.Shared + ":CREATED,UPDATED:" + s.SID + ":sharing"
 	t, err := job.NewTrigger(inst, job.TriggerInfos{
 		Domain:     inst.ContextualDomain(),
 		Type:       "@event",
 		WorkerType: "share-upload",
 		Arguments:  args,
 		Debounce:   "5s",
+		Metadata:   md,
```
nono: Idem

model/job/redis_scheduler_test.go
```diff
@@ -141,8 +141,18 @@ func TestRedisSchedulerWithTimeTriggers(t *testing.T) {
 	for _, trigger := range ts {
 		switch trigger.Infos().TID {
 		case atID:
+			assert.True(t, tat.Infos().Metadata.CreatedAt.Equal(trigger.Infos().Metadata.CreatedAt))
```
nono: [nit] I prefer `assert.Equal` to `assert.True(t, equalCondition)` as it gives a more helpful message when the assertion fails

web/jobs/jobs.go
```diff
@@ -199,6 +201,14 @@ func newTrigger(c echo.Context) error {
 		}
 	}
 
+	// Handle metadata
+	md := metadata.New()
+	if req.Metadata == nil {
+		req.Metadata = md
+	} else {
+		req.Metadata.EnsureCreatedFields(md)
+	}
```
nono: I don't see why we are using allowing the apps to give some metadata, and we are not using the JWT to create a metadata object with verified information

model/job/redis_scheduler_test.go
maximemenager: It seems that redis does not handle the monotonic clock reading (`m=...`), which leads to a not equal time and a failing test

model/job/scheduler.go
```diff
@@ -144,6 +160,10 @@ func (t *TriggerInfos) Clone() couchdb.Doc {
 		tmp := *t.CurrentState
 		cloned.CurrentState = &tmp
 	}
+	if t.Metadata != nil {
+		tmp := *t.Metadata
```
nono: We should `Clone` the metadata

web/instances/instances.go
```diff
@@ -374,11 +375,16 @@ func cleanOrphanAccounts(c echo.Context) error {
 			args = fmt.Sprintf("0 %d %d * * %d", m, h, d)
 		}
 
+		md, err := metadata.NewWithApp(acc.AccountType, "", job.DocTypeVersionTrigger)
```
nono: I wouldn't use `acc.AccountType` for `createdByApp`, as it is the stack that creates the trigger, not the konnector

docs/jobs.md
```diff
@@ -416,6 +416,10 @@ it won't create another job. It can be useful to combine it with the changes
 feed of couchdb with a last sequence number persisted by the worker, as it
 allows to have a nice diff between two executions of the worker.
 
+The `cozyMetadata` field is accepted if you want to add your own custom
+[metadata](https://github.com/cozy/cozy-doctypes/#document-metadata). In the
+other case, it will be automatically added.
+
```
nono: I think we should remove this paragraph as it is no longer true, right?

docs/jobs.md
```diff
@@ -436,6 +440,12 @@ Accept: application/vnd.api+json
         "priority": 3,
         "timeout": 60,
         "max_exec_count": 3
+      },
+      "cozyMetadata": {
+            "doctypeVersion": 1,
+            "metadataVersion": 1,
+            "createdByApp": "drive",
+            "updatedAt": "2019-05-14T12:00:37.372193145+02:00"
```
nono: idem

PR #1914, opened 2019-05-27
Allow mobile apps to install app

nono:
For cozy instances in a context where allow_install_via_a_permission is activated, the stack should allow a mobile app with the right permissions to install other apps. To do that, we need to check OAuth permissions documents on the endpoint for listing apps in the registry and installing an application.

### DIFF
--- a/web/middlewares/permissions.go
+++ b/web/middlewares/permissions.go
@@ -342,6 +342,18 @@ func AllowInstallApp(c echo.Context, appType consts.AppType, sourceURL string, v
 		if !strings.HasPrefix(sourceURL, "registry://") {
 			return ErrForbidden
 		}
+	case permission.TypeOauth:
+		// If the context allow the app install from a permission, this
+		// permission can also be used by mobile apps to install apps from the
+		// registry.
+		inst := GetInstance(c)
+		ctxSettings, err := inst.SettingsContext()
+		if err != nil || ctxSettings["allow_install_via_a_permission"] != true {
+			return ErrForbidden
+		}
+		if !strings.HasPrefix(sourceURL, "registry://") {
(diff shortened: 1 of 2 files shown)


nono:
[🍊 Débloquer les permissions pour l'assistant financier en version mobile](https://trello.com/c/lhv6z6xD/1457-%F0%9F%8D%8A-d%C3%A9bloquer-les-permissions-pour-lassistant-financier-en-version-mobile)

web/middlewares/permissions.go
```diff
@@ -342,6 +342,18 @@ func AllowInstallApp(c echo.Context, appType consts.AppType, sourceURL string, v
...
+		// If the context allow the app install from a permission, this
+		// permission can also be used by mobile apps to install apps from the
+		// registry.
+		inst := GetInstance(c)
+		ctxSettings, err := inst.SettingsContext()
+		if err != nil || ctxSettings["allow_install_via_a_permission"] != true {
+			return ErrForbidden
+		}
+		if !strings.HasPrefix(sourceURL, "registry://") {
+			return ErrForbidden
```
ptbrowne: return errForbidden('Can only install app from registry for OAuth permission')

web/middlewares/permissions.go
```diff
@@ -342,6 +342,18 @@ func AllowInstallApp(c echo.Context, appType consts.AppType, sourceURL string, v
 		if !strings.HasPrefix(sourceURL, "registry://") {
 			return ErrForbidden
 		}
+	case permission.TypeOauth:
+		// If the context allow the app install from a permission, this
```
ptbrowne:
```suggestion
		// If the context allows to install an app via a permission, this
```

PR #1916, opened 2019-05-27
Handle high number of jobs purging

maximemenager:
Some databases have a high number of jobs (250k+), leading to issues on
couch side. This commit should improve the purge by:

- Compute which jobs to delete on go-side instead of couch
- Use chunks for bulk deletes to prevent timeouts

### DIFF
--- a/model/job/broker.go
+++ b/model/job/broker.go
@@ -3,6 +3,7 @@ package job
 import (
 	"context"
 	"encoding/json"
+	"sort"
 	"time"
 
 	"github.com/cozy/cozy-stack/model/permission"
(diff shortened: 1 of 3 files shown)


model/job/broker.go
```diff
@@ -313,61 +314,86 @@ func GetQueuedJobs(db prefixer.Prefixer, workerType string) ([]*Job, error) {
...
+			StartKey: startkey,
 		}
-		err := couchdb.FindDocs(db, consts.Jobs, req, &jobs)
+		err := couchdb.GetAllDocs(db, consts.Jobs, req, &jobs)
+
 		if err != nil {
 			return nil, err
 		}
+
+		lastJob, jobs = jobs[len(jobs)-1], jobs[:len(jobs)-1]
```
nono: What happens if we have 0 jobs here?

model/job/broker.go
```diff
@@ -313,61 +314,86 @@ func GetQueuedJobs(db prefixer.Prefixer, workerType string) ([]*Job, error) {
...
+			remainingJobs = false
+			finalJobs = append(finalJobs, lastJob)
 		}
 	}
 
 	return finalJobs, nil
 }
 
+// GetJobsBeforeDate returns alls jobs queued before the specified date
+func GetJobsBeforeDate(jobs []*Job, date time.Time) ([]*Job, error) {
```
nono: s/GetJobsBeforeDate/FilterJobsBeforeDate/

nono: Why do you return an `error` if it is always nil?

model/job/broker.go
```diff
@@ -313,61 +314,86 @@ func GetQueuedJobs(db prefixer.Prefixer, workerType string) ([]*Job, error) {
...
+		}
+	}
+
+	return b, nil
+}
+
+// FilterByWorkerAndState filters a job slice by its workerType and State
+func FilterByWorkerAndState(jobs []*Job, workerType string, state State, limit int) []*Job {
+	// Step 1: Ordering
+	sort.Slice(jobs, func(i, j int) bool { return jobs[i].QueuedAt.Before(jobs[j].QueuedAt) })
```
nono: As I understand, this function is called 4 times (with different states). As such, the sort is done 4 times, which can be costly in performances. By moving the sort outside of this function, before the loop on states, we can call it only once.

model/job/broker.go
maximemenager: This is a mistake, thanks !

model/job/broker.go
maximemenager: I did not think of that case, the commits below should fix it, thanks ;)

web/jobs/jobs.go
```diff
@@ -495,19 +495,21 @@ func purgeJobs(c echo.Context) error {
...
+	}
+
 	// Step 1: We want to get all the jobs prior to the date parameter.
 	// Jobs returned are the ones we want to remove
 	d := time.Now().Add(-dur)
-	jobsBeforeDate, err := job.GetJobsBeforeDate(instance, d)
-	if err != nil && err != job.ErrNotFoundJob {
-		return err
-	}
+	jobsBeforeDate := job.FilterJobsBeforeDate(allJobs, d)
```
nono: I think there is a subtle bug:

- `job.FilterJobsBeforeDate` is modifying some of the jobs in `allJobs` because of the trick `b := jobs[:0]` to use the same underlying storage for both `allJobs` and `jobsBeforeDate` (alias `b`)
- but we are also using `allJobs` later (https://github.com/cozy/cozy-stack/pull/1916/files#diff-5b7c95bb32fcec71d666c8ab794c5371R512)

PR #1918, opened 2019-05-28
Persist updatedByApps for metadata

nono:
We have to export this field to allow the go json package to encode/decode it when metadata are sent and received from CouchDB.


nono:
[(𝐒) [DOCTYPE] Ajouter CozyMetadata au io.cozy.permissions](https://trello.com/c/8R2Ril2V/1391-%F0%9D%90%92-doctype-ajouter-cozymetadata-au-iocozypermissions)

PR #1919, opened 2019-05-28
Improve rate limit exceeded check

maximemenager:
Before this change, any kind of error returned by the CheckRateLimit
function (like access error to cache for example) was interpreted as a
rate exceeded, resulting in the blocking of the instance. This commit
ensures the error is the good one before blocking the instance.

### DIFF
--- a/web/auth/auth.go
+++ b/web/auth/auth.go
@@ -277,7 +277,7 @@ func login(c echo.Context) error {
 		// Handle 2FA failed
 		if !successfulAuthentication {
 			errCheckRateLimit = limits.CheckRateLimit(inst, limits.TwoFactorType)
-			if errCheckRateLimit != nil {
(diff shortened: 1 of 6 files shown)


web/auth/auth.go
```diff
@@ -316,7 +316,7 @@ func login(c echo.Context) error {
 				successfulAuthentication = true
 			}
 		} else { // Bad login passphrase
-			if err := limits.CheckRateLimit(inst, limits.AuthType); err != nil {
+			if err := limits.CheckRateLimit(inst, limits.AuthType); err != nil && err == limits.ErrRateLimitExceeded {
```
nono: You can replace all the `err != nil && err == limits.ErrRateLimitExceeded` by just `err == limits.ErrRateLimitExceeded `

PR #1923, opened 2019-06-04
Add available apps for intents

maximemenager:
This PR adds an API field that lists non-installed apps that handles the intent.

The `icon` field has not been added alongside the `slug` and `name` ones as it is easy to retrieve by the stack proxy: `http://<instance>/registry/<slug>/icon`.

### DIFF
--- a/model/intent/intents.go
+++ b/model/intent/intents.go
@@ -184,7 +184,7 @@ func (in *Intent) FillAvailableWebapps(inst *instance.Instance) error {
 
 	registries := inst.Registries()
 	for _, webapp := range endSlugs {
-		go func(webapp string, instance *instance.Instance) {
+		go func(webapp string) {
 			webappMan := app.WebappManifest{}
 			v, err := registry.GetLatestVersion(webapp, "stable", registries)
 			if err != nil {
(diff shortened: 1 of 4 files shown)


model/intent/intents.go
```diff
@@ -91,4 +107,124 @@ func (in *Intent) FillServices(instance *instance.Instance) error {
...
+			}
+
+			versionsChan <- webappMan
+		}(webapp, inst)
+	}
+
+	for i := 1; i <= len(endSlugs); i++ {
+		select {
+		case err := <-errorsChan:
+			inst.Logger().Errorf("intent: %s", err)
```
nono: Can you add a `WithField("nspace", "intents")`?

model/intent/intents.go
```diff
@@ -91,4 +106,123 @@ func (in *Intent) FillServices(instance *instance.Instance) error {
...
+		}
+	}
+
+	lastVersions := map[string]app.WebappManifest{}
+	versionsChan := make(chan app.WebappManifest)
+	errorsChan := make(chan error)
+
+	registries := inst.Registries()
+	for _, webapp := range endSlugs {
+		go func(webapp string, instance *instance.Instance) {
```
nono: The `instance` parameter is no longer needed, I think

PR #1924, opened 2019-06-04
Fix sending invitation mails to some contacts

nono:
When a contact is added to a sharing, the stack sends them an email to their address. This address can be in the "foo@example.net" format, or in the "Foo <foo@example.net>" format. This commit fixes sending emails when the latter is used.

### DIFF
--- a/worker/mails/mail.go
+++ b/worker/mails/mail.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"runtime"
+	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


nono:
[Bug à l'envoi des mails pour le partage](https://trello.com/c/5CzAajvt/1504-bug-%C3%A0-lenvoi-des-mails-pour-le-partage)

nono:
Arg, the CI is red because of a dependency that is currently broken: https://github.com/labstack/gommon/commit/41215b3ce6fd47274e088bae299f0cfbe6ecd25c#r33808702

Issue #1925, opened 2019-06-04
Docker image cozy/cozy-app-dev 1.2.8 does not work well when behind a corporate proxy


klehelley:
I have been following the [How to create your first Cozy application](https://docs.cozy.io/en/tutorials/app/) tutorial but when running `yarn start`, instead of the Docker container starting successfully, it fails with the following messages:

    Cozy stack (docker): checking that :8080 is free... Cozy stack (docker): 
    Cozy stack (docker): error: cozy-stack is already running on :8080
    Cozy stack (docker): kill: usage: kill [-s sigspec | -n signum | -sigspec] pid | jobspec ... or kill -l [sigspec]

I have tracked down the cause of the issue and when starting the services, the `cozy-app-dev.sh` script runs command `curl -s --max-time 1 -XGET ":8080"`. ~It looks like `curl` in this case uses the network interface IP address instead of defaulting to `localhost` or `127.0.0.1`. That IP address seems to be dynamic.~ *Edit : in fact as the FQDN/IP part of the URL is missing, it cannot match anything in `no_proxy`, and thus systematically tries to use the proxy when one is defined.*

Now I have configured Docker to use my company's proxy in `~/.docker/config.json`, with `localhost` and `127.0.0.1` being configured to not use the proxy. But because of the way it is called, `curl` sends a request to the proxy, which does respond. As this is not what `cozy-app-dev.sh` is expecting, the script ends.

It would seem that changing line 79 in that file as follows would do the trick:

    - check_not_running ":${COZY_STACK_PORT}" "cozy-stack"
    + check_not_running "localhost:${COZY_STACK_PORT}" "cozy-stack"

I have taken the time to test that change locally by cloning this repository and applying the change, and while it does fix that particular issue, it fails later (visibly the call to `cozy-stack instances add ...` line 168 of `cozy-app-dev.sh`).

    Cozy stack (docker): checking that localhost:8080 is free... Cozy stack (docker): ok
    waiting for couchdb...Cozy stack (docker): ok
    checking couchdb on http://localhost:5984/... Cozy stack (docker): ok
    Cozy stack (docker): starting cozy-stack with /data/cozy-storage...
    Cozy stack (docker): 
    creating instance cozy.tools:8080 using passphrase "cozy"... Cozy stack (docker):                            !! DEVELOPMENT RELEASE !!
    You are running a development release which may deactivate some very important
    security features. Please do not use this binary as your production server.
    
    Cozy stack (docker): error: 
[...]

PR #1926, opened 2019-06-05
Fix cozy-app-dev container start-up behind proxy (#1925)

klehelley:
When behind a corporate proxy, the start-up script inside the Docker
container used curl to query ":8080", which cannot match any `no_proxy`
item, and thus systematically goes through the corporate proxy when one
is configured, resulting in a start-up failure. This commit fixes this
query so that it succeeds in that kind of environment as long as
"localhost" is defined in `no_proxy`.


nono:
Thanks!

PR #1927, opened 2019-06-05
Document cozy-app-dev container usage behind proxy (#1925)

klehelley:
A specific Docker configuration is required for the containers to work
correctly behind a corporate proxy.

This addition to the documentation is valid as long as a Docker image containing the change applied by pull request #1926 is available (otherwise container start-up will fail even with a correct proxy configuration).


nono:
Thansk!

PR #1929, opened 2019-06-06
Allow clients to update the metadata on file uploads

nono:
When a client uploads a file, it can use the MetadataID query parameter to give the key for a metadata object that was created just before. It works for new files, but also when uploading a new version of an existing file.


nono:
[(M) [io.cozy.files] Pouvoir modifier les méta-données d'un fichier](https://trello.com/c/WwknilkH/956-pouvoir-modifier-les-m%C3%A9ta-donn%C3%A9es-dun-fichier)

PR #1931, opened 2019-06-12
Add cozyMetadata to io.cozy.files

nono:
The goal of this pull request is to add cozyMetadata to the io.cozy.files doctype. It is a doctype managed by the stack, with custom routes. The clients have few things to do: only the konnectors can be interested by the `sourceAccount` and `sourceAccountIdentifier` that can't be automatically filled by the stack.

The `cozyMetadata` for the io.cozy.files have more fields than the standard: it has some fields like `createdOn` to know the instance on which the file was created, as it can be useful for files shared between several cozy instances, and has some fields related to the last upload (like `uploadedAt`). See https://github.com/cozy/cozy-doctypes/blob/master/docs/io.cozy.files.md#cozymetadata for more details.

### DIFF
--- a/docs/files.md
+++ b/docs/files.md
@@ -281,7 +281,19 @@ A file is a binary content with some metadata.
 
 ### POST /files/:dir-id
 
-Upload a file
+Upload a file in the directory identified by `:dir-id`.
+
+The `created_at` field will be the first valid value in this list:
+
+- the datetime extracted from the EXIF for a photo
+- the `CreatedAt` parameter from the query-string
+- the `Date` HTTP header
+- the current time from the server.
+
+The `updated_at` field will be the first value in this list:
+
+- the `Date` HTTP header
+- the current time from the server.
 
 #### Query-String
 
@@ -754,7 +766,7 @@ Location: https://cozy.example.com/files/9152d568-7e7c-11e6-a377-37cbfb190b4b
             "size": 12,
             "executable": false,
             "class": "document",
-            "mime": "text/plain"
+            "mime": "text/plain",
             "cozyMetadata": {
                 "doctypeVersion": "1",
                 "metadataVersion": 1,
--- a/docs/files.md
+++ b/docs/files.md
@@ -257,6 +257,8 @@ Upload a file
 | Metadata                | a JSON with metadata on this file (_deprecated_)               |
 | MetadataID              | the identifier of a metadata object                            |
 | CreatedAt               | the creation date of the file                                  |
+| SourceAccount           | the id of the source account used by a konnector               |
+| SourceAccountIdentifier | the unique identifier of the account targeted by the connector |
 
 #### HTTP headers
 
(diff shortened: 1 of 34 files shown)


nono:
[(𝐋) 🔵 🍊 Rajouter cozyMetadata aux io.cozy.files](https://trello.com/c/zaguxJ6U/1525-%F0%9D%90%8B-%F0%9F%94%B5-%F0%9F%8D%8A-rajouter-cozymetadata-aux-iocozyfiles)

model/vfs/couchdb_indexer.go
```diff
@@ -76,28 +76,40 @@ func (c *couchdbIndexer) DiskUsage() (int64, error) {
...
-func (c *couchdbIndexer) CreateFileDoc(doc *FileDoc) error {
+func (c *couchdbIndexer) prepareFileDoc(doc *FileDoc) error {
 	// Ensure that fullpath is filled because it's used in realtime/@events
 	if _, err := doc.Path(c); err != nil {
 		return err
 	}
+	// If a valid datetime is extracted from the EXIF metadata, use it as the
+	// created_at of the file. By valid, we mean that we filter out photos
+	// taken on camera were the clock was never configured (e.g. 1970-01-01).
+	if date, ok := doc.Metadata["datetime"].(time.Time); ok && date.Year() > 1990 {
```
Crash--: 👍

web/files/files_test.go
```diff
@@ -533,6 +541,22 @@ func TestUploadWithMetadata(t *testing.T) {
...
+	assert.NoError(t, err)
+	res, obj := doUploadOrMod(t, req, "text/plain", "rL0Y20zC+Fzt72VPzMSk2A==")
+	assert.Equal(t, 201, res.StatusCode)
+	data := obj["data"].(map[string]interface{})
+	attrs := data["attributes"].(map[string]interface{})
+	fcm := attrs["cozyMetadata"].(map[string]interface{})
+	assert.Equal(t, account, fcm["sourceAccount"])
+	assert.Equal(t, identifier, fcm["sourceAccountIdentifier"])
+}
+
```
Crash--: Is it possible to add a test when we create a file but the client already provided a few cozyMetaData ? 

Because we already provide a few cozyMetaData here (https://github.com/cozy/cozy-client/blob/046810a2143d5460d63b710ec543859b9d300056/packages/cozy-client/src/CozyClient.js#L304) and I don't know if we need to keep the createdAt attribute for a file for instance?

docs/files.md
```diff
@@ -679,6 +755,19 @@ Location: http://cozy.example.com/files/9152d568-7e7c-11e6-a377-37cbfb190b4b
             "executable": false,
             "class": "document",
             "mime": "text/plain"
```
maximemenager:
```suggestion
             "mime": "text/plain",
```

web/files/files_test.go
nono: The client cannot provide a few cozyMetadata when creating a file. There are only `sourceAccount` and `sourceAccountIdentifier` that can be given. The other fields are controlled by the stack.

For example, `cozyMetadata.createdAt` means the date/time of creation of the file on the server.

Crash--: OK thanks for the clarification !

PR #1935, opened 2019-06-17
Fork go-redis to have a stable API

nono:
Currently, github.com/go-redis/redis is in the work of a new version (v7), and it introduces some breaking changes. There are no known stable version, so let's fork it to have a stable API, at least until we have go modules or the new version is stable.


nono:
[Corriger le problème de déploiement de la stack suite aux modifications de go-redis](https://trello.com/c/u5sckttf/1557-corriger-le-probl%C3%A8me-de-d%C3%A9ploiement-de-la-stack-suite-aux-modifications-de-go-redis)

PR #1936, opened 2019-06-17
Move dynamic assets to Swift

maximemenager:


pkg/statik/fs/fs.go
```diff
@@ -243,14 +220,23 @@ func registerCustomExternal(cache Cache, opt AssetOption) error {
...
+	asset := newAsset(opt, zippedDataBuf.Bytes(), unzippedData)
+
+	objectName := path.Join(asset.Context, asset.Name)
+	swiftConn := config.GetSwiftConnection()
+
+	f, err := swiftConn.ObjectCreate(DynamicAssetsContainerName, objectName, true, "", "", nil)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
```
nono: We need to check if `f.Close()` has been successful or if an error has been returned.

pkg/config/dynamic/assets.go
```diff
@@ -1,95 +1,51 @@
...
-
-// ID implements the couchdb.Doc interface
-func (a *AssetsList) ID() string { return assetsListID }
-
-// Rev implements the couchdb.Doc interface
-func (a *AssetsList) Rev() string { return a.AssetsRev }
+	objs := map[string][]*fs.Asset{}
 
-// DocType implements the couchdb.Doc interface
-func (a *AssetsList) DocType() string { return consts.Configs }
```
nono: I think we should also clean `pkg/config` and remove `Configs` from there

pkg/assets/dynamic/dynamic.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+// RemoveAsset removes a dynamic asset from Swift
+func RemoveAsset(context, name string) error {
+	return assetFS.Remove(context, name)
+}
+
+// RegisterCustomExternals ensures that the assets are in the Swift, and load
+// them from their source if they are not yet available.
+func RegisterCustomExternals(opts []model.AssetOption, maxTryCount int) error {
```
nono: I think `RegisterCustomExternals` is always called with a single `AssetOption` in `opts`, and could be simplified to accept a `model.AssetOption` instead of a slice of them.

maximemenager: The CLI indeed restricts the dynamic asset insertion to one by one, but the API endpoint gives the option to insert more than one asset in a single request.
We can keep the slice for the API, or change the logic to restrict to a single insertion on the endpoint.

PR #1938, opened 2019-06-18
Add a zip worker to create zip archives

nono:
Th new zip worker can create a zip archive from files in the VFS, and put the archive in the VFS too.


nono:
[(𝐒) Avoir un worker pour créer des archives zip](https://trello.com/c/hYyJACQP/1558-%F0%9D%90%92-avoir-un-worker-pour-cr%C3%A9er-des-archives-zip)

PR #1939, opened 2019-06-19
Allow mobile apps to declare notifications

nono:
The stack requires that apps declare their notifications before sending them. If a mobile app is linked to a web app, the stack will use the declaration from the web app manifest for it.

### DIFF
--- a/docs/notifications.md
+++ b/docs/notifications.md
@@ -53,6 +53,13 @@ notification:
 }
 ```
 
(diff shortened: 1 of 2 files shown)


nono:
[Permettre aux apps mobiles liées d'envoyer des notifications](https://trello.com/c/7aopcQwe/1574-permettre-aux-apps-mobiles-li%C3%A9es-denvoyer-des-notifications)

PR #1946, opened 2019-07-01
Fix sharing from Drive Mobile

nono:
When a sharing is created from an OAuth app that is linked to a web app, the stack should use this app slug to create the link sent in the mail.


nono:
[Partage - depuis app mobile - vers login au lieu page publique](https://trello.com/c/maTkoZLG/1590-partage-depuis-app-mobile-vers-login-au-lieu-page-publique)

PR #1947, opened 2019-07-01
Fix infinite move directory

nono:
We had a case where a goroutine was looping while renaming a directory. The root cause was not found, but at least, we should avoid an infinite loop.


nono:
[Boucle infinie sur un renommage de répertoire](https://trello.com/c/5yCXeQKp/1591-boucle-infinie-sur-un-renommage-de-r%C3%A9pertoire)

PR #1955, opened 2019-07-03
Clean the old versions of the files

nono:
When a file is changed, a new version is created, and the stack also checks if old versions should be cleaned.

1. The versions that are tagged are always kept.
2. If two versions are too close in time, the latter is cleaned (15 minutes by default, configurable)
3. When there are too many versions for a file, the old versions are cleaned, except if they were tagged (20 versions by default, configurable).


nono:
[(𝐌) [VER] Règles de conservation](https://trello.com/c/Uxuejsuz/1410-%F0%9D%90%8C-ver-r%C3%A8gles-de-conservation)

PR #1957, opened 2019-07-04
Fix the UpdatedByApp function

nono:


nono:
[Bug sur les updatedByApps des files](https://trello.com/c/rwmRg163/1622-bug-sur-les-updatedbyapps-des-files)

PR #1961, opened 2019-07-08
Add the 2FA for OIDC

maximemenager:


.eslintrc.json
```diff
@@ -0,0 +1,3 @@
+{
+    "extends": ["cozy-app"]
```
nono: It would be nice to have documentation on how to use that. And can we add this linter to CI (in `scripts/lint.sh`)?

assets/scripts/twofactor_submit.js
```diff
@@ -0,0 +1,122 @@
...
+    const avatars = document.getElementsByClassName('c-avatar')
+    for (const avatar of avatars) {
+      avatar.classList.add('c-avatar--xlarge', 'u-mv-1')
+    }
+    const footers = document.getElementsByClassName('wizard-footer')
+    for (const footer of footers) {
+      footer.classList.remove('u-pb-half')
+      footer.classList.add('u-pb-2')
+    }
+  }
```
nono: I'd prefer if we can avoid duplicating those lines. Did you ask @GoOz if some responsive utilities CSS classes have been added to cozy-ui that would help to remove this hack?

assets/templates/twofactor.html
```diff
@@ -0,0 +1,86 @@
...
+            <h2 class="password-form wizard-subtitle u-coolGrey">{{.Domain}}</h2>
+            <p class="two-factor-form wizard-header-help" id="login-two-factor-passcode-tip">{{t "Login Two factor help"}}</p>
+            <input id="redirect" type="hidden" name="redirect" value="{{.Redirect}}" />
+            <input id="csrf_token" type="hidden" name="csrf_token" value="{{.CSRF}}" />
+            <input id="two-factor-token" type="hidden" name="two-factor-token" value="{{.TwoFactorToken}}" />
+            <div class="o-field u-m-0 two-factor-form">
+              <label for="two-factor-passcode" class="c-label" aria-describedby="login-two-factor-passcode-tip">{{t "Login Two factor field"}}</label>
+              <input id="two-factor-passcode" class="wizard-input c-input-text" name="two-factor-passcode" type="text" pattern="[0-9]*" inputmode="numeric" maxlength="6" {{if .TwoFactorForm}}autofocus {{end}}autocomplete="current-password" />
+            </div>
+            {{if and (not .OAuth) .TrustedDeviceCheckBox}}
```
nono: I don't like to have logic like that in templates. Can we move that to go code?

assets/scripts/submit.js
```diff
@@ -1,159 +1,30 @@
...
-    }).then(function(response) {
-      const loginSuccess = response.status < 400
-      response.json().then(function(body) {
-        if (loginSuccess) {
-          if (body.two_factor_token) {
-            renderTwoFactorForm(body.two_factor_token)
-            return
-          }
-          submitButton.childNodes[1].innerHTML = '<svg width="16" height="16"><use xlink:href="#fa-check"/></svg>'
-          submitButton.classList.add('c-btn--highlight')
```
nono: How did you preserve this animation?

assets/scripts/twofactor_submit.js
GoOz: Indeed there is for utilities.

for the avatar I would put `class="u-mv-0-m u-mv-1"`
- `u-mv-1` adds a vertical margin of 1 in any circumstances
- `u-mv-0-m` removes vertical margin for viewports with a width below 1024px

for the footer I would put `class="u-pb-half-m u-pb-2"`
- `u-pb-2` adds a padding-bottom of 2 in any circumstances
- `u-pb-half-m` adds a padding-bottom of half a unit for viewports with a width below 1024px

web/auth/twofactor.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	if wantsJSON {
+		result := echo.Map{"redirect": redirect.String()}
+		if len(generatedTrustedDeviceToken) > 0 {
+			result["two_factor_trusted_device_token"] = string(generatedTrustedDeviceToken)
+		}
+		return c.JSON(http.StatusOK, result)
+	}
+
+	redirect = AddCodeToRedirect(redirect, inst.ContextualDomain(), sessionID)
```
nono: Can you remember me what is the purpose of `AddCodeToRedirect`?

web/auth/twofactor.go
```diff
@@ -0,0 +1,147 @@
...
+		if b, err := strconv.ParseBool(trustedDeviceCheckBoxParam); err == nil {
+			trustedDeviceCheckBox = b
+		}
+	}
+
+	return renderTwoFactorForm(c, inst, http.StatusOK, credsError, redirect, twoFactorToken, longRunSession, trustedDeviceCheckBox)
+}
+
+// twoFactor handles a the twoFactor POST request
+func twoFactor(c echo.Context) error {
```
nono: This function is a bit too long to my taste. Can you extract an helper? Maybe a func to handle the 2fa failure

web/oidc/oidc.go
```diff
@@ -94,6 +94,19 @@ func Login(c echo.Context) error {
...
+	if inst.HasAuthMode(instance.TwoFactorMail) {
+		twoFactorToken, err := lifecycle.SendTwoFactorPasscode(inst)
+		if err != nil {
+			return err
+		}
+		v := url.Values{}
+		v.Add("two_factor_token", string(twoFactorToken))
+		v.Add("trusted_device_checkbox", "false")
+		// We can not cleanly check the trusted_device option for external
+		// login. Therefore, we do not provide the checkbox
```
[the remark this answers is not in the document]
maximemenager: The trusted device checkbox allows a user to trust a device and skip the 2FA authentication. 

This is done by storing a token in the `localStorage` of the user's browser, and reading/appending it during the passphrase login form.

In the case of an external login (like OIDC for example), we cannot easily get this token because the first rendered stack page will be the 2FA page (if activated).

Even if we decide to store & check the token on the 2FA page, the user will anyway get an email with the 2FA code as we can only sent it during the OIDC Login function for security reasons.

web/auth/twofactor.go
maximemenager: Of what I understand, a generated `code` is added to the redirect. Its purpose is to be exchanged between main domain and subdomains (like apps) with a cookie.

This process allows to "transfer" a cookie for an instance configured with flat subdomains

assets/scripts/submit.js
maximemenager: It was unfortunately not preserved. I can however re-add it by going back and check the login & redirect using the JS code instead of the html form

.travis.yml
```diff
@@ -22,6 +22,9 @@ matrix:
 
 before_install:
   - docker run -d -p 5984:5984 --net=host --name couch apache/couchdb:2.3
+  - curl -sL https://deb.nodesource.com/setup_8.x | sudo bash -
+  - sudo apt install nodejs
```
nono: Can we use `nvm` and `node 8.12.0` as https://docs.travis-ci.com/user/reference/xenial/#javascript-and-nodejs-support says that they are installed by default?

.travis.yml
```diff
@@ -22,6 +22,9 @@ matrix:
...
+  - sudo npm install -g eslint@5.16.0 prettier eslint-plugin-prettier eslint-config-cozy-app
```
nono: Can we move this line to `scripts/lint.sh` (without `sudo`), as it is only for lint and not for the other builds on travis?

assets/scripts/submit.js
```diff
@@ -37,137 +42,89 @@
...
+      event.preventDefault()
+      const label = window.password.getStrength(passphraseInput.value).label
+      if (label == 'weak') {
+        return false
+      } else {
+        resetForm.submit()
+      }
+    })
+
+  resetForm &&
```
nono: s/resetForm/passphraseInput/ ?

assets/scripts/tracking-onboarding.js
```diff
@@ -1,13 +1,9 @@
...
-  }
-})(window, document);
+;(function(w, d) {
+  const data = d.currentScript.dataset
+  const tracker = w.Piwik.getTracker(data.matomoUrl, data.matomoSiteId)
+  tracker.enableHeartBeatTimer()
+  tracker.setUserId(w.location.hostname)
+  tracker.setCustomDimension(data.matomoAppId, 'Onboarding')
+  tracker.setCustomUrl('/password')
+  tracker.trackPageView()
```
nono: I'd prefer to keep the try/catch to avoid showing an error in the console when piwik is not configured

web/auth/auth.go
```diff
@@ -530,4 +469,8 @@ func Routes(router *echo.Group) {
 
 	router.POST("/access_token", accessToken)
 	router.POST("/secret_exchange", secretExchange)
+
+	// 2FA
+	router.GET("/twofactor", twoFactorForm, noCSRF)
+	router.POST("/twofactor", twoFactor, noCSRF)
```
nono: What is the rationale for using the `noCSRF` middleware?

maximemenager: It is a mistake, nice pick !

assets/scripts/submit.js
maximemenager: IIRC, this event listener is only used for reset passphrase page, I don't think we need it in the long form

(https://github.com/cozy/cozy-stack/blob/master/assets/scripts/submit.js#L167)

assets/scripts/tracking-onboarding.js
```diff
@@ -1,13 +1,13 @@
...
-  } catch (err) {
-    console.log('Cannot initialize Matomo:', err);
+    const data = d.currentScript.dataset
+    const tracker = w.Piwik.getTracker(data.matomoUrl, data.matomoSiteId)
+    tracker.enableHeartBeatTimer()
+    tracker.setUserId(w.location.hostname)
+    tracker.setCustomDimension(data.matomoAppId, 'Onboarding')
+    tracker.setCustomUrl('/password')
+    tracker.trackPageView()
+  } catch {
```
nono: I think the optional catch binding (`catch {` instead of `catch (e) {`) is not supported by Edge.

web/auth/twofactor.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	twoFactorTokenParam := c.QueryParams().Get("two_factor_token")
+	if twoFactorTokenParam == "" {
+		return c.JSON(http.StatusBadRequest, "Missing twoFactorToken")
+	}
+
+	twoFactorToken := []byte(twoFactorTokenParam)
+
+	longRunSession, err := strconv.ParseBool(c.FormValue("long-run-session"))
```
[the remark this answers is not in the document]
nono: @maximemenager did you see my question here?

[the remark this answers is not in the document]
maximemenager: Sorry, my mistake

PR #1963, opened 2019-07-08
Fix dynamic asset when loaded without /assets/ext prefix

nono:


nono:
[(𝐒) 🔵 🦌 [BAR] - ETUR le logo principal est customisé](https://trello.com/c/s9TJawGt/1521-%F0%9D%90%92-%F0%9F%94%B5-%F0%9F%A6%8C-bar-etur-le-logo-principal-est-customis%C3%A9)

PR #1965, opened 2019-07-09
Fix a panic on Swift container deletion

nono:


nono:
[Erreur à la destruction d'une instance](https://trello.com/c/AnQTz8vn/1626-erreur-%C3%A0-la-destruction-dune-instance)

PR #1976, opened 2019-07-16
Log errors on thumbnail migration

nono:
When an instance is migrated to the swift layout v3, the migration worker won't return an error if some thumbnails can't be copied. It will just log those errors.


nono:
[(𝐌) Script de migration Swift du layout v2 vers v3](https://trello.com/c/WCoSTfJ6/1620-%F0%9D%90%8C-script-de-migration-swift-du-layout-v2-vers-v3)

PR #1979, opened 2019-07-16
Make cozy-stack instances ls faster

nono:
Listing instances is done by fetching the instance information from CouchDB with pagination. By fetching 100 instances per page, we are making a lot of calls, and each call has a fixed latency that makes the
whole slow. We want to keep to pagination to avoid timeouts, but we can increase the number of items per page to 10000 to make the listing faster.


nono:
[ajouter un endpoint d'admin et cli associée instances fast-ls](https://trello.com/c/VtZhXSXm/1603-ajouter-un-endpoint-dadmin-et-cli-associ%C3%A9e-instances-fast-ls)

PR #1980, opened 2019-07-16
Use "ready" status instead of "installed" for apps

maximemenager:


web/auth/oauth.go
```diff
@@ -385,74 +385,6 @@ func authorizeSharing(c echo.Context) error {
...
-		return c.Redirect(http.StatusSeeOther, u)
-	}
-
-	app, ok, err := getApp(c, instance, c.QueryParam("slug"))
-	if !ok || err != nil {
-		return err
-	}
-
-	permissions := app.Permissions()
-	return c.Render(http.StatusOK, "authorize_app.html", echo.Map{
```
nono: Can we also delete this template?

nono: And maybe also some entries in the po files for the locales

PR #1981, opened 2019-07-16
Use a temporary fork for cron

nono:
The cron library is moving target now. We will switch back to the upstream soon, with go modules. But, for now, let's use a fork to avoid confusion.


nono:
[(𝐌) [GO-MOD] Passer sur les dépendances forkées](https://trello.com/c/JJY6Sk5r/1633-%F0%9D%90%8C-go-mod-passer-sur-les-d%C3%A9pendances-fork%C3%A9es)

PR #1982, opened 2019-07-17
Help debug the migration to swift v3

nono:


nono:
[(𝐌) Script de migration Swift du layout v1 vers v3](https://trello.com/c/R3Lj5Z9N/1619-%F0%9D%90%8C-script-de-migration-swift-du-layout-v1-vers-v3)

PR #1984, opened 2019-07-17
Remove the go-git dependency

nono:
At the beginning, we were using the go-git library to install applications from the git repository. It was not ideal as this library was not capable to handle the tricky cases like the git binary (using something else that master for the default branch for example). So, we moved to use the git binary by default, and with a fallback to go-git when the binary was not available.

Now, most applications are installed from the registry and don't need git at all, and for the developers, we can expect that they have a git binary executable from the stack. So, there is no need to keep to go-git library, and one less library is good thing: less weight, less maintenance.

### DIFF
--- a/model/app/fetcher_git.go
+++ b/model/app/fetcher_git.go
@@ -19,15 +19,9 @@ import (
 
 	"github.com/cozy/afero"
 	"github.com/cozy/cozy-stack/pkg/appfs"
-	git "github.com/cozy/go-git"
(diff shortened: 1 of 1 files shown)


nono:
[(𝐌) [GO-MOD] Passer sur les dépendances forkées](https://trello.com/c/JJY6Sk5r/1633-%F0%9D%90%8C-go-mod-passer-sur-les-d%C3%A9pendances-fork%C3%A9es)

PR #1989, opened 2019-07-18
Fix self-hosted release with go modules

nono:


nono:
[(𝐋) [GO-MOD] Mise en place des go-modules](https://trello.com/c/fEDbELWp/1634-%F0%9D%90%8B-go-mod-mise-en-place-des-go-modules)

PR #1991, opened 2019-07-23
Add pagination for /apps route

maximemenager:
This PR adds a pagination for /apps route. The legacy limit is kept, but a custom limit can now be specified if needed

### DIFF
--- a/model/app/konnector.go
+++ b/model/app/konnector.go
@@ -282,7 +282,6 @@ func GetKonnectorBySlugAndUpdate(db prefixer.Prefixer, slug string, copier appfs
 
 // ListKonnectors returns the list of installed konnectors applications.
 //
-// TODO: pagination
 func ListKonnectors(db prefixer.Prefixer) ([]Manifest, error) {
 	var docs []*KonnManifest
 	req := &couchdb.AllDocsRequest{Limit: 100}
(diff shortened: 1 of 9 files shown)


web/apps/apps.go
```diff
@@ -301,7 +301,22 @@ func listWebappsHandler(c echo.Context) error {
 	if err := middlewares.AllowWholeType(c, permission.GET, consts.Apps); err != nil {
 		return err
 	}
-	docs, err := apps.ListWebapps(instance)
+
+	// Adding the startKey if it is given in the request
+	var startKey string
+	if sk := c.QueryParam("start_key"); sk != "" {
```
nono: Can you update the documentation?

model/app/konnector.go
```diff
@@ -282,7 +282,6 @@ func GetKonnectorBySlugAndUpdate(db prefixer.Prefixer, slug string, copier appfs
 
 // ListKonnectors returns the list of installed konnectors applications.
 //
-// TODO: pagination
 func ListKonnectors(db prefixer.Prefixer) ([]Manifest, error) {
```
nono: Why did you keep this function?

model/app/konnector.go
```diff
@@ -297,4 +296,32 @@ func ListKonnectors(db prefixer.Prefixer) ([]Manifest, error) {
...
+		Limit:    limit + 1, // Also get the following document for the next key
+		StartKey: startKey,
+	}
+	err := couchdb.GetAllDocs(db, consts.Konnectors, req, &docs)
+	if err != nil {
+		return nil, "", err
+	}
+
+	nextID := ""
+	if len(docs) > 0 && len(docs) == limit+1 { // There is still documents to fetch
```
nono: s/There is/There are/

model/app/konnector.go
maximemenager: I was keeping it for the other uses (like intents), but with the same default limit (100), we can safely remove it without risking side-effects

web/apps/apps.go
```diff
@@ -301,7 +301,22 @@ func listWebappsHandler(c echo.Context) error {
...
+		startKey = sk
+	}
```
nono: Why not just `startKey := c.QueryParam("start_key")`?

web/apps/apps.go
```diff
@@ -320,15 +342,50 @@ func listKonnectorsHandler(c echo.Context) error {
...
+	// Generating links list for the next konnectors
+	links := &jsonapi.LinksList{}
+	if next != "" { // Do not generate the next URL if there are no next konnectors
+		links.Next = generateNextURL(c, limit, next)
+	}
+
+	return jsonapi.DataList(c, http.StatusOK, objs, links)
+}
+
+func generateNextURL(c echo.Context, limit int, nextID string) string {
```
nono: Can we move in this function the logic for?

```
	if next != "" { // Do not generate the next URL if there are no next konnectors
		links.Next = generateNextURL(c, limit, next)
	}
```

PR #1994, opened 2019-07-25
Use the swift container for assets for /status

nono:
The /status endpoint checks the connection to swift, among other things.
Before this commit, it was done via /info, but the Swift Restful API
implemeted by Ceph does not expose this endpoint. We will query the
container used by the stack for the dynamic assets for /status now, to
fix the compatibility with Ceph.


sblaisot:
Top 👍

PR #2004, opened 2019-07-30
Fix real-time and triggers on replication

nono:
When using the Pouch->Couch replication to push changes to the Cozy, we have to set the doctype in the stack to make the real-time and the triggers work correctly.


nono:
[(𝐌) Gérer le temps-réel et les triggers sur les routes de réplication couch](https://trello.com/c/VcDrV3Ob/1645-%F0%9D%90%8C-g%C3%A9rer-le-temps-r%C3%A9el-et-les-triggers-sur-les-routes-de-r%C3%A9plication-couch)

PR #2006, opened 2019-07-30
Create API endpoints for CLI

maximemenager:
This PR adds missings API endpoints for CLI. The CLI relies on the API, improving security by avoiding a config deployment where it is not necessary.


cmd/fixer.go
```diff
@@ -302,198 +182,6 @@ var thumbnailsFixer = &cobra.Command{
 	},
 }
 
-var contactEmailsFixer = &cobra.Command{
```
nono: I'd like to keep this fixer. It may be useful in the future

cmd/fixer.go
```diff
@@ -503,102 +191,25 @@ var contentMismatch64Kfixer = &cobra.Command{
...
-			fmt.Println("This is a dry-run, no file will be altered")
+		q := url.Values{
+			"dry_run": {strconv.FormatBool(!noDryRunFlag)},
 		}
 
 		c := newAdminClient()
 		res, err := c.Req(&request.Options{
-			Method: "GET",
-			Path:   "/instances/" + url.PathEscape(domain) + "/fsck",
+			Method:  "GET",
```
nono: It should be a POST and not a GET if the stack updates CouchDB or Swift

cmd/fixer.go
```diff
@@ -503,102 +191,25 @@ var contentMismatch64Kfixer = &cobra.Command{
...
+			Path:    "/instances/" + url.PathEscape(domain) + "/content-mismatch-fixer",
```
nono: I'd prefer to have a namespace for fixers: `"/instances/" + url.PathEscape(domain) + "/fixers/content-mismatch"`

cmd/instances.go
```diff
@@ -923,30 +921,30 @@ var instanceAppVersionCmd = &cobra.Command{
 			return cmd.Usage()
 		}
 
-		instances, err := instance.List()
+		c := newAdminClient()
+		path := fmt.Sprintf("/instances/show-app-version/%s/%s", args[0], args[1])
```
nono: We should avoid verbs in path: `/instances/with-app-version/:slug/:version`

cmd/swift.go
```diff
@@ -21,83 +19,34 @@ var flagShowDomains bool
...
 	RunE: func(cmd *cobra.Command, args []string) error {
-		type layout struct {
-			Counter int      `json:"counter"`
-			Domains []string `json:"domains,omitempty"`
+		c := newAdminClient()
+		values := url.Values{}
+		values.Add("show_domains", strconv.FormatBool(flagShowDomains))
+		res, err := c.Req(&request.Options{
+			Method:  "GET",
+			Path:    "/swift/list-layouts",
```
nono: `/swift/layouts`

docs/instance.md
```diff
@@ -84,6 +84,72 @@ Several pre-included reasons are included, but you can define your owns if neede
 - `"PAYMENT_FAILED"`: The instance waits for a payment to be done
 - `"UNKNOWN"`: The instance was blocked for an unknown reason
 
-## Modifying 
+## Modifying
 
 If you want to edit an instance, see [`instances modify manpage`](cli/cozy-stack_instances_modify.md).
+
+## Admin endpoints
```
nono: I'd prefer to have a file `docs/admin.md` with the documentation for endpoints exposed on the admin port (+ a short introduction)

web/instances/instances.go
```diff
@@ -573,6 +576,175 @@ func lsContexts(c echo.Context) error {
...
+	}
+
+	dryRun := true
+	if dryParam := c.QueryParam("dry_run"); dryParam != "" {
+		if dry, err := strconv.ParseBool(dryParam); err == nil {
+			dryRun = dry
+		}
+	}
+
+	// Directly requesting the /fsck endpoint by forging a custom echo.Context
```
nono: Yikes, it seems weird to do that. Can't we just call the right method of the VFS?

web/instances/instances.go
```diff
@@ -573,6 +576,175 @@ func lsContexts(c echo.Context) error {
...
+		Instances []string `json:"instances"`
+	}{
+		instancesAppVersion,
+	}
+
+	return c.JSON(http.StatusOK, i)
+}
+
+// contentMismatchFixer fixes the 64k bug
+func contentMismatchFixer(c echo.Context) error {
```
nono: This function is really long. Can you split it?

web/instances/instances.go
nono: And it would be nice to have the code for the fixers in their own file: `web/instances/fixers.go`

docs/admin.md
```diff
@@ -0,0 +1,239 @@
+# Admin
+
+## Introduction
+
+An admin API is avaible on the stack. It offers several endpoints to interact
```
nono: s/avaible/available/

docs/admin.md
```diff
@@ -0,0 +1,239 @@
...
+with your cozy-stack installation (E.g. interacting with instances, generating tokens, ...).
+
+:warning: Use the admin API only if you know what you are doing. The admin API
+is not authenticated, you **must** protect these endpoints.
```
nono: Well, the stack has basic auth for the admin API, so I wouldn't say that it is not authenticated (even if it remains true that these endpoints must be protected as they are very powerful).

web/instances/fixers.go
```diff
@@ -0,0 +1,201 @@
...
+		return fmt.Errorf("Cannot find instance %s", domain)
+	}
+
+	body := struct {
+		DryRun bool `json:"dry_run"`
+	}{
+		DryRun: true,
+	}
+
+	// Try to get the dru_run param from the body. If there is no body, ignore
```
nono: s/dru_run/dry_run/

web/swift/swift.go
```diff
@@ -0,0 +1,239 @@
...
+	_, err = sc.ObjectGet(swiftContainer(i), unescaped, buf, false, nil)
+	if err != nil {
+		return err
+	}
+
+	out := struct {
+		Content string `json:"content"`
+	}{
+		buf.String(),
+	}
```
nono: I'd prefer to avoid JSON serialization here. If we are trying to get an image or a PDF, it would be weird.

web/swift/swift.go
```diff
@@ -0,0 +1,239 @@
...
+}
+
+// PutObject puts an object into Swift
+func PutObject(c echo.Context) error {
+	type reqStruct struct {
+		Instance    string `json:"instance"`
+		ObjectName  string `json:"object_name"`
+		Content     string `json:"content"`
+		ContentType string `json:"content_type"`
+	}
```
nono: Idem, I'd prefer not to use JSON here. The body of the request should be the content of the file, and nothing else. For the content-type, an HTTP header is the right way to send it. For the instance, it can be the `Host` HTTP header. For the ObjectName, it can be in the path or in the query string.

web/swift/swift.go
```diff
@@ -0,0 +1,239 @@
...
+	return c.JSON(http.StatusOK, out)
+}
+
+// Routes sets the routing for the status service
+func Routes(router *echo.Group) {
+	router.GET("/layouts", ListLayouts, checkSwift)
+	router.GET("/get/:domain/:object", GetObject, checkSwift)
+	router.POST("/put", PutObject, checkSwift)
+	router.DELETE("/:domain/:object", DeleteObject, checkSwift)
+	router.GET("/ls/:domain", ListObjects, checkSwift)
```
nono: - I would prefer to use the `Host` HTTP header to give the domain
- I don't like verbs in the paths (the only HTTP verbs are GET/POST/PUT/etc. and are not in the paths)
- I want to be able in the future to add some routes for debugging the global swift containers (apps, dynamic assets)

So, I suggest to use this routes:

- GET /swift/layouts (like it is already the case)
- GET /swift/vfs/:object to get an object
- PUT /swift/vfs/:object to put an object
- DELETE /swift/vfs/:object to delete an object
- GET /swift/vfs to list all the objects of an instance VFS

web/swift/swift.go
```diff
@@ -0,0 +1,239 @@
...
+	if err != nil {
+		return err
+	}
+
+	i, err := lifecycle.GetInstance(domain)
+	if err != nil {
+		return err
+	}
+
+	fmt.Println("test", unescaped)
```
nono: Debug spotted

cmd/fixer.go
```diff
@@ -717,15 +479,12 @@ func init() {
...
 	fixerCmdGroup.AddCommand(jobsFixer)
 	fixerCmdGroup.AddCommand(md5FixerCmd)
 	fixerCmdGroup.AddCommand(mimeFixerCmd)
-	fixerCmdGroup.AddCommand(onboardingsFixer)
 	fixerCmdGroup.AddCommand(redisFixer)
 	fixerCmdGroup.AddCommand(thumbnailsFixer)
 	fixerCmdGroup.AddCommand(contactEmailsFixer)
-	fixerCmdGroup.AddCommand(linkedAppFixer)
 	fixerCmdGroup.AddCommand(contentMismatch64Kfixer)
 	fixerCmdGroup.AddCommand(orphanAccountFixer)
```
nono: You also have the new orphanAccountFixer to move to the admin endpoint

PR #2007, opened 2019-07-31
Improve the CLI for debug mode

nono:


sblaisot:
😻

cmd/instances.go
```diff
@@ -355,27 +355,54 @@ instance of the given domain. Set the quota to 0 to remove the quota.
 }
 
 var debugInstanceCmd = &cobra.Command{
-	Use:   "debug <domain> <true/false>",
+	Use:   "debug <true/false>",
```
sblaisot: Dors it break current behaviour ? Maybe you need to advertise this to the team ?

nono: No, it doesn't break. I have kept compatibility with the old way: it detects the number of arguments to know if the domain is a positional argument or a flag.

sblaisot: :+1:

sblaisot: :+1:

PR #2027, opened 2019-08-19
fix: Pass the correct _rev to account deletion job

ptbrowne:
The job for deletion needs the _rev just before the deletion to correctly work.


cmd/fixer.go
```diff
@@ -686,10 +686,11 @@ var orphanAccountFixer = &cobra.Command{
 					continue
 				}
 				acc.ManualCleaning = true
+				oldRev := acc.Rev() // The deletion job need the rev just before the deletion
```
nono: [nit] s/need/needs/

PR #2055, opened 2019-09-03
Fix renaming a file from NFC to NFD

nono:
When a file is renamed, its name is different, but it can be very subtle. For example, it can have the same unicode meaning but just with two different normalizations, like NFC and NFD. CouchDB does a unicode normalization when using strings in mango selector, so we had to make the selector more robust for this case to avoid a loop where the same sub-directories where renamed.

### DIFF
--- a/model/vfs/couchdb_indexer.go
+++ b/model/vfs/couchdb_indexer.go
@@ -226,10 +226,19 @@ func (c *couchdbIndexer) moveDir(oldpath, newpath string) error {
 		return nil
 	}
 
-	// We limit the stack to 1024 bulk updates to avoid infinite loops, as we
+	// We limit the stack to 128 bulk updates to avoid infinite loops, as we
 	// had a case in the past.
-	for i := 0; i < 1024; i++ {
-		sel := mango.StartWith("path", oldpath+"/")
+	start := oldpath + "/"
+	for i := 0; i < 128; i++ {
+		// The simple selector mango.StartWith can have some issues when
+		// renaming a folder to the same name, but with a different unicode
+		// normalization (like NFC to NFD). In that case, CouchDB would always
+		// return the same documents with this selector, as it does the
+		// comparison on a normalized string.
+		sel := mango.And(
+			mango.Gt("path", start),
+			mango.Lt("path", oldpath+"0"), // 0 is the next ascii character after /
+		)
 		req := &couchdb.FindRequest{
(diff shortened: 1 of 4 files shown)


model/vfs/couchdb_indexer.go
```diff
@@ -226,10 +226,19 @@ func (c *couchdbIndexer) moveDir(oldpath, newpath string) error {
...
+	start := oldpath + "/"
+	for i := 0; i < 128; i++ {
+		// The simple selector mango.StartWith can have some issues when
+		// renaming a folder to the same name, but with a different unicode
+		// normalization (like NFC to NFD). In that case, CouchDB would always
+		// return the same documents with this selector, as it does the
+		// comparison on a normalized string.
+		sel := mango.And(
+			mango.Gt("path", start),
+			mango.Lt("path", oldpath+"0"), // 0 is the next ascii character after /
```
taratatach: This could be assigned to an `end` variable next to `start`, don't you think?

taratatach: What happens if we have the following directories:
- `folder1/`
- `folder10/`

Will `folder10/` be selected?

tests/integration/tests/filepath_encoding.rb
```diff
@@ -0,0 +1,52 @@
...
+    # Moving inside its-self
+    folder.move_to inst, folder.couch_id rescue nil
+    # Moving inside a sub-directory
+    folder.move_to inst, sub.couch_id rescue nil
+    after = File.join Helpers.current_dir, inst.domain
+    diff = Helpers.fsdiff before, after
+    diff.must_be_empty
+    assert_equal inst.fsck, ""
+
+    # Play with NFC and NFS unicode normalization
```
taratatach: Typo: `NFS` → `NFD`

model/vfs/couchdb_indexer.go
nono: It won't be selected: `Lt` means `<`, it is Lte for `<=`.

nono: Why do you want to use a variable for `end`?

taratatach: OK, it makes sense for `Lt`.
I find the `path+"0"` value a bit magical and thought it would deserve to be documented apart from where it's used. And besides, the `start` variable was extracted but I see that's because you're re-using it later.

It can stay this way if you prefer.

PR #2077, opened 2019-09-17
Update english typography

sblaisot:
Enforce english typography in documentation.
See https://everything2.com/title/rules+of+typography


nono:
Thanks!

PR #2090, opened 2019-09-26
Add a hub for bitwarden clients

nono:
It allows the client to listen with a websocket, and the server can send them notifications about ciphers and folders.

### DIFF
--- a/web/bitwarden/bitwarden.go
+++ b/web/bitwarden/bitwarden.go
@@ -44,11 +44,11 @@ func GetProfile(c echo.Context) error {
 			"error": "invalid token",
 		})
 	}
-	settings, err := settings.Get(inst)
+	setting, err := settings.Get(inst)
 	if err != nil {
 		return err
 	}
-	profile, err := newProfileResponse(inst, settings)
+	profile, err := newProfileResponse(inst, setting)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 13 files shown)


web/bitwarden/hub.go
```diff
@@ -216,6 +264,75 @@ func buildNotification(e *realtime.Event, userID string) *notification {
 	return &msg
 }
 
+func buildFolderPayload(e *realtime.Event, userID string) map[string]interface{} {
+	var updatedAt interface{}
+	var date string
+	if doc, ok := e.Doc.(*couchdb.JSONDoc); ok {
```
taratatach: Realtime events can contain documents in another format than `realtime.JSONDoc`?

web/bitwarden/bitwarden.go
```diff
@@ -44,11 +44,11 @@ func GetProfile(c echo.Context) error {
 			"error": "invalid token",
 		})
 	}
-	settings, err := settings.Get(inst)
+	setting, err := settings.Get(inst)
```
taratatach: How come the linter didn't error out on this reassignment of `settings`?

web/bitwarden/hub.go
nono: Yes, it's a bit of technical debt in the realtime, but it's hard to fix. The documents in realtime must match the `realtime.Doc` interface, and, in practice, it depends on how the update was made, and if we are using memory (self-hosted and development image) or redis (production with several cozy-stack servers).

taratatach: If we have an interface then could we use functions to access those attributes in a standardized fashion?

web/bitwarden/bitwarden.go
nono: I don't see a linter that checks that on https://github.com/golangci/awesome-go-linters

web/bitwarden/hub.go
nono: Well, I don't see a convenient way to do that

PR #2101, opened 2019-10-02
Update assets command examples

y-lohse:
The examples are still using the deprecated version of the command. 

Our documentation 5isn't up to date either](https://docs.cozy.io/en/cozy-stack/cli/cozy-stack/), but 5looking at the source](https://github.com/cozy/cozy-stack/blob/master/docs/cli/cozy-stack.md) it should be..?

### DIFF
--- a/cmd/assets.go
+++ b/cmd/assets.go
@@ -66,7 +66,7 @@ var rmAssetCmd = &cobra.Command{
 	Aliases: []string{"remove"},
 	Short:   "Removes an asset",
 	Long:    "Removes a custom asset in a specific context",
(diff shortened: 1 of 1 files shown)


nono:
> Our documentation 5isn't up to date either](https://docs.cozy.io/en/cozy-stack/cli/cozy-stack/), but 5looking at the source](https://github.com/cozy/cozy-stack/blob/master/docs/cli/cozy-stack.md) it should be..?

The change is very recent, and `docs.cozy.io` has not been rebuilt/deployed since then. It will be updated soon, at the next rebuild.

PR #2103, opened 2019-10-02
Fix typo in completion doc

sblaisot:


nono:
Can you also fix https://github.com/cozy/cozy-stack/blob/master/cmd/completion.go#L17?

sblaisot:
> 
> 
> Can you also fix https://github.com/cozy/cozy-stack/blob/master/cmd/completion.go#L17?

Done

PR #2105, opened 2019-10-04
Make clearing the trash faster

nono:


model/vfs/vfsswift/impl_v2.go
```diff
@@ -399,6 +383,23 @@ func (sfs *swiftVFSV2) destroyFileVersions(objName string) error {
 	return nil
 }
 
+func (sfs *swiftVFSV2) EnsureErased(journal vfs.TrashJournal) error {
+	// No lock needed
+	_, err := sfs.c.BulkDelete(sfs.container, journal.ObjectNames)
+	if err == swift.Forbidden {
+		sfs.log.Infof("EnsureErased failed on BulkDelete: %s", err)
```
sblaisot: Info is probably too low for a delete error implying an inconsistency between couch and swift.

model/vfs/vfsswift/impl_v3.go
```diff
@@ -355,6 +348,23 @@ func (sfs *swiftVFSV3) DestroyFile(doc *vfs.FileDoc) error {
 	return err
 }
 
+func (sfs *swiftVFSV3) EnsureErased(journal vfs.TrashJournal) error {
+	// No lock needed
+	_, err := sfs.c.BulkDelete(sfs.container, journal.ObjectNames)
+	if err == swift.Forbidden {
+		sfs.log.Infof("EnsureErased failed on BulkDelete: %s", err)
```
sblaisot: same here, info is probably too low

worker/trash/trash.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/cozy/cozy-stack/model/job"
+	"github.com/cozy/cozy-stack/model/vfs"
+)
+
+func init() {
+	job.AddWorker(&job.WorkerConfig{
+		WorkerType:   "trash-files",
+		Concurrency:  runtime.NumCPU(),
+		MaxExecCount: 2,
+		Timeout:      10 * time.Minute,
```
sblaisot: isn't 10 minutes too loow if there was **a lot** of files in the trash ?

docs/workers.md
```diff
@@ -269,6 +269,13 @@ permission. It is done by adding this to the manifest:
 }
 ```
 
+## trash-files worker
+
+This worker is used only by the stack: when the user asks to clean the trash,
+the stack will delete the files from CouchDB and put a job for this worker. The
+deletion of files in Swift is slow, and can be done via this worker
+asynchronously.
+
```
sblaisot: Maybe you can explain here the differences between afero/swiftv1 (no worker) and swift v2/v3 (with worker) for later reference

web/jobs/jobs.go
```diff
@@ -612,3 +618,15 @@ func wrapJobsError(err error) error {
 	}
 	return err
 }
+
+// isReservedWorker returns true if the worker should only by used by the
+// stack, and the clients must not push jobs for it.
+func isReservedWorker(worker string) bool {
+	switch worker {
+	case "migrations", "thumbnail", "trash-files", "updates",
+		"share-track", "share-replicate", "share-upload", "":
```
sblaisot: I think it would be better to mark the worker as "reserved" in the definition/code of the worker itself (in a boolean property by example) rather than in a separate hardcoded blacklist.

model/instance/lifecycle/thumbs.go
```diff
@@ -0,0 +1,41 @@
...
+	case config.SchemeSwift, config.SchemeSwiftSecure:
+		switch i.SwiftLayout {
+		case 0:
+			return vfsswift.NewThumbsFs(config.GetSwiftConnection(), i.Domain)
+		case 1:
+			return vfsswift.NewThumbsFsV2(config.GetSwiftConnection(), i)
+		case 2:
+			return vfsswift.NewThumbsFsV3(config.GetSwiftConnection(), i)
+		default:
+			panic(instance.ErrInvalidSwiftLayout)
```
sblaisot: doesn't panic terminate the program ? a bit violent it seems.

model/instance/lifecycle/thumbs.go
```diff
@@ -0,0 +1,41 @@
...
+		}
+	default:
+		panic(fmt.Sprintf("instance: unknown storage provider %s", fsURL.Scheme))
```
sblaisot: same here, this should have been determined at launch, when reading hte config file, not when trying to create a container for thumbnails.
This should IMHO be a loggued error, not a panic.

web/files/files.go
```diff
@@ -707,7 +708,7 @@ func ThumbnailHandler(c echo.Context) error {
 		return jsonapi.NewError(http.StatusBadRequest, "Wrong download token")
 	}
 
-	fs := instance.ThumbsFS()
+	fs := lifecycle.ThumbsFS(instance)
```
sblaisot: related to previous comments, should be `fs, err := ...` to handle error when constructing ThumbsFS name.

worker/trash/trash.go
```diff
@@ -0,0 +1,32 @@
...
+	"time"
+
+	"github.com/cozy/cozy-stack/model/job"
+	"github.com/cozy/cozy-stack/model/vfs"
+)
+
+func init() {
+	job.AddWorker(&job.WorkerConfig{
+		WorkerType:   "trash-files",
+		Concurrency:  runtime.NumCPU(),
```
sblaisot: This worker is reponsible for deleting files in swift. It means that it mainly waits swift's response to HTTP requests and doesn't consume much CPU nor RAM, so concurrency could be increased up to 3 or 5 * `NumCPU()` I think

model/vfs/vfsafero/impl.go
```diff
@@ -237,7 +238,7 @@ func (afs *aferoVFS) CreateFile(newdoc, olddoc *vfs.FileDoc) (vfs.File, error) {
 	}, nil
 }
 
-func (afs *aferoVFS) DestroyDirContent(doc *vfs.DirDoc) error {
+func (afs *aferoVFS) DestroyDirContent(doc *vfs.DirDoc, push func(vfs.TrashJournal) error) error {
```
taratatach: We could use `_` instead of `push` here to show it's unused.

edas:
Toujours ok pour moi avec les derniers commits. En même temps je me rends compte que je n'aurais pas identifié le problème des suppressions de plus de 256 items à la relecture donc ma validation ne suffit probablement pas

nono:
> En même temps je me rends compte que je n'aurais pas identifié le problème des suppressions de plus de 256 items à la relecture

Oui, enfin le problème date d'avant la PR et je ne l'ai pas vu non plus quand ça a été introduit.

PR #2123, opened 2019-10-10
Add an option to send a notification at a given date

nono:


model/notification/center/notification_center.go
```diff
@@ -288,9 +297,48 @@ func sendMail(inst *instance.Instance, p *notification.Properties, n *notificati
...
+	}, msg)
+	if err != nil {
+		return err
+	}
+	return job.System().AddTrigger(t)
+}
+
+// Make sure the stack always fallback on mail when the other channels has
+// failed.
+func appendMailToChannels(channels []string) []string {
```
ptbrowne: `ensureAtLeastMail` ?
`ensureFallbackMail` ?

This would remove the need for a docstring and would be clearer at the callpoint IMHO.

model/notification/center/notification_center.go
```diff
@@ -240,7 +242,14 @@ func findLastNotification(inst *instance.Instance, source string) (*notification
 	return notifs[0], nil
 }
 
-func sendPush(inst *instance.Instance, p *notification.Properties, n *notification.Notification) error {
+func sendPush(inst *instance.Instance,
+	p *notification.Properties,
+	n *notification.Notification,
+	at string,
```
ptbrowne: Nit: I am a bit surprised not to see `at` as a Date type already here ?

model/notification/center/notification_center.go
```diff
@@ -288,9 +297,48 @@ func sendMail(inst *instance.Instance, p *notification.Properties, n *notificati
...
+		_, err := job.System().PushJob(inst, &job.JobRequest{
+			WorkerType: worker,
+			Message:    msg,
+		})
+		return err
+	}
+	t, err := job.NewTrigger(inst, job.TriggerInfos{
+		Type:       "@at",
+		WorkerType: worker,
+		Arguments:  at,
```
ptbrowne: I think that having `at` typed as Date lower in the call stack would prevent errors like putting an invalid date into the trigger.

tests/integration/tests/notification.rb
```diff
@@ -0,0 +1,26 @@
+require_relative '../boot'
+require 'minitest/autorun'
+require 'pry-rescue/minitest' unless ENV['CI']
```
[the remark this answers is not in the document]
nono: On charge un module qui sert à lancer un REPL quand il y a un test qui échoue, ce qui est pratique pour débugger. Mais sur les environnements de CI, lancer un REPL est inutile et bloquant, donc on ne charge le module que si la variable d'env `CI` n'est pas définie.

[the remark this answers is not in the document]
ptbrowne: > On charge un module qui sert à lancer un REPL quand il y a un test qui échoue

Nice !

tests/integration/lib/helpers.rb
```diff
@@ -24,7 +24,7 @@ def start_mailhog
       `MailHog --version`
       spawn "MailHog"
     rescue Errno::ENOENT
-      puts "MailHog is not installed (or not in the PATH)".yellow
+      # Ignored
```
ptbrowne: Why is the mailhog error ignored ? The comment could be better IMHO :D

nono: We are launching MailHog as a docker service on github actions (https://github.com/cozy/cozy-stack/pull/2123/files#diff-7327122c763912b429300af1b6158a12R17-R21), but you are right, the comment should be better

PR #2128, opened 2019-10-11
Improve migration

nono:
- Increase timeout for migration worker
- Add a --fail-fast option to cozy-stack instance fsck

### DIFF
--- a/model/vfs/vfsafero/fsck.go
+++ b/model/vfs/vfsafero/fsck.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"crypto/md5"
 	"encoding/json"
+	"errors"
 	"io"
 	"os"
 	"path/filepath"
@@ -15,7 +16,9 @@ import (
 	"github.com/spf13/afero"
 )
 
-func (afs *aferoVFS) Fsck(accumulate func(log *vfs.FsckLog)) error {
+var errFailFast = errors.New("fail fast")
+
+func (afs *aferoVFS) Fsck(accumulate func(log *vfs.FsckLog), failFast bool) error {
 	entries := make(map[string]*vfs.TreeFile, 1024)
 	tree, err := afs.BuildTree(func(f *vfs.TreeFile) {
 		if !f.IsOrphan {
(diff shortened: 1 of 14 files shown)


model/vfs/vfsafero/fsck.go
```diff
@@ -41,12 +47,16 @@ func (afs *aferoVFS) CheckFilesConsistency(accumulate func(log *vfs.FsckLog)) er
...
-	return afs.checkFiles(entries, accumulate)
+	return afs.checkFiles(entries, accumulate, failFast)
 }
 
-func (afs *aferoVFS) checkFiles(entries map[string]*vfs.TreeFile, accumulate func(log *vfs.FsckLog)) (err error) {
+func (afs *aferoVFS) checkFiles(
+	entries map[string]*vfs.TreeFile,
+	accumulate func(log *vfs.FsckLog),
+	failFast bool,
+) error {
```
sblaisot: Not sure if it was intended or not (my go skills are so bad) but `(err error)` has been replaced with `error`

model/vfs/vfsafero/fsck.go
```diff
@@ -135,13 +151,19 @@ func (afs *aferoVFS) checkFiles(entries map[string]*vfs.TreeFile, accumulate fun
...
 		}
 		delete(entries, fullpath)
 		return nil
 	})
 	if err != nil {
-		return
+		if err == errFailFast {
+			return nil
+		}
+		return err
```
sblaisot: I don't understand all these `if failfast return nil`. Wouldn't this hide errors and a risk to return that everything was ok ?

model/vfs/vfsafero/fsck.go
nono: It is intended: I try to avoid named returns as the experience has shown that they make the code harder to review and change.

model/vfs/vfsafero/fsck.go
nono: We have two things: errors and fsckLogs. Errors are for when the results are unknown (CouchDB not available for example), and fsckLogs are the messages displayed by the FSCK. When a FSCK log is found, the stack will first send it, and if the `--fail-fast` option has been used, it will stop the FSCK. To do that, we have several functions and the simpler way is to return a special error (`errFailFast`), and to transform it to nil later to make the HTTP response a 200 OK, and not a 5xx.

PR #2140, opened 2019-10-15
Fix accounts cleaning on instance deletion

nono:
When an instance is destroyed, the stack ensures that io.cozy.accounts for a konnector with on_delete_account are deleted, and the associated job has finished in success. It allows to clean the accounts on remote systems. In particular, this commit takes care to not delete aggregator accounts before the other accounts have been cleaned as they can be necessary to authenticate on the remote systems.


model/instance/lifecycle/destroy.go
```diff
@@ -39,7 +38,9 @@ func DestroyWithoutHooks(domain string) error {
 
 	// Deleting accounts manually to invoke the "account deletion hook" which may
 	// launch a worker in order to clean the account.
-	deleteAccounts(inst)
+	if err := deleteAccounts(inst); err != nil {
+		return err
```
ptbrowne: What happens then ? Do we have a pager notification ;) ?

web/apps/apps.go
```diff
@@ -294,15 +289,15 @@ func findAccountsToDelete(instance *instance.Instance, slug string) ([]deletionE
 		if err == nil && msg.Slug == slug && msg.Account != "" {
 			acc := &account.Account{}
 			if err := couchdb.GetDoc(instance, consts.Accounts, msg.Account, acc); err == nil {
-				entry := deletionEntry{acc, t}
+				entry := account.CleanEntry{Account: acc, Trigger: t}
```
ptbrowne: I have not understood why here we pass the trigger whereas in the other method, we do not pass the trigger.

model/instance/lifecycle/destroy.go
```diff
@@ -72,51 +73,38 @@ func DestroyWithoutHooks(domain string) error {
...
+		if slug == "" {
+			continue
 		}
+		man, err := app.GetKonnectorBySlug(inst, slug)
+		if err != nil {
+			return err
+		}
+		entry := account.CleanEntry{
+			Account:          acc,
+			Trigger:          nil, // We don't care, the triggers will all be deleted a bit later
```
ptbrowne: The triggers are deleted by what ? Can you explicit this here ?

ptbrowne:
I see no tests for this ? Could this be tested ? Seems important enough 😁

model/instance/lifecycle/destroy.go
nono: Not yet, but it will come in a next PR

web/apps/apps.go
nono: When we uninstall a konnector, we remove the accounts, and we also delete the associated triggers. When the stack is destroyed, we don't try to find which trigger is related to which account, we just clean the accounts, and all the triggers are deleted in bulk.

model/instance/lifecycle/destroy.go
nono: This file is about the deletion of an instance. The stack is brutal and destroy a lot of the instance in bulk (like the triggers). But the accounts are an exception, as we need to clean them on remote systems. The triggers are deleted here: https://github.com/cozy/cozy-stack/blob/master/model/instance/lifecycle/destroy.go#L51-L60

nono:
> I see no tests for this ? Could this be tested ? Seems important enough

Yup, I'm looking how to test that. I'm not very inspired, but it looks important.

model/account/accounts.go
```diff
@@ -118,6 +119,50 @@ func GetTriggers(jobsSystem job.JobSystem, db prefixer.Prefixer, accountID strin
...
+// CleanAndWait deletes the accounts. If an account is for a konnector with
+// "on_delete_account", a job is pushed and it waits for the job success to
+// continue. Finally, the associated trigger can be deleted.
+func CleanAndWait(inst *instance.Instance, toClean []CleanEntry) error {
+	jobsSystem := job.System()
+	for _, entry := range toClean {
+		acc := entry.Account
+		acc.ManualCleaning = true
+		oldRev := acc.Rev() // The deletion job needs the rev just before the deletion
+		if err := couchdb.DeleteDoc(inst, acc); err != nil {
```
taratatach: Can't we delete the document last? We wouldn't need to keep its old revision and we would still have a reference to this account in case something goes awry.

nono: Well, we could do that here. But the `on_delete.js` script called to clean the account on the remote system is also called when an account is deleted by a user. So, if we do that, we push some complexity on this script, as it would have to manage two ways to be called. And I think it would add more complexity on the JS side that it would remove on the go side.

PR #2142, opened 2019-10-16
Fix double-deletion in migration worker

sblaisot:
#2136 introduced a bug in migration worker while adding log on older container deletion. By using the `err` variable to store error on older container deletion, this triggers the deletion of the new container by the defer func after the migration has finished.

This PR fixes that bug and add a warning comment in the code.


sblaisot:
I relaunched the tests 3 times and  got 3 different errors. giving up.

PR #2143, opened 2019-10-17
Improve the Swift container deletion

nono:
When deleting a Swift container, we have to first delete all the objects inside it, and then delete the container its-self. Swift behavior on consistency makes possible that the call to delete the container can fail just after we have deleted all the objects. This commit adds a retry loop to work around this limitation.

### DIFF
--- a/model/vfs/vfsswift/swift.go
+++ b/model/vfs/vfsswift/swift.go
@@ -2,6 +2,7 @@ package vfsswift
 
 import (
 	"errors"
+	"time"
 
 	multierror "github.com/hashicorp/go-multierror"
 	"github.com/ncw/swift"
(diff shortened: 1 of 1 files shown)


model/vfs/vfsswift/swift.go
```diff
@@ -36,7 +37,21 @@ func DeleteContainer(c *swift.Connection, container string) error {
...
+	// server (at least). And, when we try to delete the container, Swift can
+	// send an error as some container servers will still have objects
+	// registered for this container. We will try several times to delete the
+	// container to work-around this limitation.
+	for i := 0; i < 5; i++ {
+		err = c.ContainerDelete(container)
+		if err == nil {
+			return nil
+		}
+		time.Sleep(5 * time.Second)
```
sblaisot: Wouldn't it be better to move sleep to the beginning of the `for` loop inside an `if i != 0 { }` ?
This would have the benefit to fail immediately on last turn instead of waiting 5 seconds after last fail to return.

nono: But it has the con of always waiting 5 seconds before the first try.

sblaisot: Not if put I side àn `if i != 0 {}`

PR #2153, opened 2019-10-21
Install couchdb via apt instead of docker for CI

nono:
We had a lot of badmatch/badarg errors on CI with CouchDB in Docker. Let's see if it works better with CouchDB installed as a debian package.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,12 +10,8 @@ on:
       - 'docs/**'
 jobs:
   test:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
     services:
-      couchdb:
-        image: apache/couchdb:2.3
-        ports:
-          - 5984:5984
       redis:
         image: redis
         ports:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -24,6 +20,13 @@ jobs:
       matrix:
         go-version: [1.12.x, 1.13.x]
     steps:
+      - name: Install CouchDB
+        run: |
+          sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 379CE192D401AB61
+          echo "deb https://apache.bintray.com/couchdb-deb bionic main" | sudo tee -a /etc/apt/sources.list
```
sblaisot: NIT: use lsb_release to make this line independant of ubuntu codename

```
echo "deb https://apache.bintray.com/couchdb-deb $(lsb_release -c -s) main" | sudo tee -a /etc/apt/sources.list
```

.github/workflows/ci.yml
```diff
@@ -24,6 +20,13 @@ jobs:
...
+          sudo apt-get update
+          echo -ne "couchdb couchdb/mode select standalone\ncouchdb couchdb/mode seen true\ncouchdb couchdb/bindaddress string 127.0.0.1\ncouchdb couchdb/bindaddress seen true\ncouchdb couchdb/adminpass password \ncouchdb couchdb/adminpass seen true\ncouchdb couchdb/adminpass_again password \ncouchdb couchdb/adminpass_again seen true\n" | sudo debconf-set-selections
+          DEBIAN_FRONTEND=noninteractive sudo apt-get install -y --force-yes couchdb
```
sblaisot: NIT: would be easilly readable with a here-document:

```
sudo debconf-set-selections <<-EOF
    couchdb couchdb/mode select standalone
    couchdb couchdb/mode seen true
    couchdb couchdb/bindaddress string 127.0.0.1
    couchdb couchdb/bindaddress seen true
    couchdb couchdb/adminpass password
    couchdb couchdb/adminpass seen true
    couchdb couchdb/adminpass_again password
    couchdb couchdb/adminpass_again seen true
EOF
```

note: leading spaces are tabs. tabs are stripped at the beginning of lines with `<<-` here-documents operator

sblaisot: unless gh actions interpret each line as a separate command :(

nono: Let's try. I have made several attempts to make it works, and I wanted to avoid this potential issue while debugging, but now that I have a version that works, we cant try to improve it.

nono: It works!

sblaisot: Nice !

PR #2157, opened 2019-10-25
Allow connection to TLS 1.0 external sites

sblaisot:
Node 12 [disable by default](https://github.com/nodejs/node/pull/23814) connection to TLS 1.0 and 1.1 https servers.

This PR updates the nsjail connector launch script to add an argument to nodejs12 to allow connection to external sites using only TLS 1.0 and/or 1.1

### DIFF
--- a/scripts/konnector-nsjail-node12-run.sh
+++ b/scripts/konnector-nsjail-node12-run.sh
@@ -22,7 +22,7 @@ else
   exit 1
 fi
 
-NODE_OPTS="--max-http-header-size=16384"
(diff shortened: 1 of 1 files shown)


nono:
The integration tests are red, but it's not related to this PR. Sometimes, they trigger an error in CouchDB (https://github.com/cozy/cozy-stack/pull/2157/checks?check_run_id=275227983#step:8:104). Most of those errors have disappeared when we have moved CouchDB from Docker to Debian package on github actions, but there is still one error, and I cannot reproduce on my PC.

So, let's merge!

PR #2168, opened 2019-11-05
Add notes with collaborative edition

nono:


docs/notes.md
```diff
@@ -164,14 +166,10 @@ Content-Type: application/vnd.api+json
         "doctypeVersion": "1",
         "metadataVersion": 1,
         "createdAt": "2019-11-05T12:38:04Z",
-        "createdByApp": "note",
```
taratatach: Why remove the app slug?

model/note/note.go
```diff
@@ -239,6 +239,10 @@ func UpdateTitle(inst *instance.Instance, file *vfs.FileDoc, title string) error
 		return nil
 	}
 
+	event := Event{"title": title, "doctype": consts.NotesDocuments}
```
taratatach: This could go into a method in the `event.go` file like `PutTelepointer`. This way we'd have all possible realtime events for notes in the same place.

model/note/step.go
```diff
@@ -147,6 +148,12 @@ func ApplySteps(inst *instance.Instance, file *vfs.FileDoc, lastVersion string,
 			return err
 		}
 	}
+	for _, s := range steps {
+		e := Event(s)
```
taratatach: This could be a method into the `event.go` file as well.

model/note/note.go
```diff
@@ -222,40 +281,173 @@ func ensureNotesDir(inst *instance.Instance) (*vfs.DirDoc, error) {
...
+	doc, err := get(inst, file)
+	if err != nil {
+		return nil, err
+	}
+
+	if doc.Title == title {
+		return file, nil
+	}
+	doc.Title = title
+	if err := saveToCache(inst, doc); err != nil {
```
taratatach: Maybe we could invalidate the cache and keep going if we fail to save the new document to cache?
It seems a bit overkill to bail out if we can't save a perfectly fine document to the cache.

model/note/note.go
```diff
@@ -274,6 +282,27 @@ func ensureNotesDir(inst *instance.Instance) (*vfs.DirDoc, error) {
...
+}
+
+func setupTrigger(inst *instance.Instance, fileID string) error {
+	sched := job.System()
+	msg := &DebounceMessage{NoteID: fileID}
+	t, err := job.NewTrigger(inst, job.TriggerInfos{
+		Type:       "@event",
+		WorkerType: "notes-save",
+		Arguments:  fmt.Sprintf("%s:UPDATED:%s", consts.NotesEvents, fileID),
+		Debounce:   "10m",
```
taratatach: 10 minutes seems like a very long time. Would performances be so bad with a 1 minute debounce time?

pkg/limits/rate_limiting.go
```diff
@@ -153,6 +155,12 @@ var configs = []counterConfig{
 		Limit:  30,
 		Period: 1 * time.Hour,
 	},
+	// JobNotesPersistType
+	{
+		Prefix: "job-notes-persist",
+		Limit:  100,
+		Period: 1 * time.Hour,
```
taratatach: Does that mean we'll persist notes on disk only once every hour? If we want to sync those files with Desktop we should expect people to think the version they have on disk is the latest one after they're done editing their note.

model/note/step.go
```diff
@@ -181,10 +195,35 @@ func saveSteps(inst *instance.Instance, steps []Step) error {
...
+		return
+	}
+	if len(steps) == 0 {
+		return
+	}
+	docs := make([]couchdb.Doc, len(steps))
+	for i := range steps {
+		docs[i] = &steps[i]
+	}
+	if err := couchdb.BulkDeleteDocs(inst, consts.NotesSteps, docs); err != nil {
```
taratatach: `couchdb.Doc` isn't an interface that `Step` implements? I would have thought we could use `steps` directly in stead of `docs` here.

model/note/note.go
```diff
@@ -365,19 +365,42 @@ func get(inst *instance.Instance, file *vfs.FileDoc) (*Document, error) {
...
+		return nil, err
+	}
+	doc, err := fromMetadata(file)
+	if err != nil {
+		return nil, err
+	}
+	if len(steps) == 0 {
+		return doc, nil
+	}
+	if version, ok := steps[0]["version"].(float64); ok {
```
taratatach: I don't understand why we recompute the doc's version here when we have one from the file's metadata. If they don't match, shouldn't we error out?

nono:
> I'm a bit surprised we're not using the versioning mechanism built for other documents here.
Would that be inefficient? Or just plain impossible?

Well, the notes are materialized by a file. It would mean that each small modification (like each letter that has been typed) would require to update a file in CouchDB an Swift. To avoid that, the stack uses a cache, but it disables the built-in versioning. Moreover, ProseMirror expects versions to be a number, not a string like in CouchDB.

docs/notes.md
```diff
@@ -305,9 +305,153 @@ Content-Type: application/vnd.api+json
...
+      "type": "file",
+      "name": "My new note.cozy-note",
+      "trashed": false,
+      "md5sum": "NjhiMzI5ZGE5ODkzZTM0MDk5YzdkOGFkNWNiOWM5NDAgIC0K",
+      "created_at": "2019-11-05T12:38:04Z",
+      "updated_at": "2019-11-05T12:38:52Z",
+      "tags": [],
+      "metadata": {
+        "title": "My new note",
+        "content": { "type": "doc", "content": [{ "type": "horizontal_rule" }] },
```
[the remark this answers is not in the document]
nono: No, if the steps are not available, it returns the latest version of the document. In most cases, we don't even have the the document as it was before those steps were applied.

model/note/note.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/cozy/cozy-stack/model/instance"
+	"github.com/cozy/cozy-stack/model/vfs"
+	"github.com/cozy/cozy-stack/pkg/consts"
+	"github.com/cozy/cozy-stack/pkg/couchdb"
+	"github.com/cozy/prosemirror-go/model"
+)
+
+// Document is the note document in memory. It is persisted to the VFS as a
+// file, but with a debounce: the intermediate states are saved in Redis.
+type Document struct {
```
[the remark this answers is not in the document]
nono: In go, I try to avoid calling the same a package and its main struct, as it is a bit weird to use `note.Note`.

model/note/note.go
nono: Why should error out? We have the steps that can be used to rebuild the latest version of the document.

model/note/note.go
nono: No, for the title, if we can't save to the cache, the change will be lost. So, it's better to return an error.

model/note/step.go
nono: > couchdb.Doc isn't an interface that Step implements?

Yes, it is.

> I would have thought we could use steps directly in stead of docs here.

No, we can't use a `[]Step` as a `[]couchdb.Doc` (even if `Step` implements the `couchdb.Doc` interface).

pkg/limits/rate_limiting.go
nono: No, we are in the rate-limiting package. It means that an instance can't create more than 100 jobs of the type `job-notes-persist` in an hour.

model/note/note.go
```diff
@@ -248,4 +249,61 @@ func UpdateTitle(inst *instance.Instance, file *vfs.FileDoc, title string) error
...
+		if result.Failed != "" {
+			inst.Logger().WithField("nspace", "notes").
+				Infof("Cannot apply a step: %s", err)
+			return ErrCannotApply
+		}
+		doc = result.Doc
+	}
+
+	olddoc := file.Clone().(*vfs.FileDoc)
+	file.Metadata["content"] = doc.ToJSON()
```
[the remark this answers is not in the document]
nono: No, the content is immutable. We are just swapping one content by another. And `olddoc` is a file, with mutable fields (and in particular `Metadata`). That's why we call `Clone` to have a copy of the file: the changes on `file` won't impact `olddoc`.

model/note/note.go
```diff
@@ -309,8 +315,11 @@ func ApplySteps(inst *instance.Instance, file *vfs.FileDoc, steps []couchdb.JSON
 		doc = result.Doc
 	}
 
+	// TODO persist the steps
+
 	olddoc := file.Clone().(*vfs.FileDoc)
 	file.Metadata["content"] = doc.ToJSON()
+	file.Metadata["revision"] = rev
```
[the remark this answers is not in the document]
nono: No, we use the revision only for the changes from ProseMirror.

model/note/note.go
```diff
@@ -258,7 +259,7 @@ func UpdateTitle(inst *instance.Instance, file *vfs.FileDoc, title string) error
 // ApplySteps takes a note and some steps, and tries to apply them. It is an
 // all or nothing change: if there is one error, the note won't be changed.
 // TODO fetch last info for file (if debounce)
-func ApplySteps(inst *instance.Instance, file *vfs.FileDoc, steps []couchdb.JSONDoc) error {
+func ApplySteps(inst *instance.Instance, file *vfs.FileDoc, lastRev string, steps []couchdb.JSONDoc) error {
```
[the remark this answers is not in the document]
nono: We can have a conflict with another file with the same filename. But it is managed.

And, for the moment, we expect that the file content has not been modified by something else that the notes application.

model/note/step.go
```diff
@@ -70,11 +84,12 @@ func ApplySteps(inst *instance.Instance, file *vfs.FileDoc, lastRev string, step
 	if !ok {
 		return ErrInvalidFile
 	}
-	revision, ok := file.Metadata["revision"].(string)
+	v, ok := file.Metadata["version"].(float64)
```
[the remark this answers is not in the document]
nono: `Metadata` is `map[string]interface{}`. When unmarshalling JSON to `map[string]interface{}`, Go uses a float64 for numbers. So, we have to first acknowledge that `version` is a float64 (not just any value that implements `interface{}`), and only after that, we can transform the float64 to an int64.

model/note/step.go
```diff
@@ -145,20 +161,30 @@ func ApplySteps(inst *instance.Instance, file *vfs.FileDoc, lastRev string, step
...
-	// TODO
-	if len(steps) == 0 || len(steps) == 1000 {
+
+	// The first step plays the role of a sentinel: if it isn't here, the
+	// version is too old. Same if we have too many steps.
+	if len(steps) == 0 || len(steps) == req.Limit {
 		return nil, ErrTooOld
 	}
+
+	steps = steps[1:] // Discard the sentinel
```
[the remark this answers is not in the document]
nono: No, we only discard the sentinel in the response to the client, we don't modify the cache or the database. Other clients can make the same call and will have the same response.

model/note/note.go
nono: Well, @edas has suggested 30 minutes. I think it is a bit too much. I've decreased to 10 minutes to see what it gives, and I will probably decrease another time later if the performances are OK.

pkg/limits/rate_limiting.go
taratatach: Ah right I missed that. Thanks.

model/note/note.go
taratatach: OK. We'll see then :)

model/note/note.go
taratatach: My bad, I missed the fact that we use early return statements before and that if we reach this line, we don't have a version and need to recompute it from the steps.

model/note/step.go
taratatach: OK. Like you said, it's been a while since last I've done go :)

model/note/note.go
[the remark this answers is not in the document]
taratatach: OK, why not.

model/note/note.go
taratatach: Hum yes. I probably wrote this before you added cache.

model/note/note.go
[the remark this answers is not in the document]
taratatach: I was thinking about title updates conflicts, from the notes application.
Since we don't change the revision of the Document when we update the title, I wonder if we could have a conflict with 2 people updating the title at the same time.

docs/notes.md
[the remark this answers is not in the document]
taratatach: OK. It's a nitpick but I feel the documentation would better represent the expected behavior if the content here had the `Hello` word added by the 2 steps returned in the successful response.

model/note/note.go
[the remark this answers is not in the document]
nono: The last one wins. The title is not managed by ProseMirror, so we don't have fancy diff/patch mechanism and 2 people editing the title at the same time is an edge case not managed by the front app.

PR #2192, opened 2019-11-15
Fix sending push notifications

nono:
When there are many OAuth clients registered on an instance, but only a few of them has a notification_device_token, the stack can miss sending some push notifications. This issue comes from the pagination on the _find request to find the notifiable clients. By including a mango
selector to find only the OAuth clients with a notification_device_token, we can avoid this issue.

### DIFF
--- a/worker/push/push.go
+++ b/worker/push/push.go
@@ -104,17 +104,19 @@ func Worker(ctx *job.WorkerContext) error {
 	if err != nil {
 		return err
 	}
+	if len(cs) > 10 {
+		ctx.Logger().Warnf("too many notifiable devices: %d", len(cs))
+		cs = cs[:10]
+	}
 	for _, c := range cs {
-		if c.NotificationDeviceToken != "" {
-			err = push(ctx, c, &msg)
-			if err != nil {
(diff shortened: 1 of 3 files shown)


worker/push/push.go
```diff
@@ -104,17 +104,19 @@ func Worker(ctx *job.WorkerContext) error {
 	if err != nil {
 		return err
 	}
+	if len(cs) > 10 {
+		ctx.Logger().Warnf("too many notifiable devices: %d", len(cs))
+		cs = cs[:10]
+	}
```
ptbrowne: Why adding this arbitrary limit ?

nono: I think we don't want to send so many pushes for the same notification

ptbrowne: Maybe we should prioritize on the latest devices, I think here the oldest devices will be used.

The logging string could also be a bit clearer that notifiables are getting sliced, "too many notifiable devices %s, capping to %s", len(cs), MAX_NOTIFIABLE_DEVICES".

Instead of a magic number, a CONSTANT would be clearer.

PR #2196, opened 2019-11-16
Send the old doc to the realtime for swift migrations

nono:
When a cozy instance is migrating to Swift layout v3, its CouchDB
documents for io.cozy.files are updated to add the internal_id
attribute. The trigger for thumbnails add a job. But as we were giving
only the new document, and not the old one, the sameImg comparison in
the thumbnail worker would fail, and a new thumbnail may be recreated.
And, as we can have a race condition between getting the Swift container
for the thumbnails and locking the VFS, it could be put in the old data
container.

### DIFF
--- a/worker/migrations/migrations.go
+++ b/worker/migrations/migrations.go
@@ -274,10 +274,12 @@ func getSrcName(inst *instance.Instance, f *vfs.FileDoc) string {
 	return srcName
 }
 
+// XXX the f FileDoc can be modified to add an InternalID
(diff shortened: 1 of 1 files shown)


sblaisot:
I don't see how this would change anything as UpdateDoc and UpdateDocWithOld are the same, with the exception of teching the old doc from the database.

nono:
The thumbnail worker makes a comparison between the old doc and the new doc to know if it needs to generate a thumbnail or can do nothing: https://github.com/cozy/cozy-stack/blob/master/worker/thumbnail/thumbnail.go#L72-L74

PR #2200, opened 2019-11-18
Add a capabilities endpoint

nono:


docs/settings.md
```diff
@@ -252,6 +252,37 @@ HTTP/1.1 204 No Content
 
 ## Instance
 
+### GET /settings/capabilities
+
+List the activated capabilities for this instance.
```
sblaisot: Nit:
*An unadvertised capability should be considered `false` and, for backward comaptibility, if you can't get a valid response from this endpoint (in particular in case of a `404 Not found` error), all capabilities should be considered `false`*

web/settings/capabilities.go
```diff
@@ -0,0 +1,55 @@
...
+	return &jsonapi.LinksList{Self: "/settings/capabilities"}
+}
+func (c *apiCapabilities) Fetch(field string) []string { return nil }
+
+func newCapabilities(inst *instance.Instance) *apiCapabilities {
+	// File versioning is enabled for all instances, except for the Swift
+	// layout v1 and v2
+	versioning := true
+	switch config.FsURL().Scheme {
+	case config.SchemeSwift, config.SchemeSwiftSecure:
```
sblaisot: For my comprehension, why a `case` and not a simple `if` ?

web/settings/settings_test.go
```diff
@@ -260,6 +260,31 @@ func TestGetPassphraseParameters(t *testing.T) {
...
+	assert.True(t, ok)
+	assert.Equal(t, "io.cozy.settings", data["type"].(string))
+	assert.Equal(t, "io.cozy.settings.capabilities", data["id"].(string))
+	attrs, ok := data["attributes"].(map[string]interface{})
+	assert.True(t, ok)
+	versioning, ok := attrs["file_versioning"].(bool)
+	assert.True(t, ok)
+	assert.True(t, versioning)
+}
+
```
sblaisot: Maybe you whould also add a negative test on an instance with `file_versioning` capability disabled ?

web/settings/settings_test.go
```diff
@@ -260,6 +260,31 @@ func TestGetPassphraseParameters(t *testing.T) {
...
+	res, err = http.DefaultClient.Do(req)
+	assert.NoError(t, err)
+	assert.Equal(t, 200, res.StatusCode)
+	var result map[string]interface{}
+	err = json.NewDecoder(res.Body).Decode(&result)
+	assert.NoError(t, err)
+	data, ok := result["data"].(map[string]interface{})
+	assert.True(t, ok)
+	assert.Equal(t, "io.cozy.settings", data["type"].(string))
+	assert.Equal(t, "io.cozy.settings.capabilities", data["id"].(string))
```
sblaisot: `_id` ?

docs/settings.md
```diff
@@ -252,6 +252,37 @@ HTTP/1.1 204 No Content
...
+Accept: application/vnd.api+json
+```
+
+#### Response
+
+```json
+{
+    "data": {
+        "type": "io.cozy.settings",
+        "id": "io.cozy.settings.capabilities",
```
sblaisot: `_id` ?

web/settings/capabilities.go
nono: With an `if` like that, it would work too:

```go
scheme := config.FsURL().Scheme
if scheme == config.SchemeSwift || scheme.SchemeSwiftSecure {
```

I've chosen the `switch case` as it is more easier to extend with a new `case`, but I don't really care.

docs/settings.md
nono: No, it is `id` and not `_id` here (as per the JSON-API spec).

web/settings/capabilities.go
sblaisot: ok, thanks for the explanation.

docs/settings.md
sblaisot: I can see `_id` [here](https://github.com/cozy/cozy-stack/pull/2200/files#diff-8727f40ae33bbd62bb4e35fbb99690d1R17) so I am wondering why it is sometimes referred to as `id` and sometimes as `_id`

nono: By convention, it's `_id` in the internal side (like when speaking with CouchDB), and `id` the jsonapi package transforms it to `id` on the serialization to match the JSON-API spec.

PR #2204, opened 2019-11-20
Improve the lock package

nono:


pkg/lock/simple_redis.go
```diff
@@ -64,45 +51,97 @@ type redisLock struct {
 	mu     sync.Mutex
 	key    string
 	token  string
```
sblaisot: Lint: I think there is one space missing before type here to align with the 3 lines below following the comment

nono: Well, you know, go developers don't discuss things like that. They use `gofmt`, and `gofmt` is always right (or, in the extreme case where `gofmt` is wrong, we should open a issue for it, not discuss it here).

`gofmt` does not try to align types before and after a comment, and as said before, `gofmt` is always right. So, let's keep it as is.

sblaisot: 👍

pkg/lock/simple_redis.go
```diff
@@ -64,45 +51,109 @@ type redisLock struct {
...
 
-func (rl *redisLock) Lock() error {
-	rl.mu.Lock()
-	defer rl.mu.Unlock()
+func (rl *redisLock) extends(writing bool) (bool, error) {
+	if rl.token == "" {
+		return false, nil
+	}
+	if (writing && rl.readers > 0) || (!writing && rl.readers < 0) {
+		return false, nil
```
taratatach: Shouldn't we return an error when trying to extend a lock we can't have since we're either getting a read lock on a write locked resource or a write lock on a read locked resource?

Not returning an error here, means we'll attempt to obtain a lock we can't obtain in `extendsOrObtainsReading()` for example.

pkg/lock/simple_redis.go
```diff
@@ -64,45 +51,109 @@ type redisLock struct {
...
+}
+
+func (rl *redisLock) obtainsWriting(token string) (bool, error) {
+	rl.mu.Lock()
+	defer rl.mu.Unlock()
+	return rl.obtains(true, token)
+}
+
+func (rl *redisLock) Lock() error {
+	ok, err := rl.extendsWriting()
```
taratatach: Why don't we retry to extend the writing lock like we do for reading locks?

pkg/lock/simple_redis.go
```diff
@@ -64,45 +51,109 @@ type redisLock struct {
...
+	}
+	if !ok {
+		return false, nil
+	}
+
+	rl.token = token
+	if writing {
+		rl.readers = -1
+	} else {
+		rl.readers++
```
taratatach: Shouldn't we make sure the `readers` count is at least 1?

nono: Do you mean before or after the increment?

pkg/lock/simple_redis.go
nono: For a reading lock, there can be several goroutines. If the lock is currently busy for writing, all the goroutines for reading will try to extend and then obtain the lock (but it will fail). When the writing goroutine release the lock, the first reading goroutine will obtain the lock, and the next goroutines can extend it.

For writing, we can't have several goroutines with the lock at the same time, so it didn't make sense to retry extending the lock. And, in fact, even the first call to extend can introduce bugs and I have preferred to not allow a goroutine to calls several times `Lock` and then calls `Unlock` several times: it must call `Unlock` before calling again `Lock`. It is also more consistent with the in-memory lock.

pkg/lock/simple_redis.go
nono: Well, with the current code, if we return an error, the `Lock()` or `RLock()` will fail with an error, without trying again later (when it should succeed). So, returning an error with the current code is not an option.

Currently, it tries to obtain the lock (even if we are sure that it will fail). We can optimize that but it's tricky. I prefer to keep the code simpler (even if it makes some useless calls to redis) instead of implementing the optimizations and taking the risk to introduce very subtle bugs in those optimizations.

PR #2209, opened 2019-11-20
Feature flags on instances

aeris:
This is to be able to manage feature flags from Cloudery and ensure
correct values at the end in case of feature removal and so no more
specified on the command line.
Settings prefixed with `feature.` will be handle specially and any
change on one setting reset any existing feature-related settings

### DIFF
--- a/cmd/settings.go
+++ b/cmd/settings.go
@@ -15,6 +15,7 @@ import (
 )
 
 var errSettingsMissingDomain = errors.New("Missing --domain flag")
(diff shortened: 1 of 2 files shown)


aeris:
Would have been better to put such logic on the server part, but currently CLI fetch settings then merge them with the provided ones before sending the overall settings to the server part for update :sob:

nono:
We don't want to put the feature flags in the settings: see #2222.

PR #2210, opened 2019-11-21
Support bookmark pagination for normal_docs

Gara64:
#2190 allowed _find queries to use bookmark. This make it possible for normal_docs as well.

### DIFF
--- a/docs/data-system.md
+++ b/docs/data-system.md
@@ -462,16 +462,19 @@ See
 The `_all_docs` endpoint sends the design docs in the response. It makes it hard
 to use pagination on it. We have added a non-standard `_normal_docs` endpoint.
 This new endpoint skip the design docs (and does not count them in the
-`total_rows`). It only accepts two parameters in the query string: `skip`
-(default: 0) and `limit` (default: 100).
+`total_rows`). It only accepts three parameters in the query string: `limit`
+(default: 100), `skip` (default: 0),  and `bookmark` (default: '').
 
 Note that the response format is a bit different, it looks more like a `_find`
-response with mango. And, like for `_find`, the limit cannot be more than 100.
+response with mango. And, like for `_find`, the limit cannot be more than 1000.
+The [pagination](https://github.com/cozy/cozy-stack/blob/master/docs/mango.md#pagination-cookbook)
+is also the same: both `bookmark` and `skip` can be used, but `bookmark` is
+recommended for performances.
 
 ### Request
 
 ```http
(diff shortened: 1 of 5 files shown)


docs/mango.md
```diff
@@ -142,17 +142,21 @@ query results to a maximum of 100 documents. This limit can be raised up to
...
+the next query to get the missing docs. It is also possible to use `skip`,
+to paginate, although this is not recommended for performances. For more details, see the
+[CouchDB documentation](https://docs.couchdb.org/en/latest/api/database/find.html#pagination).
 
 ```json
 {
     "limit": 100,
     "next": true,
     "docs": ["... first hundred docs ..."]
+    "bookmark": "g1AAAAB2eJzLYWBgYMpgSmHgKy5JLCrJTq2MT8lPzkzJBYorGKQYpVqaJRoZm1paWFiapFkamhknGpilJiampZkYJRmC9HHA9OUAdTASpS0rCwAlah76"
```
nono: [nit] Can you add a `,` at the end of the previous line?

pkg/couchdb/couchdb.go
```diff
@@ -839,23 +839,30 @@ func FindDocsRaw(db Database, doctype string, req interface{}, results interface
...
+	}
 	err := makeRequest(db, doctype, http.MethodPost, "_find", &req, &findRes)
 	if err != nil {
 		return nil, err
 	}
 	res := NormalDocsResponse{
 		Rows: findRes.Docs,
 	}
-	if len(res.Rows) < limit {
+	if skip > 0 && len(res.Rows) < limit {
```
nono: Is this version also correct?

```go
if bookmark == "" && len(res.Rows) < limit {
```

I would prefer this, as it avoids a request to CouchDB when fetching normal docs on a small database (where there are less documents than the limit). The request will have no bookmarks and no skip, and I think we can take the shortcut here.

docs/data-system.md
```diff
@@ -462,16 +462,19 @@ See
 The `_all_docs` endpoint sends the design docs in the response. It makes it hard
 to use pagination on it. We have added a non-standard `_normal_docs` endpoint.
 This new endpoint skip the design docs (and does not count them in the
-`total_rows`). It only accepts two parameters in the query string: `skip`
-(default: 0) and `limit` (default: 100).
+`total_rows`). It only accepts three parameters in the query string: `limit`
```
ptbrowne:
```suggestion
`total_rows`). It accepts three parameters in the query string: `limit`
```

PR #2213, opened 2019-11-22
Add the myself document

nono:
The myself document is an io.cozy.contacts with the me attribute set to true. It has information about the owner of the Cozy instance, like its name and email address, and can be used by konnectors to link imported documents and files to an identity.

### DIFF
--- a/docs/contacts.md
+++ b/docs/contacts.md
@@ -1,0 +1,44 @@
+[Table of contents](README.md#table-of-contents)
+
+# Contacts
+
+## Routes
+
+### POST /contacts/myself
+
+This endpoint returns the information about the io.cozy.contacts document for
+the owner of this instance, the "myself" contact. If the document does not
+exist, it is recreated with some basic fields.
+
+A permission on the `io.cozy.contacts` document is required.
+
+#### Request
+
+```http
+POST /contacts/myself HTTP/1.1
+```
+
+#### Response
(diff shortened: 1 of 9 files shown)


docs/contacts.md
```diff
@@ -0,0 +1,44 @@
...
+This endpoint returns the information about the io.cozy.contacts document for
+the owner of this instance, the "myself" contact. If the document does not
+exist, it is recreated with some basic fields.
+
+A permission on the `io.cozy.contacts` document is required.
+
+#### Request
+
+```http
+POST /contacts/myself HTTP/1.1
```
Crash--: Since we have a POST on /myself, do we have an interest of having also a get?

Currently we're requesting a find().where('mysefl', true) with an index. Is it enough from a performance point of view?

nono: > Since we have a POST on /myself, do we have an interest of having also a get?

Well, I don't see what could be the usage for that. If you want the myself document, you can call the POST route: it the document already exists it will be returned by this route. And if the document does not exist, but you are interested about it, the next step would very probably be to ask for recreating it.

> Currently we're requesting a find().where('mysefl', true) with an index. Is it enough from a performance point of view?

I think it is `where('me', true)`. And the stack does the same thing: a find request with an index for finding a document with `me: true`.

PR #2222, opened 2019-11-25
Add feature flags

nono:
**TODO**: 
- [x] add tests
- [x] check performance of crc32
- [x] update the man pages and shell completion scripts

### DIFF
--- a/cmd/feature.go
+++ b/cmd/feature.go
@@ -4,12 +4,15 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"net/url"
 	"strings"
 
 	"github.com/cozy/cozy-stack/client/request"
 	"github.com/cozy/cozy-stack/pkg/consts"
 	"github.com/spf13/cobra"
 )
+
+var flagWithSources bool
 
 var featureCmdGroup = &cobra.Command{
 	Use:     "features <command>",
@@ -30,10 +33,14 @@ cozy-stack feature show displays the feature flags that are shown by apps.
 			return cmd.Usage()
 		}
 		c := newClient(flagDomain, consts.Settings)
-		res, err := c.Req(&request.Options{
+		req := &request.Options{
 			Method: "GET",
 			Path:   "/settings/flags",
-		})
+		}
+		if flagWithSources {
+			req.Queries = url.Values{"include": {"source"}}
+		}
+		res, err := c.Req(req)
 		if err != nil {
 			return err
 		}
@@ -42,12 +49,25 @@ cozy-stack feature show displays the feature flags that are shown by apps.
 			Data struct {
 				Attributes map[string]json.RawMessage `json:"attributes"`
 			} `json:"data"`
+			Included []struct {
+				ID         string                     `json:"id"`
+				Attributes map[string]json.RawMessage `json:"attributes"`
+			} `json:"included"`
 		}
 		if err = json.NewDecoder(res.Body).Decode(&obj); err != nil {
 			return err
 		}
 		for k, v := range obj.Data.Attributes {
 			fmt.Printf("- %s: %s\n", k, string(v))
+		}
+		if len(obj.Included) > 0 {
+			fmt.Printf("\nSources:\n")
+			for _, source := range obj.Included {
+				fmt.Printf("- %s\n", source.ID)
+				for k, v := range source.Attributes {
+					fmt.Printf("\t- %s: %s\n", k, string(v))
+				}
+			}
 		}
 		return nil
 	},
@@ -223,6 +243,7 @@ If you give a null value, the flag will be removed.
 
 func init() {
 	featureShowCmd.Flags().StringVar(&flagDomain, "domain", "", "Specify the domain name of the instance")
+	featureShowCmd.Flags().BoolVar(&flagWithSources, "source", false, "Show the sources of the feature flags")
 	featureFlagCmd.Flags().StringVar(&flagDomain, "domain", "", "Specify the domain name of the instance")
 	featureSetCmd.Flags().StringVar(&flagDomain, "domain", "", "Specify the domain name of the instance")
 	featureContextCmd.Flags().StringVar(&flagContext, "context", "", "The context for the feature flags")
--- a/cmd/feature.go
+++ b/cmd/feature.go
@@ -7,12 +7,50 @@ import (
 	"strings"
 
 	"github.com/cozy/cozy-stack/client/request"
+	"github.com/cozy/cozy-stack/pkg/consts"
 	"github.com/spf13/cobra"
 )
 
 var featureCmdGroup = &cobra.Command{
-	Use:   "feature <command>",
-	Short: "Manage the feature flags",
+	Use:     "features <command>",
+	Aliases: []string{"feature"},
+	Short:   "Manage the feature flags",
+}
+
+var featureShowCmd = &cobra.Command{
+	Use:   "show",
+	Short: `Display the computed feature flags for an instance`,
+	Long: `
+cozy-stack feature show displays the feature flags that are shown by apps.
+`,
+	Example: `$ cozy-stack feature show --domain cozy.tools:8080`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if flagDomain == "" {
+			errPrintfln("%s", errMissingDomain)
+			return cmd.Usage()
+		}
+		c := newClient(flagDomain, consts.Settings)
+		res, err := c.Req(&request.Options{
+			Method: "GET",
+			Path:   "/settings/flags",
+		})
+		if err != nil {
+			return err
+		}
+		defer res.Body.Close()
+		var obj struct {
+			Data struct {
+				Attributes map[string]json.RawMessage `json:"attributes"`
+			} `json:"data"`
+		}
+		if err = json.NewDecoder(res.Body).Decode(&obj); err != nil {
+			return err
+		}
+		for k, v := range obj.Data.Attributes {
+			fmt.Printf("- %s: %s\n", k, string(v))
+		}
+		return nil
+	},
 }
 
 var featureFlagCmd = &cobra.Command{
@@ -45,12 +83,12 @@ If you give a null value, the flag will be removed.
 			return err
 		}
 		defer res.Body.Close()
-		var obj map[string]interface{}
+		var obj map[string]json.RawMessage
 		if err = json.NewDecoder(res.Body).Decode(&obj); err != nil {
 			return err
 		}
 		for k, v := range obj {
-			fmt.Printf("- %s: %v\n", k, v)
+			fmt.Printf("- %s: %s\n", k, string(v))
 		}
 		return nil
 	},
@@ -62,7 +100,8 @@ var featureSetCmd = &cobra.Command{
 	Long: `
 cozy-stack feature sets displays the feature sets coming from the manager.
 
-It can also take a list of sets that will replace the previous list (no merge).
+It can also take a space-separated list of sets that will replace the previous
+list (no merge).
 
 All the sets can be removed by setting an empty list ('').
 `,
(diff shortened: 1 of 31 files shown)


docs/settings.md
```diff
@@ -633,3 +635,142 @@ Cookie: sessionid=xxxx
...
+
+- the flags set on the instance by the command `cozy-stack feature flags`
+- the response of the manager for `GET /api/v1/flags?sets=s1,s2&context=ctx`
+  (where `s1,s2` has been set via `cozy-stack feature sets`)
+- the flags coming from the context configuration (`context.<name>.features`)
+- the flags set on the context by the command `cozy-stack feature ratio`
+- the default value (`cozy-stack feature default`).
+
+For a given flag, the stack takes the value from the source with the highest
+priority.
```
sblaisot: Please explicitely state here, in case of a json type flag, if they are merged or remplaced by higher-priority value. I guess replaced but it's better if it's explicit.

docs/settings.md
```diff
@@ -633,3 +635,142 @@ Cookie: sessionid=xxxx
...
+
+### GET /settings/flags
+
+This endpoint returns the computed list of feature flags for the given
+instance. It accepts a `include` parameter in the query string to see the
```
sblaisot: Nit: an `include` parameter

docs/settings.md
```diff
@@ -633,3 +635,142 @@ Cookie: sessionid=xxxx
 
 To use this endpoint, an application needs a valid token, but no explicit
 permission is required.
+
+## Feature flags
+
+A feature flag is a name and an associated value (boolean, number, string or a
+JSON) that can be interpreted by the apps. It can be used for giving access to
+paid features, or to enable progressively a feature on all the cozy instances
```
sblaisot: Nit: to enable a feature progressively on all the cozy instances

docs/settings.md
```diff
@@ -633,3 +635,142 @@ Cookie: sessionid=xxxx
...
+- the default value (`cozy-stack feature default`).
+
+For a given flag, the stack takes the value from the source with the highest
+priority.
+
+### GET /settings/flags
+
+This endpoint returns the computed list of feature flags for the given
+instance. It accepts a `include` parameter in the query string to see the
+details of how the flags were computed.
```
sblaisot: What are the type and acceptable values for the include parameter ?

docs/settings.md
```diff
@@ -633,3 +635,142 @@ Cookie: sessionid=xxxx
 
 To use this endpoint, an application needs a valid token, but no explicit
 permission is required.
+
+## Feature flags
+
+A feature flag is a name and an associated value (boolean, number, string or a
+JSON) that can be interpreted by the apps. It can be used for giving access to
```
sblaisot: We already have a feature flag of type `list` of `string` defined in cozy.yml :
```
- home_hidden_app:
  - app1
  - app2
```
Will they be converted to a json object containing a list of string ? Is it already the case ?

nono: JSON values can be objects or arrays. Here, this flag will be an array of strings.

docs/settings.md
nono: I thought it was explicit: it takes the value from the source with the highest priority, and doesn't look at the other sources (no merge).

docs/settings.md
nono: The query string accepts only one type: `string`. And for the acceptable values, it is not yet defined (in the example, I put `source`, but it may change).

cmd/feature.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+		return nil
+	},
+}
+
+var featureContextCmd = &cobra.Command{
+	Use:   "context <context-name>",
+	Short: `Display and update the feature flags for a context`,
+	Long: `
+cozy-stack feature defaults displays the feature flags for a context.
```
sblaisot: s/defaults/context/

cmd/feature.go
```diff
@@ -0,0 +1,194 @@
...
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/cozy/cozy-stack/client/request"
+	"github.com/spf13/cobra"
+)
+
+var featureCmdGroup = &cobra.Command{
+	Use:   "feature <command>",
```
sblaisot: `feature` in singular is not coherent with `instances` in plural.
At least call it `features` and add an `feature` alias

cmd/feature.go
```diff
@@ -0,0 +1,194 @@
...
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(args) == 0 || args[0] == "" {
+			return cmd.Usage()
+		}
+		c := newAdminClient()
+		req := request.Options{
+			Method: "GET",
+			Path:   fmt.Sprintf("/instances/feature/contexts/%s", args[0]),
+		}
+		if len(args) > 1 {
```
sblaisot: IMHO you should check that the sum of ratio is 1 to at least return a warning

cmd/feature.go
```diff
@@ -0,0 +1,194 @@
...
+	Example: `$ cozy-stack feature defaults '{"add_this_flag": true, "remove_this_flag": null}'`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		c := newAdminClient()
+		req := request.Options{
+			Method: "GET",
+			Path:   "/instances/feature/defaults",
+		}
+		if len(args) > 0 {
+			req.Method = "PATCH"
+			req.Body = strings.NewReader(args[0])
```
sblaisot: IMHO inputs should be validated. maybe not here but at least in the `/instances/feature/defaults` route

cmd/feature.go
```diff
@@ -0,0 +1,194 @@
...
+	Use:   "context <context-name>",
+	Short: `Display and update the feature flags for a context`,
+	Long: `
+cozy-stack feature defaults displays the feature flags for a context.
+
+It can also create, update, or remove flags (with a ratio and value).
+
+To remove a flag, set it to an empty array (or null).
+`,
+	Example: `$ cozy-stack feature context beta '{"set_this_flag": [{"ratio": 0.1, "value": 1}, {"ratio": 0.9, "value": 2}] }'`,
```
sblaisot: IMHO using `--domain` to `cozy-stack feature {set, flags}` and not using `--context` for `cozy-stack feature context` is hardly understandable and prone to constant errors.

cmd/feature.go
```diff
@@ -0,0 +1,194 @@
...
+var featureCmdGroup = &cobra.Command{
+	Use:   "feature <command>",
+	Short: "Manage the feature flags",
+}
+
+var featureFlagCmd = &cobra.Command{
+	Use:   "flags",
+	Short: `Display and update the feature flags for an instance`,
+	Long: `
+cozy-stack feature flags displays the feature flags that are specific to an instance.
```
sblaisot: IIUC this command let you get the instance-specifig flags only. IMHO we should also have a cozy-stack command to display the resulting flags for an instance applying defaults, context and feature-set herited flags. either with a switch to this command or with another command.

cmd/feature.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+		res, err := c.Req(&req)
+		if err != nil {
+			return err
+		}
+		defer res.Body.Close()
+		var obj map[string]interface{}
+		if err = json.NewDecoder(res.Body).Decode(&obj); err != nil {
+			return err
+		}
```
sblaisot: probably we should print here that we are showing only instance-level flags an not fully resulting flags for instance based on default, context and feature-set flags.

cmd/feature.go
```diff
@@ -0,0 +1,194 @@
...
+	},
+}
+
+var featureSetCmd = &cobra.Command{
+	Use:   "sets",
+	Short: `Display and update the feature sets for an instance`,
+	Long: `
+cozy-stack feature sets displays the feature sets coming from the manager.
+
+It can also take a list of sets that will replace the previous list (no merge).
```
sblaisot: Nit: a space-separated list

cmd/feature.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+		res, err := c.Req(&req)
+		if err != nil {
+			return err
+		}
+		defer res.Body.Close()
+		var obj map[string]json.RawMessage
+		if err = json.NewDecoder(res.Body).Decode(&obj); err != nil {
+			return err
+		}
```
sblaisot: Maybe you should print here "feature flags set for context {}:"

cmd/feature.go
```diff
@@ -0,0 +1,194 @@
...
+		res, err := c.Req(&req)
+		if err != nil {
+			return err
+		}
+		defer res.Body.Close()
+		var obj map[string]interface{}
+		if err = json.NewDecoder(res.Body).Decode(&obj); err != nil {
+			return err
+		}
```
sblaisot: Maybe printing here "Default feature flags:"

model/instance/manager.go
```diff
@@ -16,6 +16,9 @@ const (
 	ManagerPremiumURL
 	// ManagerBlockedURL is the kind for a redirection of a blocked instance.
 	ManagerBlockedURL
+	// ManagerFeatureSetsURL is the kind for an API call to list the feature
+	// flags associated to some feature sets.
+	ManagerFeatureSetsURL
```
sblaisot: I think there is a problem here as we already discussed earlier. manager_url is currently used to return a customer-accessible URL to manager frontend

To access API from the stack we have a `cloudery:<context>:url` configuration in `cozy.yml`

Can you check that you use the right endpoint ?

cmd/feature.go
nono: Well, we can have a sum of ratio < 1: if no values is selected for this flag, it will use the default from `cozy-stack features defaults`

cmd/feature.go
nono: What validation do you expect? Do you have an example of invalid input that we should reject (and is not yet rejected)?

cmd/feature.go
sblaisot: This is why I talked about a warning and not an error. I don't think it's a common use case to define ratio sum < 1, at least the admin user doing this should be warned about it I think

cmd/feature.go
sblaisot: req.Body is supposed to be valid json.
`foo{baf"baz}` is no valid json.
What do we get when monkey-typing parameters? is that a clean error or a global crash?

nono: We already check that:

```
$ cozy-stack feature defaults '{baf"baz}' 
Error: Unqualified error: invalid character 'b' looking for beginning of object key string
exit status 1
```

sblaisot: oh, ok, I didn't see that, so forget about it.

cmd/feature.go
nono: > I don't think it's a common use case to define ratio sum < 1

When discussed with Éric, it was the common case: the ratio is used on the feature flag activated (with a default for not being activated).

model/instance/manager.go
nono: Good catch, it is not the right endpoint. I'm going to fix it.

cmd/feature.go
sblaisot: ok

model/feature/flag.go
```diff
@@ -0,0 +1,270 @@
...
+
+var (
+	cacheDuration      = 12 * time.Hour
+	managerHTTPClient  = &http.Client{Timeout: 5 * time.Second}
+	errInvalidResponse = errors.New("Invalid response from the manager")
+)
+
+func getFlagsFromManager(inst *instance.Instance) (map[string]interface{}, error) {
+	cache := config.GetConfig().CacheStorage
+	cacheKey := fmt.Sprintf("flags:%s:%v", inst.ContextName, inst.FeatureSets)
```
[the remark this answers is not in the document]
nono: I don't think we need a bijection between offers/plans and stack context, only that a context always give the same feature flags. That said, I also think it would be a good thing to have this bijection.

model/feature/flag.go
```diff
@@ -0,0 +1,270 @@
...
+	var flags map[string]interface{}
+	if buf, ok := cache.Get(cacheKey); ok {
+		if err := json.Unmarshal(buf, &flags); err == nil {
+			return flags, nil
+		}
+	}
+
+	managerURL, err := inst.ManagerURL(instance.ManagerFeatureSetsURL)
```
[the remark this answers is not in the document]
nono: Correct, it has been fixed.

model/feature/flag.go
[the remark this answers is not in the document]
aeris: There is no bijection at the moment, we need in theory the full tuple partner/offer/context/plans to got the flags. In practice, features are only on plans, associated to partner.
So you need to ask flags with lists of instance plan ids, which is provided by the cloudery too on creation and subscription.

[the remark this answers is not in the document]
sblaisot: ok, we talked about it, feature set list is sufficient and there is not problem as feature sets will be ids of feature sets unique to the whole environement.
so no problem here (and cloudery can use or ignore context, whatever)

cmd/feature.go
```diff
@@ -7,12 +7,50 @@ import (
...
 
 	"github.com/cozy/cozy-stack/client/request"
+	"github.com/cozy/cozy-stack/pkg/consts"
 	"github.com/spf13/cobra"
 )
 
 var featureCmdGroup = &cobra.Command{
-	Use:   "feature <command>",
-	Short: "Manage the feature flags",
+	Use:     "features <command>",
```
sblaisot: maybe you need to update descriptions and examples to also add that S ?

cmd/feature.go
```diff
@@ -7,12 +7,50 @@ import (
...
+	Aliases: []string{"feature"},
+	Short:   "Manage the feature flags",
+}
+
+var featureShowCmd = &cobra.Command{
+	Use:   "show",
+	Short: `Display the computed feature flags for an instance`,
+	Long: `
+cozy-stack feature show displays the feature flags that are shown by apps.
```
sblaisot: probably we will someday need a `--source` or `--details` flag to get the whole computed flags sources like you have in the endpoint;

PR #2223, opened 2019-11-25
Add accounts to organization migration

Gara64:
This migrates all accounts from an instance to an encrypted cipher which belong to the cozy organization, so the password manager retrieve them automatically.

### DIFF
--- a/docs/workers.md
+++ b/docs/workers.md
@@ -306,7 +306,7 @@ The message is composed of a sharing ID and a count of the number of errors
 ## migrations
 
 The `migrations` worker can be used to migrate a cozy instance. Currently, it
-has a single option, `type`, with two supported value:
+has a single option, `type`, with two supported values:
 * `to-swift-v3`: migrate a cozy instance that has files in swift from a V1 or V2 layout to a V3 layout.
 * `accounts-to-organization`: create [ciphers](https://docs.cozy.io/en/cozy-doctypes/docs/com.bitwarden.ciphers/)
  from [accounts](https://docs.cozy.io/en/cozy-doctypes/docs/io.cozy.accounts/),
--- a/docs/workers.md
+++ b/docs/workers.md
@@ -305,9 +305,14 @@ The message is composed of a sharing ID and a count of the number of errors
 
 ## migrations
 
-The `migrations` worker can be used to migrate a cozy instance that has its
-files in swift from a V1 or V2 layout to a V3 layout. Currently, it has a
-single option, `type`, with a single supported value, `to-swift-v3`.
+The `migrations` worker can be used to migrate a cozy instance. Currently, it
+has a single option, `type`, with two supported value:
+* `to-swift-v3`: migrate a cozy instance that has files in swift from a V1 or V2 layout to a V3 layout.
+* `accounts-to-organization`: create [ciphers](https://docs.cozy.io/en/cozy-doctypes/docs/com.bitwarden.ciphers/)
+ from [accounts](https://docs.cozy.io/en/cozy-doctypes/docs/io.cozy.accounts/),
+ re-encrypted with the organization key.
+
+
 
 ### Example
 
--- a/model/account/accounts.go
+++ b/model/account/accounts.go
@@ -47,9 +47,10 @@ type OauthInfo struct {
 
 // BasicInfo holds configuration information for an user/pass account
 type BasicInfo struct {
-	Login      string `json:"login,omitempty"`
-	Password   string `json:"password,omitempty"`
-	FolderPath string `json:"folderPath,omitempty"`
+	Login                string `json:"login,omitempty"`
+	Password             string `json:"password,omitempty"` // used when no encryption
+	EncryptedCredentials string `json:"credentials_encrypted"`
+	FolderPath           string `json:"folderPath,omitempty"`
 }
 
 // ID is used to implement the couchdb.Doc interface
--- a/model/account/accounts.go
+++ b/model/account/accounts.go
@@ -49,7 +49,7 @@ type OauthInfo struct {
 type BasicInfo struct {
 	Login                string `json:"login,omitempty"`
 	Password             string `json:"password,omitempty"` // used when no encryption
-	EncryptedCredentials string `json:"credentials_encrypted"`
+	EncryptedCredentials string `json:"credentials_encrypted,omitempty"`
 	FolderPath           string `json:"folderPath,omitempty"`
 }
 
--- a/model/account/accounts.go
+++ b/model/account/accounts.go
@@ -10,21 +10,23 @@ import (
 	"github.com/cozy/cozy-stack/pkg/consts"
 	"github.com/cozy/cozy-stack/pkg/couchdb"
 	"github.com/cozy/cozy-stack/pkg/logger"
+	"github.com/cozy/cozy-stack/pkg/metadata"
 	"github.com/cozy/cozy-stack/pkg/prefixer"
 	"github.com/hashicorp/go-multierror"
 )
 
 // Account holds configuration information for an account
 type Account struct {
-	DocID       string                 `json:"_id,omitempty"`
-	DocRev      string                 `json:"_rev,omitempty"`
-	Name        string                 `json:"name"`
-	AccountType string                 `json:"account_type"`
-	FolderPath  string                 `json:"folderPath,omitempty"`
-	Basic       *BasicInfo             `json:"auth,omitempty"`
-	Oauth       *OauthInfo             `json:"oauth,omitempty"`
-	Extras      map[string]interface{} `json:"oauth_callback_results,omitempty"`
-
+	DocID         string                 `json:"_id,omitempty"`
+	DocRev        string                 `json:"_rev,omitempty"`
+	Name          string                 `json:"name"`
+	AccountType   string                 `json:"account_type"`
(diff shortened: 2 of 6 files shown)


model/account/accounts.go
```diff
@@ -47,9 +47,10 @@ type OauthInfo struct {
 
 // BasicInfo holds configuration information for an user/pass account
 type BasicInfo struct {
-	Login      string `json:"login,omitempty"`
-	Password   string `json:"password,omitempty"`
-	FolderPath string `json:"folderPath,omitempty"`
+	Login                string `json:"login,omitempty"`
+	Password             string `json:"password,omitempty"` // used when no encryption
+	EncryptedCredentials string `json:"credentials_encrypted"`
```
nono: [nit] We should probably put `omitempty` on this field too

worker/migrations/migrations.go
```diff
@@ -65,6 +74,8 @@ func worker(ctx *job.WorkerContext) error {
 		return migrateToSwiftV3(ctx.Instance.Domain)
 	case swiftV1ToV2:
 		return fmt.Errorf("this migration type is no longer supported")
+	case accountsToOrganization:
+		return migrateAccountsToOrganization(ctx.Instance.Domain)
```
nono: Can you update documentation: https://github.com/cozy/cozy-stack/blob/master/docs/workers.md#migrations ?

worker/migrations/migrations.go
```diff
@@ -80,6 +91,137 @@ func commit(ctx *job.WorkerContext, err error) error {
...
+	if err != nil {
+		return err
+	}
+	setting, err := settings.Get(inst)
+	if err != nil {
+		return err
+	}
+	// Get org key
+	var orgKey []byte
+	orgKey, err = setting.OrganizationKey()
```
nono: [nit] We can avoid the orgKey declaration with `orgkey, err := setting.OrganizationKey()`

worker/migrations/migrations.go
```diff
@@ -80,6 +91,137 @@ func commit(ctx *job.WorkerContext, err error) error {
...
+	if err == settings.ErrMissingOrgKey {
+		// Create org key if missing
+		if err := setting.EnsureCozyOrganization(inst); err != nil {
```
nono: I think `EnsureCozyOrganization` modifies the in-memory version of the setting document but does not persist it to CouchDB. Is it expected that calling `settings.UpdateRevisionDate(inst, setting)` in the loop will persist it?

worker/migrations/migrations.go
```diff
@@ -80,6 +91,137 @@ func commit(ctx *job.WorkerContext, err error) error {
...
+			return err
+		}
+		orgKey, err = setting.OrganizationKey()
+		if err != nil {
+			return err
+		}
+	} else if err != nil {
+		return err
+	}
```
nono: In fact, I think we can simplify this code to:

```go
if err := setting.EnsureCozyOrganization(inst); err != nil {
    return err
}
orgkey, err := setting.OrganizationKey()
if err != nil {
    return err
}
```

nono: In general, `EnsureSomething` methods can be called safely (they are idempotent).

worker/migrations/migrations.go
```diff
@@ -80,6 +91,137 @@ func commit(ctx *job.WorkerContext, err error) error {
...
+	var msg struct {
+		Account string `json:"account"`
+		Slug    string `json:"konnector"`
+	}
+	for _, t := range triggers {
+		if t.Infos().WorkerType != "konnector" {
+			continue
+		}
+		err := t.Infos().Message.Unmarshal(&msg)
+		if err == nil && msg.Account != "" && msg.Slug != "" {
```
nono: In go, we often prefer to use shortcuts instead of long `if` branches. Here, we can do:

```go
if err != nil || msg.Account == "" || msg.Slug == "" {
    continue
}
```

worker/migrations/migrations.go
```diff
@@ -80,6 +91,137 @@ func commit(ctx *job.WorkerContext, err error) error {
...
+				return err
+			}
+			cipher, err := createCipher(orgKey, msg.Slug, login, password, string(link))
+			if err != nil {
+				return err
+			}
+			if err := couchdb.CreateDoc(inst, cipher); err != nil {
+				return err
+			}
+			settings.UpdateRevisionDate(inst, setting)
```
nono: [nit] We can move this line outside of the loop to avoid sending several requests to CouchDB to update the same doc.

worker/migrations/migrations.go
```diff
@@ -80,6 +91,137 @@ func commit(ctx *job.WorkerContext, err error) error {
...
+		if t.Infos().WorkerType != "konnector" {
+			continue
+		}
+		err := t.Infos().Message.Unmarshal(&msg)
+		if err == nil && msg.Account != "" && msg.Slug != "" {
+			manifest, err := app.GetKonnectorBySlug(inst, msg.Slug)
+			if err != nil {
+				return err
+			}
+			vLink, err := json.Marshal(manifest.VendorLink)
```
nono: This line looks wrong. I don't see why we would want to serialize to JSON a link. Don't we want the opposite: parse a JSON message?

worker/migrations/migrations.go
```diff
@@ -80,6 +91,137 @@ func commit(ctx *job.WorkerContext, err error) error {
...
+				return err
+			}
+			link := string(vLink)
+			// Remove quotes
+			if len(link) > 0 && link[0] == '"' {
+				link = link[1:]
+			}
+			if len(link) > 0 && link[len(link)-1] == '"' {
+				link = link[:len(link)-1]
+			}
```
nono: We can use `strings.TrimPrefix` and `strings.TrimSuffix` (or maybe even `strings.Trim` if we don't want to keep quotes if there are several quotes at the beginning or end of the string).

worker/migrations/migrations.go
```diff
@@ -80,6 +91,137 @@ func commit(ctx *job.WorkerContext, err error) error {
...
+			acc := &account.Account{}
+			if err := couchdb.GetDoc(inst, consts.Accounts, msg.Account, acc); err != nil {
+				return nil
+			}
+			encryptedCreds := acc.Basic.EncryptedCredentials
+			login, password, err := account.DecryptCredentials(encryptedCreds)
+			if err != nil {
+				return err
+			}
+			cipher, err := createCipher(orgKey, msg.Slug, login, password, string(link))
```
nono: [nit] I would have thought that a function named `createSomething` will create the something in CouchDB.

worker/migrations/migrations.go
Gara64: Yes, but wouldn't it be better if the migration stops before the end?

nono: Well, it depends on your expectation. Do you plan to replay the migration if it fails?

If not, we can use a `multierror` to keep the errors while iterating (without stopping the loop when there is an error). And when the loop is done, we call `settings.UpdateRevisionDate`, and only after that, we can return the multierror.

docs/workers.md
```diff
@@ -305,9 +305,14 @@ The message is composed of a sharing ID and a count of the number of errors
 
 ## migrations
 
-The `migrations` worker can be used to migrate a cozy instance that has its
-files in swift from a V1 or V2 layout to a V3 layout. Currently, it has a
-single option, `type`, with a single supported value, `to-swift-v3`.
+The `migrations` worker can be used to migrate a cozy instance. Currently, it
+has a single option, `type`, with two supported value:
```
nono: s/value/values/

worker/migrations/migrations.go
```diff
@@ -129,51 +123,46 @@ func migrateAccountsToOrganization(domain string) error {
...
-			if err := couchdb.CreateDoc(inst, cipher); err != nil {
-				return err
-			}
-			settings.UpdateRevisionDate(inst, setting)
+		if err != nil || msg.Account == "" || msg.Slug == "" {
+			continue
+		}
+		manifest, err := app.GetKonnectorBySlug(inst, msg.Slug)
+		if err != nil {
+			errm = multierror.Append(errm, err)
```
nono: I think we should also put a `continue`, as with an empty manifest, we will accumulate several other errors for this trigger.

worker/migrations/migrations.go
```diff
@@ -129,51 +123,46 @@ func migrateAccountsToOrganization(domain string) error {
...
+		acc := &account.Account{}
+		if err := couchdb.GetDoc(inst, consts.Accounts, msg.Account, acc); err != nil {
+			errm = multierror.Append(errm, err)
+		}
+		encryptedCreds := acc.Basic.EncryptedCredentials
+		login, password, err := account.DecryptCredentials(encryptedCreds)
+		if err != nil {
+			errm = multierror.Append(errm, err)
+		}
+		cipher, err := buildCipher(orgKey, msg.Slug, login, password, string(link))
```
nono: With the new commits, `link` is a string, there is no need to cast it

worker/migrations/migrations.go
Gara64: Right, it is also true when there is an error on the account doc retrieval

nono: Yes, and we should check the other places where multierror is used. At least, the `buildCipher` error should also have its `continue`.

PR #2268, opened 2019-12-11
Improve feature flags CLI

nono:


scripts/completion/cozy-stack.bash
```diff
@@ -1368,8 +1368,16 @@ _cozy-stack_features()
 
     commands=()
     commands+=("context")
+    if [[ -z "${BASH_VERSION}" || "${BASH_VERSINFO[0]}" -gt 3 ]]; then
+        command_aliases+=("context")
+        aliashash["context"]="context"
```
sblaisot: `="ratio"` ?

scripts/completion/cozy-stack.bash
```diff
@@ -1368,8 +1368,16 @@ _cozy-stack_features()
 
     commands=()
     commands+=("context")
```
sblaisot: `commands+=("ratio")` ?

cmd/feature.go
```diff
@@ -164,17 +165,18 @@ All the sets can be removed by setting an empty list ('').
 	},
 }
 
-var featureContextCmd = &cobra.Command{
-	Use:   "context <context-name>",
-	Short: `Display and update the feature flags for a context`,
+var featureRatioCmd = &cobra.Command{
+	Use:     "context <context-name>",
```
sblaisot: s/context/ratio/ ?

nono: Good catch :+1:

PR #2282, opened 2019-12-17
Show latency information on /status

nono:


nono:
> Maybe it could be useful to add a global latency (time to do all checks) ?

The checks are made in parallel, so I don't think it is relevant.

PR #2287, opened 2019-12-18
Fix / encoding of reference IDs

nono:
The docid parameter for creating/listing/removing references was not decoded correctly: if a %2f was present, it should be interpreted as a /. This commit fixes that, but it also adds some fallback mechanisms to avoid that references that were created with the bug cannot be found and deleted.

### DIFF
--- a/web/data/references_test.go
+++ b/web/data/references_test.go
@@ -167,6 +167,115 @@ func TestRemoveReferencesHandler(t *testing.T) {
 	assert.Len(t, fdoc9.ReferencedBy, 1)
 }
 
+func TestReferencesWithSlash(t *testing.T) {
+	// Make File
+	name := "test-ref-with-slash.txt"
+	dirID := consts.RootDirID
+	filedoc, err := vfs.NewFileDoc(name, dirID, -1, nil, "", "", time.Now(), false, false, nil)
+	if !assert.NoError(t, err) {
+		return
+	}
+	f, err := testInstance.VFS().CreateFile(filedoc, nil)
+	if !assert.NoError(t, err) {
+		return
+	}
+	if err = f.Close(); !assert.NoError(t, err) {
+		return
+	}
+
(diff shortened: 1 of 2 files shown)


web/data/references_test.go
```diff
@@ -167,6 +167,115 @@ func TestRemoveReferencesHandler(t *testing.T) {
...
+	f, err := testInstance.VFS().CreateFile(filedoc, nil)
+	if !assert.NoError(t, err) {
+		return
+	}
+	if err = f.Close(); !assert.NoError(t, err) {
+		return
+	}
+
+	// Add a reference to io.cozy.apps/foobar
+	url := ts.URL + "/data/" + Type + "/io.cozy.apps%2ffoobar/relationships/references"
```
sblaisot: We should also probably add a test with %2F (upcase F)

web/files/references.go
```diff
@@ -209,6 +220,17 @@ func RemoveReferencesHandler(c echo.Context) error {
...
 	}
 
+	// XXX References with an ID that contains a / could have it encoded as %2F
+	// (legacy). We delete the references for both versions to preserve
+	// compatibility.
+	var altRef *couchdb.DocReference
+	if strings.Contains(id, "/") {
+		altRef = &couchdb.DocReference{
+			Type: doctype,
+			ID:   strings.ReplaceAll(id, "/", "%2f"),
```
sblaisot: are you sure no doc have it encoded with an upcase F ?

web/files/references.go
```diff
@@ -236,3 +264,8 @@ func RemoveReferencesHandler(c echo.Context) error {
 
 	return c.NoContent(204)
 }
+
+func getDocID(c echo.Context) string {
+	id := c.Param("docid")
+	return strings.ReplaceAll(id, "%2f", "/")
```
sblaisot: should also be done for `%2F` (upcase F)

PR #2309, opened 2020-01-08
Add a route to POST documents for the realtime

nono:
The route POST /realtime/:doctype/:id can be used to send a JSON payload that can be received in the realtime, but won't be persisted to CouchDB. For example, it can be used by konnectors to send their progress to the Cozy Home app.

### DIFF
--- a/web/realtime/realtime.go
+++ b/web/realtime/realtime.go
@@ -10,6 +10,8 @@ import (
 	"github.com/cozy/cozy-stack/model/instance"
 	"github.com/cozy/cozy-stack/model/permission"
 	"github.com/cozy/cozy-stack/pkg/consts"
+	"github.com/cozy/cozy-stack/pkg/couchdb"
+	"github.com/cozy/cozy-stack/pkg/jsonapi"
 	"github.com/cozy/cozy-stack/pkg/logger"
 	"github.com/cozy/cozy-stack/pkg/prefixer"
 	"github.com/cozy/cozy-stack/pkg/realtime"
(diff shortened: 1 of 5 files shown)


ptbrowne:
Cool, I added support for those events in cozy-realtime https://github.com/cozy/cozy-libs/pull/908

web/realtime/realtime.go
```diff
@@ -283,7 +286,34 @@ func ws(c echo.Context) error {
...
 
+// Notify is the API handler for POST /realtime/:doctype/:id: this route can be
+// used to send documents in the real-time without having to persist them in
+// CouchDB.
+func Notify(c echo.Context) error {
+	inst := middlewares.GetInstance(c)
+	doctype := c.Param("doctype")
+	id := c.Param("id")
+
+	if err := permission.CheckReadable(doctype); err != nil {
```
ptbrowne: Should we check that the doctype is Writable ?

web/realtime/realtime.go
```diff
@@ -283,7 +286,34 @@ func ws(c echo.Context) error {
...
+		return jsonapi.BadRequest(err)
+	}
+
+	var payload couchdb.JSONDoc
+	if err := c.Bind(&payload); err != nil {
```
ptbrowne: What does Bind mean in this context ?

web/realtime/realtime_test.go
```diff
@@ -166,12 +168,56 @@ func TestWSSuccess(t *testing.T) {
 	assert.Equal(t, "bar-one", payload["id"])
 }
 
+func TestWSNotify(t *testing.T) {
+	u := strings.Replace(ts.URL+"/realtime/", "http", "ws", 1)
+	c, _, err := websocket.DefaultDialer.Dial(u, nil)
```
ptbrowne: c could be renamed ws for clarity ? At first I thought it was the context.

web/realtime/realtime.go
nono: I don't think so. For example, the `io.cozy.jobs` are not writable, but I think it is OK to send real-time docs about them.

web/realtime/realtime.go
nono: It is a method from echo, out web framework. It fills the payload with the body of the request.

PR #2315, opened 2020-01-14
Migrate accounts when extension is installed

Gara64:
When the extension is installed for the first time, we want the user to get its existing account. Therefore, we run the dedicated worker on the pre-login route, only if the `extension_installed` flag is set to false.

### DIFF
--- a/web/bitwarden/bitwarden.go
+++ b/web/bitwarden/bitwarden.go
@@ -12,6 +12,7 @@ import (
 	"github.com/cozy/cozy-stack/model/bitwarden/settings"
 	"github.com/cozy/cozy-stack/model/instance"
 	"github.com/cozy/cozy-stack/model/instance/lifecycle"
+	"github.com/cozy/cozy-stack/model/job"
 	"github.com/cozy/cozy-stack/model/oauth"
 	"github.com/cozy/cozy-stack/model/permission"
 	"github.com/cozy/cozy-stack/model/session"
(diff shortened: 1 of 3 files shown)


web/bitwarden/bitwarden.go
```diff
@@ -22,14 +23,40 @@ import (
...
+		Message:    msg,
+	})
+	return err
+}
+
 // Prelogin tells to the client how many KDF iterations it must apply when
 // hashing the master password.
 func Prelogin(c echo.Context) error {
 	inst := middlewares.GetInstance(c)
+	log := inst.Logger().WithField("nspace", "pre-login")
```
nono: I'd have used `bitwarden` for the logging namespace

web/bitwarden/bitwarden.go
```diff
@@ -22,14 +23,40 @@ import (
...
 	settings, err := settings.Get(inst)
 	if err != nil {
 		return err
 	}
+	if !settings.ExtensionInstalled {
+		// This is the first time the bitwarden extension is installed: make sure
+		// the user gets the existing accounts into the vault
+		if err := migrateAccountsToCiphers(inst); err != nil {
+			log.Errorf("Account migration failed : %s", err)
```
nono: I think we can improve the log message. Maybe something like: `Cannot push job for ciphers migration: %s`

PR #2324, opened 2020-01-17
Add a deleting flag on instance

nono:
When deleting an instance, the first step is to clean the accounts. This step must not run twice in parallel if the instance deletion is runned twice. And if it fails, it must be manually managed. So, we have added a `deleting` flag on the instance to avoid those issues.

### DIFF
--- a/web/middlewares/instance.go
+++ b/web/middlewares/instance.go
@@ -18,15 +18,20 @@ func NeedInstance(next echo.HandlerFunc) echo.HandlerFunc {
 			return next(c)
 		}
 		i, err := lifecycle.GetInstance(c.Request().Host)
-		if err != nil || i.Deleting {
+		if err != nil {
 			var errHTTP *echo.HTTPError
 			switch err {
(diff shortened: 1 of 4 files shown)


web/middlewares/instance.go
```diff
@@ -18,10 +18,10 @@ func NeedInstance(next echo.HandlerFunc) echo.HandlerFunc {
 			return next(c)
 		}
 		i, err := lifecycle.GetInstance(c.Request().Host)
-		if err != nil {
+		if err != nil || i.Deleting {
 			var errHTTP *echo.HTTPError
 			switch err {
-			case instance.ErrNotFound, instance.ErrIllegalDomain:
+			case instance.ErrNotFound, instance.ErrIllegalDomain, nil:
```
ptbrowne: I think it would be clearer if this was extracted from the switch case.

```
if i.Deleting {
  err = instance.ErrNotFound
  errHTTP = echo.NewHTTPError(http.StatusNotFound, err)
}
```

nono: Good idea

PR #2335, opened 2020-01-21
feat: Add Sentry to allowed CSP url

Crash--:
Currently, we need to use a remote doctype to be able to use the Sentry API when using a webapp. Since we don't need on mobile app, let's add this URL to our allowed list of domain. 

This way, it will be easier to use Sentry for our web apps


Crash--:
🍾

Issue #1504, opened 2018-08-28
cozy-stack {apps,konnectors} CLI returns the whole manifest


ptbrowne:
When updating an application, I only need the version that has been installed and can use `show` if I need all the information.

What could be cool : 
```
$ cozy-stack konnectors update toto
toto has been updated to 1.0.0

$ cozy-stack konnectors update toto
toto is already up-to-date at 1.0.0
```

nono:
Fixed by #2340

PR #2340, opened 2020-01-22
Print useful message instead of manifest when updating apps/konnector…

mamyn0va:
This PR fixes the issue #1504.
It aims at printing a useful message on the CLI instead of manifest when updating apps/konnectors.

### DIFF
--- a/cmd/apps.go
+++ b/cmd/apps.go
@@ -362,21 +362,27 @@ func updateApp(cmd *cobra.Command, args []string, appType string) error {
 	}
 
 	c := newClient(flagDomain, appType)
-	manifest, err := c.UpdateApp(&client.AppOptions{
+	opts := &client.AppOptions{
 		AppType:   appType,
 		Slug:      args[0],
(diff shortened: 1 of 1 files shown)


cmd/apps.go
```diff
@@ -362,21 +362,27 @@ func updateApp(cmd *cobra.Command, args []string, appType string) error {
...
 	if err != nil {
 		return err
 	}
-	json, err := json.MarshalIndent(manifest.Attrs, "", "  ")
+	newManifest, err := c.UpdateApp(opts, flagSafeUpdate)
 	if err != nil {
 		return err
 	}
-	fmt.Println(string(json))
+	msg := "%s is already up-to-date at %s"
```
nono: Why this message has no `\n`, but the message for an updating app has one?

mamyn0va: It's a mistake. I forgot it.

nono:
Thanks!

PR #2345, opened 2020-01-27
Add a route /notes/:id/open

nono:
This route can be used by clients to get information about a note and build the URL where the note can be opened.

### DIFF
--- a/model/note/open.go
+++ b/model/note/open.go
@@ -6,8 +6,8 @@ import (
 
 	"github.com/cozy/cozy-stack/client/request"
 	"github.com/cozy/cozy-stack/model/instance"
-	"github.com/cozy/cozy-stack/model/permission"
 	"github.com/cozy/cozy-stack/model/sharing"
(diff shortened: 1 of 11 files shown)


model/note/open.go
```diff
@@ -0,0 +1,140 @@
...
+		doc = openLocalNote(inst, fileID, code)
+	} else {
+		doc, err = openSharedNote(inst, sharing, fileID)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// Enforce DocID and PublicName with local values
+	doc.DocID = fileID
```
taratatach: Shouldn't you use `doc.SetID(fileID)` here?

nono: We don't care. The `SetID` is used by packages like `couchdb` and `jsonapi` that needs a generic way to manipulate a lot of doctypes (well, their types in go). But, when we are accessing/setting an ID for a struct in the same module, we often prefer the direct way to just assign it.

PR #2349, opened 2020-01-29
Fix conflict on account migration

Gara64:
The setting update might fail at the end of the migration worker if the settings were updated during the process.

### DIFF
--- a/worker/migrations/migrations.go
+++ b/worker/migrations/migrations.go
@@ -233,17 +233,31 @@ func migrateAccountsToOrganization(domain string) error {
 			errm = multierror.Append(errm, err)
 		}
 	}
+	// Reload the setting in case the revision changed
+	setting, err = settings.Get(inst)
+	if err != nil {
+		errm = multierror.Append(errm, err)
+		return errm
+	}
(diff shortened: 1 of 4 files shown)


worker/migrations/migrations.go
```diff
@@ -237,6 +237,12 @@ func migrateAccountsToOrganization(domain string) error {
 	setting.ExtensionInstalled = true
 	settings.UpdateRevisionDate(inst, setting)
 	if err != nil {
+		if couchdb.IsConflictError(err) {
```
nono: :warning: It looks like the error doesn't come from saving the settings, but from before!

worker/migrations/migrations.go
```diff
@@ -237,6 +237,12 @@ func migrateAccountsToOrganization(domain string) error {
...
+			// The settings have been updated elsewere: retry
+			setting, err = settings.Get(inst)
+			setting.ExtensionInstalled = true
+			settings.UpdateRevisionDate(inst, setting)
+		}
 		errm = multierror.Append(errm, err)
```
nono: We should not happen the error if the retry has succeed.

PR #2353, opened 2020-01-29
Fix missing stamp in refresh token

Gara64:
This caused a failed attempt to refresh token


ptbrowne:
Thanks for the fix, maybe adding some tests here would help avoid future bugs :)

PR #2355, opened 2020-01-29
Print useful message instead of manifest when (un)-installing apps/konnectors

mamyn0va:
This PR aims at printing a useful message on stdout when installing or uninstalling apps or konnectors : 

```
./cozy-stack apps install --domain "cozy.tools:8080" drive
drive has been installed (1.18.23)
```

```
./cozy-stack apps uninstall --domain "cozy.tools:8080" drive
drive has been uninstalled
```


nono:
Thanks!

PR #2357, opened 2020-01-30
Upgrade goimports

Gara64:
I made `go get golang.org/x/tools/cmd/goimports`


Gara64:
Actually, we prefer not having goimports as a stack dependency

PR #2386, opened 2020-02-10
Filter accounts migration by ClientKind

Gara64:
Settings is used as a bitwarden client, therefore, it could trigger the
account migration worker as it uses the /bitwarden routes. To avoid
this, we check if the OAuth client kind is not "web", used for web apps.

### DIFF
--- a/web/bitwarden/bitwarden.go
+++ b/web/bitwarden/bitwarden.go
@@ -126,21 +126,9 @@ func SetKeyPair(c echo.Context) error {
 	if err != nil {
 		return err
 	}
-	if _, err := setting.OrganizationKey(); err == settings.ErrMissingOrgKey {
-		// The organization key should exist at this moment as it is created at the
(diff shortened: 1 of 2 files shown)


web/bitwarden/bitwarden.go
```diff
@@ -292,6 +286,20 @@ func getInitialCredentials(c echo.Context) error {
 	}
 	key := setting.Key
 
+	if _, err := setting.OrganizationKey(); err == settings.ErrMissingOrgKey {
+		// The organization key should exist at this moment as it is created at the
+		// instance creation or at the login-hashed migration.
+		log.Warningf("Organization key does not exist")
```
nono: Is it useful to repeat the same warning that we have at the prelogin step?

Gara64: I removed it with the new commit (the warning was in `SetKeyPair`)

PR #2387, opened 2020-02-12
Add support for the .url shortcuts

nono:


docs/shortcuts.md
```diff
@@ -0,0 +1,154 @@
...
+}
+```
+
+## GET /shortcuts/:id
+
+This route can be used to get information about a shortcut. A permission to
+read the file is required to use it. There are two variants, depending on the
+`Accept` header.
+
+### Request (JSON variant)
```
edas: Does this simply duplicate the /files route ?

nono: No, it also parses the content of the .url file to extract the URL

docs/shortcuts.md
```diff
@@ -0,0 +1,154 @@
...
+          },
+          "app": "photos",
+          "_type": "io.cozy.files",
+          "mime": "image/jpg"
+        }
+      }
+    }
+  }
+}
+```
```
Crash--: Can you add an example with an external link? Do we receive the external "link" in the `data.attributes.url` attribute ?

nono: Yes, the `url` is filled by parsing the content of the file, and it will be here for an external link.

PR #2396, opened 2020-02-17
fix: Flags and CozyData were passed as bytes instead of string

ptbrowne:
When using {{.Flags}} or {{.CozyData}}, bytes were outputted in the HTML.
The template engine requires a string to be passed.

### DIFF
--- a/web/apps/serve.go
+++ b/web/apps/serve.go
@@ -299,7 +299,7 @@ type serveParams struct {
 	isLoggedIn bool
 }
 
-func (s serveParams) CozyData() ([]byte, error) {
+func (s serveParams) CozyData() (string, error) {
 	flags, err := feature.GetFlags(s.instance)
 	if err != nil {
 		flags = &feature.Flags{
@@ -321,7 +321,13 @@ func (s serveParams) CozyData() ([]byte, error) {
 		},
 		"flags": flags,
 	}
-	return json.Marshal(data)
+	bytes, err := json.Marshal(data)
+
+	if err != nil {
+		return "", err
+	}
+
+	return string(bytes), nil
 }
 
 func (s serveParams) Domain() string {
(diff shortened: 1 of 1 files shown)


web/apps/serve.go
```diff
@@ -321,7 +321,13 @@ func (s serveParams) CozyData() ([]byte, error) {
 		},
 		"flags": flags,
 	}
-	return json.Marshal(data)
+	bytes, err := json.Marshal(data)
+
+	if err != nil {
+		panic(err)
```
nono: `panic` must only be used as a least resort. Here, we can return the error: `return "", err`

web/apps/serve.go
```diff
@@ -321,7 +321,13 @@ func (s serveParams) CozyData() ([]byte, error) {
...
+	}
+
+	return string(bytes), err
```
nono: We know that err is nil, and we can reflect it: `return string(bytes), nil`

web/apps/serve.go
```diff
@@ -356,12 +362,16 @@ func (s serveParams) IconPath() string {
...
-func (s serveParams) Flags() ([]byte, error) {
+func (s serveParams) Flags() (string, error) {
 	flags, err := feature.GetFlags(s.instance)
 	if err != nil {
-		return []byte("{}"), err
+		return "{}", err
+	}
+	bytes, err := json.Marshal(flags)
+	if err != nil {
+		panic(err)
```
nono: idem

web/apps/serve.go
```diff
@@ -356,12 +362,16 @@ func (s serveParams) IconPath() string {
...
 	}
-	return json.Marshal(flags)
+	return string(bytes), err
```
nono: idem

web/apps/serve.go
ptbrowne: I used panic as I saw it in the cozyclientjs function in the same file, should it be changed there also ?

nono: Yes, it should be changed: https://github.com/cozy/cozy-stack/pull/2397

PR #2401, opened 2020-02-20
Add a way to discard a subscription in realtime

### LINKED ISSUE #1921: Unsubscribe in realtime
When we subscribe to somehting in realtime, there is no way to discard this subscription later. We'll receive all updates until the socket die.

This is fine if we create a socket for each need (once we do not need something, we may close the socket) but this may not scale well if we try to have a single socket between the app and the stack.

Expected: Having an UNSUBSCRIBE message along with the SUBSCRIBE one, with the same parameters.

nono:
When a `SUBSCRIBE` order is sent to the realtime websocket, it can be convenient to cancel it later without having to close the websocket. The new `UNSUBSCRIBE` order can be sent to do that. It takes the same parameters that the `SUBSCRIBE`.

Fix #1921


nono:
No, if you have subscribed 3 times with the same parameters, you need to unsubscribed 3 times with the same parameters. It is something that I can change, I don't really know what should be the correct behavior in this case, and I've took the simpler one to implement.

PR #2417, opened 2020-03-04
Add a route to import bitwarden ciphers and folder in bulk

nono:


docs/bitwarden.md
```diff
@@ -856,6 +856,62 @@ Host: alice.example.com
 HTTP/1.1 204 No Content
 ```
 
+### POST /bitwarden/api/ciphers/import
+
+This route can be used to import several ciphers and folders in bulk.
+
+In `FolderRelationships`, the `Key` is the index of the cipher in the `Ciphers`
+list, and the `Value` is the index of the folder in the `Folders` list.
```
ptbrowne: folderRelationships ciphers key value folders ;)

web/bitwarden/ciphers.go
```diff
@@ -617,3 +626,66 @@ func ShareCipher(c echo.Context) error {
...
+	var req importCipherRequest
+	if err := json.NewDecoder(c.Request().Body).Decode(&req); err != nil {
+		return c.JSON(http.StatusBadRequest, echo.Map{
+			"error": "invalid JSON",
+		})
+	}
+
+	// Import the folders
+	folders := make([]interface{}, len(req.Folders))
+	olds := make([]interface{}, len(req.Folders))
```
ptbrowne: I don't understand the purpose of `olds` ?

web/bitwarden/bitwarden_test.go
```diff
@@ -645,6 +646,58 @@ func assertDomainsReponse(t *testing.T, res *http.Response) {
...
+}`
+	req, _ := http.NewRequest("POST", ts.URL+"/bitwarden/api/ciphers/import", bytes.NewBufferString(body))
+	req.Header.Add("Content-Type", "application/json")
+	req.Header.Add("Authorization", "Bearer "+token)
+	res, err := http.DefaultClient.Do(req)
+	assert.NoError(t, err)
+	assert.Equal(t, 204, res.StatusCode)
+	nb, err := couchdb.CountAllDocs(inst, consts.BitwardenCiphers)
+	assert.NoError(t, err)
+	assert.Equal(t, nbCiphers+2, nb)
```
ptbrowne: 2 because of index doc and design doc ?

web/bitwarden/ciphers.go
nono: The `BulkUpdateDocs` function can be used to create documents, but also to update existing ones. And when we update an existing document, the code in stack keeps both the document before and after the change for the real-time sub-system (that is also used by triggers).

Here, we are just creating documents, so we use `nil`s value for the old documents. No purpose, except to conform to the `BulkUpdateDocs` signature.

web/bitwarden/bitwarden_test.go
nono: No, 2 because 2 ciphers are sent in the JSON (1 note and 1 login).

ptbrowne: Ah yes ok.

PR #2427, opened 2020-03-10
Fix the next link for POST /files/_find

nono:
The link was wrong, because it was missing the /_find part. In the stack ignored the cursor in the query-string. This commit fixes both issues.

### DIFF
--- a/web/files/files.go
+++ b/web/files/files.go
@@ -1222,9 +1222,12 @@ func FindFilesMango(c echo.Context) error {
 	findRequest["limit"] = limit
 
 	skip := 0
-	skipF64, hasSkip := findRequest["skip"].(float64)
-	if hasSkip {
+	if skipF64, ok := findRequest["skip"].(float64); ok {
 		skip = int(skipF64)
+	}
+
+	if bookmark := c.QueryParam("cursor"); bookmark != "" {
+		findRequest["bookmark"] = bookmark
 	}
 
 	var results []vfs.DirOrFileDoc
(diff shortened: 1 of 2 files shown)


web/files/files.go
```diff
@@ -1222,11 +1222,14 @@ func FindFilesMango(c echo.Context) error {
...
 	skip := 0
-	skipF64, hasSkip := findRequest["skip"].(float64)
-	if hasSkip {
+	if skipF64, ok := findRequest["skip"].(float64); ok {
 		skip = int(skipF64)
 	}
 
+	if bookmark := c.QueryParam("cursor"); bookmark != "" {
+		findRequest["bookmark"] = bookmark
+	}
```
y-lohse: I find this strange because it seems like this route already worked. As in, if I take the `cursor` param from the `next` URL, and send it as `bookmark`, it already works..?

y-lohse: Just checked again, I can confirm the `bookmark` works even right now — sending requests with no `skip`, just a `limit` and a `bookmark`, and the results are properly paginated.

Also, we receive a `next` link even when there are no more results, but that may be more normal and not a big deal.

nono: > Also, we receive a next link even when there are no more results, but that may be more normal and not a big deal.

It should be better now

nono: About the cursor/bookmark, it is possible to send it to the stack via `bookmark` in the JSON, or via a parameter on the query-string. The first way should be preferred, but I would like that the `next` link has some meaning, so I also support the second way.

PR #2443, opened 2020-03-23
Add a sharing_domain config entry

nono:
When a member of a sharing is asked for the URL of their cozy instance, the input widget offers a suggestion on the domain. This suggestion can now be customized per context (the default is "mycozy.cloud").


nono:
> On small question though: how to deal with stack envs with no common domain between instances?

For these contexts like the self-hosted, we keep the default value (`mycozy.cloud`) as we don't have a better guess on which domain to suggest and the member of the sharing can still change the domain.

PR #2445, opened 2020-03-24
Fix event-stream extra linefeed

sblaisot:
When requesting the stack as `text/event-stream` to install or update app/konnector, the event-stream response contain an extra linefeed (`\n`) after each event.

image

This PR removes the extra linefeed.

### DIFF
--- a/web/apps/apps.go
+++ b/web/apps/apps.go
@@ -12,6 +12,7 @@ import (
 	"path"
 	"runtime"
 	"strconv"
+	"strings"
 
 	"github.com/cozy/cozy-stack/model/account"
 	"github.com/cozy/cozy-stack/model/app"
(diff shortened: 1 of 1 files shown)


web/apps/apps.go
```diff
@@ -373,7 +374,7 @@ func pollInstaller(c echo.Context, instance *instance.Instance, isEventStream bo
 		}
 		buf := new(bytes.Buffer)
 		if err := jsonapi.WriteData(buf, &apiApp{man}, nil); err == nil {
-			writeStream(w, "state", buf.String())
+			writeStream(w, "state", strings.TrimRight(buf.String(), "\n"))
```
nono: Can you use `TrimSuffix` instead of `TrimRight`?

PR #2446, opened 2020-03-25
Add context features from config to admin endpoint and cli

sblaisot:
Allow to query stack for feature flags configured in configuration file, inside a specific context

### DIFF
--- a/web/instances/feature.go
+++ b/web/instances/feature.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/cozy/cozy-stack/model/feature"
 	"github.com/cozy/cozy-stack/model/instance/lifecycle"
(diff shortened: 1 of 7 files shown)


web/instances/feature.go
```diff
@@ -71,6 +72,51 @@ func putFeatureSets(c echo.Context) error {
...
+			return ctx, nil
+		}
+	}
+
+	ctx, ok := contexts[config.DefaultInstanceContext]
+	if ok && ctx != nil {
+		return ctx, nil
+	}
+
+	return nil, fmt.Errorf("Unable to get context \"%s\" from config", context)
```
nono: [nit] You can `%q` instead of `"%s"` to print a quoted string

sblaisot: ok, thanks.

PR #2456, opened 2020-04-01
Fix support of Firefox on iOS

nono:
Firefox on iOS has a lower version number than Firefox on desktop and android. It also uses the webkit engine, and not the Gecko one. So, it makes sense to have special user-agent rules to detect it, and to not block it.


nono:
The tests are red because the debian repository for CouchDB is not accessible from the GitHub Actions (network issue).

PR #2457, opened 2020-04-01
Return updated file rev and added references

taratatach:
When adding new `referenced_by` objects to an io.cozy.files document,
we expect to get the updated list of references in return.
Since the document is updated to add new references, its revision has
changed and should therefore be returned as well so clients won't have
to make another request to get the updated revision and make new
changes to the document.

### DIFF
--- a/docs/references-docs-in-vfs.md
+++ b/docs/references-docs-in-vfs.md
@@ -113,6 +113,10 @@ Content-Type: application/vnd.api+json
 
 ```json
 {
+    "meta": {
+        "rev": "2-de8d0ba2",
+        "count": 2
+    },
     "data": [
         {
             "type": "io.cozy.playlists",
@@ -152,8 +156,23 @@ Accept: application/vnd.api+json
 #### Response
 
 ```http
-HTTP/1.1 204 No Content
+HTTP/1.1 200 OK
 Content-Type: application/vnd.api+json
+```
+
+```json
+{
+    "meta": {
+        "rev": "3-7ab812c0",
+        "count": 1
+    },
+    "data": [
+        {
+            "type": "io.cozy.playlists",
+            "id": "94375086-e2e2-11e6-81b9-5bc0b9dd4aa4"
+        }
+    ]
+}
 ```
 
 ### GET /data/:type/:doc-id/relationships/references
--- a/pkg/jsonapi/jsonapi.go
+++ b/pkg/jsonapi/jsonapi.go
@@ -102,17 +102,15 @@ func DataListWithTotal(c echo.Context, statusCode, total int, objs []Object, lin
 
 // DataRelations can be called to send a Relations page,
 // a list of ResourceIdentifier
-func DataRelations(c echo.Context, statusCode int, refs []couchdb.DocReference, total int, links *LinksList, included []Object) error {
+func DataRelations(c echo.Context, statusCode int, refs []couchdb.DocReference, meta *RelationshipMeta, links *LinksList, included []Object) error {
 	data, err := json.Marshal(refs)
 	if err != nil {
 		return InternalServerError(err)
 	}
 
 	doc := Document{
-		Data: (*json.RawMessage)(&data),
-		Meta: &RelationshipMeta{
-			Count: &total,
-		},
+		Data:  (*json.RawMessage)(&data),
+		Meta:  meta,
 		Links: links,
 	}
 
--- a/web/files/referencedby.go
+++ b/web/files/referencedby.go
@@ -33,21 +33,30 @@ func AddReferencedHandler(c echo.Context) error {
 		return WrapVfsError(err)
 	}
 
+	var newRev string
+	var refs []couchdb.DocReference
 	if dir != nil {
 		dir.AddReferencedBy(references...)
 		updateDirCozyMetadata(c, dir)
 		err = couchdb.UpdateDoc(instance, dir)
+		newRev = dir.Rev()
+		refs = dir.ReferencedBy
 	} else {
 		file.AddReferencedBy(references...)
 		updateFileCozyMetadata(c, file, false)
 		err = couchdb.UpdateDoc(instance, file)
+		newRev = file.Rev()
+		refs = file.ReferencedBy
 	}
 
 	if err != nil {
 		return WrapVfsError(err)
 	}
 
-	return c.NoContent(http.StatusNoContent)
+	count := len(refs)
+	meta := jsonapi.RelationshipMeta{Rev: newRev, Count: &count}
+
+	return jsonapi.DataRelations(c, http.StatusOK, refs, &meta, nil, nil)
 }
 
 // RemoveReferencedHandler is the echo.handler for removing referenced_by to
@@ -73,19 +82,28 @@ func RemoveReferencedHandler(c echo.Context) error {
 		return WrapVfsError(err)
 	}
 
+	var newRev string
+	var refs []couchdb.DocReference
 	if dir != nil {
 		dir.RemoveReferencedBy(references...)
 		updateDirCozyMetadata(c, dir)
 		err = couchdb.UpdateDoc(instance, dir)
+		newRev = dir.Rev()
+		refs = dir.ReferencedBy
 	} else {
 		file.RemoveReferencedBy(references...)
 		updateFileCozyMetadata(c, file, false)
 		err = couchdb.UpdateDoc(instance, file)
+		newRev = file.Rev()
+		refs = file.ReferencedBy
 	}
 
 	if err != nil {
 		return WrapVfsError(err)
 	}
 
-	return c.NoContent(http.StatusNoContent)
+	count := len(refs)
+	meta := jsonapi.RelationshipMeta{Rev: newRev, Count: &count}
+
+	return jsonapi.DataRelations(c, http.StatusOK, refs, &meta, nil, nil)
 }
--- a/web/files/referencedby_test.go
+++ b/web/files/referencedby_test.go
@@ -13,6 +13,7 @@ import (
 )
 
 var fileID1, fileID2 string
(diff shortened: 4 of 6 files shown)


docs/references-docs-in-vfs.md
```diff
@@ -113,16 +113,19 @@ Content-Type: application/vnd.api+json
...
-            "type": "io.cozy.playlists",
-            "id": "94375086-e2e2-11e6-81b9-5bc0b9dd4aa4"
-        },
-        {
-            "type": "io.cozy.playlists",
-            "id": "f2625cc0-e2d6-11e6-a0d5-cfbbfb141af0"
-        }
-    ]
+    "data": {
+        "rev": "2-de8d0ba2",
```
nono, line 117: To be more consistent with JSON-API, the `rev` should be in `meta`, not in `data`:

```json
{
    "data": [
        {
            "type": "io.cozy.playlists",
            "id": "94375086-e2e2-11e6-81b9-5bc0b9dd4aa4"
        },
        {
            "type": "io.cozy.playlists",
            "id": "f2625cc0-e2d6-11e6-a0d5-cfbbfb141af0"
        }
    ],
    "meta": {
        "rev": "2-de8d0ba2"
    }
}
```

web/files/referencedby.go
```diff
@@ -10,6 +11,53 @@ import (
 	"github.com/labstack/echo/v4"
 )
 
+// ContentType is the official mime-type for JSON-API
+const ContentType = "application/vnd.api+json"
```
nono, line 15: I would prefer to avoid re-declaring this content-type. You can use `jsonapi.ContentType` (from `pkg/jsonapi`) instead.

web/files/referencedby.go
```diff
@@ -10,6 +11,53 @@ import (
...
+	u := UpdatedReferences{DocRev: rev}
+	u.ReferencedBy = make([]couchdb.DocReference, len(refs))
+	copy(u.ReferencedBy, refs)
+
+	return u
+}
+
+// MarshalUpdatedReferences serializes an UpdatedReferences to JSON.
+// It returns a json.RawMessage that can be returned in an HTTP response.
+func MarshalUpdatedReferences(r UpdatedReferences) (json.RawMessage, error) {
```
nono, line 34: I don't see what is the purpose of this function. I think that `data, err := json.Marshal(r)` is more straightforward than `data, err := MarshalUpdatedReferences(r)`.

web/files/referencedby.go
```diff
@@ -10,6 +11,53 @@ import (
...
+const ContentType = "application/vnd.api+json"
+
+// UpdatedReferences is a container for the updated list of referenced_by object
+// for an io.cozy.files document.
+type UpdatedReferences struct {
+	DocRev       string                 `json:"rev"`
+	ReferencedBy []couchdb.DocReference `json:"referencedBy"`
+}
+
+func NewUpdatedReferences(rev string, refs []couchdb.DocReference) UpdatedReferences {
```
nono, line 24: The linter won't be happy with an exported function that has no comment.

web/files/referencedby.go
```diff
@@ -10,6 +11,53 @@ import (
...
+	if err != nil {
+		return nil, err
+	}
+
+	return b, nil
+}
+
+// WriteData can be called to write an answer with a JSON document
+// containing a single UpdatedReferences object as data into an io.Writer.
+func WriteData(c echo.Context, statusCode int, r UpdatedReferences) error {
```
nono, line 45: `WriteData` is a bit too generic for naming a function inside the `files` package for work only for references. Maybe, we can call it `WriteReferences`.

web/files/referencedby.go
```diff
@@ -10,6 +11,53 @@ import (
...
+// for an io.cozy.files document.
+type UpdatedReferences struct {
+	DocRev       string                 `json:"rev"`
+	ReferencedBy []couchdb.DocReference `json:"referencedBy"`
+}
+
+func NewUpdatedReferences(rev string, refs []couchdb.DocReference) UpdatedReferences {
+	u := UpdatedReferences{DocRev: rev}
+	u.ReferencedBy = make([]couchdb.DocReference, len(refs))
+	copy(u.ReferencedBy, refs)
```
nono, line 27: Why do you make a copy here?

web/files/referencedby_test.go
```diff
@@ -34,6 +35,12 @@ func TestAddReferencedByOneRelation(t *testing.T) {
 		return
 	}
 
+	var result struct {
+		Data struct {
+			Rev          string                 `json:"rev"`
+			ReferencedBy []couchdb.DocReference `json:"referencedBy"`
+		}
```
nono, line 42: I'm really surprised that it works with `` `json:"data"` ``. I didn't think the JSON unmarshalling was case-insensitive.

web/files/referencedby.go
nono, line 45: It's probably not necessary to give the status code in parameter, as it will always be 200.

nono:
Can you also rebase master (and https://github.com/cozy/cozy-stack/commit/13131869837eff4015d0d9b5d5095a1ffc57a277 in particular) to fix the linter?

web/files/referencedby.go
taratatach: I hesitated to do that in the first place. Will do.

web/files/referencedby.go
taratatach: Right, I forgot about this one (I added it last and didn't run the linter…)

web/files/referencedby.go
taratatach: Well, first, I had issues with pointer de-renferences when I first assigned the slice directly to the struct field.
And then I thought it would be a good idea to copy those values instead of having a reference to the original slice. But I might be wrong about slices being passed as references.

web/files/referencedby_test.go
taratatach: Ah. Well, that's unexpected :). I actually copied this struct from another test.
I can add the json tag explicitly if you think this could become a problem.

web/files/referencedby.go
```diff
@@ -10,6 +11,36 @@ import (
 	"github.com/labstack/echo/v4"
 )
 
+// ReferencesDocument is JSON-API document, identified by the mediatype
+// application/vnd.api+json
+// See http://jsonapi.org/format/#document-structure
+// It differs from jsonapi.Document because it represents a list of relationships
```
nono, line 17: I don't understand. In my mind, the response should be a `jsonapi.Document`. And a `jsonapi.Document` can have a list as its `data`, so I don't see the difference with the `ReferencesDocument`.

web/files/referencedby.go
```diff
@@ -10,6 +11,36 @@ import (
...
+// but differs from jsonapi.Relationship because it olds the revision of it
+// parent io.cozy.files document.
+type ReferencesDocument struct {
+	Data *json.RawMessage `json:"data"`
+	Meta jsonapi.Meta     `json:"meta"`
+}
+
+// WriteReferences can be called to write an answer with a JSON document
+// containing the list of added reference_by objects as data into an io.Writer.
+func WriteReferences(c echo.Context, newRev string, refs []couchdb.DocReference) error {
```
nono, line 27: I have read again the `pkg/jsonapi` package. There is a `DataRelations` function that is doing something really close to this function. I think we should modify it a bit to accept the revision in the meta.

web/files/referencedby.go
taratatach: You're right about the `data` field. But in `Document`, the `meta` field is a `RelationshipMeta` struct, with a `count`, not a `rev`.
Besides, when I looked at the methods in `jsonapi`, most of them were filling the `data` field with an `Object` which has a `type` and `attributes` and I wasn't sure this would fit our needs.

I should probably have asked you.

web/files/referencedby.go
taratatach: I saw `DataRelations` but it didn't have the `rev` field and I didn't want to modify this package.
Adding the `rev` field to `RelationshipMeta` would do the trick.

pkg/jsonapi/jsonapi.go
```diff
@@ -134,6 +134,31 @@ func DataRelations(c echo.Context, statusCode int, refs []couchdb.DocReference,
 	return json.NewEncoder(resp).Encode(doc)
 }
 
+// DataRefences can be called to send a References page,
```
nono, line 137: I would prefer to reuse the `DataRelations` function by changing the `total` parameter by a `meta` parameter.

docs/references-docs-in-vfs.md
```diff
@@ -113,6 +113,9 @@ Content-Type: application/vnd.api+json
 
 ```json
 {
+    "meta": {
+        "rev": "2-de8d0ba2"
```
nono, line 117: It looks like the `count` is also sent in the code

docs/references-docs-in-vfs.md
```diff
@@ -152,10 +155,24 @@ Accept: application/vnd.api+json
...
 ```http
-HTTP/1.1 204 No Content
+HTTP/1.1 200 OK
 Content-Type: application/vnd.api+json
 ```
 
+```json
+{
+    "meta": {
+        "rev": "3-7ab812c0"
```
nono, line 165: Idem

pkg/jsonapi/jsonapi.go
nono: If you think it is a too big change for this PR, at least we should fix the typo: s/DataRefences/Data/References/g

taratatach: It felt weird to have a very long signature and passing `nil` parameters in `referencedby.go` but if you prefer limiting the number of functions, I can do that as well.

PR #2476, opened 2020-04-20
Add a suffix to revoked sharing directories

nono:
When a sharing is revoked, the directory is kept on the recipients'
instances. If a new sharing is made for the same directory, the
recipients that accept it will have a conflict between the directories
for the old sharing and the new one. It can be confusing. To avoid that,
the directory is renammed when the sharing is revoked (with a suffix).
It will also help people understand that a sharing has been revoked.

### DIFF
--- a/model/sharing/files.go
+++ b/model/sharing/files.go
@@ -2,10 +2,12 @@ package sharing
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"net/url"
 	"os"
 	"path"
+	"path/filepath"
 	"sort"
 	"strings"
 	"time"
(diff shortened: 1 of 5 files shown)


model/sharing/files.go
```diff
@@ -322,7 +328,18 @@ func (s *Sharing) RemoveSharingDir(inst *instance.Instance) error {
...
+	suffix := inst.Translate("Tree Revoked sharing suffix")
+	parentPath := filepath.Dir(dir.Fullpath)
+	basename := fmt.Sprintf("%s (%s)", dir.DocName, suffix)
+	dir.DocName = basename
+	for i := 2; i < 100; i++ {
+		dir.Fullpath = path.Join(parentPath, dir.DocName)
+		if err = inst.VFS().UpdateDirDoc(olddoc, dir); err == nil {
+			return nil
+		}
+		dir.DocName = fmt.Sprintf("%s (%d)", basename, i)
```
Crash--, line 340: You set a limit to `(99)` for the docName's conflict? 

If I already have a folder called `Cozy-sharing (cancelled sharing)`, you will generate `Cozy-sharing (cancelled sharing) (2)` until 99 (if conflict) right? 

We have this kind of name generation on conflict on the front side, I'll use the same limit.

nono: Yes, you are right

PR #2479, opened 2020-04-21
Improve permissions checking for notes and shortcuts

nono:
When a note or a shortcut is created via their special API, the stack
checks the permission to write on the VFS. Before this commit, only a
permission on the whole io.cozy.files doctype was authorized. Now, we
also accept a permission on the directory where the note or shortcut
will be created. It is useful when a directory is shared by link for
example.

### DIFF
--- a/web/notes/notes.go
+++ b/web/notes/notes.go
@@ -41,7 +41,7 @@ func CreateNote(c echo.Context) error {
 			return err
 		}
 		fileDoc, errf := vfs.NewFileDoc(
-			doc.Title,
+			"tmp.cozy-note", // We don't care, but it can't be empty
 			dirID,
 			0,   // We don't care
 			nil, // Let the VFS compute the md5sum
--- a/web/notes/notes.go
+++ b/web/notes/notes.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"os"
 	"strconv"
+	"time"
 
 	"github.com/cozy/cozy-stack/model/note"
 	"github.com/cozy/cozy-stack/model/permission"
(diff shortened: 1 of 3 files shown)


y-lohse:
Apologies, it took me a while to find the time to try this out. It now works for shortcuts, but still not for notes. My setup:

- checked out the PR
- ran `go install`
- Go to a public page where `/permission/self` returns `files: { type: "io.cozy.files", values: ["136283bd89efae5095b0671c2f02d0b5"], verbs: ["GET", "POST", "PUT", "PATCH"]}`
- POST on `/notes` with the payload `dir_id: "136283bd89efae5095b0671c2f02d0b5", schema: {,…}, title: ""`
- I get a 403 forbidden error

nono:
@y-lohse it should be better now

web/notes/notes.go
```diff
@@ -23,17 +24,42 @@ import (
...
+	// cheap. If not, we look on more finer permissions, which is a bit more
+	// complicated and costly, but is needed for creating a note in a shared by
+	// link folder for example.
+	if err := middlewares.AllowWholeType(c, permission.POST, consts.Files); err != nil {
+		dirID, errd := doc.GetDirID(inst)
+		if errd != nil {
+			return err
+		}
+		fileDoc, errf := vfs.NewFileDoc(
+			"tmp.cozy-note", // We don't care, but it can't be empty
```
y-lohse, line 44: I may misunderstand what we do here, but if we create a temporary file to check permissions, shouldn't we delete it afterwards?

nono: The file is created in memory, but not persisted in CouchDB/swift. So there is no need to delete it, the Garbage Collector will clean the memory.

PR #2496, opened 2020-05-05
Add an option to skip OAuth state on token

nono:
When a konnector uses an OAuth provider, the stack is responsible to
manage the OAuth dance. Some providers require the state on the access
token endpoint, and some others throw en error if the state is given.
This commit adds an option to the account_types, named
skip_state_on_token to not include the state on this particular request.

### DIFF
--- a/model/account/type.go
+++ b/model/account/type.go
@@ -66,6 +66,7 @@ type AccountType struct {
 	Slug                  string            `json:"slug,omitempty"`
 	Secret                interface{}       `json:"secret,omitempty"`
 	SkipRedirectURI       bool              `json:"skip_redirect_uri_on_authorize,omitempty"`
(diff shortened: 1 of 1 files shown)


model/account/type.go
```diff
@@ -66,6 +66,7 @@ type AccountType struct {
 	Slug                  string            `json:"slug,omitempty"`
 	Secret                interface{}       `json:"secret,omitempty"`
 	SkipRedirectURI       bool              `json:"skip_redirect_uri_on_authorize,omitempty"`
+	SkipState             bool              `json:"skip_state_on_token,omitempty"`
```
Crash--, line 69: Can we upgrade the associated doc to list those new params?

nono: Done: https://github.com/cozy/cozy-stack/commit/11cb069b277c3340909c4076270d756f422714ab

PR #2506, opened 2020-05-12
feat: Add UpdatedAt param to file upload

Crash--:
Chromium based browser do not allow the use of the `Date` header
when using XHR. We now introduce a new UpdatedAt param we can use
when uploading a file or a folder to set the modification date of
the file.

(see https://github.com/cozy/cozy-client/issues/690)

### DIFF
--- a/web/files/files_test.go
+++ b/web/files/files_test.go
@@ -302,6 +302,30 @@ func TestCreateDirWithDateSuccess(t *testing.T) {
 	assert.NotContains(t, fcm, "uploadedAt")
 }
 
+func TestCreateDirWithDateSuccessAndUpdatedAt(t *testing.T) {
+	req, _ := http.NewRequest("POST", ts.URL+"/files/?Type=directory&Name=dir-with-date-and-updatedat&CreatedAt=2016-09-18T10:24:53Z&UpdatedAt=2020-05-12T12:25:00Z", strings.NewReader(""))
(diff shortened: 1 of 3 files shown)


web/files/files_test.go
```diff
@@ -508,6 +532,37 @@ func TestUploadWithDate(t *testing.T) {
 	assert.Equal(t, "2016-09-19T12:38:04Z", updatedAt)
 }
 
+func TestUploadWithCreatedAtAndUpdatedAt(t *testing.T) {
+	buf := strings.NewReader("foo")
+	req, err := http.NewRequest("POST", ts.URL+"/files/?Type=file&Name=TestUploadWithCreatedAtAndUpdatedAt&CreatedAt=2016-09-18T10:24:53Z&UpdatedAt=2020-05-12T12:25:00Z", buf)
+	req.Header.Add(echo.HeaderAuthorization, "Bearer "+token)
+	assert.NoError(t, err)
+	res, obj := doUploadOrMod(t, req, "text/plain", "rL0Y20zC+Fzt72VPzMSk2A==")
```
ptbrowne, line 540: Nit: What does `rL0Y20zC+Fzt72VPzMSk2A==` mean ?

Crash--: md5 binary hash of the "foo" string converted to base64.

PR #2512, opened 2020-05-14
Be less strict on access control for disk usage

nono:
When a directory is shared by link with a write access, the anonymous
users that go on this link can check to the disk-usage to see if the
limit of quota is reached.

### DIFF
--- a/web/middlewares/permissions.go
+++ b/web/middlewares/permissions.go
@@ -327,6 +327,21 @@ func AllowVFS(c echo.Context, v permission.Verb, o vfs.Fetcher) error {
 	return nil
 }
 
+// CanWriteToAnyDirectory checks that the context permission allows to write to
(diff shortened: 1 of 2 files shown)


web/middlewares/permissions.go
```diff
@@ -327,6 +327,21 @@ func AllowVFS(c echo.Context, v permission.Verb, o vfs.Fetcher) error {
...
+	if err != nil {
+		return err
+	}
+	for _, rule := range pdoc.Permissions {
+		if rule.Type == consts.Files && rule.Verbs.Contains(permission.POST) {
+			return nil
+		}
+	}
+	return ErrForbidden
+}
```
Crash--, line 343: Thanks to that code, I just read https://gist.github.com/posener/a303becac35835ad7bf5e15fe061893e 👍

PR #2516, opened 2020-05-18
Fix sort for deleted files

Gara64:
From the sharer side, we sort docs before replication in order to apply a bunch of rules, e.g. directories before files.
We also implemented a [rule](https://github.com/cozy/cozy-stack/commit/4f9a4cdc7134cb586d4461c2b4c05cf3c8a01b8a) on the recipient side in order to put the deleted files at the end, to efficiently deal with moves. However, the sort behavior was not the expected one, and it was also contradictory with the sort performed by the sharer. As it result, it was causing errors like [this one](https://github.com/cozy/cozy-stack/blob/master/model/sharing/files.go#L875) because the recipient was mistakenly trying to recursively rebuild the hierarchy.

Therefore, we remove this sort on the recipient side and implement it on the sharer side, with the other rules.

### DIFF
--- a/model/sharing/files_test.go
+++ b/model/sharing/files_test.go
@@ -53,7 +53,7 @@ func TestSortFilesToSent(t *testing.T) {
 	filec := map[string]interface{}{"type": "file", "name": "filec"}
 	files := []map[string]interface{}{filea, foobar, foobarbaz, dela, delb, fileb, filec, foo}
 	s.SortFilesToSent(files)
-	expected := []map[string]interface{}{delb, dela, foo, foobar, foobarbaz, filea, fileb, filec}
+	expected := []map[string]interface{}{foo, foobar, foobarbaz, filea, fileb, filec, dela, delb}
 	assert.Equal(t, expected, files)
 }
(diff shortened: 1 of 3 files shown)


model/sharing/files_test.go
```diff
@@ -53,7 +53,7 @@ func TestSortFilesToSent(t *testing.T) {
 	filec := map[string]interface{}{"type": "file", "name": "filec"}
 	files := []map[string]interface{}{filea, foobar, foobarbaz, dela, delb, fileb, filec, foo}
 	s.SortFilesToSent(files)
-	expected := []map[string]interface{}{delb, dela, foo, foobar, foobarbaz, filea, fileb, filec}
+	expected := []map[string]interface{}{foo, foobar, foobarbaz, filea, fileb, filec, dela, delb}
```
nono, line 56: Can you add a `"type" : "fil"` to `dela` and `delb` on this test?

nono: Well, in fact, _deleted docs don't have a type

PR #2536, opened 2020-06-05
Add config to allow users to choose their instance with OIDC

nono:
When the authentication via OIDC is enabled, the stack needs a way to
check that a user authenticated on the identity provider is really the
owner of the current instance. The stack was able to compute to instance
from the UserInfo response when there is a rule for the instance name at
creation. But, when the user can choose the name of their instance, we
need to rely on another strategy: the stack will take the value of the
oidc_id parameter at the instance creation, and can check it againt the
sub field of the UserInfo response.


cmd/instances.go
```diff
@@ -172,6 +173,7 @@ be used as the error message.
 			DomainAliases: flagDomainAliases,
 			Locale:        flagLocale,
 			UUID:          flagUUID,
+			OIDCID:        flagOIDCID,
```
aeris, line 176: (Nitpick) Wouldn't be better to have agnostic naming only related to oauth and not to only oidc, to allow more generic usage like Github/France Connect/whoever other oauth provider connection?

web/oidc/oidc.go
```diff
@@ -375,35 +364,76 @@ func getToken(conf *Config, code string) (string, error) {
...
+	return extractDomain(conf, params)
+}
+
+func checkDomainFromUserInfo(conf *Config, inst *instance.Instance, token string) error {
+	params, err := getUserInfo(conf, token)
+	if err != nil {
+		return err
+	}
+
+	if conf.AllowCustomInstance {
```
aeris, line 383: Same here, quite strange to have to configure specifically the behavior with a setting.
Could be better to check if OAuth ID is available on settings and if not, to fallback to ID calculation.
This way, we allow easier implementation for future feature to enroll an existing Cozy to one (and even multiple) OAuth provider just by settings the OAuth ID.

cmd/instances.go
nono: Maybe, but for the moment, I appreciate that we talk of OIDC for authentication and of OAuth for konnectors (using OAuth for both will be confusing, even if it is something that we can't avoid on the long term)

aeris: OK :)

web/oidc/oidc.go
nono: No, I really don't want to mix behaviors on the same context, as there are probably some bugs that will happen if we don't review carefully the code and modify some parts of it to manage this case.

aeris: OK. Will be easy to change in the future if needed.

PR #2543, opened 2020-06-08
Fix couchdb conflicts on directories

Gara64:
In some cases, unexpected CouchDB conflicts were produced on shared directories, leading to unexpected behaviors. For instance, after a delete on a conflicted directory, the winning revision branch was removed, but the lost one was still there, causing the directory to re-appear from the last conflicted revision.

These conflicts were caused because we force the revision list on shared documents in the [bulk_docs](https://docs.couchdb.org/en/stable/replication/protocol.html#upload-batch-of-changed-documents) replication route, as specified in the CouchDB documentation. In particular, the `new_edits` field, when set to false, avoids to generate a new revision, but can produce conflicts when a specified revision differ from a local one with the same number, e.g. `2-aa` vs `2-bb`.

See the [commit](https://github.com/cozy/cozy-stack/commit/252c0b74682ae93d6bcc01e348882bc666f7a568) description for more details about the problematic and its resolution.

### DIFF
--- a/model/sharing/indexer.go
+++ b/model/sharing/indexer.go
@@ -293,7 +293,21 @@ func (s *sharingIndexer) UpdateDirDoc(olddoc, doc *vfs.DirDoc) error {
 	}
 	doc.SetRev(s.bulkRevs.Rev)
 	docs[0]["_rev"] = s.bulkRevs.Rev
-	docs[0]["_revisions"] = s.bulkRevs.Revisions
+	if olddoc != nil {
+		/*
+			XXX We cannot directly apply the revision chain as received by the remote
+			instance because it might create a CouchDB conflict if a rev does not
+			exist on the local instance. Therefore, we start from the last local rev
+			and apply all the higher revs received from the remote.
+		*/
+		chain := revsStructToChain(s.bulkRevs.Revisions)
+		newChain := MixupChainToResolveConflict(olddoc.DocRev, chain)
+		newRevs := revsChainToStruct(newChain)
+		docs[0]["_revisions"] = newRevs
+	} else {
+		docs[0]["_revisions"] = s.bulkRevs.Revisions
+	}
+
 	if err := couchdb.BulkForceUpdateDocs(s.db, consts.Files, docs); err != nil {
 		return err
 	}
--- a/model/sharing/indexer.go
+++ b/model/sharing/indexer.go
@@ -293,20 +293,17 @@ func (s *sharingIndexer) UpdateDirDoc(olddoc, doc *vfs.DirDoc) error {
 	}
 	doc.SetRev(s.bulkRevs.Rev)
 	docs[0]["_rev"] = s.bulkRevs.Rev
+	newRevs := s.bulkRevs.Revisions
 	if olddoc != nil {
-		/*
-			XXX We cannot directly apply the revision chain as received by the remote
-			instance because it might create a CouchDB conflict if a rev does not
-			exist on the local instance. Therefore, we start from the last local rev
-			and apply all the higher revs received from the remote.
-		*/
-		chain := revsStructToChain(s.bulkRevs.Revisions)
+		// XXX We cannot directly apply the revision chain as received by the remote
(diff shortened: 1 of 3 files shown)


model/sharing/indexer.go
```diff
@@ -293,7 +293,21 @@ func (s *sharingIndexer) UpdateDirDoc(olddoc, doc *vfs.DirDoc) error {
...
 	doc.SetRev(s.bulkRevs.Rev)
 	docs[0]["_rev"] = s.bulkRevs.Rev
-	docs[0]["_revisions"] = s.bulkRevs.Revisions
+	if olddoc != nil {
+		/*
+			XXX We cannot directly apply the revision chain as received by the remote
+			instance because it might create a CouchDB conflict if a rev does not
+			exist on the local instance. Therefore, we start from the last local rev
+			and apply all the higher revs received from the remote.
+		*/
```
nono, line 302: [nit] Can you use `//` for comments please? It would be more consistent with the rest of the code base.

model/sharing/indexer.go
```diff
@@ -293,7 +293,21 @@ func (s *sharingIndexer) UpdateDirDoc(olddoc, doc *vfs.DirDoc) error {
...
+		chain := revsStructToChain(s.bulkRevs.Revisions)
+		newChain := MixupChainToResolveConflict(olddoc.DocRev, chain)
+		newRevs := revsChainToStruct(newChain)
+		docs[0]["_revisions"] = newRevs
+	} else {
+		docs[0]["_revisions"] = s.bulkRevs.Revisions
+	}
```
nono, line 309: [nit] I would prefer to write it as:

```go
newRevs := s.bulkRevs.Revisions
if olddoc != nil {
    // ...
    newRevs = revsChainToStruct(newChain)
}
docs[0]["_revisions"] = newRevs
```

I think it is a bit easier to modify later if needed. In particular, it avoids to repeat `docs[0]["_revisions"] =`.

tests/integration/tests/sharing_conflicts.rb
```diff
@@ -188,4 +190,88 @@ def assert_conflict_children(inst_a, inst_b, parent_id_a, parent_id_b, filename)
...
+    diff_ab.must_be_empty
+    diff_ac.must_be_empty
+
+    # Check there is no conflict
+    conflicts_a = Couch.new.find_conflicts inst_a.domain, 'io-cozy-files'
+    conflicts_b = Couch.new.find_conflicts inst_b.domain, 'io-cozy-files'
+    conflicts_c = Couch.new.find_conflicts inst_c.domain, 'io-cozy-files'
+    assert_empty conflicts_a
+    assert_empty conflicts_b
+    assert_empty conflicts_c
```
nono, line 275: As we have a new CLI command to check the FSCK and the conflitcts, we can rewrite this 6 lines as:

```ruby
[inst_a, inst_b, inst_c].each do |inst|
  assert_equal inst.check, []
  inst.remove
end
```

Technically, removing the instance is not needed but it helps running the integration tests on the CI. I don't remember the exact reason, but it is something like having less open databases on CouchDB.

tests/integration/lib/couch.rb
```diff
@@ -31,9 +31,10 @@ def prefix(db)
     "cozy" + Digest::SHA256.hexdigest(db)[0...32]
   end
 
-  def get_doc(domain, doctype, id)
+  def get_doc(domain, doctype, id, opts = {})
+    params = { params: opts }
```
nono, line 35: I'm curious: why this change?

Gara64: During my tests, I queried some docs with `meta: true` in order to see the conflicts as well as the rev history. It is not used in the final code, but I thought it might be useful in the future.

tests/integration/tests/sharing_conflicts.rb
Gara64: Right, I forgot it because during my tests I didn't want to remove the instances in order to be able to query the database afterwards.

PR #2550, opened 2020-06-11
Add lock on sharing reception

Gara64:
We used to have a lock when sending shared documents but not on their
reception. This could cause CouchDB conflicts when an instance receives
at the same time changes on the same document from different instances.

To illustrate a possible scenario of such conflict, let's assume a shared directory with the revisions list `{1-aa, 2-aa}` between the instances A, B and C.

At the same time, instance A receives changes from B and C, respectively with the revisions lists `{1-aa, 2-bb}` and `{1-aa, 2-cc, 3-cc}`.

Changes from B are applied: the revision `2-bb` conflicts with the local `2-aa`, and is solved by generating a new local update, producing the revisions list `{1-aa, 2-aa, 3-aa}`.

Simultaneously, the changes from C are applied: the last local revision is still seen as `2-aa` and the following revision list is forced: `{1-aa, 2-aa, 3-cc}`. Then, a CouchDB conflict is produced between `3-aa` and `3-cc`.

### DIFF
--- a/model/sharing/replicator.go
+++ b/model/sharing/replicator.go
@@ -578,6 +578,12 @@ func (s *Sharing) sendBulkDocs(inst *instance.Instance, m *Member, creds *Creden
 
 // ApplyBulkDocs is a multi-doctypes version of the POST _bulk_docs endpoint of CouchDB
 func (s *Sharing) ApplyBulkDocs(inst *instance.Instance, payload DocsByDoctype) error {
+	mu := lock.ReadWrite(inst, "sharings/"+s.SID+"/_bulk_docs")
+	if err := mu.Lock(); err != nil {
+		return err
+	}
+	defer mu.Unlock()
+
 	var refs []*SharedRef
 
 	for doctype, docs := range payload {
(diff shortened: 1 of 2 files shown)


model/sharing/replicator.go
```diff
@@ -578,6 +578,12 @@ func (s *Sharing) sendBulkDocs(inst *instance.Instance, m *Member, creds *Creden
 
 // ApplyBulkDocs is a multi-doctypes version of the POST _bulk_docs endpoint of CouchDB
 func (s *Sharing) ApplyBulkDocs(inst *instance.Instance, payload DocsByDoctype) error {
+	mu := lock.ReadWrite(inst, "sharings/"+s.SID+"/_bulk_docs")
```
nono, line 581: After thinking a bit, I would prefer to use a different lock for sending and receiving:

```go
mu := lock.ReadWrite(inst, "sharings/"+s.SID+"/apply")
```

For example, if there is a sharing between the instances A and B, and both try to replicate at the same time, they will acquire the lock when the replicate worker starts, will send their requests, but the received HTTP request cannot be processed because the lock cannot be taken (it's already taken by the sender goroutine).

nono: Oh, you have made the change while I was doing the review :+1:

Gara64: Actually, the lock has not the same name between sending `sharings/"+s.SID` and receiving `sharings/"+s.SID+"/_bulk_docs`, but I actually changed it few minutes ago by a force-push, so you might have seen the previous commit.
Should I keep the `/_bulk_docs` or do you prefer `apply`?

nono: `/_bulk_docs` is OK

PR #2558, opened 2020-06-18
Fix shared integrity

Gara64:
When a shared document has updates both from local and remote instances,
the remote revisions might be saved before the local ones in the
revisions tree. If there are several updates, this could cause integrity
issues in the tree. Here is an illustrative scenario:

1. A shared doc has the revision 1-a.
2. The doc is updated twice locally, producing the revisions 2-a and 3-a.
3. The doc is also updated twice on a remote instance, producing the revisions 4-b and 5-b.
4. The revision tree has still the 1-a revision and the following chain is saved: `[3-a, 4-b, 5-b]`. It corresponds to the last local revision (3-a) and the incoming remote revisions (4-b, 5-b).
5. The 2-a revision will be later saved in the tree, in a new branch starting from 1-a.

To solve this, at the step 4, we retrieve the revision history of the
doc, compare the incoming ones and save this revision chain:
`[2-a, 3-a, 4-b, 5-b]`

### DIFF
--- a/model/sharing/revisions.go
+++ b/model/sharing/revisions.go
@@ -5,6 +5,8 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+
+	"github.com/cozy/cozy-stack/pkg/couchdb"
 )
 
 type conflictStatus int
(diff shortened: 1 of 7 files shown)


model/sharing/revisions.go
```diff
@@ -295,6 +297,34 @@ func MixupChainToResolveConflict(rev string, chain []string) []string {
...
 
+// addMissingRevsToChain includes the missing doc revisions to the chain, i.e.
+// all the doc revisions between the highest revision saved in the
+// revisions tree, and the lowest revision of the chain.
+// This can occur when both local and remote updates are made to the same doc,
+// with the remote ones saved before the local ones.
+func addMissingRevsToChain(db couchdb.Database, ref *SharedRef, chain []string) ([]string, error) {
+	refHighestGen := ref.Revisions.Generation()
+	chainLowestGen := RevGeneration(chain[0])
+	if refHighestGen < chainLowestGen-1 {
```
nono, line 308: [nit] I prefer to avoid long if:

```go
if refHighestGen >= chainLowestGen-1 {
  return chain, nil
}
```

model/sharing/shared.go
```diff
@@ -263,14 +263,15 @@ func updateRemovedForFiles(inst *instance.Instance, sharingID, dirID string, rul
 // UpdateShared updates the io.cozy.shared database when a document is
 // created/update/removed
 func UpdateShared(inst *instance.Instance, msg TrackMessage, evt TrackEvent) error {
-	mu := lock.ReadWrite(inst, "shared")
+	evt.Doc.Type = msg.DocType
+	sid := evt.Doc.Type + "/" + evt.Doc.ID()
+
+	mu := lock.ReadWrite(inst, "shared/"+sid)
```
nono, line 269: If we update the lock name here, we should also look at the other calls to the same lock, like https://github.com/cozy/cozy-stack/blob/master/model/sharing/upload.go#L342

Gara64: Right, I updated the lock in 2 more places. I just left this one https://github.com/cozy/cozy-stack/blob/master/model/sharing/setup.go#L206

nono:
Can you look why the integration tests are red?

PR #2559, opened 2020-06-19
Fix multiple conflicts on the same file

nono:
When a shared file is updated often by several members of the sharing,
it will lead to conflicts. And if several conflicts happen on the same
file, the stack can try to use a new name for the old content that was
already used. It fails and it aborts the upload of the new content. With
this commit, the stack will try harder to avoid already used names for
conflicts.

### DIFF
--- a/model/sharing/revisions_test.go
+++ b/model/sharing/revisions_test.go
@@ -347,17 +347,6 @@ func TestIndexerCreateBogusPrevRev(t *testing.T) {
 	assert.Equal(t, indexer.bulkRevs.Rev, rev)
 }
 
-func TestConflictName(t *testing.T) {
-	assert.Equal(t, "foo (2)", conflictName("foo", false))
-	assert.Equal(t, "foo (3)", conflictName("foo (2)", false))
-	assert.Equal(t, "foo (1000)", conflictName("foo (999)", false))
-	assert.Equal(t, "foo () (2)", conflictName("foo ()", false))
-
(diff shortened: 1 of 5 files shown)


model/sharing/revisions_test.go
```diff
@@ -347,17 +347,6 @@ func TestIndexerCreateBogusPrevRev(t *testing.T) {
...
-	assert.Equal(t, "foo (2)", conflictName("foo", false))
-	assert.Equal(t, "foo (3)", conflictName("foo (2)", false))
-	assert.Equal(t, "foo (1000)", conflictName("foo (999)", false))
-	assert.Equal(t, "foo () (2)", conflictName("foo ()", false))
-
-	assert.Equal(t, "bar (2)", conflictName("bar", true))
-	assert.Equal(t, "bar (2).txt", conflictName("bar.txt", true))
-	assert.Equal(t, "bar (3).txt", conflictName("bar (2).txt", true))
-}
-
```
Gara64, line 360: It is not possible to mock an indexer to keep the test?

nono: Well, technically, it is possible. But it's a lot of work for not a lot of value.

Gara64: Ok, I think it's a bit of a shame to lose it, as it also helps to understand the behavior. Maybe at least a comment in the function header, would help, like: 
```
A conflicted file `foo` will be renamed foo (2), then foo (3), etc
```

Issue #1610, opened 2018-11-01
OAuth register endpoint should have CORS enabled


ptbrowne:
If I try to instantiate an OAuth client from a webpage, it fails because there are no "Access-Control-Allow-Origin" served from the `register` endpoint.

I think this should be relaxed as the `register` endpoint should be accessible from any webpage. 

```
Access to fetch at 'https://recette.cozy.works/auth/register' from origin 'http://localhost:1234' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
_fails.js:8 Uncaught (in promise) TypeError: Failed to fetch
```

nono:
> I think this should be relaxed as the register endpoint should be accessible from any webpage.

Why? Do you think of a use-case for this?

I'm really not cleared about if it's a good idea or not to do that, and what are the security implications of that. I've tried a quick search on Internet but didn't find anything about CORS and the register endpoint.

ptbrowne:
It is also good to test the OAuth part of our applications with no mobile apps.

y-lohse:
👍 on this issue — and possibly disabling CORS on all routes.

Aside from using the stack on pages that are not served by a cozy (which would be useful for documentation examples and code sharing / collaboration, for example via codesandbox.io), this would also help our transition on mobile to the new WebView layer, which does implement CORS (the previous cordova version did not).

CORS are only active on a standard browser anyway and can already be bypassed. The real protection comes from the tokens embeded in the requests — except on the routes related to login/registrations of course. From what I know, it's pretty standard practice to enable CORS on API servers.

nono:
> and possibly disabling CORS on all routes

Do you mean enabling or disabling? We have CORS on most routes, and we can't disable CORS on them, or else our web apps will be broken (the apps are on a separated origin from the API routes, https://foo-drive.mycozycloud vs https://foo.mycozy.cloud and need CORS).

> CORS are only active on a standard browser anyway and can already be bypassed. The real protection comes from the tokens embeded in the requests — except on the routes related to login/registrations of course. From what I know, it's pretty standard practice to enable CORS on API servers.

It's true, but I don't see what you want here. All the endpoint of cozy-stack protected by a token have CORS enabled. It's just the routes related to login/registrations (`/auth`) that don't have that.

y-lohse:
[...]

PR #2564, opened 2020-06-22
feat(auth): Register client preflight

sebn:
So one can register a client using XMLHttpRequest or Fetch APIs.

### DIFF
--- a/web/auth/auth.go
+++ b/web/auth/auth.go
@@ -536,6 +536,7 @@ func Routes(router *echo.Group) {
 
 	// Register OAuth clients
 	router.POST("/register", registerClient, middlewares.AcceptJSON, middlewares.ContentTypeJSON)
+	router.OPTIONS("/register", registerClientPreflight)
 	router.GET("/register/:client-id", readClient, middlewares.AcceptJSON, checkRegistrationToken)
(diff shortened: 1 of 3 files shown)


nono:
Hello, can you describe why do you want it?

See https://github.com/cozy/cozy-stack/issues/1610 for reasons why I don't want it. But I may change my opinion if there is a good use case AND I can find something in the OAuth RFCs to make it safer than today.

sebn:
Hey @nono 👋 
Thanks for the fast answer!

Sorry, I forgot to look for an existing issue and missed this one :)

My reasons are mostly the same as the ones mentioned in #1610.
I'm writing an app in Elm and I wanted to test it on my Cozy although it's not published yet.
@Crash-- told me it could be easier to use oauth for this than to go through the publication process.
So I was trying to write the registration logic in Elm... Which means client-side and relying on XMLHttpRequest indeed:
https://github.com/elm/http/blob/master/src/Elm/Kernel/Http.js#L24

To be honest, I wondered whether this could be a security issue...
And decided to make a PR and wait for feedback :P

So feel free to close this, I'll eventually find another way.

nono:
Yup, it would be better to find another way (like a small server part or maybe a script that register the client and generates an HTML page with the access token in a data attribute).

PR #2580, opened 2020-07-24
feat: allow connectors to be run in cozy-app-dev image

doubleface:
This way, connectors can be run in the cozy-app-dev image without any special setup, but this add
75Mo to the cozy-app-dev image.
Another solution would be to create another docker image specific for this goal.

### DIFF
--- a/scripts/Dockerfile
+++ b/scripts/Dockerfile
@@ -65,6 +65,11 @@ RUN set -eux; apt-get update \
   && mv ./rel/couchdb /usr/local \
   && cd / \
   && rm -rf /usr/src/couchdb \
+  && curl -fsSL --compressed https://nodejs.org/dist/v12.9.1/node-v12.9.1-linux-x64.tar.xz \
+  && tar -xJf "node-v12.9.1-linux-x64.tar.xz" -C /usr/local --strip-components=1 --no-same-owner \
+  && rm "node-v12.9.1-linux-x64.tar.xz" \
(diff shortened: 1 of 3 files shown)


scripts/Dockerfile
```diff
@@ -65,6 +65,11 @@ RUN set -eux; apt-get update \
   && mv ./rel/couchdb /usr/local \
   && cd / \
   && rm -rf /usr/src/couchdb \
+  && curl -fsSL --compressed https://nodejs.org/dist/v12.9.1/node-v12.9.1-linux-x64.tar.xz \
```
nono, line 68: The debian package for nodejs is too old?

scripts/Dockerfile
```diff
@@ -65,6 +65,11 @@ RUN set -eux; apt-get update \
...
+  && tar -xJf "node-v12.9.1-linux-x64.tar.xz" -C /usr/local --strip-components=1 --no-same-owner \
+  && rm "node-v12.9.1-linux-x64.tar.xz" \
+  && ln -s /usr/local/bin/node /usr/local/bin/nodejs \
+  && node --version \
```
nono, line 72: I think we can remove this line

scripts/Dockerfile
doubleface: It is just the same version as we have in production

scripts/Dockerfile
doubleface: node --version is  not mandatory but I supposed it was nice to have the node version in the build logs

PR #2583, opened 2020-07-31
Add lock on accounts migration

Gara64:
This lock is useful to prevent concurrent workers to be run at the same
time if several authentication requests are made at once

### DIFF
--- a/worker/migrations/migrate-accounts-to-organisation.go
+++ b/worker/migrations/migrate-accounts-to-organisation.go
@@ -12,6 +12,7 @@ import (
 	"github.com/cozy/cozy-stack/pkg/consts"
 	"github.com/cozy/cozy-stack/pkg/couchdb"
 	"github.com/cozy/cozy-stack/pkg/crypto"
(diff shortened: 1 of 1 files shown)


worker/migrations/migrate-accounts-to-organisation.go
```diff
@@ -185,12 +186,22 @@ func migrateAccountsToOrganization(domain string) error {
 	if err != nil {
 		return err
 	}
+	mu := lock.ReadWrite(inst, "migrate-accounts")
```
sblaisot, line 189: `migrate-accounts` seems a bit too generic.
I would have named it like the migration : ` accounts-to-organization-migration` to avoid future collision.

nono: I have exact the opposite opinion: it is a good thing to also reuse the same lock for other migrations on the accounts to avoid CouchDB conflicts on those documents

sblaisot: ok, that makes sense too.

PR #2600, opened 2020-08-17
Allow to configure a mail server per context 

nono:


cozy.example.yaml
```diff
@@ -318,10 +318,16 @@ contexts:
...
-    noreply_address: noreply@cozy.beta
-    noreply_name: My Cozy Beta
-    reply_to: support@cozy.beta
+    # Use a different config for mails for this context
+    mail:
+      # If the host is set to "-", no mail will be sent on this context
+      host: smtp.cozy.beta
+      port: 465
+      username: {{.Env.COZY_BETA_MAIL_USERNAME}}
+      password: {{.Env.COZY_BETA_MAIL_PASSWORD}}
```
sblaisot, line 327: Contexts are publicly accessible so it is not an option to store a password in contexts. please move it outside of context

cozy.example.yaml
```diff
@@ -318,10 +318,16 @@ contexts:
...
+      noreply_address: noreply@cozy.beta
```
sblaisot, line 328: moving this from context.name to context.name.mail should keep backward compatibility for older config scheme

sblaisot:
I think it would be better to have a `mail:` namespace in the config with a `default:` subnamespace and an overwrite per context like this have been done for registry spaces or clouderies

cozy.example.yaml
nono: Good catch! I could remove the mail entry from the HTTP response on the public route, but it is probably safer to not store mail configuration here.

cozy.example.yaml
nono: It is the case: the stack looks at both `context.name.noreply_address` and `context.name.mail.noreply_address` for example. I have documented only the new way, but there is no need to change the current configuration.

nono:
> I think it would be better to have a mail: namespace in the config with a default: subnamespace and an overwrite per context like this have been done for registry spaces or clouderies

It was also my first reaction, but it looks difficult to preserve the compatibility with the current config format and I didn't want to have to change the config at the same time as we deploy the new code. And when I saw the `noreply_address` in the context, I went on putting this configuration in the `contexts:` section.

cozy.example.yaml
sblaisot: Oh, I didn't saw that. Thanks.

nono:
I've put the configuration for the mail server of a contexts inside `mail.contexts.name`, and kept the default directly inside `mail`. It is not consistent with the registries, but it allows to preserve the backward compatibility without too much complexity.

PR #2606, opened 2020-08-19
Be explicit about sameSite for cookies

nono:
The session cookies will have a lax policy for sameSite, and the CSRF
cookies will have a strict policy.

See https://hacks.mozilla.org/2020/08/changes-to-samesite-cookie-behavior/

### DIFF
--- a/model/session/session.go
+++ b/model/session/session.go
@@ -216,6 +216,7 @@ func (s *Session) ToCookie() (*http.Cookie, error) {
 		Domain:   CookieDomain(inst),
 		Secure:   !build.IsDevRelease(),
 		HttpOnly: true,
+		SameSite: http.SameSiteLaxMode,
 	}, nil
 }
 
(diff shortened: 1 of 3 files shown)


taratatach:
Should we make a request to add `mycozy.cloud` to the [Public Suffix list](https://publicsuffix.org/) to make sure `alice.mycozy.cloud` and `bob.mycozy.cloud` are not considered as Same-Site?

nono:
> Should we make a request to add mycozy.cloud to the Public Suffix list to make sure alice.mycozy.cloud and bob.mycozy.cloud are not considered as Same-Site?

Absolutely not! We are putting the session cookies on `mycozy.cloud` to share the cookies between `alice.mycozy.cloud` and `alice-drive.mycozy.cloud`. It has not always been the case, but with browsers like Safari and Firefox that push to not send third party cookies, we needed to do move the cookies to `mycozy.cloud` to avoid reliability issues.

taratatach:
> We are putting the session cookies on `mycozy.cloud` to share the cookies between `alice.mycozy.cloud` and `alice-drive.mycozy.cloud`.

Isn't this a reliability when it comes to third-party applications running on the Cozy?

nono:
> Isn't this a liability when it comes to third-party applications running on the Cozy?

The cookies are marked as httpOnly, which means that the apps can't read the cookies.

PR #2607, opened 2020-08-20
Update Cozy-ui

nono:
It fixes an issue with recent Chrome where text is not centered inside buttons.

### DIFF
--- a/assets/templates/login.html
+++ b/assets/templates/login.html
@@ -58,7 +58,7 @@
             <input id="csrf_token" type="hidden" name="csrf_token" value="{{.CSRF}}" />
             <input id="two-factor-trusted-device-token" type="hidden" name="two-factor-trusted-device-token" value="" />
             <div class="o-field password-form u-m-0">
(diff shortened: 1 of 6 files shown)


assets/templates/login.html
```diff
@@ -58,7 +58,7 @@ <h2 class="password-form wizard-subtitle u-coolGrey">{{.Domain}}</h2>
             <input id="csrf_token" type="hidden" name="csrf_token" value="{{.CSRF}}" />
             <input id="two-factor-trusted-device-token" type="hidden" name="two-factor-trusted-device-token" value="" />
             <div class="o-field password-form u-m-0">
-              <label for="password" class="c-label">{{t "Login Password field"}}</label>
+              <label for="password" class="c-label c-label--block u-mt1">{{t "Login Password field"}}</label>
```
Crash--, line 61:
```suggestion
              <label for="password" class="c-label c-label--block u-mt-1">{{t "Login Password field"}}</label>
```

nono: Good catch :+1:

PR #2615, opened 2020-09-02
Add a route to show the avatar of a sharing member

nono:


docs/public.md
```diff
@@ -8,4 +8,11 @@ These routes are public: no authentication is required for them.
 
 ### GET /public/avatar
 
-Returns an image chosen by the user as its avatar, or the default avatar.
+Returns an image chosen by the user as their avatar. If no image has been
+chosen, a fallback will be used, depending of the `default` parameter in the
+query-string:
+
+- `cloud`: a default image that shows the Cozy Cloud logo, but it can be
```
y-lohse, line 15: Should we name it something more generic than `cloud` since it's not necessarily a cloud? Eg. `default` or `fallback`?

nono: Yeah, I was not very inspired. I have started with `default` but I found very weird to have `default=default` in a query-string.

y-lohse: Right 😁 . Maybe we can name the parameter `fallback`, support `fallback=initials` or `fallback=404`, and if there's no provided `fallback` we use the default automatically?

PR #2633, opened 2020-09-09
Support SoftDelete and Restore routes for bitwarden

paultranvan:
Bitwarden now [supports](https://github.com/bitwarden/server/pull/684) a trash system, where (soft) deleted ciphers are actually updated with a `deletedDate` attribute. Then, the client filters the ciphers having this attribute to display them in a Trash.
The restore simply consists of removing this attribute.

This PR adds the soft delete and restore routes.

### DIFF
--- a/web/bitwarden/ciphers.go
+++ b/web/bitwarden/ciphers.go
@@ -123,6 +123,7 @@ type cipherResponse struct {
 	Card           map[string]interface{} `json:"Card,omitempty"`
 	Identity       map[string]interface{} `json:"Identity,omitempty"`
 	Date           time.Time              `json:"RevisionDate"`
+	DeletedDate    *time.Time             `json:"DeletedDate,omitempty"`
 	Edit           bool                   `json:"Edit"`
 	UseOTP         bool                   `json:"OrganizationUseTotp"`
 }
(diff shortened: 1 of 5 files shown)


nono (commented):
Can you also add the documentation for the 2 new routes on https://github.com/cozy/cozy-stack/blob/master/docs/bitwarden.md?

web/bitwarden/ciphers.go
```diff
@@ -529,6 +535,82 @@ func DeleteCipher(c echo.Context) error {
...
+			"error": err.Error(),
+		})
+	}
+	cipher.Metadata.ChangeUpdatedAt()
+	cipher.DeletedDate = &cipher.Metadata.UpdatedAt
+	if err := couchdb.UpdateDoc(inst, cipher); err != nil {
+		return c.JSON(http.StatusInternalServerError, echo.Map{
+			"error": err.Error(),
+		})
+	}
```
nono, line 572: I think we should also update the revision date like for UpdateCipher:

```go
_ = settings.UpdateRevisionDate(inst, setting)
```

paultranvan: Done with e4de808

nono (approved):

PR #2634, opened 2020-09-10
Let recipients choose between shortcut or sync

nono:
When someone receives a mail for a sharing for the first time, and
accepts the sharing, the stack will ask them if they want to synchronize
the content of the sharing or just add a shortcut to it. For later
sharings, there is no choice as the shortcut has already been added to
their Cozy when they can accept the sharing.


joel-costa:
@nono Recettage : 
J'ai essayé de pointer les corrections sur les screenshots ci-dessous.
Je relink aussi les zeplins :
- https://zpl.io/awRQlJj
- https://zpl.io/29Qk3jl

image
image
image
image

PR #2644, opened 2020-09-15
Improve UI of the authorize sharing page

nono:
* https://zpl.io/awRQlJj
* https://zpl.io/29Qk3jl

### DIFF
--- a/assets/styles/stack.css
+++ b/assets/styles/stack.css
@@ -36,18 +36,6 @@ a:focus {
   min-height: 100%;
 }
 
-/* Utilities */
-/* TODO https://github.com/cozy/cozy-ui/issues/1555 */
-.u-br-silver {
-  border: solid 1px var(--silver);
-}
-.u-br-grey {
-  border: solid 1px rgba(0, 0, 0, 0.08);
-}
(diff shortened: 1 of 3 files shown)


assets/styles/stack.css
```diff
@@ -415,3 +431,10 @@ li.filetype-zip:before {
     padding: 0.5em;
   }
 }
+
+@media (min-width: 32rem) {
+  .with-sticky-footer footer {
+    margin: 0 calc(-1 * (100vw - 32rem) / 2);
+    padding: 1rem calc((100vw - 32rem) / 2 + 1rem);
```
nono, line 438: @ptbrowne I'm interested to know if you a less hacky way to add the box-shadow on the full width of the page but still restricts the content to a maximal width of 32rem.

ptbrowne: With two divs ? One with for the border, the other for the content ?

nono: Maybe, but I didn't succeed to place the div with the border at the right place (y position), both when the buttons are on the same line and when they are one above the other, and to have some other side effects that makes the whole page a mess.

PR #2660, opened 2020-09-23
Detect thumbnails with no file

nono:
When checking the file system of an instance in Swift layout v3, the
stack will also warns for thumbnails where the file that have been used
to generate it is missing.


sblaisot (approved):

sblaisot:
LGTM, thanks

PR #2674, opened 2020-09-29
Allow to choose the app to redirect on konnectors OAuth

nono:


web/accounts/oauth.go
```diff
@@ -57,9 +56,12 @@ func start(c echo.Context) error {
...
 }
 
-func redirectToApp(c echo.Context, acc *account.Account, clientState string) error {
+func redirectToApp(c echo.Context, acc *account.Account, clientState, slug string) error {
 	instance := middlewares.GetInstance(c)
-	u := instance.SubDomain(consts.HomeSlug)
+	if slug == "" {
+		slug = consts.HomeSlug
+	}
+	u := instance.SubDomain(slug)
```
sblaisot, line 64: What if given slug is an app that has not been installed on this cozy?
Don't we need to check this and fallback to `home` app if `slug` app is missing?

Crash-- (approved):

web/accounts/oauth.go
nono: I consider that it is the role of the caller to check if the app exists.

PR #2700, opened 2020-10-14
Fix a panic when notifying on a sharing revocation

nono:


ptbrowne:
Should we add a test in those cases ?

nono:
Well, I don't know how to reproduce, and it could be way harder to add a test than just fix the issue.

PR #2710, opened 2020-10-19
Include the sharing member in GET /permissions/self

nono:
When the GET /permissions/self route is called with a sharecode, the
stack will include the information about the linked member in the
response. In particular, if the instance field is filled, it means that
this member had a link to this sharing in their Cozy instance.


Crash-- (approved):
👍

PR #2716, opened 2020-10-23
Emphasize the data loss when reseting the password

paultranvan:
Before
image

After 
image


nono (approved):
LGTM, but don't forget to update the french locale on transifex. In theory, we should update only the english wording in the repository as it is the only file that transifex imports to its database, but in practice, it's a OK to take a shortcut and also update other locales as long as we also do the change on transifex.

PR #2738, opened 2020-11-09
Check carbonCopy and electronicSafe metadata

nono:
The metadata map for io.cozy.files cannot have a carbonCopy or
electronicSafe key unless the app/konnector that adds it has a
permission on io.cozy.certified.carbonCopy /
io.cozy.certified.electronicSafe.

When a file with such metadata attribute is shared, this attribute will
be removed on the instances of the other members: this attribute is only
valid localy, we don't want to trust other instances as it is easy to
cheat for a self-hosted.

### DIFF
--- a/docs/files.md
+++ b/docs/files.md
@@ -643,7 +643,7 @@ Send a metadata object that can be associated to a file uploaded after that,
 via the `MetadataID` query parameter.
 
 **Note:** a special permission on `io.cozy.certified.carbonCopy` is required to
-send a request with `carbonCopy` as key in the `attributes` map. Safe for
+send a request with `carbonCopy` as key in the `attributes` map. Same for
 `electronicSafe`.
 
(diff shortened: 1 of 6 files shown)


docs/files.md
```diff
@@ -642,6 +642,10 @@ more informations about the references field.
 Send a metadata object that can be associated to a file uploaded after that,
 via the `MetadataID` query parameter.
 
+**Note:** a special permission on `io.cozy.certified.carbonCopy` is required to
+send a request with `carbonCopy` as key in the `attributes` map. Safe for
```
LucsT, line 646: Typo here , Safe -> Same

LucsT (approved):
Ok for me. There is a typo in the doc.
I have a konnector ready to test that.

We probably need some strings for this new permissions in home App.

PR #2743, opened 2020-11-10
Fix files wildcard permission

paultranvan:
The files permissions are handled differently from the other doctypes, and it wasn't possible to request `io.cozy.files` docs when using a `io.cozy.files.*` permission. This should fix this behaviour.

### DIFF
--- a/model/vfs/permissions.go
+++ b/model/vfs/permissions.go
@@ -29,11 +29,11 @@ func Allows(fs VFS, pset permission.Set, v permission.Verb, fd Fetcher) error {
 	// First pass, we iterate over the rules, check if we have an easy match
 	// keep a short list of useful rules and allowed IDs.
 	for _, r := range pset {
-		isFileDoctype := false
-		if r.Type == consts.Files || permission.TrimWildcard(r.Type) == consts.Files {
-			isFileDoctype = true
+		typ := permission.TrimWildcard(r.Type)
+		if typ != consts.Files && !strings.HasPrefix(consts.Files, typ+".") {
+			continue
 		}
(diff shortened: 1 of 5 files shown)


model/vfs/permissions.go
```diff
@@ -29,7 +29,11 @@ func Allows(fs VFS, pset permission.Set, v permission.Verb, fd Fetcher) error {
 	// First pass, we iterate over the rules, check if we have an easy match
 	// keep a short list of useful rules and allowed IDs.
 	for _, r := range pset {
-		if r.Type != consts.Files || !r.Verbs.Contains(v) {
+		isFileDoctype := false
+		if r.Type == consts.Files || permission.TrimWildcard(r.Type) == consts.Files {
+			isFileDoctype = true
+		}
+		if !isFileDoctype || !r.Verbs.Contains(v) {
```
nono, line 36: I would prefer:

```go
typ := TrimWildcard(r.Type)
if typ != consts.Files && strings.HasPrefix(consts.Files, typ+".") {
  continue
}
if !r.Verbs.Contains(v) {
  continue
}
```

In particular, it would be safer if we may have doctypes like `io.cozy.files.foo.bar.baz` in the future.

web/files/files_test.go
```diff
@@ -612,6 +612,7 @@ func TestUploadWithMetadata(t *testing.T) {
 }
 
 func TestUploadWithSourceAccount(t *testing.T) {
+	fmt.Printf("token : %v\n", token)
```
nono, line 615: Can you remove this debug line?

paultranvan: Wops, done

nono (approved):

PR #2744, opened 2020-11-10
Add an initial_sync flag on sharings

nono:


Crash--:
Is it really needed since we have : `initial_number_of_files_to_sync`? It does almost the same thing

nono:
We can see `initial_sync` as `initial_number_of_files_to_sync > 0 && active`, so it's not really needed but it can help to make things simpler on the front-end. I'm OK to close this PR without merging.

Crash--:
Oh, I didn't have the check on `active` in mind. So OK to have this "complexity" done by the stack. Thanks for the explanation :)

PR #2755, opened 2020-11-16
Fix getting Cozy URL from contact when no https://

nono:


Crash-- (approved):
👍. Will still work locally, thanks!

PR #2762, opened 2020-11-18
Add a route to send mail to the support

nono:


docs/jobs.md
```diff
@@ -348,6 +348,39 @@ allowed):
 }
 ```
 
+### POST /jobs/support
+
+Send a mail to the support. It requires a permission on `io.cozy.support` (a
```
Crash--, line 353: This support's mail is defined by the `mail-reply-to` flag ? Or is it something in the config? Context or something else ?

nono: > This support's mail is defined by the mail-reply-to flag?

Yes, it is.

> Or is it something in the config?

Yes, if you prefer, you can also use the equivalent `mail.reply_to` in the config (https://github.com/cozy/cozy-stack/blob/master/cozy.example.yaml#L173)

> Context or something else

And it can be overloaded by context: https://github.com/cozy/cozy-stack/blob/master/cozy.example.yaml#L342

Crash--: OK thanks ! What do you think about adding this information to the doc ? 

Send a mail to the support (email address defined by [mail.reply_to](https://github.com/cozy/cozy-stack/blob/master/cozy.example.yaml#L173)  or overwritten by the [--mail-reply-to flag](https://docs.cozy.io/en/cozy-stack/cli/cozy-stack_serve/) ). 

It requires a permission on `io.cozy.support` (a permission on `io.cozy.jobs:POST:sendmail:worker`, or larger, is also accepted
to ease the transition from sending manually a mail to the support via the
sendmail queue).

nono: If you ask, it is probably something to add to the doc: https://github.com/cozy/cozy-stack/commit/098acce3e518f9cffc139e41a7aa056ba99391fc

Crash--: Thanks!

PR #2768, opened 2020-11-19
docs: Add docs about SMS notification

Crash--:
Adding a few lines about SMS notification. I haven't used it yet, but it will help answer some questions from partners ;). But I'm not sure that ./worker#sms is the best place to do it.

I'm wondering, currently for push notification configuration, we're setting a document in the `secrets/io-cozy-account_types` database. Can we do the same for SMS provider?

### DIFF
--- a/docs/workers.md
+++ b/docs/workers.md
@@ -59,7 +59,16 @@ To use this worker from a client-side application, you should use
 ## sms worker
 
 The `sms` worker can be used to send SMS notifications to a user, via
-[the notifications API](./notifications.md).
+[the notifications API](./notifications.md). 
+
+The phone number used is the `primary phone number` registered in 
+the contact that has the flag `me: true` (see [contact doctype](https://github.com/cozy/cozy-doctypes/blob/master/docs/io.cozy.contacts.md) )
+
+Note: If an error happens during the send of the sms (for instance no phone number on the cozy or SMS provider not available), the content will be send by email as fallback.
+
+To use the sms worker, you need to configure your SMS provider:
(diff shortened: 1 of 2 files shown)


nono:
> I'm wondering, currently for push notification configuration, we're setting a document in the secrets/io-cozy-account_types database. Can we do the same for SMS provider?

Well, we could but I don't think it is a good idea. For the push notifications, it can be configured via the config file (https://github.com/cozy/cozy-stack/blob/master/cozy.example.yaml#L270), but if you want to overload per context, you have to use the account types. I don't understand why it works this way. For me, notification channels are known by the stack, and should be configured in the config file ([mail](https://github.com/cozy/cozy-stack/blob/master/cozy.example.yaml#L168-L193), push, and SMS). The account types should be reserved for external services in relation to the apps registry.

nono (approved):

docs/workers.md
```diff
@@ -59,7 +59,16 @@ To use this worker from a client-side application, you should use
 ## sms worker
 
 The `sms` worker can be used to send SMS notifications to a user, via
-[the notifications API](./notifications.md).
+[the notifications API](./notifications.md). 
+
+The phone number used is the `primary phone number` registered in 
+the contact that has the flag `me: true` (see [contact doctype](https://github.com/cozy/cozy-doctypes/blob/master/docs/io.cozy.contacts.md) )
+
+Note: If an error appends during the send of the sms (for instance no phone number on the cozy or SMS provider not available), the content will be send by email as fallback.
```
nono, line 67: s/appends/happens/

PR #2781, opened 2020-11-25
fix: Do not return err if m.Notifications is nil

Crash--:
Since err can be nil when m.Notifications is nil, we're not sending
any error at the end in that case. We returned a 201 OK whereas we're
not creating anything.

Later in the code, we check if m.Notifications[p.category] exists. If
not we are returning an error. Let's use this error.

I had this case when the app manifest was not well build for the notifications part (having the notifications object part of the permissions object, not a new one)


nono:
> Later in the code, we check if m.Notifications[p.category] exists. If
not we are returning an error. Let's use this error.

In Go, if we try to access an item from a nil map, it panics, which is not what we want

nono (approved):

PR #2785, opened 2020-11-27
Fix lack of support of wildcard permission in several cases

paultranvan:
The permission logic is splitted in many places in the stack and it some cases we didn't handled the wildcard permission logic, in particular for files.

This also removes unused code in web/jobs, where we implemented a specific logic to handle jobs permissions for
apps to manage konnectors. This does not seem useful anymore as this is handled directly through harvest.

### DIFF
--- a/web/jobs/jobs.go
+++ b/web/jobs/jobs.go
@@ -11,8 +11,6 @@ import (
 
 	"github.com/justincampbell/bigduration"
 
-	"github.com/cozy/cozy-stack/model/app"
-	"github.com/cozy/cozy-stack/model/instance"
 	"github.com/cozy/cozy-stack/model/job"
 	"github.com/cozy/cozy-stack/model/permission"
 	"github.com/cozy/cozy-stack/pkg/config/config"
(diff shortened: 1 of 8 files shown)


worker/exec/konnector.go
```diff
@@ -370,7 +370,7 @@ func (w *konnectorWorker) ensurePermissions(inst *instance.Instance) error {
 	}
 	value := consts.Konnectors + "/" + w.slug
 	for _, rule := range perms.Permissions {
-		if rule.Type == consts.Files && rule.Selector == couchdb.SelectorReferencedBy {
+		if permission.MatchType(rule, consts.Files) && rule.Selector == couchdb.SelectorReferencedBy {
```
nono, line 373: I don't think we want to support wildcard permission with a referenced_by selector. I would have let this line of code like it was.

web/jobs/jobs.go
```diff
@@ -320,15 +316,8 @@ func getTriggerState(c echo.Context) error {
 	if err != nil {
 		return wrapJobsError(err)
 	}
-
 	if err = middlewares.Allow(c, permission.GET, t); err != nil {
-		// For konnector trigger, we have a specific logic to check for
-		// permissions: if the application permissions and the konnector
-		// permissions intersect on one or more doctypes, we give the right to the
-		// application to read the trigger state.
```
nono, line 328: Can you add a paragraph about this change in https://github.com/cozy/cozy-stack/blob/master/docs/important-changes.md please?

paultranvan: Done with 64b8bf1a0ed2fcefc09e3369d5840bf713db3fd9

nono (approved):

PR #2798, opened 2020-12-10
Fix a race condition when installing app with services

nono:
When the stack receive several requests to install the same application
at the same time, a race condition happens and the triggers for this app
may be created several times. The fix is to create the triggers only
after creating the CouchDB document for the application, as it will
return a conflict in that case.


paultranvan (approved):
Nice!

PR #2812, opened 2020-12-17
Fix services execution with --appdir

nono:
When a stack is started with the --appdir option, it should take the .js
file for a service from the given appdir, not from where the application
is normally installed.


paultranvan (approved):
:clap:

PR #2813, opened 2020-12-18
Add tinycode support

nono:
When creating a permission for a sharing by link, it is now possible to
create a tinycode made of 6 digits by using the `tiny=true` parameter in
the query-string when the sharing has a TTL that doesn't exceed 1 hour.

### DIFF
--- a/web/permissions/permissions.go
+++ b/web/permissions/permissions.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"strconv"
 	"strings"
 	"time"
 
@@ -105,6 +106,7 @@ func createPermission(c echo.Context) error {
 	instance := middlewares.GetInstance(c)
 	names := strings.Split(c.QueryParam("codes"), ",")
 	ttl := c.QueryParam("ttl")
+	tiny, _ := strconv.ParseBool(c.QueryParam("tiny"))
 
 	parent, err := middlewares.GetPermission(c)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


web/permissions/permissions.go
```diff
@@ -128,6 +130,17 @@ func createPermission(c echo.Context) error {
 		return err
 	}
 
+	var expiresAt *time.Time
+	if ttl != "" {
+		if d, errd := bigduration.ParseDuration(ttl); errd == nil {
+			ex := time.Now().Add(d)
+			expiresAt = &ex
+			if d.Hours() > 1.0 {
+				tiny = false
```
Crash--, line 139: Since we can not respect what the dev expect, is it possible to add a log to explain why? Even if it's written in the documentation I think it's a good to have it logged

```suggestion
     if tiny == true {
        log('Tiny can not be set to true since duration is > 1h')
     }       
     tiny = false
```

web/permissions/permissions_test.go
```diff
@@ -681,6 +708,31 @@ func TestCannotFindToken(t *testing.T) {
...
+	assert.Equal(t, res1.StatusCode, http.StatusOK)
+
+	token, _ := permission.GetTokenFromShortcode(testInstance, perm.ShortCodes["elise"])
+	assert.Equal(t, perm.Codes["elise"], token)
+}
+
+func TestTinyShortCodeInvalid(t *testing.T) {
+	_, codes, _ := createTestTinyCode(token, "fanny", "24h")
+	code := codes["fanny"]
+	assert.Len(t, code, 12)
```
Crash--, line 733: In fact, I'm wondering if we should not drop the shortcode creation (by returning a 400 error code) if tiny is set and duration > 1h ?

nono: I don't think so. I have hesitated when coding it, but I think we may want to change the limit and it should not break the apps: they will have a shortcode instead of a tinycode but it should not break anything as the format of the response is the same.

PR #2827, opened 2020-12-28
Explain how to use the --deleting flag for `instances modify` command

nicofrand:


nono:
Thanks for the PR. The manpages are generated, I have also updated the source code for this change.

PR #2851, opened 2021-01-15
Add a route to generate PDF preview images

nono:
**TODO** add a caching layer (Swift or local FS)

### DIFF
--- a/model/vfs/preview.go
+++ b/model/vfs/preview.go
@@ -1,0 +1,79 @@
+package vfs
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"os/exec"
+
+	"github.com/cozy/cozy-stack/pkg/config/config"
+	"github.com/cozy/cozy-stack/pkg/logger"
+)
+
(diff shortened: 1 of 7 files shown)


model/vfs/preview.go
```diff
@@ -0,0 +1,79 @@
...
+	"os"
+	"os/exec"
+
+	"github.com/cozy/cozy-stack/pkg/config/config"
+	"github.com/cozy/cozy-stack/pkg/logger"
+)
+
+// ServePDFPreview will send the preview image for a PDF.
+func ServePDFPreview(w http.ResponseWriter, req *http.Request, fs VFS, doc *FileDoc) error {
+	name := fmt.Sprintf("%s-preview.jpg", doc.ID())
```
sblaisot, line 17: Maybe you can add a check on file mimetype to ensure it is a PDF.

nono: Nope, the check https://github.com/cozy/cozy-stack/pull/2851/files#diff-977f264df65962f14320bbe0d18e0b81db367cf1e2f64f48572df5ccee81cbe3R342 is enough

PR #2855, opened 2021-01-18
Add secret injection for remote doctypes

nono:


docs/remote.md
```diff
@@ -169,6 +169,33 @@ The requests are logged as the `io.cozy.remote.requests` doctype, with the
 doctype asked, the parameter (even those that have not been used, like `comment`
 in the previous example), and the application that has made the request.
 
+## Secrets
+
+It is possible to make the stack inject a secret in a request. For example, if
+we want to make the stack add a `token`, we can use `{{secret_token}}` in the
```
Crash--, line 175: I understand the choice of adding `secret_` in the request template to show that this information is "secret" and will be added by the cozy-stack during the execution. 
But at first I found it disconcerting to have in the doctype the normal name and in the template the `secret_`name. Is it something we already do for `io-cozy-account_types`?

Crash-- (approved):
Just a small question, but looks good thanks

docs/remote.md
nono: I don't think it is comparable, there is no template for `io-cozy-account_types`. For example, if a konnector is retrieving data from a service with OAuth, the konnector can only ask the stack to make the OAuth requests and only the `grant_type` parameter tells how to use `client_id`, `client_secret`, and co.

PR #2856, opened 2021-01-19
Improve cozy-stack check triggers

nono:
The command will now detect triggers for an application service that
have no name in their message.

### DIFF
--- a/web/instances/checks.go
+++ b/web/instances/checks.go
@@ -187,6 +187,15 @@ func checkTriggers(c echo.Context) error {
 					"message": fmt.Sprintf("%s", t.Message),
 				})
 			}
+			if slug, _ := msg["slug"].(string); slug == "" {
+				results = append(results, map[string]interface{}{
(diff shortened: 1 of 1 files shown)


web/instances/checks.go
```diff
@@ -164,6 +164,32 @@ func checkTriggers(c echo.Context) error {
...
+				results = append(results, map[string]interface{}{
+					"type":    "invalid_message",
+					"_id":     t.TID,
+					"trigger": t.Type,
+					"worker":  t.WorkerType,
+					"message": fmt.Sprintf("%s", t.Message),
+				})
+				continue
+			}
+			if name, _ := msg["name"].(string); name == "" {
```
sblaisot, line 181: shouldn't we also check for `slug` also?

sblaisot (approved):

web/instances/checks.go
nono: Yup, good idea!

PR #2882, opened 2021-02-01
Add _design docs routes

paultranvan:
This adds the following routes:
- `GET /data/:doctype/_design_docs`: access a design doc
- `DELETE /data/:doctype/_design/:ddoc`: delete a design doc
- `POST /data/:doctype/_design/:ddoci/copy`: copy a design doc

This will be useful in order to optimize the index creation from the client side: we used to create index with no name, relying on CouchDB to generate an id. We now want to create named index, by copying existing index to avoid recomputing existing ones.

### DIFF
--- a/pkg/couchdb/couchdb.go
+++ b/pkg/couchdb/couchdb.go
@@ -281,21 +281,19 @@ func buildCouchRequest(db Database, doctype, method, path string, reqjson []byte
 }
 
 func handleResponseError(db Database, resp *http.Response) error {
-	log := logger.WithDomain(db.DomainName()).WithField("nspace", "couchdb")
-
-	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
-		var body []byte
-		body, err := ioutil.ReadAll(resp.Body)
-		if err != nil {
-			err = newIOReadError(err)
-			log.Error(err.Error())
-		} else {
-			err = newCouchdbError(resp.StatusCode, body)
-			log.Debug(err.Error())
-		}
-		return err
+	if resp.StatusCode >= 200 && resp.StatusCode < 300 {
+		return nil
 	}
-	return nil
+	log := logger.WithDomain(db.DomainName()).WithField("nspace", "couchdb")
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		err = newIOReadError(err)
+		log.Error(err.Error())
+	} else {
+		err = newCouchdbError(resp.StatusCode, body)
+		log.Debug(err.Error())
+	}
+	return err
 }
 
 func makeRequest(db Database, doctype, method, path string, reqbody interface{}, resbody interface{}) error {
@@ -859,9 +857,7 @@ func DefineIndexes(g *errgroup.Group, db Database, indexes []*mango.Index) {
 
 // Copy copies an existing doc to a specified destination
 func Copy(db Database, doctype, path, destination string) (map[string]interface{}, error) {
-	headers := make(map[string]string)
-	headers["Destination"] = destination
-	headers["Content-Type"] = "application/json"
+	headers := map[string]string{"Destination": destination}
 	// COPY is not a standard HTTP method
 	req, err := buildCouchRequest(db, doctype, "COPY", path, nil, headers)
 	if err != nil {
--- a/pkg/couchdb/couchdb.go
+++ b/pkg/couchdb/couchdb.go
@@ -251,9 +251,56 @@ func dbNameHasPrefix(dbname, dbprefix string) (bool, string) {
 	return true, strings.Replace(dbname, dbprefix, "", 1)
 }
 
+func buildCouchRequest(db Database, doctype, method, path string, reqjson []byte, headers map[string]string) (*http.Request, error) {
+	if doctype != "" {
+		path = makeDBName(db, doctype) + "/" + path
+	}
+	req, err := http.NewRequest(
+		method,
+		config.CouchURL().String()+path,
+		bytes.NewReader(reqjson),
+	)
+	// Possible err = wrong method, unparsable url
+	if err != nil {
+		return nil, newRequestError(err)
+	}
+	req.Header.Add("Accept", "application/json")
+	if len(reqjson) > 0 {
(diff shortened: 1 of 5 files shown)


web/data/data.go
```diff
@@ -398,6 +398,62 @@ func normalDocs(c echo.Context) error {
...
+		return err
+	}
+	return proxy(c, "_design_docs")
+}
+
+func copyDesignDoc(c echo.Context) error {
+	doctype := c.Param("doctype")
+	ddoc := c.Param("designdocid")
+
+	c.Request().Method = "COPY"
```
nono, line 426: It's quite ugly to have to modify the incoming request :-/

web/data/data.go
```diff
@@ -398,6 +398,62 @@ func normalDocs(c echo.Context) error {
...
+		return err
+	}
+	return proxy(c, "_design/"+ddoc)
+}
+
+func deleteDesignDoc(c echo.Context) error {
+	ddoc := c.Param("designdocid")
+	doctype := c.Param("doctype")
+
+	if err := middlewares.AllowWholeType(c, permission.DELETE, doctype); err != nil {
```
nono, line 446: I think we should `CheckWritable` here

web/data/data.go
```diff
@@ -398,6 +398,62 @@ func normalDocs(c echo.Context) error {
 	return c.JSON(http.StatusOK, res)
 }
 
+func getDesignDoc(c echo.Context) error {
+	doctype := c.Get("doctype").(string)
```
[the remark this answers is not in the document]
paultranvan: This method was actually extracted from replication.go, so I kept it like this, but agreed.

web/data/data.go
paultranvan: Indeed, but what else could we do as the COPY is not supported by go, and with no possible override from echo? :/

web/data/data.go
```diff
@@ -398,6 +398,62 @@ func normalDocs(c echo.Context) error {
...
+	ddoc := c.Param("designdocid")
+
+	if err := middlewares.AllowWholeType(c, permission.GET, doctype); err != nil {
```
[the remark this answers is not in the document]
paultranvan: Ok, as this route was already used for replication, there is no risk to add this check, e.g. for desktop?

web/data/data.go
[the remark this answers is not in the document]
nono: Maybe we can make the request to CouchDB without using the `proxy` method

web/data/data.go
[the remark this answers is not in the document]
nono: It is not possible to replicate a doctype that is not readable (the replication uses other routes that check that).

web/data/data.go
```diff
@@ -398,6 +398,62 @@ func normalDocs(c echo.Context) error {
...
+	ddoc := c.Param("designdocid")
+
+	c.Request().Method = "COPY"
+	header := c.Request().Header
+
+	newDdoc := header.Get("Destination")
+	if newDdoc == "" {
+		return c.JSON(http.StatusBadRequest, "You must set a Destination header")
+	}
+	if err := permission.CheckReadable(doctype); err != nil {
```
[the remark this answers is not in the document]
paultranvan: Actually, this would block the possibility to query design docs for io.cozy.files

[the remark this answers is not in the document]
paultranvan: Maybe we could simply remove this check? I guess it's ok to manipulate index, even if it's a "protected" doctype,  as long as the querier has the rights on the doctype ?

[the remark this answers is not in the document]
nono: I think we should keep the `CheckReadable`: it is OK to add a design doc for a doctype you can read

pkg/couchdb/couchdb.go
```diff
@@ -839,6 +857,30 @@ func DefineIndexes(g *errgroup.Group, db Database, indexes []*mango.Index) {
 	}
 }
 
+// Copy copies an existing doc to a specified destination
+func Copy(db Database, doctype, path, destination string) (map[string]interface{}, error) {
+	headers := make(map[string]string)
+	headers["Destination"] = destination
+	headers["Content-Type"] = "application/json"
```
nono, line 864: Why do you add  a `Content-Type: application/json` here? I don't see a body in the request

pkg/couchdb/couchdb.go
```diff
@@ -839,6 +857,30 @@ func DefineIndexes(g *errgroup.Group, db Database, indexes []*mango.Index) {
 	}
 }
 
+// Copy copies an existing doc to a specified destination
+func Copy(db Database, doctype, path, destination string) (map[string]interface{}, error) {
+	headers := make(map[string]string)
```
nono, line 862: Maybe we can use the shorter `headers := map[string]string{"Destination": destination}`

pkg/couchdb/couchdb.go
```diff
@@ -251,21 +251,63 @@ func dbNameHasPrefix(dbname, dbprefix string) (bool, string) {
...
+		if err != nil {
+			err = newIOReadError(err)
+			log.Error(err.Error())
+		} else {
+			err = newCouchdbError(resp.StatusCode, body)
+			log.Debug(err.Error())
+		}
+		return err
+	}
+	return nil
```
nono, line 298: [nit] I prefer to write this function with a short if (less indentation levels):

```go
func handleResponseError(db Database, resp *http.Response) error {
	if resp.StatusCode >= 200 && resp.StatusCode < 300 {
		return nil
	}

	log := logger.WithDomain(db.DomainName()).WithField("nspace", "couchdb")
	body, err := ioutil.ReadAll(resp.Body)
	if err != nil {
		err = newIOReadError(err)
		log.Error(err.Error())
	} else {
		err = newCouchdbError(resp.StatusCode, body)
		log.Debug(err.Error())
	}
	return err
}
````

paultranvan: Right, looks better :+1:

pkg/couchdb/couchdb.go
paultranvan: Right, it's a test residual, my bad

nono (approved):

PR #2890, opened 2021-02-10
Make CSP configurable per context

nono:
Note: the config per context is only available via the `csp_allowlist key`
in the config file (not via the deprecated `csp_whitelist`).

### DIFF
--- a/cozy.example.yaml
+++ b/cozy.example.yaml
@@ -295,6 +295,10 @@ csp_allowlist:
   # style:  https://allowed.domain.com/
   # font:   https://allowed.domain.com/
 
+  contexts:
+    beta:
+      img: https://allowed2.domain.com/
+
 # It can useful to disable the CSP policy to debug and test things in local
 # disable_csp: true
 
--- a/cozy.example.yaml
+++ b/cozy.example.yaml
@@ -295,6 +295,8 @@ csp_allowlist:
   # style:  https://allowed.domain.com/
   # font:   https://allowed.domain.com/
 
+  # It is also possible to configure the CSP policy per context. The values are
+  # cumulative with the global csp allowlist.
   contexts:
     beta:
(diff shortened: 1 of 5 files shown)


cozy.example.yaml
```diff
@@ -295,6 +295,10 @@ csp_allowlist:
   # style:  https://allowed.domain.com/
   # font:   https://allowed.domain.com/
 
+  contexts:
```
sblaisot, line 298: Does the context CSP replace the default CSP or are they cumulative? IMHO this should be documented at least with a comment here.

pkg/config/config/config.go
```diff
@@ -663,6 +664,27 @@ func UseViper(v *viper.Viper) error {
 		}
 	}
 
+	cspAllowList := map[string]string{}
+	cspPerContext := map[string]map[string]string{}
+	cspList := v.GetStringMap("csp_allowlist")
+	for key, value := range cspList {
+		if val, ok := value.(string); ok {
+			cspAllowList[key] = val
+		} else if val, ok := value.(map[string]interface{}); ok && key == "contexts" {
```
sblaisot, line 673: probably you should first check for `key == "contexts"` and only try to decode if it maches.
This would avoid trying to decode something not intended to be decoded here.

sblaisot (approved):

cozy.example.yaml
nono: They are cumulative. And yes, I should add a comment here.

pkg/config/config/config.go
nono: I don't understand what you want to do. Is it an optimization for returning faster an error if the config is invalid?

sblaisot: finally no need to change anything here

PR #2892, opened 2021-02-12
Add context settings for the error page

nono:


cozy.example.yaml
```diff
@@ -350,6 +350,9 @@ contexts:
     noreply_address: noreply@cozy.beta
     noreply_name: My Cozy Beta
     reply_to: support@cozy.beta
+    # Configure the error page
+    support_address: support@cozy.beta
+    hide_button_on_app_not_found: true
```
sblaisot, line 355: Why a context property configured in the stack config and requiring a stack restart on change instead of a (dynamic) context/ratio feature flag?

sblaisot: Using a feature flag can also be useful to test the feature on a single instance or define it environment-wide

PR #2893, opened 2021-02-12
Add optional pagination for GET /instances

nono:


sblaisot (approved):

sblaisot:
IIUC we don't handle the `skip` here so in practice this would work for feching instance list in page order but we can't return to the previous page.

Is it possible to handle skip also?

PR #2958, opened 2021-03-22
fix: Pass encrypt key when force updating passphrase

ptbrowne:
Otherwise, key was not replaced in the bitwarden settings document and then
the vault was unusable


nono (approved):

ptbrowne:
Thanks, the session has been preserved after I changed the password 🎉

PR #2966, opened 2021-03-26
Forbid the deletion of stack indexes/views

nono:
The stack needs a few CouchDB design docs for its own indexes and views.
The front applications will no longer be able to delete them with the
DELETE /data/:doctype/_design/:ddoc route.


paultranvan (approved):

Crash-- (approved):
I think it's a good fix as a first step 👍

PR #2977, opened 2021-04-14
Improve the filenames for notes

nono:
The filename of a note is generated from its title. We should avoid some
characters like ? to ensure that the filename is compatible with the
windows and macOS file systems, so that the file can be synchronized
with the desktop client safely.

### DIFF
--- a/model/note/note.go
+++ b/model/note/note.go
@@ -261,8 +261,15 @@ func titleToFilename(inst *instance.Instance, title string, updatedAt time.Time)
 		name = inst.Translate("Notes New note")
 		name += " " + updatedAt.Format(time.RFC3339)
 	}
-	name = strings.ReplaceAll(name, "/", "-")
-	name = strings.ReplaceAll(name, ":", "-")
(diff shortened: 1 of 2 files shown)


model/note/note.go
```diff
@@ -261,8 +261,11 @@ func titleToFilename(inst *instance.Instance, title string, updatedAt time.Time)
 		name = inst.Translate("Notes New note")
 		name += " " + updatedAt.Format(time.RFC3339)
 	}
-	name = strings.ReplaceAll(name, "/", "-")
-	name = strings.ReplaceAll(name, ":", "-")
+	// Create file with a name compatible with Windows/macOS to avoid
+	// synchronization issues with the desktop client
+	r := strings.NewReplacer("/", "-", ":", "-", "<", "-", ">", "-",
```
taratatach, line 266: We should also replace `\` as it's forbidden on Windows.

taratatach:
Should we go all the way and make sure filenames won't be too long? They can't be more than 255 characters long on Linux and macOS (and 256 on Windows).

PR #2980, opened 2021-04-14
docs: Update onboarded_redirection doc

Crash--:
I had an hard time to understand that the first word was the slug of the app. I first though that it has to be a specified route of the Home application


nono (approved):

nono:
Oh yeah, long time ago, this application was named cozy-collect (and it was renamed to cozy-home when it became the default app, instead of drive).

PR #3002, opened 2021-04-27
Allow to put in maintenance a konnector on an environment

nono:
With this PR, it will now be possible to put a konnector for only one environment (aka a stack). Before, it was only possible to do that by updating the konnector on the apps registry, which impacts all the environments.

We have 2 CLI commands to manipulate this: `cozy-stack konnectors maintenance` and `cozy-stack konnectors deactivate-maintenance`. And we have modified two routes to use this information: `GET /registry/maintenance` and `GET /registry/:app`.

### DIFF
--- a/client/apps.go
+++ b/client/apps.go
@@ -201,6 +201,32 @@ func (c *Client) UninstallApp(opts *AppOptions) (*AppManifest, error) {
 	return readAppManifest(res)
 }
 
+// ActivateMaintenance is used to activate the maintenance for a konnector
+func (c *Client) ActivateMaintenance(slug string, opts map[string]interface{}) error {
+	data := map[string]interface{}{"attributes": opts}
+	body, err := writeJSONAPI(data)
+	if err != nil {
+		return err
+	}
+	_, err = c.Req(&request.Options{
+		Method:     "PUT",
+		Path:       "/konnectors/maintenance/" + slug,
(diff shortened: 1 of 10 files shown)


doubleface (approved):

doubleface (approved):

Crash--:
I guess we do that for environnements that are not the Cozy's ones. But like this quickly, I would have expected to see these changes on the registry side and not on the stack side, right? 

So now, the application has to get the maintenance list from the registry and the maintenance list from the stack and merge them to know which konnectors are in maintenance?

nono:
> I guess we do that for environnements that are not the Cozy's ones.

No, it is mostly for putting a konnector on maintenance on one of the Cozy's environment without impacting the others or the self-hosted (things like IPs blacklisted on OVH but not on Gandi).

> But like this quickly, I would have expected to see these changes on the registry side and not on the stack side, right?

We can do that by creating new virtual spaces on the registry and overriding the maintenance status, but it is a lot of manual work and it is quite fragile.

> So now, the application has to get the maintenance list from the registry and the maintenance list from the stack and merge them to know which konnectors are in maintenance?

No, in theory, the application could still work as is. This PR updates the stack to fetch both statuses on the endpoints used by the application.

Crash-- (approved):

PR #3017, opened 2021-05-18
Clean unused OAuth clients

nono:
If 1 hour after an OAuth client has been registered, and no access code has been generated for this client, it will be deleted. It will help avoiding unused OAuth clients in the clients page of settings.

@sblaisot FYI, there is a new worker (`clean-clients`).


pkg/limits/rate_limiting.go
```diff
@@ -199,6 +201,12 @@ var configs = []counterConfig{
 		Limit:  30,
 		Period: 1 * time.Hour,
 	},
+	// JobThumbnail
```
sblaisot, line 204: Probably this comment is a copy/paste junk

nono: Yes, good catch!

PR #3051, opened 2021-06-17
Use cozy.localhost instead of cozy.tools

nono:
By using http://cozy.localhost:8080/ as the default instance for
developement, we don't need JS polyfill for window.crypto.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ run:
 
 ## instance: create an instance for local development
 instance:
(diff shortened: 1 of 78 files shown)


Crash-- (commented):
Nice! Was looking for the polyfill but you already makes the modification few days ago. Missed see this change https://github.com/cozy/cozy-stack/commit/d19007435bb3f68796b32676bfcea04ae0124f35

PR #3077, opened 2021-07-09
Fix sending mails with attachments

nono:


Crash-- (approved):
You rock! Works well, thanks! 👍

PR #3079, opened 2021-07-12
Add a route to upload an image for a note

nono:


acezard (approved):

nono:
We want to go a different direction where the images are saved inside the note (tar archive) instead of being other files in the VFS. Cf https://github.com/cozy/cozy-stack/pull/3087

PR #3086, opened 2021-07-21
Allow to share ciphers between Cozys

nono:


model/bitwarden/organization.go
```diff
@@ -0,0 +1,108 @@
...
+const (
+	// OrgMemberInvited is used when the member is invited but has not yet
+	// accepted the invitation.
+	OrgMemberInvited = 0
+	// OrgMemberAccepted is used when the member is accepted but the owner has
+	// not yet confirmed that the fingerprint is OK.
+	OrgMemberAccepted = 1
+	// OrgMemberConfirmed is used when the member is confirmed, and has access
+	// to the organization key to decrypt/encrypt ciphers.
+	OrgMemberConfirmed = 2
```
Ldoppea, line 26: What if the user reject the invitation? Is this a new status? Or is the member removed from the organisation?

web/bitwarden/organization.go
```diff
@@ -1,64 +1,93 @@
...
-		Email:          string(email),
+		ID:             org.ID(),
+		Identifier:     nil, // Not supported by us
+		Name:           org.Name,
+		Key:            m.Key,
+		Email:          m.Email,
 		Plan:           "TeamsAnnually",
-		PlanType:       5, // TeamsAnnually plan
-		Seats:          2,
+		PlanType:       9,  // TeamsAnnually plan
```
Ldoppea, line 89: This number may be subject to future changes. Shouldn't we recreate [this enum](https://github.com/bitwarden/jslib/blob/1016bbfb9eb28c220de8d2ab86d1f2757328f254/common/src/enums/planType.ts) in go or maybe use a constant?

model/bitwarden/organization.go
nono: I don't know yet, but my guess would be that the member is removed from the organization.

web/bitwarden/organization.go
nono: No, we don't really care. It's like `Seats`: the client is expecting a value, but it doesn't do anything from it. It is normally checked server-side.

Ldoppea:
Commit https://github.com/cozy/cozy-stack/pull/3086/commits/a2339aa7f688c09339868a186e9bf0f4677490e7 breaks the notification service as it is guarded by the following code in `jslib/src/services/notifications.service.ts`:
```ts
private async processNotification(notification: NotificationResponse) {
  // ...

  const isAuthenticated = await this.userService.isAuthenticated();
  const payloadUserId = notification.payload.userId || notification.payload.UserId;
  const myUserId = await this.userService.getUserId();
  if (isAuthenticated && payloadUserId != null && payloadUserId !== myUserId) {
    return;
  }
```

With this commit, `payloadUserId !== myUserId` will always be `true`.

Side effect : created ciphers are not propagated between clients in realtime even if they are not in a shared folder.

nono:
I had a git conflict. I made a rebase (from master) and a push-force to resolve it.

PR #3098, opened 2021-08-12
Debug app install on CI

nono:


nono:
It is a network issue between Azure and OVH, when sometimes the downloads are very slow (about 30 kB/s).

PR #3148, opened 2021-09-21
Add a FromCloudery flag for PATCH /instances/:domain

nono:
It allows to avoid calling the cloudery when the email or public name is
changed and the change is coming from the cloudery.

### DIFF
--- a/model/instance/lifecycle/patch.go
+++ b/model/instance/lifecycle/patch.go
@@ -206,7 +206,7 @@ func managerUpdateSettings(inst *instance.Instance, changes map[string]interface
 		return
 	}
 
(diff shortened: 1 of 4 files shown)


model/instance/lifecycle/patch.go
```diff
@@ -189,7 +189,9 @@ func Patch(i *instance.Instance, opts *Options) error {
 		}
 	}
 
-	managerUpdateSettings(i, clouderyChanges)
+	if opts.FromCloudery {
```
sblaisot, line 192: shouldn't this logic be the other way?
if FromCloudery is true, we should not notify cloudery

sblaisot (approved):

PR #3149, opened 2021-09-22
Allow to filter fields for mango queries on files 

paultranvan:
Note we must keep mandatory fields for JSON-API to allow fields filtering.

This PR is bigger than expected mainly because the CreatedAt and UpdatedAt fields are now pointers to be able to omit them when empty.


nono:
Don't forget to update the `Clone` method of vfs.DirDoc and vfs.FileDoc

web/files/files.go
```diff
@@ -1245,8 +1245,14 @@ func FindFilesMango(c echo.Context) error {
...
 
-	// drop the fields, they can cause issues if not properly manipulated
-	delete(findRequest, "fields")
+	if reqFields, ok := findRequest["fields"].([]interface{}); ok {
+		// Those fields are necessary for the JSON-API response
+		fields := []string{"_id", "_rev", "type", "class"}
+		for _, v := range reqFields {
+			fields = append(fields, v.(string))
+		}
+		findRequest["fields"] = fields
```
nono, line 1254: CouchDB is OK if we have twice the same field?

paultranvan: Yes, I tested, it's ok. But I could add a unit test for this

model/vfs/metadata.go
```diff
@@ -91,11 +91,12 @@ func NewMetaExtractor(doc *FileDoc) *MetaExtractor {
 	var e MetaExtractor
 	switch doc.Mime {
 	case "image/jpeg":
-		e = NewExifExtractor(doc.CreatedAt, true)
+		fmt.Printf("doc : %+v\n", doc)
```
nono, line 94: debug spotted

paultranvan: Hence the draft PR ;)

nono:
For the `Clone` methods, I also think we should copy the `UpdatedAt` and `CreatedAt`:

```go
updatedAt := *d.UpdatedAt
cloned.UpdatedAt = &updatedAt
```

model/vfs/file.go
```diff
@@ -70,10 +70,20 @@ func (f *FileDoc) DocType() string { return consts.Files }
...
+	if f.Tags != nil {
+		cloned.Tags = make([]string, len(f.Tags))
+		copy(cloned.Tags, f.Tags)
+	}
+	createdAt := *f.CreatedAt
+	cloned.CreatedAt = &createdAt
+	updatedAt := *f.UpdatedAt
+	cloned.UpdatedAt = &updatedAt
+	executable := *f.Executable
+	cloned.Executable = &executable
```
nono, line 86: You also need to protect against nil for createdAt, updatedAt, and executable

model/vfs/file.go
```diff
@@ -24,21 +24,21 @@ type FileDoc struct {
...
 	ByteSize   int64    `json:"size,string"` // Serialized in JSON as a string, because JS has some issues with big numbers
-	MD5Sum     []byte   `json:"md5sum"`
-	Mime       string   `json:"mime"`
-	Class      string   `json:"class"`
-	Executable bool     `json:"executable"`
+	MD5Sum     []byte   `json:"md5sum,omitempty"`
+	Mime       string   `json:"mime,omitempty"`
+	Class      string   `json:"class,omitempty"`
+	Executable *bool    `json:"executable,omitempty"`
 	Trashed    bool     `json:"trashed"`
```
nono, line 40: Why `Trashed` was kept as is (no `omitempty`)?

paultranvan: I feel like `trashed` is more or less always relevant to be returned (at least as relevant as type, class, etc). Also, it would imply to turn `trashed` as a pointer, which has many more impact in the code.

paultranvan:
https://github.com/cozy/cozy-stack/pull/3152 is simpler and has less regresssion-potential

PR #3150, opened 2021-09-22
feat: Add ReadOnly property to organization.GetCollections()'s return

Ldoppea:
Bitwarden's protocol return the ReadOnly property only when in a
FullSync context

They don't need it on `GetCollections` because it is used only by
Export methods

As we use this method for realtime sync, we need to have the ReadOnly
property

____

This is WIP because I wan't to discuss about those edits first.

Those changes imply that `CollectionResponse` and `CollectionDetailsResponse` now have the same content. So maybe there is no need to keep `CollectionDetailsResponse` ? (although it fit the class name used on Bitwarden's protocol)

Also this may impact the Export method. I think it ignores the ReadOnly property but I should test it to confirm that

Finally I did not edit tests nor run them. Current tests seems to cover only a local organization not a shared one. Should we edit tests to cover the ReadOnly field?


nono:
> So maybe there is no need to keep CollectionDetailsResponse?

Yes, we can remove `CollectionDetailsResponse` in that case. And, please add a comment on `CollectionResponse` to explain why we deviate from the bitwarden protocol.

> Current tests seems to cover only a local organization not a shared one.

Yes, the unit tests don't cover the shared organizations. We only have an integration test here:

https://github.com/cozy/cozy-stack/blob/master/tests/integration/tests/bitwarden_cli.rb

> Should we edit tests to cover the ReadOnly field?

Yes, please.

PR #3152, opened 2021-09-22
Allow to filter fields for mango queries on files

nono:
I was not very pleased by #3149, as it makes changes on a lot of files, and looks like it can easily introduce a regression. Here, I'm making another proposition, with changes more focused in `web/files`.

### DIFF
--- a/web/files/paginated.go
+++ b/web/files/paginated.go
@@ -3,6 +3,7 @@ package files
 // Links is used to generate a JSON-API link for the directory (part of
 import (
 	"encoding/json"
+	"time"
 
 	"github.com/cozy/cozy-stack/model/instance"
 	"github.com/cozy/cozy-stack/model/note"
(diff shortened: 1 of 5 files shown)


web/files/paginated.go
```diff
@@ -363,3 +364,71 @@ func (f *file) IncludePath(fp vfs.FilePather) {
...
+	*vfs.FileDoc
+	file *file
+	// We may want to hide some fields from the JSON response if the fields has
+	// not been requested to CouchDB, as they are blank
+	CreatedAt  *time.Time `json:"created_at,omitempty"`
+	UpdatedAt  *time.Time `json:"updated_at,omitempty"`
+	Executable *bool      `json:"executable,omitempty"`
+	// Hide the internal_vfs_id and referenced_by
+	InternalID   *interface{} `json:"internal_vfs_id,omitempty"`
+	ReferencedBy *interface{} `json:"referenced_by,omitempty"`
```
paultranvan, line 405: Why do we need this for files and not directories?

paultranvan (approved):

web/files/paginated.go
nono: For `internal_vfs_id`, this attribute only exists for files. For `referenced_by`, good question, I don't know. It is how it works today, but, yeah, looks weird. Cf https://github.com/cozy/cozy-stack/blob/b027bf9b02fe23caf0fe93bfc8aba95ba3b8ea25/web/files/paginated.go#L283

PR #3173, opened 2021-10-04
Use 401 Unauthorized for bad bearer tokens

nono:
**This is breaking change!** We will now use 401 instead of 400 as the HTTP
code for the response when a bearer token in the request is invalid
(expired for example). See RFC 6750.

This is a draft pull request, not to be merged yet, as we want to give time to clients to test it and update their code if needed.

### DIFF
--- a/docs/important-changes.md
+++ b/docs/important-changes.md
@@ -4,15 +4,21 @@
 
 This section will list important changes to the stack or its usage, and migration procedures if any is needed.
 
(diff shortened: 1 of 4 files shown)


nono:
@Crash-- ping, what is the status for this PR?

nono:
Ping @Crash--

nono:
Ping @Crash--

Crash--:
I think we can merge this one now :trollface:

PR #3233, opened 2021-12-03
Add an option for not creating the administrative folder

nono:


cozy.example.yaml
```diff
@@ -360,6 +360,9 @@ contexts:
     # Tells if the photo folder should be created or not during the instance
     # creation
     init_photos_folder: true
+    # Tells if the administrative folder should be created or not during the
+    # instance creation
+    init_administrative_folder: true
```
sblaisot, line 365: I think it should be made clear here in comment that `true` is the default value.

sblaisot (approved):
Nice!

PR #3249, opened 2021-12-14
Fix opening shared office documents

nono:
When a sharing member has been invited with several email addresses, and
has accepted and revoked the sharing from one email addresse, and later
accepted the sharing from another email address, their instance will be
listed twice in the sharing (with the status revoked and ready). If this
member creates an office document, and someone else in the sharing tries
to open it, it may fail as the sharing status may have been seen as
revoked (it depends of the order of sharing members). We are now
ignoring the not-ready members to avoid this issue.

### DIFF
--- a/model/sharing/open_file.go
+++ b/model/sharing/open_file.go
@@ -303,6 +303,12 @@ func (o *FileOpener) PrepareRequestForSharedFile() (*PreparedRequest, error) {
 			if i == 0 {
 				continue // Skip the owner
 			}
+			// XXX Skip the not-ready members, as an instance can be listed
+			// several times in members;w ith different email addresses and
+			// status.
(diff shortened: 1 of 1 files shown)


model/sharing/open_file.go
```diff
@@ -303,6 +303,12 @@ func (o *FileOpener) PrepareRequestForSharedFile() (*PreparedRequest, error) {
 			if i == 0 {
 				continue // Skip the owner
 			}
+			// XXX Skip the not-ready members, as an instance can be listed
+			// several times in members;w ith different email addresses and
```
sblaisot, line 307:
```suggestion
			// several times in members; with different email addresses and
```

sblaisot (approved):

PR #3254, opened 2021-12-31
fix(notification): do not log error when account has no device with push notif

trollepierre:


model/notification/center/notification_center.go
```diff
@@ -217,6 +217,9 @@ func makePush(inst *instance.Instance, p *notification.Properties, n *notificati
 				if err == nil {
 					return nil
 				}
+				if err == "No device with push notification" {
```
trollepierre, line 220: maybe we should use a variable to contain the error below and this string?
maybe we should gather the 2 if ?

trollepierre:
not an error

PR #3276, opened 2022-01-24
Add a hack for chromium emulators

nono:


Crash--:
It should be fixed in the next Chromium version https://bugs.chromium.org/p/chromium/issues/detail?id=1289553

PR #3325, opened 2022-03-10
doc: Attempt to warn about the use of _all_docs

Crash--:
Let's make it explicit that _all_docs is not the way to 
go to fetch the full list of documents since it can be 
subject to performance issue on large dataset.

I know that some use cases are legitimate to make 
a request on _all_docs. With this edit, I try to warn 
that you can do that only if you know that the dataset 
is not pretty large.

### DIFF
--- a/docs/data-system.md
+++ b/docs/data-system.md
@@ -407,7 +407,66 @@ header to prevent conflict on deletion:
 
 -   If no id is provided in URL, an error 400 is returned
 
-## List all the documents
+## List all the documents (recommended & paginated way)
+
+We have added a non-standard `_normal_docs` endpoint. Since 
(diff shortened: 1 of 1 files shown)


docs/data-system.md
```diff
@@ -407,12 +407,29 @@ header to prevent conflict on deletion:
 
 -   If no id is provided in URL, an error 400 is returned
 
-## List all the documents
+## List all the documents (recommended & paginated way)
+
+We have added a non-standard `_normal_docs` endpoint. Since 
+`_all_docs` endpoint sends the design docs in the response. 
+Ant that :
```
nono, line 414: Ant?

nono (approved):

PR #3344, opened 2022-03-21
Add a route /public/prelogin

nono:


web/public/public.go
```diff
@@ -41,10 +42,30 @@ func Avatar(c echo.Context) error {
 	return echo.NewHTTPError(http.StatusNotFound, "Page not found")
 }
 
+// Prelogin returns information that could be useful to show a login page (like
+// in the flagship app).
+func Prelogin(c echo.Context) error {
+	inst := middlewares.GetInstance(c)
+	publicName, err := inst.PublicName()
+	if err != nil {
+		publicName = ""
```
Ldoppea, line 51: Is it possible to have no public name ? Should I set a default value in the login UI?

Ldoppea (approved):

web/public/public.go
nono: Yes, it's possible (even if it should be really rare in practice).

PR #3354, opened 2022-03-29
Fix webapp icons when using --appdir

nono:


trollepierre (approved):
validation test in success

PR #3367, opened 2022-04-11
Multi couchdbs

nono:


docs/config.md
```diff
@@ -95,6 +95,45 @@ Magick, konnectors and services for example). And they can take several GB for
...
+couchdb:
+  url: http://cluster1:5984/
+  clusters:
+    - url: http://cluster1:5984/
+      instance_creation: true
+    - url: http://cluster2:5984/
+      instance_creation: false
+```
+
+Then, we can restart the cozy-stack and every thing will work the same, except
```
sblaisot, line 123:
```suggestion
Then, we can restart the cozy-stack and everything will work the same, except
```

docs/config.md
```diff
@@ -95,6 +95,45 @@ Magick, konnectors and services for example). And they can take several GB for
...
+```yaml
+couchdb:
+  url: http://cluster1:5984/
+  clusters:
+    - url: http://cluster1:5984/
+      instance_creation: false
+    - url: http://cluster2:5984/
+      instance_creation: true
+```
+
```
sblaisot, line 136: I would explain here what happens if multiple clusters have `instance_creation` set to true. Is this round-robin? first always wins? and if round-robin, is it reset to first cluster at each stack restart?

sblaisot: It would also be nice to state here what would happen in case of cluster list change (changing url, or removing a cluster in the middle of the list).
I suspect couch cluster is stored as int representing list index so moving the list could result in unexpected behaviour like rendering instances as broken. Maybe it would be nice to add a cluster unique "name" or "slug" in the definition to identify the cluster?

model/instance/lifecycle/create.go
```diff
@@ -253,6 +262,24 @@ func CreateWithoutHooks(opts *Options) (*instance.Instance, error) {
...
 
+func ChooseCouchCluster(clusters []config.CouchDBCluster) (int, error) {
+	index := -1
+	var count uint32 = 0
+	for i, cluster := range clusters {
+		if !cluster.Creation {
+			continue
+		}
+		count++
+		if rand.Uint32() <= math.MaxUint32/count {
```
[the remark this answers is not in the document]
sblaisot: I don't get how below test could pass, There must be something I don't understand.

model/job/broker.go
```diff
@@ -76,6 +76,7 @@ type (
 	Job struct {
 		JobID       string      `json:"_id,omitempty"`
 		JobRev      string      `json:"_rev,omitempty"`
+		Cluster     int         `json:"cluster,omitempty"`
```
sblaisot, line 79: This should probably be names `CouchCluster` or `DBCluster` to avoid future collision.

model/job/scheduler.go
```diff
@@ -60,6 +60,7 @@ type (
 	TriggerInfos struct {
 		TID          string                 `json:"_id,omitempty"`
 		TRev         string                 `json:"_rev,omitempty"`
+		Cluster      int                    `json:"cluster,omitempty"`
```
sblaisot, line 63: Probably this should be named `CouchCluster` or `DBCluster` to avoid future collision.

sblaisot (commented):
I'm not confortable with the term `cluster`. I thing "database url" is a more suitable term, should it be a single couchDB server or a couchDB cluster. I think cozy-stack should be kept agnostic to database servers topology.

sblaisot:
Also I don't see:
- cmdline update to handle new field
- impact on `show db_prefix` cmdline
- corresponding admin endpoints

sblaisot:
It would also be nice to be able to force instance creation on a given cluster (cmdline + creation endpoint), even if that cluster is declared 'instance_creation: false` in configuration like we can do with `swift_layout`. This would allow to test creation on the new cluster before switching all creation to this cluster.

nono:
> I'm not confortable with the term cluster. I thing "database url" is a more suitable term, should it be a single couchDB server or a couchDB cluster. 

I'm not totally satisfied with "cluster", but I don't want to use "database URL", as it means something else in CouchDB lingo (`http://host:5984/database/`, not just `http://host:5984/`). And "shards" is also taken by CouchDB.

For the tooling, yes, we want to add it but in another PR. I would like to start testing this part as soon as possible.

model/instance/lifecycle/create.go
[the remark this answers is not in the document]
nono: > On the first round (first cluster defined with creation enabled), count is 1 so math.MaxUint32/count equals to math.MaxUint32 and the condition is always true

Correct

> resulting in always choosing the first cluster with creation enabled

No, the loop doesn't stop, and another cluster can be chosen at a later iteration:

1. On the first eligible cluster, the condition is always true and we take it
2. On the second eligible cluster, it is fifty-fifty: the chances are equal that the first cluster is kept or that the second cluster is preferred
3. On the third eligible cluster, the condition is true 1/3 of the time, and the 3 clusters have the same probability to have been selected after this iteration
4. etc.

docs/config.md
nono: > I would explain here what happens if multiple clusters have instance_creation set to true.

:+1: I have noted that in the commit message of https://github.com/cozy/cozy-stack/pull/3367/commits/542320097172c63b8d55245f7958e100ee026a4b but I forgot it in the doc.

> Maybe it would be nice to add a cluster unique "name" or "slug" in the definition to identify the cluster?

I have hesitated about that, but it makes harder the migration from a single cluster, so I've taken the path of using the list index.

PR #3378, opened 2022-04-21
Add encrypted file attribute

paultranvan:
This will be useful to identify client-side encrypted files


nono (commented):
I have a few questions:

- Is it possible to share these files between two Cozy instances? If yes, I think we need some changes to keep the `encrypted` flag:
    * https://github.com/cozy/cozy-stack/blob/master/model/sharing/indexer.go#L242-L256
    * https://github.com/cozy/cozy-stack/blob/master/model/sharing/files.go#L626-L635
    * https://github.com/cozy/cozy-stack/blob/master/model/sharing/files.go#L1237-L1252
- How these files will be managed when a user asks for an export for its Cozy data?
- When a user moves its Cozy from one address/provider to another, should these files be moved automatically? Or should we do like for the passwords (ie add explanations for the user to export them and later reimport them)?

paultranvan:
> Is it possible to share these files between two Cozy instances? 

For now, we do not want to allow the sharing of encrypted files, it will be disabled in the interface

> How these files will be managed when a user asks for an export for its Cozy data?

For now, I guess the files will remain encrypted

> When a user moves its Cozy from one address/provider to another, should these files be moved automatically? Or should we do like for the passwords (ie add explanations for the user to export them and later reimport them)?

Mhh, I don't think we thought about it yet. ideally it should be moved automatically, but I guess there are good reasons if it's not the case for passwords. I guess the manual actions for the user consist of export/import through Cozy Pass ?

nono:
> I guess the manual actions for the user consist of export/import through Cozy Pass ?

Yes, we have screens to explain how to export from one Cozy and reimport to the new Cozy via Cozy Pass.

nono (approved):

PR #3400, opened 2022-05-20
Add a new route for login from the flagship app

nono:


docs/auth.md
```diff
@@ -164,6 +164,90 @@ Set-Cookie: ...
...
+Content-Type: application/json
+```
+
+```json
+{
+  "two_factor_token": "123123123123"
+}
+```
+
+Then, the client can retry by sending the two-factor token and code:
```
Ldoppea, line 219: If the entered two-factor code is wrong here we get a 401 response (like when the password is not valid) and we should retry with the same token. This is different than for `/auth/session_code` which would return a new `two_factor_token` with a 403 response.

Maybe we should document that specificity?

web/auth/flagship.go
```diff
@@ -164,3 +172,87 @@ func confirmFlagship(c echo.Context) error {
...
+		})
+	}
+
+	if inst.HasAuthMode(instance.TwoFactorMail) {
+		if len(args.TwoFactorPasscode) == 0 || len(args.TwoFactorToken) == 0 {
+			twoFactorToken, err := lifecycle.SendTwoFactorPasscode(inst)
+			if err != nil {
+				return err
+			}
+			return c.JSON(http.StatusAccepted, echo.Map{
```
Ldoppea, line 210: Why do we return `http.StatusAccepted` instead of `http.StatusForbidden` in `CreateSessionCode`? Should we homogenise those scenario?

Ldoppea (approved):
The code seems to be ok. I still have some edge cases to test, but nominal scenario work well (login with/without 2FA, login with flagship certification OK/KO etc).

docs/auth.md
nono: I have fixed the code to return a 403 in that case.

web/auth/flagship.go
nono: Yep, it has been fixed to `http.StatusForbidden`.

PR #3415, opened 2022-06-03
Update the app when calling /apps/:slug

nono:
When someone is using only the flagship app, we still need a way to keep
the webapps up-to-date. When the flagship app opens a webapp, it makes 3
calls to the stack:

- GET /apps/:slug/open
- GET /apps/:slug
- GET /apps/:slug/download

It seems that updating the app (if a new version is available without
needing a consent from the user) on the second call, GET /apps/:slug,
is the best choice as this call in made in background.


Ldoppea (approved):
Maybe we should add this info in https://github.com/cozy/cozy-stack/blob/master/docs/apps.md#get-appsslug ?

PR #3422, opened 2022-06-08
Accept session_code and id_token for konnectors with OAuth

nono:


web/accounts/oauth.go
```diff
@@ -28,10 +34,6 @@ func (a *apiAccount) Links() *jsonapi.LinksList {
 }
 
 func start(c echo.Context) error {
-	if !middlewares.IsLoggedIn(c) {
```
doubleface, line 31: Will it still work in the browser without session code ?

nono: Yes, it will still work without a session code if the user is logged in

doubleface (approved):
Tested in local.

PR #3430, opened 2022-06-13
Forward warning for unoptimal index

nono:
With CouchDB 3+, when an index can be used for a _find request, but
CouchDB still need to look to many documents that will not appear in
the response, a warning is emitted. This warning is now forwarded by the
stack to the client. The warning for no index at all is still
transformed to an error by the stack.

### DIFF
--- a/docs/couchdb-quirks.md
+++ b/docs/couchdb-quirks.md
@@ -34,6 +34,24 @@ reasons. In general, you can follow these two rules of thumb:
    omit the `sort` operator on the query (except if you want the `descending`
    order).
 
+### Warnings for slow requests
+
+When requesting a mango index, CouchDB can use an index. But there are also
+cases where no index can be used, or where the index is not optimal. Let's
+see the different scenarios:
+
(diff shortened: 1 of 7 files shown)


Crash-- (approved):
Nice. 

Issue created on the CC side https://github.com/cozy/cozy-client/issues/1189

docs/couchdb-quirks.md
```diff
@@ -34,6 +34,24 @@ reasons. In general, you can follow these two rules of thumb:
    omit the `sort` operator on the query (except if you want the `descending`
    order).
 
+### Warnings for slow requests
+
+When requesting a mango index, CouchDB can use an index. But there are also
+cases where no index can be used, or where the index is not optimal. Let's
+see the different scenarios:
+
+- CouchDB does use an index, it will respond with a warning, and cozy-stack
```
paultranvan, line 43: I'm not sure I understand this: if CouchDB does use an index, why would we have a warning? Do you have an example of such situation?

nono: It was a typo

PR #3444, opened 2022-06-23
fix: Revert redirection to store on google danse error

doubleface:
We have the same access_denied message when user cancels BI webview and in some cases, there is no store application available.

Then, we go back to a redirection to home/harvest which will handle 
- BI webview cancel case
- Google Oauth danse cancel case

To se in another PR

This reverts commit b8a9f46cf1405712e2b9c0f39a56c523e4bc7ec6.

### DIFF
--- a/web/accounts/oauth.go
+++ b/web/accounts/oauth.go
@@ -121,13 +121,6 @@ func redirect(c echo.Context) error {
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


nono:
Has been fixed by https://github.com/cozy/cozy-stack/pull/3445

PR #3458, opened 2022-07-07
Fix bulk save events

paultranvan:
When docs were created in bulk, no event was fired, because we missed to
set the ID and/or the Type.
Therefore, the @event triggers were blind on such updates.


nono (approved):
LGTM

PR #3475, opened 2022-07-21
Add COZY_FIELDS to service worker

paultranvan:
Just like the konnectors, services now have the possiblity to retrieve
custom fields through the COZY_FIELDS env var.


sblaisot:
I think you missed this : https://github.com/cozy/cozy-stack/blob/master/scripts/konnector-nsjail-node16-run.sh#L135-L144

Variable should be added in all konnector launchers in the `scripts` directory (and applied on our servers too)

PR #3489, opened 2022-08-17
Avoid recreating account with same connection_id

nono:
On the OAuth flow for BI konnectors, the stack should avoid creating a
new account when there is already one account with the same
connection_id.

### DIFF
--- a/web/accounts/oauth.go
+++ b/web/accounts/oauth.go
@@ -165,12 +165,41 @@ func redirect(c echo.Context) error {
 		}
 	}
 
-	if err := couchdb.CreateDoc(i, acc); err != nil {
-		return err
+	if connID := c.QueryParam("connection_id"); connID != "" {
+		if existingAccount, err := findAccountWithSameConnectionID(i, connID); err == nil {
+			acc = existingAccount
+		}
+	}
+
+	if acc.ID() == "" {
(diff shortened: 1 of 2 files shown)


web/accounts/oauth.go
```diff
@@ -165,14 +165,43 @@ func redirect(c echo.Context) error {
...
 	}
 
 	c.Set("instance", i.WithContextualDomain(c.Request().Host))
 	return redirectToApp(c, i, acc, clientState, slug, "")
 }
 
+func findAccountWithSameConnectionID(inst *instance.Instance, connectionID string) (*account.Account, error) {
+	var accounts []*account.Account
+	req := &couchdb.AllDocsRequest{Limit: 1000}
+	err := couchdb.GetAllDocs(inst, consts.Accounts, req, &accounts)
```
Crash--, line 187: When I first look to implement this feature, I saw that our couchdb package was not supporting the $in operator (https://github.com/cozy/cozy-stack/blob/master/pkg/couchdb/mango/query.go) I know that this $in operator is pretty slow, but I never expected that `all_docs` then a mapping was quicker than a mango query with a $in.

nono: If we had millions+ of accounts, mango would be faster, but with a very small number of accounts, all_docs wins by a lot.

PR #3498, opened 2022-09-01
Remove the deprecated routes for PDF icons & previews

nono:
The routes have been removed a first time (#3497), but they were still
used, so we put them back. Let's remove them again!


nono:
@Crash-- ping, do you know what apps still use the old routes?

nono:
Ping @Crash--

PR #3506, opened 2022-09-08
docs: Try to make clearer the secret in remote doctype

Crash--:
Try to make clearer about the fact that `token` is stored in  the secret database, but need to use `secret_token`  for the remote doctype declaration.


nono (approved):

nono:
Ping @Crash--

nono:
Ping @Crash--

nono (approved):

PR #3518, opened 2022-09-21
Add client_os & last_refreshed_at to OAuth clients

nono:


web/auth/oauth.go
```diff
@@ -877,6 +884,11 @@ func accessToken(c echo.Context) error {
 				"error": "invalid refresh token",
 			})
 		}
+
+		// Update the last_refreshed_at field of the OAuth client
+		client.LastRefreshedAt = time.Now()
```
paultranvan, line 889: Could this be added when the token is first created as well? So we can always rely on this field even though it was not  refreshed yet

web/auth/oauth.go
```diff
@@ -352,6 +353,12 @@ func authorize(c echo.Context) error {
 		}
 	}
 
+	// Fill the client_os of the OAuth client
+	rawUserAgent := c.Request().UserAgent()
+	ua := user_agent.New(rawUserAgent)
+	params.client.ClientOS = ua.OS()
+	_ = couchdb.UpdateDoc(instance, params.client)
```
paultranvan, line 360: Out of curiosity, why the `_` ? Is it required to get the returned value?

web/auth/oauth.go
nono: We have `cozyMetadata.created_at`. It is not enough?

web/auth/oauth.go
nono: In go, nope, but our linter is not happy when a function returns an error and we don't use it. Assigning the error to `_` is a way to tell to the linter that it is OK.

web/auth/oauth.go
paultranvan: We can use this, but the client might not be aware that the `LastRefreshedAt` won't be there before a week, so I feel like it could prevent future mistakes

nono: OK, if you prefer, I add it.

paultranvan (approved):
Thanks!

PR #3519, opened 2022-09-22
Fix session timeout on OAuth for banks

nono:
When a short session is created on the OAuth flow for banks, it can be reused if the user makes a second flow. But, as it is a short session, it can easily timeout between the beginning and the end of the flow. This fix consists of creating a new short session in that case.


Crash-- (approved):
Thanks

PR #3520, opened 2022-09-23
feat: Add file copy route

taratatach:
We add a new `POST /files/:file-id/copy` to duplicate a file on the
VFS within the same directory and with a localized "(copy)" suffix
added to the original name.

e.g. duplicating `/my-dir/my file.pdf` will create
`/my-dir/my file (copy).pdf` on a Cozy with an `en` locale.

In case a document with the same name as the copy already exists,
we'll use the same mechanism as for handling sharing conflicts: a
"(<n>)" suffix will be added to the name where `<n>` is an incremental
number starting with `2`.
If enough conflicting documents exist, we end up returning a
`409 Conflict` response.

If there isn't enough available disk space on the Cozy to duplicate
the file, we'll return a `413 Payload Too Large` response as we would
if the file had been uploaded.

Finally, if the file to duplicate does not exist, we'll respond with
`404 Not Found`.


web/files/files.go
```diff
@@ -293,6 +293,75 @@ func UploadMetadataHandler(c echo.Context) error {
...
+	newdoc := olddoc.Clone().(*vfs.FileDoc)
+	newdoc.DocID = ""
+	newdoc.DocRev = ""
+	newdoc.DocName = olddoc.DocName + " (copy)"
+	newdoc.Metadata.RemoveCertifiedMetadata()
+	newdoc.CozyMetadata = nil
+	newdoc.ReferencedBy = nil
+	newdoc.InternalID = ""
+	newdoc.CreatedAt = time.Now()
+	newdoc.UpdatedAt = time.Now()
```
nono, line 325: [nitpick] I would prefer `newdoc.UpdateAt = newdoc.CreatedAt` to have exactly the same time for both

web/files/files.go
```diff
@@ -293,6 +293,75 @@ func UploadMetadataHandler(c echo.Context) error {
...
+	newdoc.RestorePath = ""
```
nono, line 326: I don't see what you want to do here. We have two cases:

- the file wasn't in the trash, and `RestorePath` is already blank
- the file was in the trash, and we should either block the action or keep the `RestorePath`

web/files/files.go
```diff
@@ -293,6 +293,75 @@ func UploadMetadataHandler(c echo.Context) error {
...
+		return WrapVfsError(err)
+	}
+	defer content.Close()
+
+	file, err := fs.CreateFile(newdoc, nil)
+	if err != nil {
+		return WrapVfsError(err)
+	}
+
+	// TODO: delegate content copy to VFS?
```
nono, line 346: Yes, please

web/files/files_test.go
```diff
@@ -764,6 +764,59 @@ func TestUploadWithSourceAccount(t *testing.T) {
...
+	// 4. fetch copy and check its content
+	res, resbody := download(t, "/files/download/"+copyID, "")
+	assert.Equal(t, 200, res.StatusCode)
+	assert.True(t, strings.HasPrefix(res.Header.Get("Content-Disposition"), "inline"))
+	println("Content-Disposition", res.Header.Get("Content-Disposition"))
+	assert.True(t, strings.Contains(res.Header.Get("Content-Disposition"), `filename="`+fileName+`(copy)"`))
+	assert.True(t, strings.HasPrefix(res.Header.Get("Content-Type"), "text/plain"))
+	assert.NotEmpty(t, res.Header.Get("Etag"))
+	assert.Equal(t, res.Header.Get("Etag")[:1], `"`)
+	assert.Equal(t, res.Header.Get("Content-Length"), fmt.Sprint(len(fileContent)))
```
nono, line 815: [nit] I would have used `strconv.Itoa` instead of `fmt.Sprint`

web/files/files_test.go
```diff
@@ -764,6 +764,59 @@ func TestUploadWithSourceAccount(t *testing.T) {
...
+	data := obj["data"].(map[string]interface{})
+	fileID = data["id"].(string)
+	fileAttributes := data["attributes"].(map[string]interface{})
+
+	var body map[string]interface{}
+
+	// 2. Send file copy request
+	res, _ = httpPost(ts.URL+"/files/"+fileID+"/copy", "")
+	assert.Equal(t, 201, res.StatusCode)
+	assert.NoError(t, json.NewDecoder(res.Body).Decode(&body))
```
nono, line 786: You can use `require.NoError` when you want the test to stop in case of error. It is not used often as `require` has been added several years after `assert` and I also discovered it long time after it was added. But it can help to avoid lots of error when something got wrong in the first few steps.

model/vfs/vfsswift/impl_v3.go
```diff
@@ -259,6 +249,65 @@ func (sfs *swiftVFSV3) CreateFile(newdoc, olddoc *vfs.FileDoc, opts ...vfs.Creat
...
+		newname = vfs.ConflictName(sfs, newdoc.DirID, newname, true)
+		exists, err = sfs.Indexer.DirChildExists(newdoc.DirID, newname)
+		if err != nil {
+			return err
+		}
+		if exists {
+			return os.ErrExist
+		}
+	}
+	newdoc.DocName = newname
```
nono, line 281: I'm not sure if it is needed, but I would have added:

```go
newdoc.ResetFullPath()
```

The `file` has a `fullpath` field, which is not saved in CouchDB but computed from the path of its parent, and kept as cache. When we change the name or the parent, we often reset the `fullpath` to avoid keeping a cached value with an incorrect path.

model/vfs/vfs.go
```diff
@@ -870,3 +879,56 @@ func OptionsAllowCreationInTrash(opts []CreateOptions) bool {
...
+	diskQuota := fs.DiskQuota()
+
+	maxsize = fs.MaxFileSize()
+	newsize = doc.ByteSize
+	if diskQuota > 0 {
+		diskUsage, err := fs.DiskUsage()
+		if err != nil {
+			return 0, 0, 0, err
+		}
+		maxsize = diskQuota - diskUsage
```
nono, line 893: I think we have lost a case in extracting `CheckAvailableDiskSpace`. If the instance has a 50GB disk quota, with 1GB taken, `maxsize` should be 5GB when using VFS Swift, not 49GB (`diskQuota - diskUsage`).

model/vfs/vfsafero/impl.go
```diff
@@ -255,6 +245,85 @@ func (afs *aferoVFS) CreateFile(newdoc, olddoc *vfs.FileDoc, opts ...vfs.CreateO
...
+
+	// XXX: we use the internal openFile method as we already have a VFS lock
+	content, err := afs.openFile(olddoc)
+	if err != nil {
+		return err
+	}
+	defer content.Close()
+
+	hash := md5.New()
+	extractor := vfs.NewMetaExtractor(newdoc)
```
nono, line 304: I would prefer to keep the metadata from `olddoc`, and not trying to extract them again.

model/vfs/vfsafero/impl.go
```diff
@@ -255,6 +245,85 @@ func (afs *aferoVFS) CreateFile(newdoc, olddoc *vfs.FileDoc, opts ...vfs.CreateO
...
+		hash:    hash,
+		meta:    extractor,
+	}
+
+	_, err = io.Copy(newfile, content)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
nono, line 324: The sequence:

```go
	if err != nil {
		return err
	}

	return nil
```

can be simplified to:

```go
	return err
```

web/files/files.go
taratatach: The idea was to regenerate the restore path based on the directory and new name. I must have forgotten about it. I'm gonna fix thsi.

web/files/files_test.go
taratatach: I've actually been using it a lot in the other tests. I'll update this one !

web/files/files.go
```diff
@@ -293,6 +293,47 @@ func UploadMetadataHandler(c echo.Context) error {
 	return jsonapi.Data(c, http.StatusCreated, &m, nil)
 }
 
+// FileCopyHandler handles POST requests on /files/:file-id/copy
+//
+// It is used to duplicate the given file and its metadata except for
+// relationships.
+func FileCopyHandler(c echo.Context) error {
+	instance := middlewares.GetInstance(c)
```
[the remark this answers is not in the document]
taratatach: I kinda copy/pasted it from another function in the file but I get your point. I'll use `inst` instead.

model/vfs/vfsswift/impl_v3.go
```diff
@@ -259,6 +249,65 @@ func (sfs *swiftVFSV3) CreateFile(newdoc, olddoc *vfs.FileDoc, opts ...vfs.Creat
...
+	}
+	newdoc.InternalID = NewInternalID()
+
+	// Copy the file
+	srcName := MakeObjectNameV3(olddoc.DocID, olddoc.InternalID)
+	dstName := MakeObjectNameV3(newdoc.DocID, newdoc.InternalID)
+	headers := swift.Metadata{
+		"creation-name":  olddoc.Name(),
+		"created-at":     olddoc.CreatedAt.Format(time.RFC3339),
+		"dissociated-of": olddoc.ID(),
```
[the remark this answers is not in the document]
taratatach: Ah sorry, it's a copy/paste error. I changed it into `copied-from`.

model/vfs/vfsswift/impl_v3.go
taratatach: It seems necessary indeed. I call the method after cloning the old document but I didn't think about calling it again here. I'll fix that.

model/vfs/vfsafero/impl.go
taratatach: Couldn't we have metadata that we don't want to copy though?

model/vfs/vfsafero/impl.go
taratatach: Totally. This part has changed though and the sequence is now relevant.

model/vfs/vfs.go
taratatach: I'll take a look at this.

model/vfs/vfsafero/impl.go
nono: Our PO says that we should keep them

web/files/files.go
taratatach: Ah I thought `RestorePath` contained the name of the document but it's only its old parent path. Since the parent of the copy should be the same as the original one we don't need to do anything here.

model/vfs/vfsswift/impl_v3.go
```diff
@@ -259,6 +249,65 @@ func (sfs *swiftVFSV3) CreateFile(newdoc, olddoc *vfs.FileDoc, opts ...vfs.Creat
...
+	if lockerr := sfs.mu.Lock(); lockerr != nil {
+		return nil, lockerr
+	}
+	defer sfs.mu.Unlock()
+
+	newsize, maxsize, capsize, err := vfs.CheckAvailableDiskSpace(sfs, olddoc)
+	if err != nil {
+		return nil, err
+	}
+	if newsize > maxsize {
```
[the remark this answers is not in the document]
taratatach: Right. It is some leftover of the "missing case" issue around `CheckAvailableDiskSpace`. That's fixed.

nono (approved):

PR #3525, opened 2022-10-03
Fix orphan triggers on app updates

nono:
When a webapp is updated, and its manifest contains services, it can happen that a new trigger is created but the old one is not removed. It was the case, for example, when the type of the trigger is changed (@every -> @monthly).


paultranvan (approved):
Nice test :+1:

PR #3529, opened 2022-10-12
Fix race condition on realtime

nono:
https://github.com/cozy/cozy-stack/pull/3492 has introduced a race condition on the list of channels of a topic. And the goroutine leak was still here (in a different way). I hope to do better this time.


pkg/realtime/mem_realtime.go
```diff
@@ -47,103 +56,120 @@ func (h *memHub) subscribe(sub *Subscriber, key string) {
...
 	}
 
-	delete(it.subs, &sub.Channel)
-	if len(it.subs) == 0 {
+	h.removeTopic(sub, key)
+
+	w := &toWatch{sub, ""}
+	select {
+	case it.unsubscribe <- w:
+		if running := <-it.running; !running {
```
taratatach, line 92: What happens when this condition is not met? The topic is never removed?

PR #3530, opened 2022-10-13
Improve remote doctypes

nono:


web/remote/remote.go
```diff
@@ -30,7 +33,8 @@ func remoteGet(c echo.Context) error {
 
 func remotePost(c echo.Context) error {
 	doctype := c.Param("doctype")
-	if err := middlewares.AllowWholeType(c, permission.POST, doctype); err != nil {
+	slug, err := allowWholeType(c, permission.POST, doctype)
```
taratatach, line 51: nit: I don't see how this could be done otherwise but it feels a bit weird to get the slug of the requesting app from the permission rather than the context itsellf

docs/remote.md
```diff
@@ -151,6 +151,26 @@ JSON or XML are accepted. Other content-types are blocked the time to evaluate
 if they are useful and their security implication (javascript is probably not
 something we want to allow).
 
+### GET `/remote/_all_doctypes`
+
+This endpoint list all the known remote doctypes. A permission on
```
taratatach, line 156: nit: lists (missing an s at the end)

docs/remote.md
```diff
@@ -151,6 +151,26 @@ JSON or XML are accepted. Other content-types are blocked the time to evaluate
 if they are useful and their security implication (javascript is probably not
 something we want to allow).
 
+### GET `/remote/_all_doctypes`
```
taratatach, line 154: Why the leading `_all_`? Does it carry special meaning?

model/remote/list.go
```diff
@@ -0,0 +1,72 @@
...
+		}
+		for _, entry := range entries {
+			if entry.Type != "dir" {
+				continue
+			}
+			if entry.Name == "docs" || strings.HasPrefix(entry.Name, ".") {
+				continue
+			}
+			doctypes = append(doctypes, entry.Name)
+		}
```
taratatach, line 54: It seems risky to me to rely on the fact that all non docs, non hidden files in this repository are remote doctype definitions.
It might not be easy to change that today but what do you think about adding an extra check for the presence of at least two `.` (dots) in the name for example?

model/remote/list.go
```diff
@@ -0,0 +1,72 @@
...
+			if !entry.IsDir() {
+				continue
+			}
+			name := entry.Name()
+			if name == "docs" || strings.HasPrefix(name, ".") {
+				continue
+			}
+			doctypes = append(doctypes, name)
+		}
+	}
```
paultranvan, line 70: We'll have to keep this in mind if we ever change the cozy-doctypes repo organization

paultranvan (approved):

model/remote/remote.go
```diff
@@ -97,6 +97,11 @@ type Request struct {
 	ContentType   string            `json:"content_type"`
 	Variables     map[string]string `json:"variables"`
 	CreatedAt     time.Time         `json:"created_at"`
+	CozyMetadata  CozyMetadata      `json:"cozyMetadata"`
+}
+
+type CozyMetadata struct {
+	CreatedByApp string `json:"createdByApp,omitempty"`
```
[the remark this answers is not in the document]
nono: I don't think so (but I could be wrong).

docs/remote.md
nono: The stack has a `/data/_all_doctypes` endpoint (https://docs.cozy.io/en/cozy-stack/data-system/#list-the-known-doctypes), and I copied the `_all_` prefix from there.

model/remote/list.go
nono: Yes, but it was already the case before this PR. If the organization of cozy-doctypes repo is changed, the stack won't find the remote doctypes definition.

model/remote/list.go
nono: We have a few remote doctypes with a name that contains only one dot (like `com.inokufu`).

PR #3546, opened 2022-10-25
feat: Allow passing file copy name and parent dir

taratatach:
We want to allow users to decide a file copy's name and the directory
in which it should be created.
This can be done by passing the `Name` and `dir_id` query-string
parameters in the request.

We now check for the permission to create the copy after building the
new CouchDB document as the permissions might be different in the
selected directory than the duplicated file's directory.

We still handle potential name conflicts.


web/files/files.go
```diff
@@ -307,13 +307,23 @@ func FileCopyHandler(c echo.Context) error {
 		return WrapVfsError(err)
 	}
 
+	copyName := c.QueryParam("Name")
+	if copyName == "" || filepath.Ext(copyName) != filepath.Ext(olddoc.DocName) {
```
nono, line 311: So, if a user wants to copy `FOO.JPG` to `bar.jpg`, it won't be accepted?

web/files/files.go
```diff
@@ -307,13 +307,23 @@ func FileCopyHandler(c echo.Context) error {
...
+	}
+
+	newdoc := vfs.CreateFileDocCopy(olddoc, copyName)
+
+	if newDirID := c.QueryParam("dir_id"); newDirID != "" {
+		newdoc.DirID = newDirID
+		newdoc.ResetFullpath()
+	}
+
 	err = checkPerm(c, permission.POST, nil, olddoc)
```
nono, line 322: We should check the permission on `newdoc`. If an app has a permission on a folder, it should not be able to create a copy of a file in this filter at the root of the Cozy.

web/files/files.go
```diff
@@ -307,13 +307,23 @@ func FileCopyHandler(c echo.Context) error {
 		return WrapVfsError(err)
 	}
 
+	copyName := c.QueryParam("Name")
```
nono, line 310: Documentation!

web/files/files.go
```diff
@@ -307,13 +307,23 @@ func FileCopyHandler(c echo.Context) error {
...
+	if copyName == "" || filepath.Ext(copyName) != filepath.Ext(olddoc.DocName) {
+		copyName = fileCopyName(inst, olddoc.DocName)
+	}
+
+	newdoc := vfs.CreateFileDocCopy(olddoc, copyName)
+
+	if newDirID := c.QueryParam("dir_id"); newDirID != "" {
```
nono, line 317: s/dir_id/DirID/ ?

web/files/files.go
taratatach: Ah, I hadn't thought `filepath.Ext` could return different values for the same extension. Do you think this function is reliable enough to just strip the extension in the given name and use the original one?

web/files/files.go
taratatach: Ah good catch. That was the intention but I forgot to replace `olddoc` with `newdoc` after moving the check.

web/files/files.go
nono: Which route uses `dir_id` in the query-string?

taratatach: Not in the query-string but `PATCH /files/:file-id` accepts `dir_id` as an patch attribute. But if your intention is to have all query-string parameters be in PascalCase then it makes sense to use `DirId` instead (or `DirID` in this case).

nono (approved):

PR #3549, opened 2022-10-27
WIP: add route to send CCC logs

taratatach:


web/apps/apps.go
```diff
@@ -29,17 +30,59 @@ import (
...
+	"github.com/sirupsen/logrus"
 )
 
 // JSMimeType is the content-type for javascript
 const JSMimeType = "application/javascript"
 
 const typeTextEventStream = "text/event-stream"
 
+type AppLog struct {
+	Time  time.Time    "json:timestamp,string"
```
nono, line 47: It should be ``Time time.Time `json:"timestamp"`  ``

web/apps/apps.go
```diff
@@ -29,17 +30,59 @@ import (
...
+	Level logrus.Level "json:level,string"
```
nono, line 48: I would prefer to declare `Level` as a `string` to avoid the `UnmarshalJSON` method, and just doing the `ParseLevel` later.

web/apps/apps.go
```diff
@@ -214,6 +257,36 @@ func installHandler(installerType consts.AppType) echo.HandlerFunc {
...
 
+// logsHandler handles all POST /:slug/logs requests and forwards to sent log
+// lines to the server logger.
+func logsHandler(appType consts.AppType) echo.HandlerFunc {
+	return func(c echo.Context) error {
+		inst := middlewares.GetInstance(c)
+		slug := c.Param("slug")
+		if err := middlewares.AllowWholeType(c, permission.POST, consts.AppLogs); err != nil {
+			return err
+		}
```
nono, line 268: I would also check that the token comes from the flagship app (`middlewares.AllowMaximal`) or that the slug is the same as the one of the token app (`middlewares.AllowFromKonnector`). I don't want that konnector `foo` creates logs for konnector `bar` by mistake (or malicious behavior).

web/apps/apps.go
```diff
@@ -214,6 +257,36 @@ func installHandler(installerType consts.AppType) echo.HandlerFunc {
...
+
+		body, err := ioutil.ReadAll(c.Request().Body)
```
nono, line 270: You don't need to read the body before parsing it as JSON. You can just do:

```go
err := json.NewDecoder(c.Request().Body).Decode(&logs)
```

web/apps/apps.go
```diff
@@ -214,6 +257,36 @@ func installHandler(installerType consts.AppType) echo.HandlerFunc {
...
+		if err != nil {
+			return err
+		}
+
+		var logs []AppLog
+		if err := json.Unmarshal(body, &logs); err != nil {
+			return err
+		}
+
+		logger := logger.WithDomain(inst.Domain).WithNamespace(slug)
```
nono, line 280: The server-side konnectors have their logs with the slug in a `slug` field. We can use `konnectors` or `apps` for the namespace.

web/apps/apps.go
```diff
@@ -214,6 +223,46 @@ func installHandler(installerType consts.AppType) echo.HandlerFunc {
...
+		for _, log := range logs {
+			level, err := logrus.ParseLevel(log.Level)
+			if err != nil {
+				return echo.NewHTTPError(http.StatusBadRequest)
+			}
+			logger.WithTime(log.Time).Log(level, log.Msg)
+		}
+
+		return nil
```
nono, line 262: If you return nothing, the http code `204 No Content` is adapted:

```go
return c.NoContent(http.StatusNoContent)
```

web/apps/apps.go
```diff
@@ -214,6 +223,46 @@ func installHandler(installerType consts.AppType) echo.HandlerFunc {
...
+		if err := json.Unmarshal(body, &logs); err != nil {
+			fmt.Println("Bad json: ", err)
+			return echo.NewHTTPError(http.StatusBadRequest)
+		}
+
+		logger := logger.WithDomain(inst.Domain).WithNamespace(slug)
+		for _, log := range logs {
+			level, err := logrus.ParseLevel(log.Level)
+			if err != nil {
+				return echo.NewHTTPError(http.StatusBadRequest)
```
nono, line 257: I would prefer to have an explicit error which tells to the caller that it is the `level` that has this issue. You can use a JSON-API error for that:

```go
return jsonapi.InvalidAttribute("level", err)
```

web/apps/apps.go
taratatach: I saw other functions of that module returning `nil` so I thought that was appropriate.

nono: Which functions?

nono: This package is old, maybe we can improve some functions

taratatach: The `updateHandler` does that.

nono: Well, the `updateHandler` is a particular case: it is doing EventStream (aka Server-Side Events)

web/apps/apps.go
```diff
@@ -214,6 +222,69 @@ func installHandler(installerType consts.AppType) echo.HandlerFunc {
...
+			if err != nil {
+				return wrapAppsError(err)
+			}
+		} else {
+			// If logs are not sent by the flagship app, check that it's sent by
+			// a konnector or an app with the logs permission and get its slug
+			// from the permission.
+			pdoc, err := middlewares.GetPermission(c)
+			if err != nil {
+				fmt.Println("No permissions for ", appType, " ", slug)
```
nono, line 247: You can remove this debug line

nono (commented):
Please add documentation for this new route

PR #3562, opened 2022-11-07
feat: Add client-side konnector tokens build route

taratatach:
The flagship app should be able to retrieve tokens for client-side
konnectors so it can run them with their own token instead of the
flagship token.

Using specific konnector tokens means separate permission sets and an
accurate client name for konnector requests to cozy-stack.

### DIFF
--- a/docs/auth.md
+++ b/docs/auth.md
@@ -1036,6 +1036,50 @@ authentication cookie.
 The passcode can be sent to the instance's owner via email — more transport
 shall be added later.
 
+### POST /auth/tokens/konnectors/:slug
+
+This endpoint can be used by the flagship application in order to create a
+token for the konnector with the given slug. This token can then be used by the
+client-side konnector to make requests to cozy-stack.
+The flagship app will need to use its own access token to request the konnector 
+token.
+
+#### Request
+
+```http
+POST /auth/tokens/konnectors/impots HTTP/1.1
+Host: cozy.example.org
+Accept: application/json
+Authorization: Bearer eyJpc3Mi...
+```
+
+#### Response
+
+```http
+HTTP/1.1 201 Created
+Content-Type: application/json
+```
+
+```json
+"OWY0MjNjMGEtOTNmNi0xMWVjLWIyZGItN2I5YjgwNmRjYzBiCg"
+```
+
+### FAQ
+
+> What format is used for tokens?
+
+The access tokens are formatted as [JSON Web Tokens (JWT)](https://jwt.io/),
+like this:
+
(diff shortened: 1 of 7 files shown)


web/auth/auth.go
```diff
@@ -614,6 +614,7 @@ func Routes(router *echo.Group) {
 
 	// Flagship app
 	router.POST("/session_code", CreateSessionCode)
+	router.POST("/:slug/access_token", impersonatedAccessToken)
```
nono, line 617: I would prefer to avoid the `access_token` naming. `access_token` is generally used for OAuth clients, and a konnector doesn't have an OAuth client. And I would like to have `konnectors` somewhere in the route. Maybe `POST /auth/tokens/konnectors/:slug`.

web/auth/oauth.go
```diff
@@ -914,6 +914,25 @@ func accessToken(c echo.Context) error {
 	return c.JSON(http.StatusOK, out)
 }
 
+func impersonatedAccessToken(c echo.Context) error {
```
nono, line 917: I don't like the `impersonate`. It implies that the token will be used by something else that is not the konnector. But, here, the goal is to provide the token to the flagship app, so that the client-side konnector can use it.

web/auth/oauth.go
```diff
@@ -914,6 +914,25 @@ func accessToken(c echo.Context) error {
...
+		return c.JSON(http.StatusForbidden, err)
+	}
+
+	_, err := app.GetBySlug(inst, slug, consts.KonnectorType)
+	if err != nil {
+		return c.JSON(http.StatusNotFound, err)
+	}
+
+	token := inst.BuildKonnectorToken(slug)
+	fmt.Println("Token: ", token)
```
nono, line 931: Don't forget to remove the debug println

nono (commented):
We need documentation:

- the new route should be documented in `docs/auth.md`
- add a line or two in `docs/flagship.md`

docs/auth.md
```diff
@@ -1036,6 +1036,50 @@ authentication cookie.
...
+
+#### Response
+
+```http
+HTTP/1.1 201 Created
+Content-Type: application/json
+```
+
+```json
+OWY0MjNjMGEtOTNmNi0xMWVjLWIyZGItN2I5YjgwNmRjYzBiCg
```
nono, line 1064: If it is JSON, we should have quotes around the value (`"OW...Cg"`)

taratatach: OK. I wasn't sure there would be quotes around a single string.

nono (approved):

PR #3564, opened 2022-11-09
fix: Save konnector folder path into account

taratatach:
When running a konnector for the first time, we compute the path of
the folder that will be used to save the files fetched by the
konnector if necessary and create it if it's missing.

We should also save that path as the associated Account's default folder
path so it's reused later.

### DIFF
--- a/model/account/accounts.go
+++ b/model/account/accounts.go
@@ -18,20 +18,26 @@ import (
 
 // Account holds configuration information for an account
 type Account struct {
-	DocID             string                 `json:"_id,omitempty"`
-	DocRev            string                 `json:"_rev,omitempty"`
-	Name              string                 `json:"name"`
-	AccountType       string                 `json:"account_type"`
-	DefaultFolderPath string                 `json:"defaultFolderPath,omitempty"`
-	FolderPath        string                 `json:"folderPath,omitempty"` // Legacy
-	Token             string                 `json:"token,omitempty"`      // Used by bi-aggregator
-	UserID            string                 `json:"user_id,omitempty"`    // Used by bi-aggregator-user
-	Basic             *BasicInfo             `json:"auth,omitempty"`
-	Oauth             *OauthInfo             `json:"oauth,omitempty"`
-	Extras            map[string]interface{} `json:"oauth_callback_results,omitempty"`
-	Relationships     map[string]interface{} `json:"relationships,omitempty"`
-	Data              map[string]interface{} `json:"data,omitempty"`
-	Metadata          *metadata.CozyMetadata `json:"cozyMetadata,omitempty"`
+	DocID         string                 `json:"_id,omitempty"`
+	DocRev        string                 `json:"_rev,omitempty"`
+	Relationships map[string]interface{} `json:"relationships,omitempty"`
+	Metadata      *metadata.CozyMetadata `json:"cozyMetadata,omitempty"`
+
+	AccountType       string                   `json:"account_type"`
+	Name              string                   `json:"name"`                        // Filled during creation request
+	FolderPath        string                   `json:"folderPath,omitempty"`        // Legacy. Replaced by DefaultFolderPath
+	DefaultFolderPath string                   `json:"defaultFolderPath,omitempty"` // Computed from other attributes if not provided
+	Identifier        string                   `json:"identifier,omitempty"`        // Name of the Basic attribute used as identifier
+	Basic             *BasicInfo               `json:"auth,omitempty"`
+	Oauth             *OauthInfo               `json:"oauth,omitempty"`
+	Extras            map[string]interface{}   `json:"oauth_callback_results,omitempty"`
+	Data              map[string]interface{}   `json:"data,omitempty"`
+	State             string                   `json:"state,omitempty"`
+	TwoFACode         string                   `json:"twoFACode,omitempty"`
+	MutedErrors       []map[string]interface{} `json:"mutedErrors,omitempty"`
+	Token             string                   `json:"token,omitempty"`   // Used by bi-aggregator
+	UserID            string                   `json:"user_id,omitempty"` // Used by bi-aggregator-user
+
 	// When an account is deleted, the stack cleans the triggers and calls its
 	// konnector to clean the account remotely (when available). It is done via
 	// a hook on deletion, but when the konnector is removed, this cleaning is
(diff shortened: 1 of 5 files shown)


worker/exec/konnector.go
```diff
@@ -375,6 +375,14 @@ func (w *konnectorWorker) ensureFolderToSave(ctx *job.WorkerContext, inst *insta
 		dir.CozyMetadata.SourceAccount = acc.ID()
 		_ = couchdb.UpdateDoc(inst, dir)
 	}
+
+	// 6. Ensure that the account knows the folder path
+	if acc.FolderPath != folderPath {
+		acc.FolderPath = folderPath
+		_ = couchdb.UpdateDoc(inst, acc)
+		fmt.Println("Saved folder path into account")
```
nono, line 383: Can you remove the `Println` please?

nono (approved):
LGTM

model/account/accounts.go
```diff
@@ -18,20 +18,25 @@ import (
...
-	Relationships     map[string]interface{} `json:"relationships,omitempty"`
-	Data              map[string]interface{} `json:"data,omitempty"`
-	Metadata          *metadata.CozyMetadata `json:"cozyMetadata,omitempty"`
+	DocID         string                 `json:"_id,omitempty"`
+	DocRev        string                 `json:"_rev,omitempty"`
+	Relationships map[string]interface{} `json:"relationships,omitempty"`
+	Metadata      *metadata.CozyMetadata `json:"cozyMetadata,omitempty"`
+
+	AccountType       string                   `json:"account_type"`
+	Name              string                   `json:"name"`                        // Filled with the value of
```
nono, line 27: I don't understand the comment `Filled with the value of`. Is it truncated?

model/account/accounts.go
```diff
@@ -239,6 +248,33 @@ func PushAccountDeletedJob(jobsSystem job.JobSystem, db prefixer.Prefixer, accou
...
+			doc.M["name"] = login
+		} else {
+			doc.M["name"] = ""
+		}
+		return
+	}
+
+	if name, ok := auth[identifier].(string); ok {
+		doc.M["name"] = name
+	}
```
nono, line 275: It is a bit weird: you are forcing `doc.M["name"]` to a blank string when there are no auth or no identifier, but not here when the field pointed by the identifier is missing. Is it voluntary?

web/data/accounts.go
```diff
@@ -153,6 +153,7 @@ func createAccount(c echo.Context) error {
 	}
 
 	account.Encrypt(doc)
+	account.ComputeName(doc)
```
nono, line 156: Should we also recompute the name when the account is updated?

worker/exec/konnector_test.go
```diff
@@ -330,16 +330,21 @@ echo "{\"type\": \"toto\", \"message\": \"COZY_URL=${COZY_URL}\"}"
...
-		acc := &account.Account{DefaultFolderPath: "/Administrative/toto"}
-		assert.NoError(t, couchdb.CreateDoc(inst, acc))
+
+		acc := &account.Account{}
 		defer func() { _ = couchdb.DeleteDoc(inst, acc) }()
+
+		// Folder is created from DefaultFolderPath
+		wg.Add(1)
+		acc.DefaultFolderPath = "/Administrative/toto"
+		require.NoError(t, couchdb.CreateDoc(inst, acc))
```
nono, line 340: I don't understand why the defer for deleting the doc is before the creation of it.

worker/exec/konnector_test.go
```diff
@@ -361,5 +367,45 @@ echo "{\"type\": \"toto\", \"message\": \"COZY_URL=${COZY_URL}\"}"
 		assert.Contains(t, dir.CozyMetadata.CreatedOn, inst.Domain)
 		assert.Len(t, dir.CozyMetadata.UpdatedByApps, 1)
 		assert.Equal(t, dir.CozyMetadata.SourceAccount, acc.ID())
+
+		// Folder is created from Konnector name and Account name
+		wg.Add(1)
+		acc.DefaultFolderPath = ""
+		acc.Name = "account-1"
+		require.NoError(t, couchdb.GetDoc(inst, consts.Accounts, acc.ID(), acc))
```
nono, line 375: Why do you set some fields and then load the document from CouchDB? I would have expected to first load the doc from CouchDB and then set some fields, or set some fields and then save the doc.

worker/exec/konnector_test.go
```diff
@@ -361,5 +367,45 @@ echo "{\"type\": \"toto\", \"message\": \"COZY_URL=${COZY_URL}\"}"
...
+			"account":        acc.ID(),
+		})
+		require.NoError(t, err)
+
+		j = job.NewJob(inst, &job.JobRequest{
+			Message:    msg,
+			WorkerType: "konnector",
+		})
+
+		config.GetConfig().Konnectors.Cmd = tmpScript
```
nono, line 389: Maybe we could add a defer to reset this configuration value to its old value at the end of this test?

model/account/accounts.go
taratatach: I assumed the result of the type assertion would be an empty string in this case. But I might even not need to force this value at all?!
`""` is the default value of a string field, right?

web/data/accounts.go
taratatach: My idea was not to update the account name since the folder path depends on it when a recreation is needed. I'm not sure I fully understand all the different situations we have to handle though.

worker/exec/konnector_test.go
taratatach: Since I'm using the same `acc` variable I decided to call the add the `defer` and then manipulate the account. Is that a bad habit?

worker/exec/konnector_test.go
taratatach: Hum that's a mistake. It should be a call to `UpdateDoc` here.

worker/exec/konnector_test.go
taratatach: Is that change really saved?

model/account/accounts.go
taratatach: That was somehow truncated, yes.

model/account/accounts.go
nono: Right, the default value is `""`. You only need to assign to a blank string if there was a previous value that you want to remove.

worker/exec/konnector_test.go
nono: It changes a global variable of the config package, it is not persisted on disk.

worker/exec/konnector_test.go
nono: Well, as we ignore the error on `DeleteDoc`, it works, but I don't think it is a good practice. I try to put the defer for cleaning just after the line that create the thing. Here, the deferred `DeleteDoc` would come just after `CreateDoc`.

nono (approved):

PR #3567, opened 2022-11-15
Fix a conflict case for io.cozy.sharings doc

nono:
When a sharing is revoked, it may happen that a conflict (409 from CouchDB) is received when updating the io.cozy.sharings document after deleting its triggers. After that, an error trigger_on_inactive_sharing is reported by cozy-stack check sharings.

And the fix is quite simple: when we receive a 409, we reload the document from CouchDB and try again.

### DIFF
--- a/model/sharing/sharing.go
+++ b/model/sharing/sharing.go
@@ -553,17 +553,31 @@ func (s *Sharing) RevokeByNotification(inst *instance.Instance) error {
 			return err
 		}
 	}
-	s.Credentials = nil
-	s.Active = false
 
-	for i, m := range s.Members {
-		if i > 0 && m.Instance != "" {
-			s.Members[i].Status = MemberStatusRevoked
+	var err error
+	for i := 0; i < 3; i++ {
+		s.Triggers = Triggers{}
+		s.Credentials = nil
+		s.Active = false
+
+		for i, m := range s.Members {
+			if i > 0 && m.Instance != "" {
+				s.Members[i].Status = MemberStatusRevoked
+				break
+			}
+		}
+
+		err := couchdb.UpdateDoc(inst, s)
+		if err == nil || !couchdb.IsConflictError(err) {
+			break
+		}
+
+		// In case of conflict (409 from CouchDB), reload the document and try again
+		if errb := couchdb.GetDoc(inst, consts.Sharings, s.ID(), s); errb != nil {
 			break
 		}
 	}
-
-	return couchdb.UpdateDoc(inst, s)
+	return err
 }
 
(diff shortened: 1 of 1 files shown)


model/sharing/sharing.go
```diff
@@ -553,17 +553,31 @@ func (s *Sharing) RevokeByNotification(inst *instance.Instance) error {
...
-	for i, m := range s.Members {
-		if i > 0 && m.Instance != "" {
-			s.Members[i].Status = MemberStatusRevoked
+	var err error
+	for i := 0; i < 3; i++ {
+		s.Triggers = Triggers{}
+		s.Credentials = nil
+		s.Active = false
+
+		for i, m := range s.Members {
```
taratatach, line 563: No issues with reusing the same variable `i` here?

model/sharing/sharing.go
```diff
@@ -553,17 +553,31 @@ func (s *Sharing) RevokeByNotification(inst *instance.Instance) error {
...
+			if i > 0 && m.Instance != "" {
```
taratatach, line 564: The idea is to not revoke the owner of the sharing?

model/sharing/sharing.go
```diff
@@ -553,17 +553,31 @@ func (s *Sharing) RevokeByNotification(inst *instance.Instance) error {
...
 	}
-	s.Credentials = nil
-	s.Active = false
 
-	for i, m := range s.Members {
-		if i > 0 && m.Instance != "" {
-			s.Members[i].Status = MemberStatusRevoked
+	var err error
+	for i := 0; i < 3; i++ {
+		s.Triggers = Triggers{}
```
taratatach, line 559: What's the idea here? Haven't we removed all the triggers already at that point?

taratatach:
Out of curiosity, why do we have a `trigger_on_inactive_sharing` error when we have a sharing with no triggers? I would have expected something like `sharing_without_triggers`.

nono:
We have the triggers, but also a reference to them in the `io.cozy.sharings` document. When the conflict blocks the `io.cozy.sharings` doc update, the reference is still here. And the stack just look at the reference for the `trigger_on_inactive_sharing` error (not that the triggers really exist or not).

model/sharing/sharing.go
nono: No, we are using `:=` which declares a new variable (and mask the previous one for this range).

model/sharing/sharing.go
nono: Yes, it's the idea.

taratatach (approved):

PR #3568, opened 2022-11-15
Do not retry upload on Internal Server Error

nono:
When uploading a file for a sharing, it may happen that the io.cozy.shared document is corrupted (bad chain of revisions), and the upload fails with an Internal Server Error. In that case, it is useless to retry, as it will fail again, but it is better to skip this file to allow uploading the next files sooner.


taratatach (approved):
Do we have some kind of alerting system to catch these cases?

nono:
> Do we have some kind of alerting system to catch these cases?

Nope, it's just me that looks manually to the logs.

PR #3574, opened 2022-11-23
Add connectivity check route

sblaisot:
This PR adds a public unauthenticated connectivity check route `/connection_check` that respond with HTTP 204 No Content for the flagship application to check for connectivity


nono:
Can you add a bit of documentation please?

nono (approved):

PR #3575, opened 2022-11-24
Add logs for client-side konnectors stats

nono:


web/jobs/jobs.go
```diff
@@ -620,16 +621,27 @@ func patchJob(c echo.Context) error {
...
 	}
+
+	log := inst.Logger().
+		WithField("job_id", j.ID()).
+		WithField("worker_id", "client").
+		WithNamespace("jobs")
+	msg := &exec.KonnectorMessage{}
+	if err := j.Message.Unmarshal(&msg); err == nil {
+		log = log.
+			WithField("slug", msg.Konnector).
```
sblaisot, line 632: Current konnector log also contain konnector version in `version` field.
but maybe it's harder to get it for client-side konnectors.

sblaisot (commented):
Konnector logs currently also include a `bi_webhook` boolean field that could probably be always set to `false` here

nono:
> Konnector logs currently also include a bi_webhook boolean field that could probably be always set to false here

Logs for server-side konnectors don't have a `bi_webhook` field most of the time. This field is only present when it has a value (ie true).

sblaisot (approved):

Crash-- (approved):

PR #3583, opened 2022-12-06
Use interact token for preview URL of sharing

nono:
When a user goes to the preview URL for a Cozy to Cozy sharing, and if their Cozy URL is known (ie they have accepted a sharing), then we should allow this user to interact with the sharing. For a note, it means that they will be allowed to edit it. For a folder, they can upload new files.

Note: we also need to update the webapps to use the interact token.

### DIFF
--- a/web/apps/serve.go
+++ b/web/apps/serve.go
@@ -18,6 +18,7 @@ import (
 	"github.com/cozy/cozy-stack/model/intent"
 	"github.com/cozy/cozy-stack/model/permission"
 	"github.com/cozy/cozy-stack/model/session"
+	"github.com/cozy/cozy-stack/model/sharing"
 	"github.com/cozy/cozy-stack/pkg/appfs"
 	"github.com/cozy/cozy-stack/pkg/assets"
 	"github.com/cozy/cozy-stack/pkg/config/config"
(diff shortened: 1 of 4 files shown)


web/apps/serve.go
```diff
@@ -342,6 +337,56 @@ func buildServeParams(
...
+	// sharings. When it is used for a Cozy to Cozy sharing, it can be for a
+	// preview token, and we need to upgrade it to an interact token if the
+	// member has a known Cozy URL. We do this upgrade when serving the preview
+	// page, not when sending the invitation link by mail, because we want the
+	// same link to work (and be upgraded) after the user has accepted the
+	// sharing.
+	token, pdoc, err := permission.GetTokenAndPermissionsFromShortcode(inst, sharecode)
+	if err != nil || pdoc.Type != permission.TypeSharePreview {
+		return sharecode
+	}
```
taratatach, line 365: Shouldn't we at least log in the error in this case? Otherwise we might miss something

web/apps/serve.go
```diff
@@ -342,6 +337,56 @@ func buildServeParams(
...
+	sharingID := strings.Split(pdoc.SourceID, "/")
```
taratatach, line 366: nit: Since the actual sharing ID is stored in the second part of the split, I think we should assign it directly to `sharingID`.

taratatach (approved):

web/apps/serve.go
nono: I think it would be too much noise, with revoked sharing and stuff like that.

PR #3590, opened 2022-12-14
feat: Allow hiding back button on passphrase reset

taratatach:
When the passphrase reset page is opened from a mobile native
application into the phone's browser, we don't want users to click on
the provided back button as it would not redirect them to the native
application but the login page inside the browser and this would be
confusing.

We add a `hideBackButton` query-string parameter that native
applications can set to `true` so we know not to display the back
button on the passphrase reset page.


nono (commented):
Please update the doc

nono (approved):

PR #3597, opened 2022-12-20
feat: Add sharing trees consistency check

taratatach:
Each check has its own function.

### DIFF
--- a/model/sharing/setup.go
+++ b/model/sharing/setup.go
@@ -221,7 +221,7 @@ func (s *Sharing) InitialCopy(inst *instance.Instance, rule Rule, r int) error {
 	}
 	defer mu.Unlock()
 
-	docs, err := findDocsToCopy(inst, rule)
+	docs, err := FindMatchingDocs(inst, rule)
 	if err != nil {
 		return err
 	}
@@ -237,8 +237,8 @@ func (s *Sharing) InitialCopy(inst *instance.Instance, rule Rule, r int) error {
 	return couchdb.BulkUpdateDocs(inst, consts.Shared, refs, olds)
 }
 
-// findDocsToCopy finds the documents that match the given rule
-func findDocsToCopy(inst *instance.Instance, rule Rule) ([]couchdb.JSONDoc, error) {
+// FindMatchingDocs finds the documents that match the given rule
+func FindMatchingDocs(inst *instance.Instance, rule Rule) ([]couchdb.JSONDoc, error) {
 	var docs []couchdb.JSONDoc
 	if rule.Selector == "" || rule.Selector == "id" {
 		if rule.DocType == consts.Files {
--- a/model/sharing/sharing.go
+++ b/model/sharing/sharing.go
@@ -17,6 +17,7 @@ import (
 	"github.com/cozy/cozy-stack/model/instance"
 	"github.com/cozy/cozy-stack/model/job"
 	"github.com/cozy/cozy-stack/model/permission"
+	"github.com/cozy/cozy-stack/model/vfs"
 	"github.com/cozy/cozy-stack/pkg/consts"
 	"github.com/cozy/cozy-stack/pkg/couchdb"
 	"github.com/cozy/cozy-stack/pkg/couchdb/mango"
@@ -1046,6 +1047,7 @@ func CheckSharings(inst *instance.Instance) ([]map[string]interface{}, error) {
 				"type":  "invalid_rules",
 				"error": err.Error(),
 			})
+			return nil
 		}
 
 		accepted := false
@@ -1055,9 +1057,61 @@ func CheckSharings(inst *instance.Instance) ([]map[string]interface{}, error) {
 			}
 		}
 
-		checks = append(checks, s.checkSharingTriggers(inst, accepted)...)
-		checks = append(checks, s.checkSharingMembers()...)
-		checks = append(checks, s.checkSharingCredentials()...)
+		membersChecks, validMembers := s.checkSharingMembers()
+		checks = append(checks, membersChecks...)
+
+		triggersChecks := s.checkSharingTriggers(inst, accepted)
+		checks = append(checks, triggersChecks...)
+
+		credentialsChecks := s.checkSharingCredentials()
+		checks = append(checks, credentialsChecks...)
+
+		if len(membersChecks) == 0 && len(triggersChecks) == 0 && len(credentialsChecks) == 0 {
+			if !s.Owner || !s.Active || s.Initial || s.ReadOnly() {
+				return nil
+			}
+
+			isSharingReady := false
+			for _, m := range s.Members {
+				if m.Status == MemberStatusReady {
+					isSharingReady = true
+					break
+				}
+			}
+			if !isSharingReady {
+				return nil
+			}
+
+			rule := s.FirstFilesRule()
+			if rule == nil {
+				return nil
+			}
+
+			ownerDocs, err := FindMatchingDocs(inst, *rule)
+			if err != nil {
+				checks = append(checks, map[string]interface{}{
+					"id":    s.SID,
+					"type":  "missing_matching_docs_for_owner",
+					"error": err.Error(),
+				})
+				return nil
+			}
+
+			for _, m := range validMembers {
+				ms, err := FindSharing(m, s.ID())
+				if err != nil {
+					checks = append(checks, map[string]interface{}{
+						"id":     s.SID,
+						"type":   "missing_sharing_for_member",
+						"member": m.Domain,
+						"error":  err.Error(),
+					})
+					continue
+				}
+
+				checks = append(checks, s.checkSharingTreesConsistency(inst, ownerDocs, m, ms)...)
+			}
+		}
 
 		return nil
 	})
@@ -1169,22 +1223,25 @@ func (s *Sharing) checkSharingTriggers(inst *instance.Instance, accepted bool) (
 	return checks
 }
 
-func (s *Sharing) checkSharingMembers() (checks []map[string]interface{}) {
+func (s *Sharing) checkSharingMembers() (checks []map[string]interface{}, validMembers []*instance.Instance) {
 	if len(s.Members) < 2 {
 		checks = append(checks, map[string]interface{}{
 			"id":         s.SID,
 			"type":       "not_enough_members",
 			"nb_members": len(s.Members),
 		})
-		return checks
+		return checks, nil
 	}
 
+	var ownerDomain string
 	for i, m := range s.Members {
 		if m.Status == MemberStatusRevoked && !s.Active {
 			continue
 		}
+
 		isFirst := i == 0
 		isOwner := m.Status == MemberStatusOwner
+
 		if isFirst != isOwner {
 			checks = append(checks, map[string]interface{}{
 				"id":     s.SID,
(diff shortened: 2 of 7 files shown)


model/sharing/setup.go
```diff
@@ -238,7 +238,7 @@ func (s *Sharing) InitialCopy(inst *instance.Instance, rule Rule, r int) error {
 }
 
 // findDocsToCopy finds the documents that match the given rule
```
nono, line 240: Don't forget to also update the comment

model/sharing/sharing.go
```diff
@@ -1056,206 +1056,315 @@ func CheckSharings(inst *instance.Instance) ([]map[string]interface{}, error) {
...
+	}
+
+	for _, d := range domains {
+		inst, err := instance.GetFromCouch(d)
+		if err != nil {
+			checks = append(checks, map[string]interface{}{
+				"id":     s.SID,
+				"type":   "missing_instance_for_member",
+				"member": d,
+			})
```
nono, line 1334: I think we should ignore the error if the Cozy are not in the same environment: `cozy.rocks` vs `mycozy.cloud`

model/sharing/sharing.go
```diff
@@ -1056,206 +1056,315 @@ func CheckSharings(inst *instance.Instance) ([]map[string]interface{}, error) {
...
+		}
+
+		instDocs, err := FindMatchingDocs(inst, *rule)
```
nono, line 1337: Does it work? I don't the transformation (XOR) on the id

taratatach: I missed that part. It should be good now.

model/sharing/sharing.go
taratatach: Is there a way to know if two Cozy are in the same environment? I can check if they have the same domain but that's not enough.

model/sharing/sharing.go
```diff
@@ -1056,206 +1056,337 @@ func CheckSharings(inst *instance.Instance) ([]map[string]interface{}, error) {
...
+			return checks
+		}
+
+		// Convert member doc ids to their owner's counterpart
+		memberKey := s.Credentials[i].XorKey
+		for _, doc := range memberDocs {
+			doc.SetID(XorID(doc.ID(), memberKey))
+		}
+
+		checkedMemberDocs := make([]bool, len(memberDocs))
```
nono, line 1342: I would have made a `ownerDocsByID` map. Then, I can iterate on the memberDocs to find if they have a match in this map:

- if not, I can add a check error
- if yes, I can check that the two documents have the same revision, name, parent id (+checksum for file) and remove the doc from the map

At the end, I can list the doc remaining in the map and add an error for them (as they are only on the owner's Cozy).

model/sharing/sharing.go
nono: I would use the same domain for checking that. In theory, the best way to find if two Cozy instances are in the same env is to look if they can be both found in CouchDB, but we need something more to detect errors here.

taratatach: OK. Let's keep it simple.

model/sharing/sharing.go
```diff
@@ -1056,206 +1083,423 @@ func CheckSharings(inst *instance.Instance) ([]map[string]interface{}, error) {
...
+		for _, ownerDoc := range ownerDocs {
+			if memberDoc, found := memberDocsByID[ownerDoc.ID()]; found {
+				if ownerDoc.Rev() != memberDoc.Rev() {
+					if ownerDoc.ShortRev() > memberDoc.ShortRev() && isMemberDiskFull {
+						checks = append(checks, map[string]interface{}{
+							"id":       s.SID,
+							"type":     "disk_quota_exceeded",
+							"instance": m.Instance,
+						})
+						return checks
```
nono, line 1412: When we have this error, I think we should continue to look at the other members to have a full picture.

A better way would be to have a fail fast flag, like it is implemented for `check fs`, but it is more work.

model/sharing/sharing.go
```diff
@@ -1056,206 +1083,423 @@ func CheckSharings(inst *instance.Instance) ([]map[string]interface{}, error) {
...
+							"id":        s.SID,
+							"type":      "invalid_doc_rev",
+							"member":    m.Instance,
+							"ownerDoc":  ownerDoc,
+							"memberRev": memberDoc.Rev(),
+						})
+					}
+				} else {
+					// It's unnecessary to run these checks if both docs don't
+					// have the same revision in the first place.
```
nono, line 1431: No, I think it would still be useful to check that (at least for `dir_id` and `name`) even if the revision is the same

model/sharing/sharing.go
```diff
@@ -1056,206 +1083,423 @@ func CheckSharings(inst *instance.Instance) ([]map[string]interface{}, error) {
...
+						checks = append(checks, map[string]interface{}{
+							"id":         s.SID,
+							"type":       "invalid_doc_name",
+							"member":     m.Instance,
+							"ownerDoc":   ownerDoc,
+							"memberName": memberDoc.M["name"],
+						})
+					}
+
+					if ownerDoc.M["dir_id"] != memberDoc.M["dir_id"] {
```
nono, line 1443: 2 cases to manage:

1. the id has been XORed, so we need to redo a XOR to compare them
2. I'm not sure, but I think it will give errors for files/folders that are directly inside the shared directory (not in a sub-directory)

pkg/couchdb/couchdb.go
```diff
@@ -92,6 +93,19 @@ func (j *JSONDoc) Rev() string {
 	return ""
 }
 
+// ShortRev returns the index of the document's revision
```
nono, line 96: We have a function `RevGeneration` that can be useful for that

model/sharing/sharing.go
```diff
@@ -1056,206 +1083,423 @@ func CheckSharings(inst *instance.Instance) ([]map[string]interface{}, error) {
...
-				}
-				if s.Credentials[i-1].Client == nil {
-					checks = append(checks, map[string]interface{}{
-						"id":     s.SID,
-						"type":   "missing_oauth_client",
-						"member": i,
-						"owner":  true,
-					})
+		if ms.ReadOnly() {
+			checks = append(checks, map[string]interface{}{
```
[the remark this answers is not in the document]
taratatach: Per the specs, we should signal when there's a difference between the trees and the member is read-only.
I'm using the checks to signal this as it seemed simpler than creating another kind of signal.

This should allow us to see what share of the errors come from a read-only member.

model/sharing/sharing.go
taratatach: Right. The idea was to fail fast for the member being checked. I went a bit too fast here.
I'll take a look at the fail fast flag implementation as well.

model/sharing/sharing.go
taratatach: Hum, I'm not sure I get what you mean here. When the revision is the same, we run the other checks. Did you mean we should look at `dir_id` and `name` even if the revisions are **not** the same?

model/sharing/sharing.go
taratatach: Ah right, I forgot ids were XORed. And you're probably right about 2.

pkg/couchdb/couchdb.go
taratatach: Ah thanks! I didn't find it.

model/sharing/sharing.go
```diff
@@ -1056,206 +1083,423 @@ func CheckSharings(inst *instance.Instance) ([]map[string]interface{}, error) {
...
+		isOwnerDiskFull = usage >= inst.VFS().DiskQuota()
+	}
+
+	for i, m := range s.Members {
+		member, err := instance.GetFromCouch(m.Instance)
+		if err != nil {
+			// If the member's instance cannot be found and doesn't share the
+			// owner's instance domain, they're probably on different
+			// environments so we simply skip this member.
+			if !strings.HasSuffix(m.Instance, inst.Domain) {
```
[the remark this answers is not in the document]
taratatach: I probably missed something. I'll take another look at the struct.

model/sharing/sharing.go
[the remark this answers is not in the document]
nono: But you always return a check for the member in read-only, even if there are no tree difference.

model/sharing/sharing.go
nono: My bad, forget that, I didn't read correctly.

model/sharing/sharing.go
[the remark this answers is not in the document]
taratatach: Right. That wasn't the goal in the first place. I'll fix it.

model/sharing/sharing.go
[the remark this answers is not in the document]
taratatach: That was a misinterpretation of the comment for the `Domain` attribute. I'm gonna use `SlugAndDomain()` instead.

model/sharing/sharing.go
```diff
@@ -1056,206 +1058,592 @@ func CheckSharings(inst *instance.Instance) ([]map[string]interface{}, error) {
...
+}
+
+// getParentFileSharing returns the first sharing found accepting the root of
+// the given sharing.
+//
+// Since have a sharing within another one will generate unexpected behavior,
+// the goal is to find these situations.
+func getParentFileSharing(inst *instance.Instance, sharing *Sharing) (*Sharing, error) {
+	if sharing.DocType() != consts.Files {
+		return nil, nil
```
nono, line 1158: I don't like functions that can return `nil, nil`. I prefer to add a third return parameter like `(*Sharing, bool, error)` and `return nil, false, nil` or `return something, true, nil`.

I know when reading code, I make easily the assumption that after a block like:

```go
foo, err := doSomething()
if err != nil {
    return err
}
```

that `foo` is not nil. And it was the cause of a few bugs that could have been avoided with the third return parameter trick.

model/sharing/sharing.go
```diff
@@ -1056,206 +1058,592 @@ func CheckSharings(inst *instance.Instance) ([]map[string]interface{}, error) {
...
+	ownerRule := s.FirstFilesRule()
+
+	memberRule := ms.FirstFilesRule()
+	if memberRule == nil {
+		checks = append(checks, map[string]interface{}{
+			"id":     s.SID,
+			"type":   "missing_files_rule_for_member",
+			"member": m.Domain,
+		})
+		return nil
```
nono, line 1441: It is weird to return nil here, just after assigning to `checks`

web/instances/checks.go
```diff
@@ -232,7 +232,9 @@ func checkSharings(c echo.Context) error {
 		return wrapError(err)
 	}
 
-	results, err := sharing.CheckSharings(i)
+	skipFSConsistency, _ := strconv.ParseBool(c.QueryParam("SkipFSConsistency"))
```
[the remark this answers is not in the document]
taratatach: Because we're looking specifically at the FS objects. I wanted to make a difference with the other checks. But if you can think of a better alternative, I'm open to suggestions :).

model/sharing/sharing.go
```diff
@@ -1056,206 +1058,592 @@ func CheckSharings(inst *instance.Instance) ([]map[string]interface{}, error) {
 			}
 		}
 
-		// Check triggers
-		if s.Active && accepted {
-			if s.Triggers.TrackID == "" && len(s.Triggers.TrackIDs) == 0 {
+		checks, validMembers := s.checkSharingMembers()
```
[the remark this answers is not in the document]
taratatach: Ah I missed that. I thought the linter would tell me about reassignments but since Go allows reassignment in narrower contexts, that makes sense.

model/sharing/sharing.go
taratatach: You're right. It's a mistake introduced when moving around some checks. This should return `checks`.

model/sharing/sharing.go
taratatach: Is it the same reason why the linter allows me to return only `nil` from `CheckSharings()` while it's expecting a tuple as output?

nono: No, the compiler doesn't allow that. There are no `return nil` for `CheckSharings`, only for the anonymous function passed as a parameter to `couchdb.ForEach` inside `CheckSharings`.

taratatach: Right. I read a bit fast.

taratatach: So you'd write something like this?

```go
parentSharing, ok, err := getParentFileSharing(m, ms)
if err != nil {
	return err
} else if ok {
	checks = append(checks, map[string]interface{}{
		"id":             ms.SID,
		"type":           "sharing_in_sharing",
		"parent_sharing": parentSharing.SID,
	})
	continue
}
```

nono: Yes

model/sharing/sharing.go
```diff
@@ -1055,206 +1057,635 @@ func CheckSharings(inst *instance.Instance) ([]map[string]interface{}, error) {
...
+// wasUpdatedRecently returns true if the given document's latest update, given
+// by its `cozyMetadata.updatedAt` attribute, happened less than 5 minutes ago.
+// If the attribute is missing or does not represent a valid date, we consider
+// the latest update happened before that.
+func wasUpdatedRecently(doc couchdb.JSONDoc) bool {
+	cozyMetadata, ok := doc.M["cozyMetadata"].(map[string]interface{})
+	if !ok || cozyMetadata == nil {
+		return false
+	}
+	if updatedAt, ok := cozyMetadata["updatedAt"].(time.Time); ok {
```
nono, line 1687: I'm a bit surprised to have a `time.Time` here. I would have expected it to be a string, as the JSON was unmarshaled to a `map[string]interface{}`, not to a `struct` that would say that its type is a `time.Time`.

nono (commented):
LGTM

model/sharing/sharing.go
taratatach: I was surprised as well and the first implementation was parsing a string to get a time value but I got type errors.
It makes sense when you follow the trail though:
1. https://github.com/cozy/cozy-stack/pull/3597/files#diff-d96f65ea3d1c149ba02f26b9d45c1a92b040deda89c41acb212a1947b776ac21R252-R256
2. https://github.com/cozy/cozy-stack/blob/master/model/sharing/files.go#L1272 & https://github.com/cozy/cozy-stack/blob/master/model/sharing/files.go#L1311
3. https://github.com/cozy/cozy-stack/blob/e15da402ff79315503d1f659a3dcb83c1a3f0f4f/model/vfs/cozy_metadata.go#L118

PR #3602, opened 2022-12-28
Add a command to clean sessions

nono:


web/instances/instances.go
```diff
@@ -360,6 +360,22 @@ func createSessionCode(c echo.Context) error {
...
 
+func cleanSessions(c echo.Context) error {
+	domain := c.Param("domain")
+	inst, err := lifecycle.GetInstance(domain)
+	if err != nil {
+		return err
+	}
+
+	if err := couchdb.DeleteDB(inst, consts.Sessions); err != nil {
+		return err
```
sblaisot, line 371: If there is no sessions database, then the sessionsLogin database will not be delete.
Maybe you can use a cumulative error handler and only return after atempting to delete both databases? or ignore 404 errors?

sblaisot (approved):

PR #3604, opened 2022-12-28
Increase the number of iterations for PBKDF2

nono:
The OWASP recommends at least 310,000 iterations for PBKDF2. Cf https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html

We also remove the exception for lower iterations when the instance was created from Edge, as this browser now supports PBKDF2 via the subtle crypto API (the polyfill is no longer needed in production).

### DIFF
--- a/docs/important-changes.md
+++ b/docs/important-changes.md
@@ -4,6 +4,10 @@
 
 This section will list important changes to the stack or its usage, and migration procedures if any is needed.
 
+## December 2023: Iterations for PBKDF2 increased
+
+We have increased the number of PBKDF2 iterations for new users to 650_000, and removed the exception for Edge as it now supports PBKDF2 via the subtle crypto API.
+
 ## December 2020: Jobs permissions
 
 We used to have a specific permission logic for jobs, in order to allow apps to direclty manage konnectors. More specifically, an app had the right to access a trigger state or remove it, if there was a doctype in common between the app permissions and the konnector manifest.
--- a/pkg/crypto/pbkdf2.go
+++ b/pkg/crypto/pbkdf2.go
@@ -9,22 +9,15 @@ import (
 
 // DefaultPBKDF2Iterations is the number of iterations used to hash the
 // passphrase on the client-side with the PBKDF2 algorithm.
-const DefaultPBKDF2Iterations = 100000
-
-// EdgePBKDF2Iterations is the number of iterations used to hash the passphrase
-// on the client-side with the PBKDF2 algorithm when the browser is Edge. As
-// Edge doesn't support PBKDF2 in its window.crypto, we have to rely on a
-// slower JS alternative, and to reduce the number of PBKDF2 iterations to
-// avoid freezing the browser.
-const EdgePBKDF2Iterations = 10000
+const DefaultPBKDF2Iterations = 650000
 
 // MinPBKDF2Iterations is the recommended minimum number of iterations for
 // hashing with PBKDF2.
-const MinPBKDF2Iterations = 5000
+const MinPBKDF2Iterations = 50000
 
 // MaxPBKDF2Iterations is the recommended maximal number of iterations for
 // hashing with PBKDF2.
-const MaxPBKDF2Iterations = 1000000
+const MaxPBKDF2Iterations = 5000000
 
 // hashedPassLen is the length of the hashed password (in bytes).
 const hashedPassLen = 32
--- a/tests/system/lib/bitwarden/organization.rb
+++ b/tests/system/lib/bitwarden/organization.rb
@@ -72,7 +72,7 @@ class Bitwarden
       master_key = PBKDF2.new do |p|
         p.password = inst.passphrase
         p.salt = "me@" + inst.domain.split(':').first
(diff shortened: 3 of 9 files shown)


taratatach (approved):

nono:
@Ldoppea can you check that we have no hard-coded 100 000 iterations in clients please?

Ldoppea:
I found 3 references to this hard-coded value:
- In [CozyPassMobile/SetPasswordPageViewModel](https://cs.github.com/cozy/cozy-pass-mobile/blob/3d388ab0b2519cde27f557a7dbcb7adcd4483a13/src/App/Pages/Accounts/SetPasswordPageViewModel.cs?q=100000#L161)
  - This page should not be used in Cozy context as we don't create any account nor set password from the App
- In [CozyPassMobile/RegisterPageViewModel](https://cs.github.com/cozy/cozy-pass-mobile/blob/3d388ab0b2519cde27f557a7dbcb7adcd4483a13/src/App/Pages/Accounts/RegisterPageViewModel.cs?q=100_000#L147)
  - Same, should not be used
- In [CozyPassWeb/browserStorage.service.ts](https://github.com/cozy/cozy-pass-web/blob/master/src/app/browser/browserStorage.service.ts#L74-L76)
  - ⚠️  This should be problematic, as we completely override the server's value.
  - There is a trello card linked in comment, this seems to be related to an old cozy-stack bug. I'll have to check if this is still the case or not

Crash--:
@Ldoppea we also use it here https://github.com/cozy/cozy-flagship-app/blob/918c583f2f48b4dd769c1d802f9fc7a2c2e29aac/src/libs/functions/passwordHelpers.js#L8 We dont pass the value here https://github.com/cozy/cozy-flagship-app/blob/918c583f2f48b4dd769c1d802f9fc7a2c2e29aac/src/screens/login/OnboardingScreen.js#L147 when calling https://github.com/cozy/cozy-flagship-app/blob/918c583f2f48b4dd769c1d802f9fc7a2c2e29aac/src/screens/login/components/OnboardingPasswordView.js#L129

Ldoppea:
I just made some tests on cozy-pass-web. The browserStorage object has the correct value, so I suppose the cozy-stack bug described in Trello has been fixed. Do you have more history on this @nono ?
If effectively fixed, I can remove the hack easily.
As of today the only impact is in when calling `refreshIdentityToken()`. It is also used in settings and lockScreen, but both pages have been disabled from cozy-pass-web

Ldoppea:
> @Ldoppea we also use it here https://github.com/cozy/cozy-flagship-app/blob/918c583f2f48b4dd769c1d802f9fc7a2c2e29aac/src/libs/functions/passwordHelpers.js#L8 We dont pass the value here https://github.com/cozy/cozy-flagship-app/blob/918c583f2f48b4dd769c1d802f9fc7a2c2e29aac/src/screens/login/OnboardingScreen.js#L147 when calling https://github.com/cozy/cozy-flagship-app/blob/918c583f2f48b4dd769c1d802f9fc7a2c2e29aac/src/screens/login/components/OnboardingPasswordView.js#L129

Good catch, I'll check if we can have the cozy-stack value in this scenario

nono:
> I just made some tests on cozy-pass-web. The browserStorage object has the correct value, so I suppose the cozy-stack bug described in Trello has been fixed. Do you have more history on this @nono ?

I don't remember, but the trello card is saying that it has been fixed. So, I think we should remove the hack.

Ldoppea:
I'm stuck on the Flagship app's onboarding scenario. After the instance creation I need the `DefaultPBKDF2Iterations` value in order to create the password.

@nono I expected to retrieve the PBKDF2 iterations number from `/public/prelogin` route (we already do that in the login page). But this route returns `KdfIterations:0` if the registration is not done yet.

Do you have an idea on how I can retrieve this value before the registration?

nono:
We don't have an endpoint that gives this number. And, I think that it is the client that should decide of the number of iterations on that scenario.

Ldoppea:
> We don't have an endpoint that gives this number. And, I think that it is the client that should decide of the number of iterations on that scenario.

So should I just change the default from 100000 to 650000 [here](https://github.com/cozy/cozy-flagship-app/blob/918c583f2f48b4dd769c1d802f9fc7a2c2e29aac/src/libs/functions/passwordHelpers.js#L8)?

nono:
Yes, if this constant is only used for onboarding (not login), it should be set to 650000.

Ldoppea:
PR on Flagship app: https://github.com/cozy/cozy-flagship-app/pull/606

Tests have been done on cozy-pass-web and cozy-pass-mobile and they correctly use the cozy-stack's value (650000)

Ldoppea:
Cozy-keys-browser also correctly works ✅

Ldoppea:
@nono the Flagship App's fix has been merged cozy/cozy-flagship-app#606

nono:
Thanks for the update, but for the moment, I don't have time for this subject.

taratatach (approved):

PR #3605, opened 2022-12-29
Allow getting a sharing with its share-interact token

nono:


zatteo (approved):
Tested successfully with https://github.com/cozy/cozy-notes/pull/358

PR #3606, opened 2022-12-31
Refactor the some tests and handle the `--short` flag

Peltoche:
This refactoring goals are:
- Taking into account the `--short` flag and skip the tests with dependencies
- Handling the errors with `require.NoError` in order to have more detailed errors messages
- Replacing the use of `MainTest` by the use of subtests. This task have several advantages: 
  - It allow to have different init in the same packages. 
  - It allow to skip several test at once if necessary. 
  - It allow the use of `t.Cleanup` which ensure the tests cleanup even in case of panic. 
  - It avoid the use of globals which can be source of dataraces. 
  - It make more explicit the depedances between the subtests inside the logs.

### DIFF
--- a/cmd/cmd_test.go
+++ b/cmd/cmd_test.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"bytes"
-	"fmt"
 	"net/http/httptest"
 	"net/url"
 	"os"
@@ -12,7 +11,6 @@ import (
 
 	"github.com/cozy/cozy-stack/client"
 	"github.com/cozy/cozy-stack/client/request"
-	"github.com/cozy/cozy-stack/model/instance"
 	"github.com/cozy/cozy-stack/model/instance/lifecycle"
 	"github.com/cozy/cozy-stack/model/stack"
 	"github.com/cozy/cozy-stack/pkg/config/config"
@@ -21,32 +19,24 @@ import (
 	"github.com/cozy/cozy-stack/web"
 	"github.com/labstack/echo/v4"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
-var (
-	testInstance *instance.Instance
-	testClient   *client.Client
-)
+func TestCmd(t *testing.T) {
+	if testing.Short() {
+		t.Skip("a redis is required for this test: test skipped due to the use of --short flag")
+	}
 
-func TestMain(m *testing.M) {
 	config.UseTestFile()
 
-	if _, err := couchdb.CheckStatus(); err != nil {
-		fmt.Println("This test need couchdb to run.")
-		os.Exit(1)
-	}
+	_, err := couchdb.CheckStatus()
+	require.NoError(t, err, "This test need couchdb to run.")
 
-	_, err := stack.Start()
-	if err != nil {
-		fmt.Println(err)
-		os.Exit(1)
-	}
+	_, err = stack.Start()
+	require.NoError(t, err)
 
 	tempdir, err := os.MkdirTemp("", "cozy-stack")
-	if err != nil {
-		fmt.Println("Could not create temporary directory.")
-		os.Exit(1)
-	}
+	require.NoError(t, err)
 
 	config.GetConfig().Fs.URL = &url.URL{
 		Scheme: "file",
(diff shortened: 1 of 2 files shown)


nono:
> Taking into account the --short flag and skip the tests with dependencies

Most of the tests of cozy-stack are using dependencies, so I don't think it makes sense to use the short flag for that.

Peltoche:
> Most of the tests of cozy-stack are using dependencies, so I don't think it makes sense to use the short flag for that.

Even with almost no tests handling the `--short` flags seems to be a good idea to handle. It allows the dev to have a really quick feedback loop by running a quick sanity check in a single command. This could be useful for example inside a pre-commit hook for example. 

> Most of the tests of cozy-stack are using dependencies

I have seen, you have a lot of integration tests. One day I would like to propose you some ideas that could help you write more unit tests. 

At Leboncoin we try to follow the "test pyramid", I don't know if you are familiar with. I not there is [a blog article from octo](https://blog.octo.com/la-pyramide-des-tests-par-la-pratique-1-5/) and [an article from our beloved martin fowler (in java)](https://martinfowler.com/articles/practical-test-pyramid.html). Any search with `unit vs integration vs e2e tests` or `test pyramid` would give you a plenty of article going in the same direction . 

This means that we have ~90% of our tests runs with the `--short` flag and we can run them all the time in a few seconds. Some people run them automatically on save. This make us really confident during the development. The last 10% are the equivalent of your tests were we are checking that everything go well together but it all the code is already tests by the unit test.

nono:
> At Leboncoin we try to follow the "test pyramid", I don't know if you are familiar with. I not there is [a blog article from octo](https://blog.octo.com/la-pyramide-des-tests-par-la-pratique-1-5/) and [an article from our beloved martin fowler (in java)](https://martinfowler.com/articles/practical-test-pyramid.html). Any search with unit vs integration vs e2e tests or test pyramid would give you a plenty of article going in the same direction .

I know test pyramid, but there more shapes, and it looks like test pyramid is less popular than diamond, honeycomb or trophies shapes. Cf https://martinfowler.com/articles/2021-test-shapes.html and https://www.crispy-engineering.com/p/why-test-diamond-model-makes-sense for example. But the real reason that we have so many integration tests is more basic: we have CouchDB and we want to test with it, as it is often a cause for bugs.

To have a fast feedback loop, I am launching only the tests for the package(s) I'm working on, and it often takes less than 5 seconds to run, and the linter which is quite fast with cache. Every thing else runs on the CI (GitHub Actions).

> Except for the --short flag, what do you think of this kind of refactoring?

As many things for the tests in this repository, I have mixed feelings. We could and should do better, but for existing tests, I appreciate that they are quite regular (they looks the same and use the same patterns), and I would take too much time to update them all. I prefer to use my time for other things, like new feature or tooling.

So, yeah, the refactored code is better than before if we just look at this package. But it does no longer look like the tests for the other packages. I'm not sure I see enough value to justify reviewing the PR for changing all the packages (and I'm sure I won't want to update the tests myself).

Peltoche:
> Cf https://martinfowler.com/articles/2021-test-shapes.html and https://www.crispy-engineering.com/p/why-test-diamond-model-makes-sense for example.

Thanks for the links !


> I'm not sure I see enough value to justify reviewing the PR for changing all the packages (and I'm sure I won't want to update the tests myself).

I'm ready to make the changes myself and I kind of used to make this kind of project at a way bigger scale. I estimate the work to ~4H (5mn by review X ~45 `TestMain`). I don't know how many time you have taken for the review but I guess it's way less as there is no logic change. After it's up to you to choose if your ready to invest this time or not. I can only give you my explanation on why I think it worth the case for several reason:
- It will allow me to enter into the code smoothly
- It will scope the data inside the their tests, making them easier to read
- It will allow to clean the code correctly by applying the defer and using `t.cleanup` (#3622)
- It will allow to use the logger from the tests and printing only the log in error
- It will allow you to use `t.Tempdir` which is automatically cleaned
- It will allow you to use the test name (`t.Name`) inside the host
- It will allow you to setup the `--short` flag
- It will allow you to remove a lot of global variables
- It could allow you to run the test in parallel soon :crossed_fingers:

nono:
OK, you feel really motivated. Let's do this.

Peltoche:
Great! :+1: 

I just need one last information: What are you preferred PR size for this kind of projects? 

I try to keep the PR as small as possible in order to keep the reviews short but if I spam you with reviews the context switch can cost you a lot of time and can be tedious and that's really not the goal. Maybe you prefere to make a big switch in order to keep your tests consistent?

So you prefer that I make a PR by file? I group some together? I put everything in a big PR?

nono:
Let's say something in between, like:

- 1 PR for tests in `worker/*`` packages
- 1 PR for tests in `web/*` packages
- 1 PR for tests in `model/*` packages
- 1 PR for the remaining (`pkg/*`, `cmd`, etc.) and probably also clean `testutils`

Is it OK for you?

Peltoche:
This review will be replaces by the script like #3637 .

PR #3607, opened 2022-12-31
Refactor the pkg/cache tests and handle the `--short` flag

Peltoche:
This refactoring goals are:

- Use the same test for all the implementations: this allow to ensure that all implementations have the same behavior.
- Taking into account the --short flag and skip the tests with dependencies
- Handling the errors with require.NoError in order to have more detailed errors messages
- Replacing the use of MainTest by the use of subtests. This task have several advantages: \
  - It allow to have different init in the same packages. 
  - It allow to skip several test at once if necessary. 
  - It allow the use of t.Cleanup which ensure the tests cleanup even in case of panic. 
  - It avoid the use of globals which can be source of dataraces. 
  - It make more explicit the depedances between the subtests inside the logs.

### DIFF
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -1,55 +1,72 @@
 package cache
 
 import (
+	"context"
 	"sort"
 	"testing"
 	"time"
 
 	"github.com/go-redis/redis/v8"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


Peltoche:
This review have been propose a little quickly. First we need to see if this kind of review interest the team or not. 

Let's focus on #3606 first. If everything goes well I would open this review again.

PR #3608, opened 2023-01-01
Replace some multiline assert() use by a require call

Peltoche:
The following pattern:

```go
if !assert.NoError(t, err) {
  return
}
```

does the following steps:
- check if there is an error
- log the error with the assert package
- exit the test immediately

All those steps can be done with a single line:

```go
require.NoError(t, err)
```


nono (approved):
Thanks for the work. We were using this pattern before the `require` package was added and we have been too lazy to change that later.

PR #3610, opened 2023-01-01
Add the golangci-lint config file

Peltoche:
This config file take the exact same configuration than found inside the
Makefile. Using a config file will allow a more fine grained
configuration and also permit the direct use of golangci-lint


nono:
Closed by #3627

PR #3614, opened 2023-01-02
Use `strings.ReplaceAll` instead of `strings.Replace(*, *, -1)`

Peltoche:
Those replacements have been sugested by the `gocritic` linter.


nono:
Good catch, those lines were from a time when `strings.ReplaceAll` wasn't supported by the version of Go we were using.

nono (approved):

PR #3617, opened 2023-01-02
Fix the gocritic `elseIf` suggestions

Peltoche:
Those fix have be done following the suggestions from the gocritic linter

### DIFF
--- a/model/sharing/member.go
+++ b/model/sharing/member.go
@@ -543,10 +543,11 @@ func (s *Sharing) FindMemberByInboundClientID(clientID string) (*Member, error) 
 				return &s.Members[i+1], nil
 			}
 		}
-	} else {
-		if s.Credentials[0].InboundClientID == clientID {
-			return &s.Members[0], nil
-		}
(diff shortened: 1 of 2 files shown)


nono:
Well, I prefer my version. It helps to understand the code: the first if and the second if are not for the same thing, and merging them makes it harder for me to see what the code does.

Peltoche:
> Well, I prefer my version. It helps to understand the code: the first if and the second if are not for the same thing, and merging them makes it harder for me to see what the code does.

Understandable. After a reread I agree that the `if/else if` pattern doesn't match. I would like to propose you a second solution with an early return.

nono:
It's better. I still have a light preference for the original code, but if the goal is to enable gocritic, I can accept this PR.

Peltoche:
> if the goal is to enable gocritic, I can accept this PR.

The commit is ready, it just require to merge this PR and #3610 because it require to have the config file in order to remove some sub-lint. I can push it now if you want to take a look.

nono (approved):

PR #3619, opened 2023-01-02
Fix the gocritic `singleCaseSwitch` suggestions

Peltoche:
Those fix have be done following the suggestions from the gocritic linter

This commit fix the followings errors:
```
model/job/broker.go:198:2: singleCaseSwitch: should rewrite switch statement to if statement (gocritic)
model/job/scheduler.go:228:2: singleCaseSwitch: should rewrite switch statement to if statement (gocritic)
model/vfs/metadata.go:176:2: singleCaseSwitch: should rewrite switch statement to if statement (gocritic)
```

### DIFF
--- a/model/job/scheduler.go
+++ b/model/job/scheduler.go
@@ -228,8 +228,9 @@ func (t *TriggerInfos) Fetch(field string) []string {
 	switch field {
 	case "worker":
 		return []string{t.WorkerType}
+	default:
+		return nil
 	}
-	return nil
 }
 
 func createTrigger(t Trigger) error {
--- a/model/sharing/shared.go
+++ b/model/sharing/shared.go
@@ -99,8 +99,9 @@ func (s *SharedRef) Fetch(field string) []string {
 			keys = append(keys, key)
 		}
 		return keys
+	default:
+		return nil
 	}
-	return nil
 }
 
 // FindReferences returns the io.cozy.shared references to the given identifiers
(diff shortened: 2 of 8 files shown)


model/job/scheduler.go
```diff
@@ -306,5 +307,7 @@ func GetTriggerState(db prefixer.Prefixer, triggerID string) (*TriggerState, err
 	return &state, nil
 }
 
-var _ couchdb.Doc = &TriggerInfos{}
-var _ permission.Fetcher = &TriggerInfos{}
+var (
+	_ couchdb.Doc        = &TriggerInfos{}
+	_ permission.Fetcher = &TriggerInfos{}
+)
```
nono, line 313: I would prefer to keep the short version (2 lines) instead of the longer version (4 lines) here

model/sharing/shared.go
```diff
@@ -560,5 +561,7 @@ func CheckShared(inst *instance.Instance) ([]*CheckSharedError, error) {
 	return checks, err
 }
 
-var _ couchdb.Doc = &SharedRef{}
-var _ permission.Fetcher = &SharedRef{}
+var (
+	_ couchdb.Doc        = &SharedRef{}
+	_ permission.Fetcher = &SharedRef{}
+)
```
nono, line 567: idem

nono (approved):
Thanks!

PR #3623, opened 2023-01-02
Fix the gocritic `badCall` suggestions

Peltoche:
Those fixes are following the suggestions from the gocritic linter

The fixed errors was:
```
cmd/config.go:45:15: badCall: suspicious Join on 1 argument (gocritic)
			filename = filepath.Join(utils.AbsPath(args[0]))
			          ^
cmd/config.go:125:15: badCall: suspicious Join on 1 argument (gocritic)
		filename := filepath.Join(utils.AbsPath(args[0]))
```

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -8,7 +8,6 @@ import (
 	"io"
 	"os"
 	"path"
-	"path/filepath"
 
 	"github.com/cozy/cozy-stack/client/request"
 	"github.com/cozy/cozy-stack/model/account"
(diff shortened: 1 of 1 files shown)


nono:
It's weird. I would have expected the `filepath.Join` to have been called on several items, and only after a refactoring, only one item would have been left. But it is not the case, it was `filepath.Join` on a single item from the start. By reading the code, my hypothesis is that is a way to clean the path. As such, I would prefer `filename = filepath.Clean(utils.AbsPath(args[0]))`.

Peltoche:
>  By reading the code, my hypothesis is that is a way to clean the path

I was wondering too.

nono (approved):

PR #3625, opened 2023-01-02
Add a lock around fs.Copier when updating apps

nono:
When an app is installed or updated, it may happen that several instances are doing this operation at the same time. There was already a lock to avoid several installs/updates of an app on the same instance, but we want to avoid a too large lock, which can create contention.

This commit adds a new lock to avoid FS write operations for installing/updating an app from several instances at the same time (ie just after a new version has been published). The happy path avoids the lock to ensure best performances.

It should help to reduce the pressure on Swift after a new version of an app is published.

### DIFF
--- a/model/app/fetcher_http.go
+++ b/model/app/fetcher_http.go
@@ -16,7 +16,9 @@ import (
 	"time"
 
 	"github.com/cozy/cozy-stack/pkg/appfs"
+	"github.com/cozy/cozy-stack/pkg/lock"
 	"github.com/cozy/cozy-stack/pkg/logger"
+	"github.com/cozy/cozy-stack/pkg/prefixer"
 	"github.com/cozy/cozy-stack/pkg/utils"
 	"github.com/labstack/echo/v4"
 )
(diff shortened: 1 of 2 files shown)


model/app/fetcher_http.go
```diff
@@ -107,7 +109,25 @@ func (f *httpFetcher) Fetch(src *url.URL, fs appfs.Copier, man Manifest) (err er
...
 
 func fetchHTTP(src *url.URL, shasum []byte, fs appfs.Copier, man Manifest, prefix string) (err error) {
-	exists, err := fs.Start(man.Slug(), man.Version(), man.Checksum())
+	// Happy path: it exists and we don't need to acquire the lock.
+	exists, err := fs.Exist(man.Slug(), man.Version(), man.Checksum())
+	if err != nil || exists {
+		return err
+	}
+
+	// For the lock key, we use the checksum when available, and else, we
```
taratatach, line 118: Why do we use the checksum in the first place? Can we have multiple versions of the same app being installed at the same time?

pkg/appfs/copier.go
```diff
@@ -147,6 +157,14 @@ func NewAferoCopier(fs afero.Fs) Copier {
...
+func (f *aferoCopier) Exist(slug, version, shasum string) (bool, error) {
+	appDir := path.Join("/", slug, version)
+	if shasum != "" {
+		appDir += "-" + shasum
+	}
+	return afero.DirExists(f.fs, appDir)
+}
+
 func (f *aferoCopier) Start(slug, version, shasum string) (bool, error) {
 	f.appDir = path.Join("/", slug, version)
```
taratatach, line 169: You could use `f.Exist()` here, couldn't you?

model/app/fetcher_http.go
nono: Yes, we can have the same app+version with two checksums being installed/updated at the same time if the app is present in several contexts with different spaces for the apps registry.

pkg/appfs/copier.go
nono: `f.appDir` is not defined in `f.Exist()` and it is used later to create the directory (`dir := path.Dir(f.appDir)`).

taratatach: Then why not assigning the path of the directory to f.appDir in Exist like is done for the Swift copier?

PR #3627, opened 2023-01-03
Add some linters to golangci-lint

Peltoche:
- `govet` examines Go source code and reports suspicious constructs, such as Printf calls whose arguments do not align with the format string
-  `nolintlint` reports ill-formed or insufficient nolint directives
- `gocritic` provides diagnostics that check for bugs, performance and style issues.
  Extensible without recompilation through dynamic rules.
  Dynamic rules are written declaratively with AST patterns, filters, report message and optional suggestion.
- `tenv` tenv is analyzer that detects using os.Setenv instead of t.Setenv since Go1.17


Peltoche:
This review require #3617 and #3610  to merged in order to pass the CI

PR #3633, opened 2023-01-04
ADR system proposition

Peltoche:
I have started the test migration project #3606 and encounter an issue: On project/migration requiring severa commit/PR it is really difficult to give a complete context for each commit/PR. 

In order to solve that I propose you to implement a [simple ADR system](https://adr.github.io/). This documentation tool have several benefits:
- It allows to easily link a commit to a full explanation on why this is done by writing the ADR link into the description.
- It allows contributors to propose a consequent change and exchange with the core team before any code is written. This "fail fast" system can avoid the frustration of having a big PR refused.
- It allows to keep a detailed record about decisions history. If someone ask why this decision have been made it's possible to redirect him to the corresponding ADR.
- It allows to make implicit rules explicit by writing them.

### DIFF
--- a/.adr-dir
+++ b/.adr-dir
@@ -1,0 +1,1 @@
+./docs/architecture/decisions
--- a/docs/architecture/decisions/0001-record-architecture-decisions.md
+++ b/docs/architecture/decisions/0001-record-architecture-decisions.md
@@ -1,0 +1,19 @@
+# 1. Record architecture decisions
+
+Date: 2023-01-04
+
+## Status
+
+Accepted
+
+## Context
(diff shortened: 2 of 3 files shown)


nono:
I think ADR are a good practices, but it is not the right timing for us.

For the past few years, the maintainers team for cozy-stack was just me. We are now building a team and I would like that the process are discussed and chosen by the whole team. And we also have a bit of internal documentation (when it includes client names and things like that).

Peltoche:
Completely understandable, thanks for the explanation. :+1:

PR #3645, opened 2023-01-08
Refacto tests `/pkg`

Peltoche:


pkg/hooks/hooks_test.go
```diff
@@ -2,28 +2,21 @@ package hooks
...
 
 func TestHooks(t *testing.T) {
-	if testing.Short() {
-		t.Skip("an instance is required for this test: test skipped due to the use of --short flag")
-	}
-
 	config.UseTestFile()
-	config.GetConfig().Hooks = "../../tests/hooks"
-	os.Exit(m.Run())
+	config.GetConfig().Hooks = "./testdata"
```
nono, line 15: Why did you move the hooks from `tests/hooks` to `pkg/hooks/testdata`?

nono:
LGTM but there are some conflicts that must be resolved before merging

pkg/hooks/hooks_test.go
Peltoche: Based on the description of `go help test` the `/testdata` seems to be a more correct/standard place to handle this kind of file. 

```
go help test
...
The go tool will ignore a directory named "testdata", making it available
to hold ancillary data needed by the tests.
...
```

As explained in [the testing package doc](https://pkg.go.dev/testing#hdr-Fuzzing) this is also a good place for the Fuzz testing data feed.  :slightly_smiling_face:

nono: By the way, you can remove the hooks from `tests/hooks`, not just copy them

nono (approved):

Peltoche:
:confetti_ball: And this is the end of the refactoring! :confetti_ball:

PR #3660, opened 2023-01-09
Add a `context.Context` arg to the `pkg/cache` package

Peltoche:


nono:
What's the plan? I see a lot of `context.TODO()`.

Peltoche:
I would like to propose you to take advantage of the `context.Context` for all your I/Os. This would allows you several benefits better explained by the package documentation first line:

> Package context defines the Context type, which carries deadlines, cancellation signals, and other 
request-scoped values across API boundaries and between processes.

As you codebase start to be really huge it's not possible to add the arg everywhere in one commit or in one PR.

So the goal was to implement `context.Context` on all the lib requiring it and slowly propagating the `ctx` arg until we reach the http handlers. This seems to be the more classical approach and in order to help this pattern the standard librairy have implemented [context.TODO()](https://pkg.go.dev/context#TODO) which is an alias for `context.Background()`:

> TODO returns a non-nil, empty Context. Code should use context.TODO when it's unclear which Context to use or it is not yet available (because the surrounding function has not yet been extended to accept a Context parameter).

Using this method instead of `context.Background` allow to easily find where the context propagation still need works. The goal is of course to replace them by a valid context as soons as possible.

nono:
> So the goal was to implement context.Context on all the lib requiring it and slowly propagating the ctx arg until we reach the http handlers.

My issue is here. I really don't think we should propagate the context from http handlers to the `pkg` calls.

Let's take the `Remove` function from the `pkg/asset` package. As I understand, what you want to do is something like:

```go
func Remove(ctx context.Context, name, instanceCtx string) error {
	err := dynamic.RemoveAsset(ctx, instanceCtx, name)
	if err == nil {
		key := fmt.Sprintf("dyn-assets:%s/%s", instanceCtx, name)
		cache := config.GetConfig().CacheStorage
		cache.Clear(ctx, key)
	}
	return err
}
```

If `ctx` has a deadline that expires after `dynamic.RemoveAsset` but before `cache.Clear` (or the client has closed the http connection that cancels `ctx`), the cache won't be invalided but it should have been done.

And with CouchDB that has no transaction mechanisms, this kind of issue can be really bad.

Peltoche:
In your case above you can use `context.Background` for your `cache.Clear` method. You will have no cancel/deadline and be sure to make the call without interruptions. 

```go
func Remove(ctx context.Context, name, instanceCtx string) error {
	err := dynamic.RemoveAsset(ctx, instanceCtx, name)
	if err == nil {
		key := fmt.Sprintf("dyn-assets:%s/%s", instanceCtx, name)
		cache := config.GetConfig().CacheStorage
		cache.Clear(context.Background(), key)
	}
	return err
}
```

nono:
> In your case above you can use context.Background for your cache.Clear method.

Sure, but it means that you cannot do the work mechanically, like you started to do. You have to think for every place where contexts are manipulated. And it takes time to think. My fear is that we start the migration and see that it is very long to make it right and abort.

> BTW maybe you should clean your cache before your source of trust because. If your app crash between the two calls you prefer having you document removed from the cache and still inside the couchdb than having the opposite. That said you sill have a really little chance to have an another call pushing a new version in the cache between the two calls.

Well, for this particular example, I would say I prefer to remove the doc from CouchDB and still have the cache. The cache will expire and, for this example, it is not a big deal if we serve the old content a bit longer.

> That said you sill have a really little chance to have an another call pushing a new version in the cache between the two calls.

I'm far more worried by race conditions like this one than by crashes. But, as the adage says: your mileage may vary ;-)

nono:
Also, for the cache package, I don't think we should add a `ctx` parameter to most functions. I don't see a case where the caller may want to cancel the call or have a deadline/timeout (in addition to the one already configured by the redis client).

Peltoche:
> Sure, but it means that you cannot do the work mechanically, like you started to do. You have to think for every place where contexts are manipulated. And it takes time to think.

I can if I put placeholder values doing nothing (`context.TODO()`) this would allow to add the arguments where it needs to be and after you can start connecting the context between them.

For the connection part where we start reusing the context across functions and start having a real interest it seems that in 80%/90% of the time this is a no brainer work, you only have to forward the context. You can start asking questions when you haves consecutive writes (like in your example). In that cas it would require more intention indeed. I can skip those by keeping `context.TODO` during the first passe and then we check one by one what to do? 

> My fear is that we start the migration and see that it is very long to make it right and abort.

This migration will take time indeed but I can take the subject if you want. IMHO even a partial migration would bring some benefice for the handler migrated but if you think this subject too big, not the right time or anything it's your choice.

> Also, for the cache package, I don't think we should add a ctx parameter to most functions. I don't see a case where the caller may want to cancel the call or have a deadline/timeout (in addition to the one already configured by the redis client).

IHMO for the deadline part you should give the possibility to the caller and let him choose with it's context. It could be useful for the getters at least. The other point is more prospective: the context is useful for many other things than the deadline/cancel feature. You can use it to spread a `UniqueID` header across all your http calls  in order to trace you requests inside the logs, it's often a requirement for the tracing and you can use it to share contextual stuff. At the moment you use any of those capabilities but having a context everywhere could open those possibilities for you. 


In any case I'm kind of sorry to start this kind of (big) project out of the blue and directly with a PR, this is kind of brutal and maybe you want to take the time to assess before. For the moment the only channel I found to exchange with you was the PR discussions but it require some code and clearly for the last two projects it was not the correct way to propose and exchange about a prodject. Do you have a preferred channel of communication for this talking about ideas/projects/advices? issue? irc? email?

nono:
Hi, for your information, I have not yet decided if I want to advance on this project, and what would be a good way to do that. But, it's not lost, I just need more time to think about it.

Peltoche:
No worry. I'm thinking on my side how I can advance this subject more smoothly. I propose you to close this PR and move the discussion inside and issue.

PR #3661, opened 2023-01-11
fix: Return member with interact permissions

taratatach:
When returning permissions from an interact sharecode, we'll return
the associated sharing member in the `included` part of the response
like we already do with preview sharecodes.


zatteo:
It works! 🎉

zatteo (approved):

PR #3663, opened 2023-01-11
Add the `context.Context` arg support for the `pkg/couchdb` package

Peltoche:
A `context.TODO()` have been used as a placeholder for all the calls,
even if there is a real context available.

This is due to the fact that the call migration is done programatically
(with `sed`) without taking into account each call environment. This
should be fixed in a following step.

### DIFF
--- a/model/account/accounts.go
+++ b/model/account/accounts.go
@@ -1,6 +1,7 @@
 package account
 
 import (
+	"context"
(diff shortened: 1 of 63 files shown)


Peltoche:
This reviews is too big, too quick, too early for the moment I think. First we need to see what #3660 will finish.

PR #3666, opened 2023-01-12
Fix memory leaks in redis locks

nono:
We have many `redisLock` structs, and using one random number generator and logger entry for each `redisLock` takes a lot of RAM. We can use a single global variable to avoid that.


taratatach:
Could you explain why having the logger and the random generator in the `RedisLock` struct is a source of memory leaks?

sblaisot (approved):

taratatach (approved):

PR #3678, opened 2023-01-18
feat: Add cozyMetadata to new Accounts if missing

taratatach:
If the client requesting the creation of an Account does not provide
any `cozyMetadata`, we'll generate it.
The idea is to minimize the amount of data partners have to provide to
create Accounts.

### DIFF
--- a/web/data/accounts.go
+++ b/web/data/accounts.go
@@ -4,12 +4,15 @@ import (
 	"encoding/json"
 	"errors"
 	"net/http"
+	"strings"
 
 	"github.com/cozy/cozy-stack/model/account"
+	"github.com/cozy/cozy-stack/model/oauth"
 	"github.com/cozy/cozy-stack/model/permission"
 	"github.com/cozy/cozy-stack/pkg/consts"
 	"github.com/cozy/cozy-stack/pkg/couchdb"
 	"github.com/cozy/cozy-stack/pkg/jsonapi"
+	"github.com/cozy/cozy-stack/pkg/metadata"
 	"github.com/cozy/cozy-stack/web/middlewares"
 	"github.com/labstack/echo/v4"
 )
@@ -117,6 +120,12 @@ func updateAccount(c echo.Context) error {
 
 	account.Encrypt(doc)
 
+	if doc.M["cozyMetadata"] == nil {
+		// This is not the expected type for a JSON doc but it should work since it
+		// will be marshalled when saved.
+		doc.M["cozyMetadata"] = CozyMetadataFromClaims(c)
+	}
+
 	errUpdate := couchdb.UpdateDoc(instance, &doc)
 	if errUpdate != nil {
 		return fixErrorNoDatabaseIsWrongDoctype(errUpdate)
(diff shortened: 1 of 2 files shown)


web/data/accounts.go
```diff
@@ -155,6 +156,21 @@ func createAccount(c echo.Context) error {
...
+		data, err := json.Marshal(cm)
+		if err != nil {
+			return err
+		}
+
+		doc.M["cozyMetadata"] = new(map[string]interface{})
+		err = json.Unmarshal(data, doc.M["cozyMetadata"])
+		if err != nil {
+			return err
+		}
```
nono, line 171: Why not just `doc.M["cozyMetadata"] = metadata.New()`?

nono:
And maybe we could also adds the slug of app that have made the request, like we do for files: https://github.com/cozy/cozy-stack/blob/master/web/files/files.go#L2197-L2201

web/data/accounts.go
taratatach: Well, we expect `doc.M["cozyMetadata"]` to be of type `map[string]interface{}` while `metadata.New()` returns a type `*CozyMetadata`.
Maybe that's overkill since we save `doc` right after that but I wanted to stay consistent with the usage we have of this attribute in other places (e.g. https://github.com/cozy/cozy-stack/blob/master/model/sharing/sharing.go#L1702)

taratatach:
Do we have claims in all requests made by apps? On another note, my interpretation was that we still recommend the client provides the `cozyMetadata` and that Harvest would do so. Thus this would not be necessary.

But adding it is a no-brainer so let's do it.

I'll take a look at the BI webhooks originated creations.

nono:
> Do we have claims in all requests made by apps?

Yes, the claims are extracted from the token and are needed to have permissions. So, if a request wouldn't have claims, it would mean that this request would have been rejected when we check the permission for creating/updating the account.

web/data/accounts.go
```diff
@@ -167,3 +174,45 @@ func createAccount(c echo.Context) error {
...
+	if claims := c.Get("claims"); claims != nil {
+		cl := claims.(permission.Claims)
+		switch cl.Audience {
+		case consts.AppAudience, consts.KonnectorAudience:
+			slug = cl.Subject
+		case consts.AccessTokenAudience:
+			if perms, err := middlewares.GetPermission(c); err == nil {
+				if cli, ok := perms.Client.(*oauth.Client); ok {
+					slug = oauth.GetLinkedAppSlug(cli.SoftwareID)
+					// Special case for cozy-desktop: it is an OAuth app not linked
```
nono, line 193: Well, I don't think we need this special case for cozy-desktop, as this app doesn't create any io.cozy.accounts

nono (commented):
We are adding cozyMetadata to new accounts. What about existing accounts that are updated? I think it would be nice to add cozyMetadata if they don't have it.

web/data/accounts.go
taratatach: It doesn't at the moment but I thought it would be nice to handle this special case right away. I fear we might not think about this part if we start creating accounts from Desktop some day.

taratatach:
I can add the `cozyMetadata` in the update actions matching the create actions that add it. This will be limited but it will be consistent.

nono (approved):

PR #3680, opened 2023-01-18
Keep order for permissions on /registry

nono:


pkg/registry/registry.go
```diff
@@ -368,29 +369,24 @@ func (a *appsList) fetch(r *registryFetchState, fetchAll bool) error {
...
-			valB := vj[sortBy].(int)
-			less = valA < valB
-			if !sortBySlug && !less {
-				equal = valA == valB
-			}
-		}
+		valA := a.list[i][sortBy]
+		valB := a.list[j][sortBy]
+		cmp := bytes.Compare([]byte(valA), []byte(valB))
+		equal := cmp == 0
```
taratatach, line 385: This could be true even even in the case where `sortBy` is `"slug"` when it would have been false before.
I'm still having a hard time figuring out if that's an issue or not as I'm not sure I understand what this lambda is meant to do.

taratatach, line 385: Since this `equal` variable is only used to modify the value of `less` based on the second comparison of slugs, I would suggest not assigning `cmp == 0` to it and do the comparison directly in the next `if`.

taratatach (approved):

pkg/registry/registry.go
nono: The `equal` and `less` variables are here to make the code more readable, but I agree that they could be removed. I have a light preference for keeping them, but if you prefer that we remove them and just use `cmp`, I can do that. What do you think?

PR #3685, opened 2023-01-23
Fix downgrading app

nono:
When an app is installed on version x.y.z from the stable canal of the apps registry, and the stack is asked to update it to the version x.y.z-dev.shasum from the dev canal of the apps registry, the downgrade was failing because the manifest sent to the apps registry was for the x.y.z version, without the shasum. The stack is now forcing the version for this case.


taratatach (approved):
Should we do that for the first install of the app as well?

nono:
No, it is not required, it already works for the first install

PR #3687, opened 2023-01-25
Improve logs sent by the flagship app

nono:
- accept logs sent for an uninstalled application
- use the jobs namespace
- accept a job_id parameter in the query string
- accept a version parameter in the query string.

### DIFF
--- a/web/apps/apps.go
+++ b/web/apps/apps.go
@@ -228,17 +228,8 @@ func logsHandler(appType consts.AppType) echo.HandlerFunc {
 	return func(c echo.Context) error {
 		inst := middlewares.GetInstance(c)
 
-		var slug string
-		err := middlewares.AllowMaximal(c)
-		if err == nil {
-			// If logs are sent by the flagship app, get the slug from the
-			// request params.
-			slug = c.Param("slug")
-			_, err := app.GetBySlug(inst, slug, appType)
-			if err != nil {
-				return wrapAppsError(err)
-			}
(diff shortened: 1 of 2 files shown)


web/apps/apps.go
```diff
@@ -228,17 +228,8 @@ func logsHandler(appType consts.AppType) echo.HandlerFunc {
...
-			// request params.
-			slug = c.Param("slug")
-			_, err := app.GetBySlug(inst, slug, appType)
-			if err != nil {
-				return wrapAppsError(err)
-			}
-		} else {
+		slug := c.Param("slug")
+		if err := middlewares.AllowMaximal(c); err != nil {
 			// If logs are not sent by the flagship app, check that it's sent by
```
taratatach, line 233: This seems unnecessary if we only allow flagship apps to send logs.

nono: Yes, but I would like to keep it as it can be useful for debugging

taratatach: OK. Maybe update the comment then, since we won't get into this block if logs are not sent by the flagship app.

nono: I don't understand: we go in this block if logs are not sent by the flagship app, it's the purpose of this block

taratatach: Aaah I'm sorry, I misread the condition… This is totally fine!

taratatach (approved):

PR #3689, opened 2023-01-25
Fix cozy-stack tools unxor command

nono:


taratatach (approved):
Could you explain why it didn't work please?

nono:
The configuration was not loaded when running the command directly. Now, it will call an admin endpoint of the stack, where the configuration is already loaded.

PR #3690, opened 2023-01-25
Fix creation of a note with same title

nono:
When a note is created, and another note has the same title, the filename of the new note should be "title (n)", not "new note - date".

### DIFF
--- a/model/note/note.go
+++ b/model/note/note.go
@@ -367,6 +367,8 @@ func writeFile(inst *instance.Instance, doc *Document, oldDoc *vfs.FileDoc) (fil
 			if err != nil {
 				return nil, err
 			}
+			fileDoc.DocName = oldDoc.DocName
+			fileDoc.ResetFullpath()
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


model/note/note.go
```diff
@@ -418,8 +420,17 @@ func forceRename(inst *instance.Instance, old *vfs.FileDoc, file *vfs.FileDoc) (
...
+	basename := tmp.DocName
+	for i := 2; i < 100; i++ {
+		err := inst.VFS().UpdateFileDoc(old, tmp)
+		if err == nil {
+			break
+		} else if !errors.Is(err, os.ErrExist) {
+			return nil, err
+		}
+		filename := strings.TrimSuffix(path.Base(basename), path.Ext(basename))
+		tmp.DocName = fmt.Sprintf("%s (%d).cozy-note", filename, i)
```
taratatach, line 432: With this algorithm, we could end up with a note named `my note (2) (3) (4).cozy-note`. What do you think about using the same logic as the one handling file conflicts? See https://github.com/cozy/cozy-stack/blob/master/model/vfs/vfs.go#L936-L962

nono: I don't understand how we could obtain `my note (2) (3) (4).cozy-note`

taratatach: With this scenario :

1. create note `my note.cozy-note`
2. create note `my note.cozy-note` → it's renamed into `my note (2).cozy.note`
3. create note `my note.cozy-note` → we try to rename it into `my note (2).cozy-note` (:negative_squared_cross_mark: ) →  it's renamed into `my note (2) (3).cozy-note`

taratatach: You're right. I must have been tired. `my note (2) (2).cozy-note` is still possible though. We might not find it a problem but the result would be different when using the copy feature.

PR #3691, opened 2023-01-25
Add an admin endpoint to get the last activity date of a Cozy

nono:


docs/admin.md
```diff
@@ -83,6 +83,31 @@ Content-Type: application/vnd.api+json
 }
 ```
 
+### GET /instances/:domain/last-activity
+
+It returns an approximate date of when the instance was last used by their
+owner (automatic jobs like connectors doesn't count). It looks at the sessions
```
taratatach, line 89:
```suggestion
owner (automatic jobs like connectors don't count). It looks at the sessions
```

web/instances/instances.go
```diff
@@ -377,6 +379,75 @@ func cleanSessions(c echo.Context) error {
...
+			}
+		}
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	return c.JSON(http.StatusCreated, echo.Map{
+		"last-activity": last.Format("2006-01-02"),
```
taratatach, line 447: Do we limit the precision for privacy issues?

taratatach (approved):

web/instances/instances.go
nono: Well, we don't need a good precision on that endpoint, and the current sources for this are not precise. For example, one source if when an OAuth token is renewed, and as an OAuth token is valid for one week, the precision is really not good. And, yes, we don't want to track users for very precise events for privacy issues.

PR #3692, opened 2023-01-26
feat: Add --path flag to admin export command

taratatach:
This new flag accepts a local directory path so that an admin can
make an export of the given instance.

When passing this flag, the cli will download all the export's
archives from the Cozy and save them with their archive name or, if
that name is missing from the download response, as
`<instance domain> - part<00n>.zip` in the directory pointed by the
flag.
If a file with the same name already exists, it is overwritten.

The command returns when all the archives have been downloaded.

### DIFF
--- a/client/admin_client.go
+++ b/client/admin_client.go
@@ -1,0 +1,46 @@
+package client
+
+import (
+	"time"
+
+	"github.com/cozy/cozy-stack/client/request"
+	"github.com/cozy/cozy-stack/pkg/config/config"
+	"github.com/cozy/cozy-stack/pkg/consts"
+	"github.com/cozy/cozy-stack/pkg/tlsclient"
+)
+
+type AdminClient struct {
+	Client
+}
+
+func (ac *AdminClient) NewInstanceClient(domain string, scopes ...string) (*Client, error) {
+	token, err := ac.GetToken(&TokenOptions{
+		Domain:   domain,
+		Subject:  "CLI",
+		Audience: consts.CLIAudience,
+		Scope:    scopes,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	httpClient, clientURL, err := tlsclient.NewHTTPClient(tlsclient.HTTPEndpoint{
+		Host:      config.GetConfig().Host,
+		Port:      config.GetConfig().Port,
+		Timeout:   5 * time.Minute,
+		EnvPrefix: "COZY_HOST",
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	c := &Client{
+		Scheme:     clientURL.Scheme,
+		Addr:       clientURL.Host,
+		Domain:     domain,
+		Client:     httpClient,
+		Authorizer: &request.BearerAuthorizer{Token: token},
+	}
+
+	return c, nil
+}
--- a/client/instances.go
+++ b/client/instances.go
@@ -4,13 +4,21 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"mime"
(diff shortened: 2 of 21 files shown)


client/admin_client.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+type AdminClient struct {
+	Client
+}
+
+func (ac *AdminClient) NewInstanceClient(domain string, scopes ...string) (*Client, error) {
+	// For the CLI client, we rely on the admin APIs to generate a CLI token.
+	// We may want in the future to rely on OAuth to handle the permissions with
+	// more granularity.
```
nono, line 19: I don't understand this comment. Can you explain me what are you thinking about?

client/instances.go
```diff
@@ -453,27 +457,73 @@ func (c *Client) Updates(opts *UpdatesOptions) error {
...
+		}
+		channel, err := ic.RealtimeClient(RealtimeOptions{
+			DocTypes: []string{"io.cozy.jobs"},
+		})
+		if err != nil {
+			return err
+		}
+		defer func() {
+			channel.Close()
+		}()
```
nono, line 489: [nit] you can compress the 3 lines in `defer channel.Close()`

client/instances.go
```diff
@@ -453,27 +457,73 @@ func (c *Client) Updates(opts *UpdatesOptions) error {
...
+
+		defer res.Body.Close()
```
nono, line 491: I don't understand why we close the response body only if we wait for the result. We should do it on both cases (waiting the result, or not).

model/move/doc.go
```diff
@@ -223,16 +224,19 @@ func (e *ExportDoc) CleanPreviousExports(archiver Archiver) error {
...
-		bucketSize = minimalPartsSize
-		if usage, err := i.VFS().DiskUsage(); err == nil && usage > bucketSize {
-			factor := math.Sqrt(float64(usage) / float64(minimalPartsSize))
-			bucketSize = int64(factor * float64(bucketSize))
+	var bucketSize int64
+	if opts.LocalPath == "" {
+		// The size of the buckets can be specified by the options. If it is not
+		// the case, it is computed from the disk usage. An instance with 4x more
+		// bytes than another instance will have 2x more buckets and the buckets
+		// will be 2x larger.
```
nono, line 232: We should keep the system of buckets even when doing the export from the CLI.

nono (commented):
> The exported archive can be found at `<given path>/<instance domain>/<random id>.tar.gz`.

It should be zip, not tar.gz.

client/instances.go
taratatach: I thought it wasn't necessary as long as we were not reading it. I moved the defer call right after getting the response (if there aren't any errors).

model/move/doc.go
taratatach: I didn't quite understand how the bucket system worked. This now fixed now.

client/admin_client.go
taratatach: I just copy/pasted the comment from https://github.com/cozy/cozy-stack/blob/master/cmd/root.go#L55-L58 since I copy/pasted the logic.
This might not be relevant anymore though.

taratatach:
I mistook the export archive (tar + gzipped) for the actual Cozy export archives (zipped) at first. This is now corrected.

client/admin_client.go
nono: Hu, yes, let's remove this comment

nono (approved):

nono (approved):

PR #3693, opened 2023-01-26
Use `gavv/httpexpect` for the `/web/bitwarden` tests

Peltoche:
This migration bring several advantages:
- It allow to handle the body close automatically
- It remove the need to handle a lot of useless bytes/buffer managment
- The body validation is attached to the request and so the error messages are greatly improved
- Later we will be able to setup a more verbose logger if we catch the use of a custom flag.

### DIFF
--- a/web/bitwarden/bitwarden_test.go
+++ b/web/bitwarden/bitwarden_test.go
@@ -1,18 +1,15 @@
 package bitwarden
 
 import (
-	"bytes"
 	"encoding/base64"
 	"encoding/json"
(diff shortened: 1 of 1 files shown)


Peltoche:
Nop sorry. I should stop marking them as WIP, I always forgot to mark them as ready

nono:
For the integration tests, it's OK, it's not related to this PR: https://github.com/cozy/cozy-stack/commit/aa067f34c540651170a73371130e559d8eb2f987

PR #3694, opened 2023-01-26
Fix for thumbnails handling

nono:
We had some old thumbnails that were incorrectly saved in Swift. When encountering one of them, we will delete it and retry generating a new one.

### DIFF
--- a/model/vfs/vfsswift/thumbs_v2_v3.go
+++ b/model/vfs/vfsswift/thumbs_v2_v3.go
@@ -142,6 +142,14 @@ func (t *thumbsV2) ServeThumbContent(w http.ResponseWriter, req *http.Request, i
 
 	ctype := o["Content-Type"]
 	if ctype == echo.MIMEOctetStream {
+		// We have some old images where the thumbnail has not been correctly
+		// saved in Swift. We should delete the thumbnail to allow another try.
+		if l, err := f.Length(t.ctx); err == nil && l > 0 {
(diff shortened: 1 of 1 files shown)


model/vfs/vfsswift/thumbs_v2_v3.go
```diff
@@ -142,6 +142,14 @@ func (t *thumbsV2) ServeThumbContent(w http.ResponseWriter, req *http.Request, i
 
 	ctype := o["Content-Type"]
 	if ctype == echo.MIMEOctetStream {
+		// We have some old images where the thumbnail has not been correctly
+		// saved in Swift. We should delete the thumbnail to allow another try.
+		if l, err := f.Length(t.ctx); err == nil && l > 0 {
+			_ = t.RemoveThumbs(img, vfs.ThumbnailFormatNames)
```
paultranvan, line 148: Why the `_ = ` if we don't use it?

paultranvan (approved):

model/vfs/vfsswift/thumbs_v2_v3.go
nono: The linter is not happy when an error is returned and we don't do anything of it. A trick to tell the linter that we do that on purpose is to assign the error to `_`. It is a way to express that the function returns an error, that we have not missed it, and that we want to do nothing of the error.

taratatach (approved):

PR #3707, opened 2023-02-01
Use `gavv/httpexpect` for the `/web/permissions` tests

Peltoche:


nono:
It looks like the linter is not happy by your changes

Peltoche:
Yep, I pushed without the last patch. Should be good now

nono (approved):

PR #3716, opened 2023-02-03
Use `gavv/httpexpect` for the `/web/auth` tests

Peltoche:


web/auth/auth_test.go
```diff
@@ -34,40 +29,36 @@ import (
...
+	var clientID string
+	var clientSecret string
+	var registrationToken string
+	var altClientID string
+	var altRegistrationToken string
+	var csrfToken string
+	var code string
+	var refreshToken string
+
+	// TODO: Remove this print use to reomve an error for unused variables
```
nono, line 59: TODO spotted: do you want to fix it before we merge the PR?

Peltoche: Yep, this would be cool to remove the todo but I don't have the knowledge to do it. Do you thinks you can give me some explanation that I can put in this place?

nono: Humm, I'm not sure to understand what is blocking. The variables in the print are used later. For example, `clientID` and `clientSecret` are set in the `RegisterClientSuccessWithJustMandatoryFields` subtest, and used in the `ReadClientNoToken` subtest. If you remove the print line, what error(s) do you see?

Peltoche: Ha damn sorry! Wrong todo! I was thinking about an another TODO asking for a comment why a `time.Sleep` was required in a test. 

Indeed this line can be removed and I think I forgotten some other in the previous PR. I remove this one now and I will make a PR removing the other in a few minutes

nono:
The CI is red, but it is not your fault: the install of the couchdb package fails on GitHub Actions this morning on most runs. It looks like it is an availability issue of the CouchDB repository for debian packages.

nono (approved):

PR #3732, opened 2023-02-08
Fix the cozy-stack fix service-triggers command

nono:
This command wasn't recreating services in some cases where the trigger_id was blank.

### DIFF
--- a/web/instances/fixers.go
+++ b/web/instances/fixers.go
@@ -362,37 +362,42 @@ func serviceTriggersFixer(c echo.Context) error {
 			if service.TriggerOptions == "" {
 				continue
 			}
-			if service.TriggerID != "" {
-				_, err := jobsSystem.GetTrigger(inst, service.TriggerID)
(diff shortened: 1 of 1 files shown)


web/instances/fixers.go
```diff
@@ -362,37 +362,41 @@ func serviceTriggersFixer(c echo.Context) error {
...
+					var msg serviceMessage
+					if err := json.Unmarshal(infos.Message, &msg); err != nil {
+						continue
 					}
-					service.TriggerID = triggerID
+					if msg.Name != name {
+						continue
+					}
+					service.TriggerID = infos.TID
 					update = true
```
taratatach, line 384: Isn't a `break` missing here to stop looping when we've found the appropriate trigger ID to use?

taratatach (approved):

web/instances/fixers.go
nono: We could add it as an optimization, but it's not a big deal if we don't have it

PR #3734, opened 2023-02-08
Fix adding name to contacts on sharings

nono:
We need to fill both the fullname and the displayName, as it is what the contacts app expects.

### DIFF
--- a/model/contact/contacts.go
+++ b/model/contact/contacts.go
@@ -139,12 +139,19 @@ func (c *Contact) PrimaryCozyURL() string {
 	return url
 }
 
-// FillFullnameIfMissing can be used to add a fullname if there was none.
-func (c *Contact) FillFullnameIfMissing(fullname string) {
-	name, ok := c.Get("fullname").(string)
-	if !ok || len(name) == 0 {
-		c.M["fullname"] = fullname
+// AddNameIfMissing can be used to add a name if there was none.
+func (c *Contact) AddNameIfMissing(db prefixer.Prefixer, name string) error {
+	was, ok := c.Get("displayName").(string)
+	if ok && len(was) > 0 {
+		return nil
 	}
+	was, ok = c.Get("fullname").(string)
+	if ok && len(was) > 0 {
+		return nil
+	}
+	c.M["displayName"] = name
+	c.M["fullname"] = name
+	return couchdb.UpdateDoc(db, c)
 }
 
 // AddCozyURL adds a cozy URL to this contact (unless the contact has already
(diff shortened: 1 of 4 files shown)


model/contact/contacts.go
```diff
@@ -139,12 +139,19 @@ func (c *Contact) PrimaryCozyURL() string {
...
-func (c *Contact) FillFullnameIfMissing(fullname string) {
-	name, ok := c.Get("fullname").(string)
-	if !ok || len(name) == 0 {
-		c.M["fullname"] = fullname
-	}
+// AddNameIfMissing can be used to add a name if there was none.
+func (c *Contact) AddNameIfMissing(db prefixer.Prefixer, name string) error {
+	was, ok := c.Get("displayName").(string)
+	if ok && len(was) > 0 {
+		return nil
```
taratatach, line 146: Are we sure we'll always have a `fullname` if `displayName` is present? Otherwise, this early return will prevent filling in the missing `fullname`.

model/sharing/oauth.go
```diff
@@ -435,6 +435,16 @@ func (s *Sharing) ProcessAnswer(inst *instance.Instance, creds *APICredentials)
 			}
 			ac.Credentials.AccessToken = token
 
+			// Update the contact to fill the name if missing
+			if email := s.Members[i+1].Email; email != "" {
+				if c, err := contact.FindByEmail(inst, email); err == nil {
```
taratatach, line 440: This might be unrelated but is `FindByEmail()` resilient to case differences? Since we could have uppercase letters in user input, we could miss the contact.

model/contact/contacts.go
nono: No, but I would prefer to avoid filling a `fullname` if `displayName` is present, as it is really unclear what would be the correct way to fill it.

model/sharing/oauth.go
nono: `FindByEmail()` is case sensitive, but we don't care here, because we will use the exact same string: the member email has been taken from the contact, and we use it to find the contact later.

model/contact/contacts.go
taratatach: So, doesn't that mean we only really care about the `displayName` and we should always avoid filling in the `fullname`?

nono: No, I think we should fill `fullname` if `displayName` is missing. As I understand, a contact can have a `fullname`, a first name, a last name, a pseudo, etc. And `displayName` is a way to take the name that is the more meaningful between all the variants. It is persisted in CouchDB and not computed at runtime to allow indexing it and making requests sorting by it.

taratatach: OK. I still find a bit weird to have a display name without a full name but if we don't have the data, I guess there's nothing we can do about it.

taratatach (approved):

PR #3741, opened 2023-02-13
Redirect to store for not installed apps

nono:


web/errors/errors.go
```diff
@@ -157,13 +157,6 @@ func HTMLErrorHandler(err error, c echo.Context) {
 			linkURL = "https://manager.cozycloud.cc/v2/cozy/remind"
 		case app.ErrNotFound:
 			illustration = "/images/desert.svg"
-			linkURL = i.DefaultRedirection().String()
-			link = "Error Application not found Action"
-			if ctxSettings, ok := i.SettingsContext(); ok {
-				if hide, ok := ctxSettings["hide_button_on_app_not_found"].(bool); ok && hide {
-					link = ""
-				}
-			}
```
taratatach, line 166: Could you give me a bit more context about this change please? I don't understand what is the expected result. Shouldn't we redirect to the store as well?

nono: The only case I found where we were executing this code was for app that wasn't in the apps registry. Now, there is also the case where the store is not installed. But, in both cases, we shouldn't redirect to the store or show a link to it.

taratatach (approved):

PR #3751, opened 2023-02-15
Add a trigger for saving OO documents

nono:
When a document is edited with OnlyOffice, the OnlyOffice server keeps a cache of the document and will send the document to save to the stack only if the user asks for a forcesave (ctrl-s shortcut), or after some time with no tab opened on the document. In our experience, it can be long and the users can expect the document to be saved to the VFS, and then synchronized via the sharings / desktop, but OO still has the document in its cache and doesn't send the save.

So, we decided to add a new trigger that will ask the OO server to forcesave often to ensure a better synchronisation between OO and the Cozy.

### DIFF
--- a/model/office/trigger.go
+++ b/model/office/trigger.go
@@ -1,0 +1,111 @@
+package office
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"io"
+	"net/url"
+	"strings"
+
+	"github.com/cozy/cozy-stack/model/instance"
+	"github.com/cozy/cozy-stack/model/job"
+	jwt "github.com/golang-jwt/jwt/v4"
+	"github.com/labstack/echo/v4"
+)
+
+// SendSaveMessage is used by the trigger for asking OO to save the document in
+// the Cozy.
+type SendSaveMessage struct {
+	Key string `json:"key"`
+}
+
+func setupTrigger(inst *instance.Instance, key string) error {
+	sched := job.System()
+	infos := job.TriggerInfos{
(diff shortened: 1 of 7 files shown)


model/office/trigger.go
```diff
@@ -0,0 +1,111 @@
...
+		Type:       "@every",
+		WorkerType: "office-save",
+		Arguments:  "10m",
+	}
+	msg := &SendSaveMessage{Key: key}
+	t, err := job.NewTrigger(inst, infos, msg)
+	if err != nil {
+		return err
+	}
+	return sched.AddTrigger(t)
```
taratatach, line 35: Shouldn't we check if a trigger already exists to avoid creating multiple triggers in case a document is opened in multiple tabs?

model/office/trigger.go
```diff
@@ -0,0 +1,111 @@
...
+type commandResponse struct {
+	Key   string `json:"key"`
+	Error int    `json:"error"`
+}
+
+func SendSave(inst *instance.Instance, msg SendSaveMessage) error {
+	if _, err := GetStore().GetDoc(inst, msg.Key); err != nil {
+		// By returning the ErrBadTrigger code, the stack will know that it
+		// must delete the trigger.
+		return job.ErrBadTrigger{Err: err}
```
taratatach, line 57: Is this how the trigger will be removed once the user is done editing?

nono:
This PR has some strong drawbacks, and a new way has been prefered: https://github.com/cozy/cozy-stack/pull/4138

PR #3759, opened 2023-02-16
Fix the cozy-stack fix service-triggers command

nono:
This command was recreating missing triggers for app services, but the triggers were recreated without the service name in the message. So, we need to fix the script, and to add a way to also fix the triggers that miss the name.

### DIFF
--- a/web/instances/fixers.go
+++ b/web/instances/fixers.go
@@ -344,6 +344,7 @@ func serviceTriggersFixer(c echo.Context) error {
 
 	var toDelete []job.Trigger
 	recreated := 0
+	updated := 0
 
 	for slug, triggers := range byApps {
 		manifest, err := app.GetWebappBySlug(inst, slug)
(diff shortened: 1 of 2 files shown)


web/instances/fixers.go
```diff
@@ -381,27 +382,37 @@ func serviceTriggersFixer(c echo.Context) error {
...
+					if msg.Name == "" {
+						fixTriggerName(inst, trigger, msg, name)
+						updated++
+					}
+				}
 			}
 
 			if recreate {
-				triggerID, err := app.CreateServiceTrigger(inst, slug, service)
+				triggerID, err := app.CreateServiceTrigger(inst, slug, name, service)
```
taratatach, line 405: I don't understand why the service does not have a `name` here while the struct shows it should

taratatach (approved):

web/instances/fixers.go
nono: `name` is not exported as it starts with a lowercase letter, and as such, it is not filled when we unmarshal JSON. I was tricked too by this!

PR #3765, opened 2023-02-19
Cache refacto

Peltoche:
The `/pkg/cache/cache.go.Cache` struct seems to contains two separate
behavior based on the presence (or not) of a redis client.

If the redis client is present we are using the redis client, if not
we are using an in-memory implementation.

The goal of the next commits is to separate those two logic in two
separate implementations in order to increase the readability.

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -12,32 +12,55 @@ import (
 	"github.com/redis/go-redis/v9"
 )
 
+// Cache is a rudimentary key/value caching store backed by redis. It offers a
+// Get/Set interface as well a its gzip compressed alternative
+// GetCompressed/SetCompressed
+type Cache interface {
+	CheckStatus(ctx context.Context) (time.Duration, error)
+	Get(key string) ([]byte, bool)
+	MultiGet(keys []string) [][]byte
+	Keys(prefix string) []string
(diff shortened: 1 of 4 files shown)


pkg/cache/cache.go
```diff
@@ -1,173 +1,43 @@
...
 }
 
-// SetCompressed works like Set but compress the asset data before storing it.
-func (c Cache) SetCompressed(key string, data []byte, expiration time.Duration) {
-	dataCompressed := new(bytes.Buffer)
-	gw := gzip.NewWriter(dataCompressed)
-	defer gw.Close()
-	if _, err := io.Copy(gw, bytes.NewReader(data)); err != nil {
-		return
+// Init instantiate a Cache Client.
```
nono, line 32: I would prefer to call this function `New`. For me, `Init` is called for its side effect on the package, while `New` is called to return an object.

pkg/cache/cache.go
```diff
@@ -1,173 +1,43 @@
...
+//
+// The backend selection is done based on the `client` argument. If a client is
+// given, the redis backend is chosed, if nil is provided the inmemory backend would
```
nono, line 35: s/chosed/chosen/

nono (approved):

PR #3777, opened 2023-02-21
Split the current global lock function into two separate implementations

Peltoche:


model/app/fetcher_http.go
```diff
@@ -118,7 +118,7 @@ func fetchHTTP(src *url.URL, shasum []byte, fs appfs.Copier, man Manifest, prefi
 
 	// For the lock key, we use the checksum when available, and else, we
 	// fallback on the app name.
-	mu := lock.ReadWrite(prefixer.GlobalPrefixer, "app-"+man.Slug()+"-"+man.Checksum())
+	mu := config.GetConfig().Lock.ReadWrite(prefixer.GlobalPrefixer, "app-"+man.Slug()+"-"+man.Checksum())
```
nono, line 121: Maybe we can add an helper in the `config` package to make those lines shorter and less tedious to read:

```go
config.Lock().ReadWrite(...)
```

pkg/lock/lock.go
```diff
@@ -4,19 +4,28 @@ import (
...
 )
 
-// ReadWrite returns the read/write lock for the given name.
-// By convention, the name should be prefixed by the instance domain on which
-// it applies, then a slash and the package name (ie alice.example.net/vfs).
-func ReadWrite(db prefixer.Prefixer, name string) ErrorRWLocker {
-	cli := config.GetConfig().Lock.Client()
-	if cli != nil {
-		return getRedisReadWriteLock(cli, db.DBPrefix()+"/"+name)
+// LockGetter return a lock on a resource matching the given `name`.
```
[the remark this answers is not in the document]
Peltoche: Totally agree.

model/app/installer.go
```diff
@@ -275,7 +274,7 @@ func (i *Installer) run() (err error) {
 	if i.man == nil {
 		panic("Manifest is nil")
 	}
-	mu := lock.ReadWrite(i.db, "app-"+i.man.Slug())
+	mu := config.GetConfig().Lock.ReadWrite(i.db, "app-"+i.man.Slug())
```
nono, line 277: We still have `config.GetConfig().Lock` instead of the helper `config.Lock()` here

Peltoche: erf, forgotten to move all the calls indeed... Doing it now.

nono (approved):

PR #3778, opened 2023-02-21
Improve sharing logs

nono:
Adding some logs for bulk operations can be helpful for debugging sharing issues.

### DIFF
--- a/pkg/couchdb/bulk.go
+++ b/pkg/couchdb/bulk.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/cozy/cozy-stack/pkg/config/config"
+	"github.com/cozy/cozy-stack/pkg/couchdb/revision"
 	"github.com/cozy/cozy-stack/pkg/logger"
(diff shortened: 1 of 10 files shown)


pkg/couchdb/bulk.go
```diff
@@ -368,7 +371,47 @@ func BulkForceUpdateDocs(db prefixer.Prefixer, doctype string, docs []map[string
...
+	}
+
+	var messages []string
+	for len(extracted) > 0 {
+		nb := len(extracted)
+		// We limit the number of ids per log to avoid the line width limit.
+		if nb > 50 {
+			nb = 50
+		}
+		msg := strings.Join(extracted[:nb], " ")
```
taratatach, line 408: Maybe we could use another separator as there can be spaces in the extracted strings?!

taratatach (approved):

pkg/couchdb/bulk.go
nono: No, spaces is practical for searching in logs

PR #3779, opened 2023-02-22
Remove upper case in errors following go code review comments

Peltoche:
This review apply the advices found at:
https://github.com/golang/go/wiki/CodeReviewComments#error-strings

This review have been generated with by the following command:

```bash
sed -i 's/errors.New("\([A-Z]\)/errors.New("\L\1/g' **/*.go
sed -i 's/fmt.Errorf("\([A-Z]\)/fmt.Errorf("\L\1/g' **/*.go
````

### DIFF
--- a/client/apps.go
+++ b/client/apps.go
@@ -256,7 +256,7 @@ func makeAppsPath(appType, path string) string {
 	case consts.Konnectors:
 		return "/konnectors/" + path
 	}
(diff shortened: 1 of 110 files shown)


Peltoche:
To early to do that. There is too much tests checking the errors string.

nono:
And, we don't want to do that, as some error messages are returned by the API or the CLI, and as such, changing them will be a breaking change.

PR #3784, opened 2023-02-23
Move some errors from `/web/iodc` to the err1.13 format

Peltoche:


web/oidc/oidc.go
```diff
@@ -32,6 +32,13 @@ import (
 	"github.com/labstack/echo/v4"
 )
 
+var (
+	ErrInvalidToken         = errors.New("invalid token")
+	ErrInvalidConfiguration = errors.New("invalid configuration")
+	ErrAuthenticationFailed = errors.New("the authentication has failed")
```
nono, line 38: For this error, the string is used as a key for localization. So, we need to keep the same string, or update the locale files.

nono (changes requested):

web/oidc/oidc.go
Peltoche: ok, I fix this

Peltoche: I have updated the locals for `ErrAuthenticationFailed`. The others have not changed. Everything should be good

nono (approved):

PR #3785, opened 2023-02-23
Clean up the `/pkg/assets/dynamic.NewOsFS` function

Peltoche:
This PR bring several ameliorations:
- It restrict the OSFS access to only its own folder be setting up the root directory.
- It remove the dependency to the config package


nono:
There is a git conflict on this PR

nono (approved):

PR #3794, opened 2023-02-28
Optimize counting the number of files in a sharing

nono:
When the stack sends a request for creating a sharing, it includes the number of files in the sharing. Counting the number of files may have been slow because we were walking the VFS to count the files. We have optimized it by reusing a CouchDB view.

### DIFF
--- a/model/sharing/oauth.go
+++ b/model/sharing/oauth.go
@@ -16,7 +16,6 @@ import (
 	"github.com/cozy/cozy-stack/model/instance"
 	"github.com/cozy/cozy-stack/model/oauth"
 	"github.com/cozy/cozy-stack/model/permission"
-	"github.com/cozy/cozy-stack/model/vfs"
 	"github.com/cozy/cozy-stack/pkg/consts"
 	"github.com/cozy/cozy-stack/pkg/couchdb"
 	"github.com/cozy/cozy-stack/pkg/jsonapi"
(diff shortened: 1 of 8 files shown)


model/sharing/oauth.go
```diff
@@ -123,31 +122,34 @@ func (m *Member) CreateSharingRequest(inst *instance.Instance, s *Sharing, c *Cr
...
+					continue
+				}
+				if dir != nil {
+					if _, nb, err := fs.DirSizeAndCount(dir); err == nil {
+						count += nb
 					}
-					return nil
-				})
+				} else {
+					count += file.ByteSize
```
taratatach, line 143: I believe this should be `count++` since we're counting files and not their size.

model/vfs/couchdb_indexer.go
```diff
@@ -186,22 +193,31 @@ func (c *couchdbIndexer) DirSize(doc *DirDoc) (int64, error) {
...
 	if err != nil {
-		return 0, err
+		return 0, 0, err
 	}
 	if len(resp.Rows) == 0 {
-		return 0, nil
+		return 0, 0, nil
 	}
 
 	// Reduce of _sum should give us a number value
```
taratatach, line 202:
```suggestion
	// Reduce of _stats should give us `sum` and `count` fields with number values

taratatach (changes requested):

taratatach (approved):

PR #3796, opened 2023-03-01
Don't let the CI skip some tests

Peltoche:


nono (commented):
Oops, I forgot we had `-short` for unit-tests. I would like to keep the `make unit-tests`, even if it not used in CI and we can remove `-short` for it.

There are a bunch of tests failing with this PR, is it because of the removal of `-p 1` ?

Peltoche:
I don't really know where the errors come from at the moment. I will investigate this and will propose some fix.

About the `-short` and the `make unit-tests` comment, IMHO depending how we define unit/integration/E2E tests the two are coherent together. The issue is that we don't want the CI to run only the unit-tests but all the tests. 

This comment is not coherent with my review, I just think about it. I will update ma review.

Peltoche:
Some of the errors a due to unexisting instances. I think some review similar as https://github.com/cozy/cozy-stack/pull/3797/commits/cd583ab87e5353cd03d17f4951ee20df3c770c4e should help a lot

Peltoche:
@nono  The test inside the commit  115ad72e3ed2074a3e4d39ef61ef7ab553e85531 seems to bring an deadlock. More precisely you add a `wg.Add(1)` then you call `wg.Wait()` without any new `wg.Done()` so you waitgroup will be locked forever. 

I don't know well enough the business logic to propose a solution here.

nono:
-> https://github.com/cozy/cozy-stack/pull/3805

nono:
For `TestFiles/ThumbnailPDFs`, it looks like a configuration issue for ImageMagick. Cf https://stackoverflow.com/questions/52998331/imagemagick-security-policy-pdf-blocking-conversion

nono:
For `TestPermissions` (in `model/vfs`), we are using the same contexter for `makeAferoFS` and `makeSwiftFS`, and it doesn't work well because they are trying to use the same CouchDB databases.

For AferoFS, you can try:

```
	db := &contexter{0, "afero.testvfs.example.org", "afero.testvfs.example.org", "cozy_beta"}
```

and for SwiftFS:

```
	db := &contexter{0, "swift.testvfs.example.org", "swift.testvfs.example.org", "cozy_beta"}
```

Peltoche:
Damn, I was so near for the `TestPermissions`! I did try something like this but it doesn't seems to work. I will try again

Peltoche:
> For `TestFiles/ThumbnailPDFs`, it looks like a configuration issue for ImageMagick. Cf https://stackoverflow.com/questions/52998331/imagemagick-security-policy-pdf-blocking-conversion

This is kind of problematic. Soo I will tcheck if a migration to https://pkg.go.dev/github.com/davidbyttow/govips/v2#readme-a-lightning-fast-image-processing-and-resizing-library-for-go would be possible. This should reduce this kind of errors and should improve the perfs

nono:
Nah, I don't think using vips will make those kind of errors disappear. It won't be the same, but we will have new errors. Vips is not reading PDFs its-self, it is delegating this stuff to PDFium or ImageMagick.

Peltoche:
> Nah, I don't think using vips will make those kind of errors disappear. It won't be the same, but we will have new errors. Vips is not reading PDFs its-self, it is delegating this stuff to PDFium or ImageMagick.

At first glance it seems to be a separate project and quicker but I will make a documented issue to propose a migration.

nono:
I have fixed some tests for `model/job`: https://github.com/cozy/cozy-stack/pull/3822

Peltoche:
> time="2023-03-07T14:10:28Z" level=error msg="imagemagick failed: exit status 1" domain=testfiles_lxuaewjtfb.cozy.local file_id=a417223ecfd474d83fe16edbf525b03f job_id=a417223ecfd474d83fe16edbf525b722 nspace=jobs stderr="convert: no decode delegate for this image format `' @ error/constitute.c/ReadImage/746.\nconvert: no images defined `jpg:-' @ error/convert.c/ConvertImageCommand/3342.\n" worker_id=thumbnail/1

I can't find a way to make imagemagick works, I have tried to reinstall or even install from scratch but it seems that a side library is missing but can't find which one. 

If this is the only test failing I propose to skip it and handle it in a new PR in order to enable the CI to tests properly all the other PR.

>  time="2023-03-07T14:07:33Z" level=error msg="Could not commit installer process: manifest types are not the sames: expected 2, got 1. Are you sure of cozy-bad-type type ? (konnector/webapp)" domain= nspace=apps slug=cozy-bad-type type=install version_start=
panic: failed start the git server (output: ""): signal: interrupt

This one I don't know how to fix it at all...


That's almose the end :muscle:

nono:
I fix the issue with `git server` on https://github.com/cozy/cozy-stack/pull/3827, but `model/app` tests are failing later.

web/files/files_test.go
```diff
@@ -3099,7 +3099,7 @@ func TestFiles(t *testing.T) {
 		// Wait for tiny thumbnail generation
 		time.Sleep(time.Second)
 
-		tiny := links.Value("tiny").String().NotEmpty().Raw()
+		// tiny := links.Value("tiny").String().NotEmpty().Raw()
```
nono, line 3102: Shouldn't we call `t.Skip("TODO: Fix those tests")` instead of commenting out the code?

Peltoche: After re-reading the doc on `t.Skip` it seems a good idea indeed. I was thinking that it would skip all the test and don't report previous errors but it's not the case. 

I change this

Peltoche:
>  panic: failed to run the git server (output: "[4774] Could not bind to 0.0.0.0: Address already in use\n[4774] Could not bind to ::: Address already in use\nfatal: unable to allocate any listen sockets on port 9418\n"): exit status 128

Seems that we have the two `Go tests` jobs battling for the port `4774`. I will try to find a way to start them on different ports.

nono:
> Seems that we have the two Go tests jobs battling for the port 4774. I will try to find a way to start them on different ports.

I think it is the conflict for port 4774 happens in the same Go tests suite. My theory is that the git daemon started for konnectors tests may be stopped after the git daemon for webapps tests is started. But, finding an available port works too in this case.

model/app/installer_test.go
```diff
@@ -132,3 +138,16 @@ git checkout master`
 		t.Log("did upgrade", localVersion)
 	}
 }
+
+func findAvailablePort(t *testing.T) string {
+	t.Helper()
+
+	// ":0" means: find my any available port
```
nono, line 145: [nit] s/my// (or s/my/me/)

model/app/installer_test.go
```diff
@@ -71,8 +74,8 @@ func manifestKonnector() string {
 }`
 }
 
-func serveGitRep(dir string) context.CancelFunc {
-	localGitDir = dir
+func serveGitRep(t *testing.T) (context.CancelFunc, string) {
```
nono, line 77: [nit] I would have swapped the returned variables: string first, context.CancelFunc second.

nono (approved):

PR #3799, opened 2023-03-02
Add some cache to the CI job

Peltoche:
This cache option follow the official github documentation: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-go#caching-dependencies

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,18 +48,13 @@ jobs:
           EOF
           DEBIAN_FRONTEND=noninteractive sudo apt-get install -y --force-yes couchdb=${{ matrix.couchdb-version }}*
           echo "COZY_COUCHDB_URL=http://admin:password@localhost:5984/" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


nono:
GitHub Actions doesn't look happy by your change ;-)

Peltoche:
> GitHub Actions doesn't look happy by your change ;-)

Fixed! The `setup-go`  step handle the `go install` (with a cache now) but as we checkout the file after the `go.sum` was missing. I have move the `checkout` step before the `setup-go` and remove the manual install and we are good to go.

nono:
No, we are not good. The jobs were not launched:

```
The workflow is not valid. .github/workflows/ci.yml (Line: 53, Col: 9): 'uses' is already defined
```

Peltoche:
Arg! Indeed... A `-` was missing...

nono:
LGTM, but waiting for #3804, as it seems to be required before merging

nono (approved):

PR #3800, opened 2023-03-02
Upload the coverage to codecove after the tests

### LINKED ISSUE #3798: Discussion: Code coverage integration?
Hi @nono 

It would be great to have some coverage information. This could help us to know if a part of the codebase need more tests. 

What do you think of adding a coverage integration like https://about.codecov.io/?

Peltoche:
This review is based on the following documentation: https://about.codecov.io/blog/getting-started-with-code-coverage-for-golang/

It also add a check for dataraces

Fix #3798

### DIFF
--- a/.github/workflows/go-tests.yml
+++ b/.github/workflows/go-tests.yml
@@ -63,6 +63,8 @@ jobs:
           go-version: ${{ matrix.go-version }}
           cache: true
 
-      - name: Unit tests
-        run: |
-          go test -p 1 -timeout 15m -cover ./...
+      - name: Run tests
+        run: go test -p 1 -timeout 5m -coverprofile=coverage.out -covermode=atomic ./...
+
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v3


nono:
It looks like we have some data races detected

Peltoche:
Arg! Indeed... A `-` was missing...

Peltoche:
Yep, I have remove the `-race` flag and put back the `-p 1` flag in order to keep the job as similar as possible. I will try to add the `-race` flag in an another PR.

Peltoche:
And now we fall back on the same errors than #3796. We will need to merge it before this one

nono:
By the way, I've just seen a new post about code coverage: https://go.dev/blog/integration-test-coverage (not yet read).

Peltoche:
> By the way, I've just seen a new post about code coverage: https://go.dev/blog/integration-test-coverage (not yet read).

Yep, seen it. Not sure that's useful for us. Maybe for the Ruby tests but I don't know how the coverage tool is able to merge two separate coverage reports.

nono (approved):

PR #3805, opened 2023-03-02
Fix worker/exec tests

nono:
Some tests in worker/exec were not running on the CI, because of -short flag, and we missed that they were failing. Let's fix them!

### DIFF
--- a/worker/exec/konnector_test.go
+++ b/worker/exec/konnector_test.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path"
 	"strings"
 	"sync"
 	"testing"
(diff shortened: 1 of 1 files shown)


worker/exec/konnector_test.go
```diff
@@ -162,7 +167,7 @@ echo "{\"type\": \"manifest\", \"message\": \"$(ls ${1}/manifest.konnector)\" }"
 			assert.Equal(t, "manifest", doc2.M["type"])
 
 			msg2 := doc2.M["message"].(string)
-			assert.True(t, strings.HasPrefix(msg2, t.TempDir()))
+			assert.True(t, strings.HasPrefix(msg2, "/tmp"))
```
Peltoche, line 170: You can use `os.TempDir()` instead. It was the value before the (incorrect) migration to `t.TempDir()`

worker/exec/konnector_test.go
```diff
@@ -410,6 +431,20 @@ echo "{\"type\": \"toto\", \"message\": \"COZY_URL=${COZY_URL}\"}"
 		var updatedAcc account.Account
 		err = couchdb.GetDoc(inst, consts.Accounts, acc.ID(), &updatedAcc)
 		require.NoError(t, err)
-		assert.Equal(t, updatedAcc.DefaultFolderPath, "/Administrative/my-konnector-1/account-1")
+		assert.Equal(t, updatedAcc.DefaultFolderPath, "/Administrative/Trainline/account-1")
 	})
 }
+
+func loadLocale() error {
```
Peltoche, line 438: You can take `t *testing.T` as an argument and mark the function as an helper with `t.Helper()`

taratatach (approved):

worker/exec/konnector_test.go
```diff
@@ -410,6 +431,20 @@ echo "{\"type\": \"toto\", \"message\": \"COZY_URL=${COZY_URL}\"}"
...
+	locale := consts.DefaultLocale
+	assetsPath := config.GetConfig().Assets
+	if assetsPath != "" {
+		pofile := path.Join("../..", assetsPath, "locales", locale+".po")
+		po, err := os.ReadFile(pofile)
+		if err != nil {
+			return fmt.Errorf("Can't load the po file for %s", locale)
+		}
```
Peltoche, line 446: And then failing directly instead of returning an error. This would simplify a little the test readability and could allow you to use `t.Cleanup` if you need to clean some stuff

PR #3802, opened 2023-03-02
Add the datarace detection to the CI tests

Peltoche:

PR #3806, opened 2023-03-02
Fix some dataraces inside `/model/job`

Peltoche:
Fix some issues from #3802


model/job/mem_scheduler.go
```diff
@@ -24,7 +24,7 @@ type memScheduler struct {
 
 	ts    map[string]Trigger
 	thumb *ThumbnailTrigger
-	mu    sync.RWMutex
+	mu    *sync.RWMutex
```
nono, line 27: I would prefer not to make it a pointer

Peltoche: ok, fix it

nono (approved):

PR #3821, opened 2023-03-07
Move the vault logic into a new package

Peltoche:
Previously the vault logic was living inside the `/pkg/config` package in a global alongs with other config stuff.

This PR goal is to make a clear separation between the vault logic, where some important secrets are handled and the config logic. It also remove the global variable and move to the same initialization logic than the cache or lock package: It create an object and put it into the `Config` object.

### DIFF
--- a/model/account/credentials.go
+++ b/model/account/credentials.go
@@ -63,7 +63,7 @@ func EncryptCredentialsWithKey(encryptorKey *keymgmt.NACLKey, login, password st
 // EncryptCredentialsData takes any json encodable data and encode and encrypts
 // it using the vault public key.
 func EncryptCredentialsData(data interface{}) (string, error) {
-	encryptorKey := config.GetVault().CredentialsEncryptorKey()
+	encryptorKey := config.GetKeyring().CredentialsEncryptorKey()
 	if encryptorKey == nil {
 		return "", errCannotEncrypt
 	}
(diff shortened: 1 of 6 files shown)


nono:
I'm a bit annoyed by this change. The name "vault" is more often used for the bitwarden vault, not this object. It's not your fault, the confusion was already there, but promoting this config part as a new "vault" package that is not the bitwarden vault will increase confusion. I think we should find another name for this thing.

nono:
I don't find a good suggestion. Maybe a key ring, but it's two words and I'm not convinced.

Peltoche:
Totally agree with you. Personaly I find the package name `keyring` totally ok. It could take even more sens if we merge the `keymgmt` package inside this one.

`keymgmt` is only used by the vault and inside `/model/account` for the credentials encryption but I think it would make sens if the `account` package use the new `keyring` in order to fetch its encryption/decryption key.

Peltoche:
I see that the package is the only `/pkg/keymgmt` user except for two function used to encrypt/decrypt credentials in `model/account`. Should I merge the `keymgmt` package inside `keyring` and make the `model/account` package use the `keyring` package? 

An another possibility would be to move the NACL into the `crypto` package.  What do you think of it?

nono (approved):

PR #3825, opened 2023-03-07
Remove some restrictive ImageMagick policies for the PDF generation

Peltoche:
https://stackoverflow.com/questions/52998331/imagemagick-security-policy-pdf-blocking-conversion

### DIFF
--- a/.github/workflows/go-tests.yml
+++ b/.github/workflows/go-tests.yml
@@ -47,6 +47,12 @@ jobs:
           EOF
           DEBIAN_FRONTEND=noninteractive sudo apt-get install -y --force-yes couchdb=${{ matrix.couchdb-version }}*
           echo "COZY_COUCHDB_URL=http://admin:password@localhost:5984/" >> $GITHUB_ENV
+      
+        # This step is required to let ImageMagick generate PDF files.
+      - name: Fix ImageMagick config
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/go-tests.yml
```diff
@@ -47,6 +47,12 @@ jobs:
           EOF
           DEBIAN_FRONTEND=noninteractive sudo apt-get install -y --force-yes couchdb=${{ matrix.couchdb-version }}*
           echo "COZY_COUCHDB_URL=http://admin:password@localhost:5984/" >> $GITHUB_ENV
+      
+        # This step is required to let ImageMagick generate PDF files.
+      - name: Fix ImageMagick config
+        run: |
+          sudo sed -i '/disable ghostscript format types/,+6d' /etc/ImageMagick-6/policy.xml
+          cat /etc/ImageMagick-6/policy.xml
```
nono, line 55: I think the `cat` here is for debugging, right?

Peltoche: Yep, this is still a WIP but I can't find a way to move back a PR into a draft.

BTW it seems that there is some other libs missing. I also test the commit on #3796 in order to check what it does when the tests are running

Peltoche: Just found the button! :confetti_ball:

PR #3835, opened 2023-03-10
feat: Add Hourly trigger

Crash--:
In order to be able to run a trigger hourly.
It was already possible to do it from the manifest but not from the API.

### DIFF
--- a/model/job/scheduler.go
+++ b/model/job/scheduler.go
@@ -141,6 +141,8 @@ func fromTriggerInfos(infos *TriggerInfos) (Trigger, error) {
 		return NewAtTrigger(infos)
 	case "@in":
 		return NewInTrigger(infos)
+	case "@hourly":
+		return NewHourlyTrigger(infos)
 	case "@daily":
 		return NewDailyTrigger(infos)
 	case "@weekly":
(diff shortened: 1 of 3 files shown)


model/job/scheduler.go
```diff
@@ -141,6 +141,8 @@ func fromTriggerInfos(infos *TriggerInfos) (Trigger, error) {
 		return NewAtTrigger(infos)
 	case "@in":
 		return NewInTrigger(infos)
+	case "hourly":
```
nono, line 144: s/hourly/@hourly/

nono (commented):
Do you have a use case for `@hourly` trigger?

I think it wasn't added on purpose to avoid encouraging triggers with such frequency, IIRC.

Crash--:
We have a few `hourly` trigger in a few konnectors (https://github.com/search?q=org%3Akonnectors%20hourly&type=code). 

The idea is to remove a lot of code on the front app (harvest / cozy-client / flagship) in order to avoid creating cron but use this randomness directly from the stack (https://github.com/cozy/cozy-client/pull/1325). 

But I agree, I'm not sure we should continue to have an hourly trigger. Maybe we should change the frequency on those 2 konnectors.

nono (approved):

PR #3836, opened 2023-03-11
Move the globals from `pkg/limits.rate_limiting` to a struct with met…

Peltoche:
…hods

This allow to move the global inside a scoped struct and setup the Counter once at startup.

### DIFF
--- a/pkg/config/config/config.go
+++ b/pkg/config/config/config.go
@@ -23,6 +23,7 @@ import (
 	"github.com/cozy/cozy-stack/pkg/cache"
 	build "github.com/cozy/cozy-stack/pkg/config"
 	"github.com/cozy/cozy-stack/pkg/keyring"
+	"github.com/cozy/cozy-stack/pkg/limits"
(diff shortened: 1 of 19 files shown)


pkg/config/config/config.go
```diff
@@ -808,13 +815,13 @@ func UseViper(v *viper.Viper) error {
...
-		OauthStateStorage:   oauthStateRedis,
-		Realtime:            realtimeRedis,
-		CacheStorage:        cacheStorage,
+		Lock:              lock.New(lockRedis.Client()),
+		SessionStorage:    sessionsRedis,
+		DownloadStorage:   downloadRedis,
+		Limiter:           limits.NewRateLimiter(rateLimitingRedis.Client()),
+		OauthStateStorage: oauthStateRedis,
+		Realtime:          realtimeRedis,
+		CacheStorage:      cache.New(cacheRedis.Client()),
```
nono, line 824: Why the change on `CacheStorage`?

nono (approved):

PR #3845, opened 2023-03-13
Remove the `dyn-assets` folders

Peltoche:
Those folder are created by the `/pkg/assets` package during some tests


Peltoche:
Require #3847

PR #3850, opened 2023-03-15
Add delegated code for OIDC in the flagship app

nono:


web/oidc/oidc.go
```diff
@@ -245,34 +245,49 @@ func createSessionAndRedirect(c echo.Context, inst *instance.Instance, redirect,
 // a valid token for OIDC.
 func AccessToken(c echo.Context) error {
 	inst := middlewares.GetInstance(c)
-	conf, err := getGenericConfig(inst.ContextName)
-	if err != nil || !conf.AllowOAuthToken {
```
taratatach, line 249: Why don't we need to check if the instance's context allows the creation of OAuth tokens when using a delegated code?

web/oidc/oidc.go
```diff
@@ -343,6 +343,11 @@ func AccessToken(c echo.Context) error {
 			"error": "invalid scope",
 		})
 	}
+	if out.Scope == "*" {
+		if !client.Flagship || !client.CertifiedFromStore {
```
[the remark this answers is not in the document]
nono: Nope, but I have removed `|| !client.CertifiedFromStore` here

web/oidc/oidc.go
nono: I don't see a good reason to not allow the flagship app to work for a context (which would be the consequence of always checking AllowOAuthToken here). But, when an `access_token` or `id_token` is given to this endpoint, it weaken security. It is acceptable when some conditions are met (like the instance URL cannot be guessed), but it should really be something we need to activate per context after checking those conditions.

PR #3858, opened 2023-03-22
Improve views and indexes update

nono:
When a Cozy was created before the set of indexes or views for an instance was last changed, it has a different IndexViewsVersion. This number can be used by the stack when a Cozy is used to check if the views and indexes must be updated.

Before this commit, the update process was the same as the creation process. We iterate on all views and indexes, trying to create them, and if it fails, look if the view/index exists and has the right definition, and if necessary update it.

With this commit, the update tries to opmitize this process, by first looking what views and indexes exists, and only try to create/update it when missing or not the right definition. It should help to lower the load on the CouchDB cluster when we release a stack with an increased IndexViewsVersion.

### DIFF
--- a/pkg/couchdb/views.go
+++ b/pkg/couchdb/views.go
@@ -1,0 +1,290 @@
+package couchdb
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/url"
+	"reflect"
+	"time"
+
+	"github.com/cozy/cozy-stack/pkg/couchdb/mango"
+	"github.com/cozy/cozy-stack/pkg/logger"
+	"github.com/cozy/cozy-stack/pkg/prefixer"
+	"golang.org/x/sync/errgroup"
+)
+
+// View is the map/reduce thing in CouchDB
+type View struct {
+	Name    string      `json:"-"`
+	Doctype string      `json:"-"`
+	Map     interface{} `json:"map"`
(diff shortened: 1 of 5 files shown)


nono:
Waiting for https://issues.apache.org/jira/browse/INFRA-24371 for the CI...

pkg/couchdb/views.go
```diff
@@ -0,0 +1,290 @@
...
+		view := views[i]
+		ddoc := &DesignDoc{
+			ID:    "_design/" + view.Name,
+			Lang:  "javascript",
+			Views: map[string]*View{view.Name: view},
+		}
+		exists := false
+		for _, old := range ddocsByDoctype[view.Doctype] {
+			if equalViews(old, ddoc) {
+				exists = true
```
taratatach, line 139: nit: we could break here

pkg/couchdb/views.go
```diff
@@ -0,0 +1,290 @@
...
+				Reduce: "_count",
+				Options: map[string]interface{}{
+					"def": map[string]interface{}{
+						"fields": defFields,
+					},
+				},
+			}
+			ddoc.Views = map[string]*View{name: view}
+			if equalViews(old, ddoc) {
+				exists = true
```
taratatach, line 185: Same here.

taratatach (approved):

PR #3862, opened 2023-03-29
fix: Delete triggers for `client` trigger

Crash--:
When we delete an account related to a client side konnector or when we uninstall a client side konnector, we want the stack to delete orphaned triggers.

So we have to not ignore `client` trigger.

### DIFF
--- a/model/account/accounts.go
+++ b/model/account/accounts.go
@@ -142,7 +142,7 @@ func GetTriggers(jobsSystem job.JobSystem, db prefixer.Prefixer, accountID strin
 
 	var toDelete []job.Trigger
 	for _, t := range triggers {
-		if t.Infos().WorkerType != "konnector" {
+		if !(t.Infos().WorkerType == "konnector" || t.Infos().WorkerType == "client") {
 			continue
 		}
 


Crash--:
@doubleface can you test it?

acezard (approved):

Crash--:
@nono I didn't have the time to check the remaining error " Konnector failure: jobs: abort ". Do you want to take the subject on this PR? Or can we merge it and create a new one later?

nono:
I don't have any context for this PR. Maybe you can ask @taratatach to look at it

Crash--:
I tried to explain our context in the PR description, but it seems that is was not enough, let's restart: 

As part of CLISK (client side konnector), when I uninstall a konnector or delete an account, the associated trigger was not deleted.

For a server-side Konnector, the stack takes care of going through orphaned triggers aka triggers whose account no longer exists to delete them. Either when deleting an account or uninstalling a konnector. 

Initially, I thought of making the client delete the associated trigger when deleting the account, but then came the case of uninstalling the konnector. And this uninstallation can be done via web / store, I can't tell them to manage client-side konnectors and how to handle them. 

So, I propose to manage orphaned triggers that come from CLISK or server-side konnector, in the same way.

This works well, it's confirmed, by @doubleface. 

However, during the test of uninstalling a CLISK, we realized that even before this PR we have the error "Konnector failure: jobs: abort ". And this error is something I don't have time to look into.

I'll ask @taratatach to take a look at this issue

model/account/accounts.go
```diff
@@ -140,7 +140,7 @@ func GetTriggers(jobsSystem job.JobSystem, db prefixer.Prefixer, accountID strin
 
 	var toDelete []job.Trigger
 	for _, t := range triggers {
-		if t.Infos().WorkerType != "konnector" {
+		if !(t.Infos().WorkerType == "konnector" || t.Infos().WorkerType == "client") {
```
taratatach, line 143: nit: I find this kind of condition particularly hard to read as the exclamation mark can be easily missed here.
Would you be okay with using `!= && !=` instead?

Crash--: My brain works differently, I've hard time to understand `!= && !=` but you're in charge here, you do whatever you want :). I'm not maintaining this code, so if it's easier for you to change, let's change it :)

Crash--:
@taratatach Since this commit fixes a pretty serious problem we have, I wonder if we shouldn't merge it, so we don't continue to pollute any cozy with bad data, and we don't have to behave badly because of it. 

The error during the uninstall does not cause any side effect today, as far as I know. It seems to be "just" a warning.

model/account/accounts.go
taratatach: OK :). I would prefer that then.

```suggestion
		if t.Infos().WorkerType != "konnector" && t.Infos().WorkerType != "client" {
```

taratatach (approved):

taratatach:
The suspicious logs issue is fixed by https://github.com/cozy/cozy-stack/pull/3886

model/account/accounts.go
taratatach: This will be taken care of in https://github.com/cozy/cozy-stack/pull/3887 instead.

PR #3863, opened 2023-03-30
Add support for XDG Base Directory specification for the config file

Peltoche:
As precised in https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html the config files should be hosted inside the folder `$XDG_CONFIG_HOME`.

In order to allow a retro-compatibility with the previous path we use this file only when it's available.

### DIFF
--- a/pkg/config/config/config.go
+++ b/pkg/config/config/config.go
@@ -46,6 +46,8 @@ var Paths = []string{
 	".",
 	".cozy",
 	"$HOME/.cozy",
+	"$HOME/.config/cozy",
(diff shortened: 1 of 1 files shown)


nono (changes requested):
If we want to support XDG, we should do it correctly. In particular:

> If $XDG_CONFIG_HOME is either not set or empty, a default equal to $HOME/.config should be used.

Peltoche:
Fixed

nono (approved):

PR #3871, opened 2023-04-04
Fix keyring initialization

nono:
The keyring is initialized in the pkg/config/config package. It was initialized a second time in the model/stack package, but with the stub implementation. It was a mistake!

Cf https://github.com/cozy/cozy-stack/pull/3821


taratatach (commented):
Is it a good thing to keep this call to the stub in `pkg/keyring/service.go` ? https://github.com/Peltoche/cozy-stack/blob/master/pkg/keyring/service.go#L38

nono:
Yes, it is useful when developing with a local stack

taratatach (approved):

PR #3886, opened 2023-04-20
fix: Push OnDeleteAccount job only if needed

taratatach:
Konnectors can declare in their manifest a script to be executed when
associated `io.cozy.accounts` documents are deleted.

However, when deleting an account, `cozy-stack` would push a job to
run an OnDeleteAccount script even when the konnector had not declared
one.
This would not prevent the account deletion from succeeding but we'd
execute a job unecessarily and add noise to our logs since the job
would fail.

We'll now check if an OnDeleteAccount script is declared for the
konnector associated with the deleted account before pushing a job.


model/account/accounts.go
```diff
@@ -331,8 +332,17 @@ func init() {
 
 			createSoftDeletedAccount(db, old)
 
-			_, err = PushAccountDeletedJob(jobsSystem, db, old.ID(), old.Rev(), konnector)
-			return err
+			// Execute the OnDeleteAccount if the konnector has declared one
+			man, err := app.GetKonnectorBySlug(db, konnector)
+			if err != nil {
```
nono, line 337: Could we have a case where we delete an account for a konnector that has already been uninstalled? I don't see one, but I'm not confident on that.

nono (approved):

PR #3892, opened 2023-04-26
Fix adding a contact by Cozy URL to a sharing with delegation

nono:
If Alice creates a sharing with Bob in read/write mode, and with the open_sharing set tot true, Bob can add recipients to this sharing. If Bob has a contact with a Cozy URL but no email, adding this contact to the sharing was failing. Bob's Cozy was trying to send the invitation, but it can't create the shortcut as it requires the Xor key for this new recipient (only available on Alice's Cozy), and it can't send an email as the email address is not known. We have fixed this issue by making the owner's Cozy creating the shortcut in that case.

### DIFF
--- a/web/sharings/sharings.go
+++ b/web/sharings/sharings.go
@@ -363,7 +363,21 @@ func AddRecipientsDelegated(c echo.Context) error {
 				} else {
 					states[email] = state
 				}
+
+				// If we have an URL for the Cozy, we can create a shortcut as an invitation
+				if cozy != "" {
+					var perms *permission.Permission
+					if s.PreviewPath != "" {
+						if perms, err = s.CreatePreviewPermissions(inst); err != nil {
+							return wrapErrors(err)
+						}
+					}
+					if err = s.SendInvitations(inst, perms); err != nil {
+						return wrapErrors(err)
+					}
+				}
 			}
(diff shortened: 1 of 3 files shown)


web/sharings/sharings.go
```diff
@@ -363,7 +363,21 @@ func AddRecipientsDelegated(c echo.Context) error {
 				} else {
 					states[email] = state
 				}
+
+				// If we have an URL for the Cozy, we can create a shortcut as an invitation
+				if cozy != "" {
```
taratatach, line 368: nit: Maybe we should move this block before updating `states` to avoid overriding `states[""]` in case we have multiple recipients without emails.
It might not be used today but it could become the source of bugs if we ever care about it.

We can even move the block before even trying to get the state?!

taratatach (approved):

web/sharings/sharings.go
nono: > Maybe we should move this block before updating states to avoid overriding states[""] in case we have multiple recipients without emails.

It doesn't change anything, it would still update the `states` after the block (no return for the success case). And, we are using the instance URL as key when the URL is missing, so it's not an issue when there are multiple recipients without emails.

PR #3900, opened 2023-05-03
Create the `Logger` abstraction around the use of logrus

Peltoche:
This interface aims to bring two main advantages:

- This open the possibility to have several different implementation
  for a logger and switch between them when needed. For example we will
  be able to create:
    - A logger for the tests based on `testing.T.Logf` which will attach
  the logs to a given test and improve the errors readability.
    - A NoOp logger logging nothing when we don't want to polute the tests
  with useless logger. We can a
    - A mock logger when we need to test whate have been send to the
    logger.

- It allow to standardise and greatly reduce the number of available methods
  for the loggers. We don't have access to unwanted methods from the lib.

### DIFF
--- a/model/job/worker.go
+++ b/model/job/worker.go
@@ -88,7 +88,7 @@ type (
 		context.Context
 		Instance *instance.Instance
 		job      *Job
-		log      *logger.Entry
+		log      logger.Logger
 		id       string
 		cookie   interface{}
 		noRetry  bool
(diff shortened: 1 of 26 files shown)


model/job/worker.go
```diff
@@ -114,16 +114,18 @@ func (w *WorkerConfig) Clone() *WorkerConfig {
...
+	entry := logger.WithDomain(job.Domain).WithNamespace("jobs")
 
 	if job.ForwardLogs {
 		hook := realtime.LogHook(job, realtime.GetHub(), consts.Jobs, job.ID())
-		log.AddHook(hook)
+		entry.AddHook(hook)
 	}
 
+	log := entry.
+		WithNamespace(job.WorkerType).
```
nono, line 125: I don't think we want to override the namespace set a few lines before (`WithNamespace("jobs")`)

web/apps/apps.go
```diff
@@ -259,21 +258,23 @@ func logsHandler(appType consts.AppType) echo.HandlerFunc {
...
-			level, err := logrus.ParseLevel(log.Level)
+			level, err := logger.ParseLevel(log.Level)
 			if err != nil {
 				return jsonapi.InvalidAttribute("level", err)
 			}
-			logger.WithTime(log.Time).Log(level, log.Msg)
+
+			l := l.WithTime(log.Time).
+				WithField("slug", slug).
+				WithField("job_id", c.QueryParam("job_id"))
```
nono, line 271: Why don't you keep setting the slug, job_id, and version outside of the loop?

PR #3902, opened 2023-05-03
Prepare work on B2B customizations

nono:


taratatach (approved):
What's the idea behind the addition of the `cirrus` class?

EDIT: OK, I've seen the theme changes that go with it but I don't get the name :)

nono:
The name of the class comes from `cirrus.css` (https://github.com/cozy/cozy-stack/blob/master/assets/styles/cirrus.css), the set of CSS styles specific to the stack. Cirrus is the name of some clouds (https://en.wikipedia.org/wiki/Cirrus_cloud).

PR #3903, opened 2023-05-03
Handle the IPv4/IPv6 dual stack mode and support several host

Peltoche:


web/server.go
```diff
@@ -139,38 +145,81 @@ func ListenAndServe() (*Servers, error) {
...
+				Addr:              addr,
+				ReadHeaderTimeout: ReadHeaderTimeout,
+			})
+		}(addr)
+	}
+
+	s.servers = append(s.servers, e)
+
+	// Need some time to start the goroutines.
+	time.Sleep(50 * time.Millisecond)
```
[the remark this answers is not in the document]
Peltoche: This can avoid panic when calling `servers.Shutdown` before the last server is started. This is a case I have encountered inside the tests for example. 

That's really not a "must have" and I can move the sleeps back inside the tests. I'm not a fan of the sleep solution either.

nono:
I have found why the integration tests are failing: echo only accepts a single listener per `Echo` object. Cf https://github.com/labstack/echo/blob/master/echo.go#L782-L786. The second call to `StartServer` is silently ignored.

web/server.go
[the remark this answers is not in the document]
Peltoche: Correction: This is not necessary anymore, the panic come from the situation where we consume the first event in `servers.errs` by calling `servers.Wait()` then we call directly `Shutdown` and we never consume the remaining errors events when the other servers are closing. As I tried to close the channel and a server tried to push an event at the same time, it panics. It will solve this in another PR

Peltoche:
> I have found why the integration tests are failing: echo only accepts a single listener per `Echo` object. Cf https://github.com/labstack/echo/blob/master/echo.go#L782-L786. The second call to `StartServer` is silently ignored.

Erf, I did ask myself about that too at the beginning but it seems to work well so... I don't understand why it seems for me locally.

Peltoche:
```
func (e *Echo) StartServer(s *http.Server) (err error) {
	e.startupMutex.Lock()
	if err := e.configureServer(s); err != nil {
		e.startupMutex.Unlock()
		return err
	}
	if s.TLSConfig != nil {
		e.startupMutex.Unlock()
		return s.Serve(e.TLSListener)
	}
	e.startupMutex.Unlock()
	return s.Serve(e.Listener)
}
```

That's a datarace. if the first goroutine have enought time to run the list line (`s.Serve(e.Listener)`) everything goes well. If not `e.Listener` will be overriden by the second gorouting  (the piece of code you have shown above).

cozy.example.yaml
```diff
@@ -11,6 +11,15 @@
 # variable. The template is evaluated at startup of the stack.
 
 # server host - flags: --host
+#
+# Specified on which ip/domain the server will listen to. It also define which
+# IP version will be available (IPv4 and/or IPv6). The server will be available
+# only on the IP version you have defined in host. For domains it will favorise
+# the IPv4 version if available.
+#
+# There is two values with custome behavious:
```
taratatach, line 20:
```suggestion
# There are two values with custom behaviour:
```

cozy.example.yaml
```diff
@@ -11,6 +11,15 @@
...
+# - "localhost" -> will listen both on "127.0.0.1" on IPv4 and "[::1]" on IPv6
```
taratatach, line 21:
```suggestion
# - "localhost" -> will listen on both "127.0.0.1" on IPv4 and "[::1]" on IPv6
```

cozy.example.yaml
```diff
@@ -11,6 +11,15 @@
...
+# - "0.0.0.0" and "[::] -> will listen both on "0.0.0.0" on IPv4 and "[::]" on IPv6
```
taratatach, line 22:
```suggestion
# - "0.0.0.0" and "[::]" -> will listen on both "0.0.0.0" on IPv4 and "[::]" on IPv6
```

taratatach (approved):

PR #3904, opened 2023-05-04
Send new registration mail for flagship app

nono:


web/auth/magic_link.go
```diff
@@ -204,6 +204,12 @@ func magicLinkFlagship(c echo.Context) error {
 		client.ClientID = client.CouchID
 	}
 
+	if err := session.SendNewRegistrationNotification(inst, client.ClientID); err != nil {
+		return c.JSON(http.StatusInternalServerError, echo.Map{
+			"error": err.Error(),
+		})
+	}
+
```
taratatach, line 212: Does it make sense to send this new registration notification by e-mail since the magic link itself is sent by e-mail?

web/oidc/oidc.go
```diff
@@ -358,6 +358,12 @@ func AccessToken(c echo.Context) error {
 		client.ClientID = client.CouchID
 	}
 
+	if err := session.SendNewRegistrationNotification(inst, client.ClientID); err != nil {
+		return c.JSON(http.StatusInternalServerError, echo.Map{
+			"error": err.Error(),
+		})
+	}
+
```
taratatach, line 366: Is this access token generated only once, when authenticating with OIDC?

Peltoche (commented):
LGTM except the comment from @taratatach

web/auth/magic_link.go
nono: We are OK with that.

PR #3905, opened 2023-05-04
Add a route to get the note as raw text

nono:


model/note/note_test.go
```diff
@@ -27,3 +27,35 @@ foobar **bold**
...
+- bar
+- baz
+`
+
+	expected := `My title
+
+hello world
+
+- foo
+
```
taratatach, line 46: Any idea why we end up with these empty lines in lists?

model/note/text.go
```diff
@@ -0,0 +1,104 @@
...
+			state.Text(*node.Text)
+		},
+		"bulletList": func(state *markdown.SerializerState, node, _parent *model.Node, _index int) {
+			state.RenderList(node, "", func(_ int) string { return "- " })
+		},
+		"orderedList": func(state *markdown.SerializerState, node, _parent *model.Node, _index int) {
+			state.RenderList(node, "", func(i int) string { return fmt.Sprintf("%d.", i+1) })
+		},
+		"listItem": func(state *markdown.SerializerState, node, _parent *model.Node, _index int) {
+			state.RenderContent(node)
```
taratatach, line 28: Maybe we should use `state.RenderInline()` to avoid empty lines in lists?

model/note/text.go
```diff
@@ -0,0 +1,104 @@
...
+		},
+		"rule": func(state *markdown.SerializerState, node, _parent *model.Node, _index int) {
+			state.Write("---")
+			state.CloseBlock(node)
+		},
+		"hardBreak": func(state *markdown.SerializerState, node, _parent *model.Node, _index int) {
+			state.Write("\n")
+		},
+		"image": func(state *markdown.SerializerState, node, _parent *model.Node, _index int) {
+			// Nothing
```
taratatach, line 45: Do we have a notion of alt text for images in Notes? We could render something like `[My amazing cat]`

taratatach (approved):

model/note/note_test.go
nono: Yes, it's because the schema allows to have paragraphs and blocks of code inside a list item

model/note/text.go
nono: No, the prosemirror plugin from atlaskit doesn't have that

model/note/text.go
nono: I think it doesn't work when we have nested nodes inside the `listItem`

model/note/text.go
taratatach: OK, too bad.

model/note/text.go
nono: Maybe, but I'm not really inspired by this...

PR #3914, opened 2023-05-10
Refactor the `--appdir` flag to use the new `RegistryFs`

Peltoche:
This refactoring will greatly facilitate the dev for the futur `--konndir` flag

### DIFF
--- a/model/app/registry_fs.go
+++ b/model/app/registry_fs.go
@@ -1,0 +1,182 @@
+package app
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path"
+
+	"github.com/cozy/cozy-stack/pkg/appfs"
+	"github.com/cozy/cozy-stack/pkg/consts"
+	"github.com/cozy/cozy-stack/pkg/utils"
+
+	"github.com/spf13/afero"
+)
+
+var (
+	ErrInvalidDirPath         = errors.New("invalid dir path")
+	ErrInvalidIndexPath       = errors.New("invalid index path")
+	ErrInvalidManifestPath    = errors.New("invalid manifest path")
+	ErrUnexpectedResourceType = errors.New("unexpected resource type")
+)
+
+var RegistryFS *LocalRegistry
+
+type LocalResource struct {
+	Type consts.AppType
+	Dir  string
+}
+
+// LocalRegistry is a map of slug -> LocalResource used during the development
+// of webapps that are not installed in the Cozy but served from local
+// directories.
+type LocalRegistry struct {
+	resources map[string]LocalResource
+	fs        afero.Fs
+}
+
+// NewLocalRegistry instantiates a new [LocalResources].
+func NewLocalRegistry(fs afero.Fs) *LocalRegistry {
+	return &LocalRegistry{
+		resources: map[string]LocalResource{},
+		fs:        fs,
+	}
+}
+
+// ListApps returns all the application's slugs handled by the registry.
+func (r *LocalRegistry) ListApps() []string {
+	res := []string{}
+
+	for slug, resource := range r.resources {
+		if resource.Type == consts.WebappType {
+			res = append(res, slug)
+		}
+	}
+
+	return res
+}
+
+// ListKonnss returns all the konnectors's slugs handled by the registry.
+func (r *LocalRegistry) ListKonns() []string {
+	res := []string{}
+
+	for slug, resource := range r.resources {
+		if resource.Type == consts.KonnectorType {
+			res = append(res, slug)
+		}
+	}
+
+	return res
+}
+
+// Contains returns true if an app with the given slug is stored inside
+// this registry.
+func (r *LocalRegistry) Contains(slug string) bool {
+	_, ok := r.resources[slug]
+	return ok
+}
+
+// FileServer returns an `appfs.FileServer` serving the app matching
+// the given `slug`
+func (r *LocalRegistry) FileServer(slug string) appfs.FileServer {
+	resource := r.resources[slug]
+
+	return appfs.NewAferoFileServer(r.fs, func(_, _, _, file string) string {
+		return path.Join(resource.Dir, file)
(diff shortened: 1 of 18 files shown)


model/app/registry_fs.go
```diff
@@ -0,0 +1,156 @@
...
+
+// LocalRegistry is a map of slug -> LocalResource used during the development
+// of webapps that are not installed in the Cozy but served from local
+// directories.
+type LocalRegistry struct {
+	resources map[string]LocalResource
+	fs        afero.Fs
+}
+
+// NewLocalRegistry instantiate a new [LocalResources].
```
taratatach, line 38:
```suggestion
// NewLocalRegistry instantiates a new [LocalRegistry].
```

model/app/registry_fs.go
```diff
@@ -0,0 +1,156 @@
...
+func (r *LocalRegistry) FileServer(slug string) appfs.FileServer {
+	resource := r.resources[slug]
+
+	return appfs.NewAferoFileServer(r.fs, func(_, _, _, file string) string {
+		return path.Join(resource.Dir, file)
+	})
+}
+
+func (r *LocalRegistry) Add(slug string, res LocalResource) error {
+	absDir := utils.AbsPath(res.Dir)
```
taratatach, line 71: Isn't it weird not to pass the `fs` here? I believe different types of `fs` might return different absolute paths, no?

model/app/registry_fs.go
```diff
@@ -0,0 +1,156 @@
...
+
+	err := utils.DirExists(r.fs, absDir)
+	if err != nil {
+		return fmt.Errorf("%w: %w", ErrInvalidDirPath, os.ErrNotExist)
```
taratatach, line 75: Shoudn't we wrap `err` rather than `os.ErrNotExist` here?

model/app/registry_fs.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	var manifestPath string
+	var indexPath string
+
+	switch res.Type {
+	case consts.WebappType:
+		manifestPath = path.Join(absDir, WebappManifestName)
+		indexPath = path.Join(absDir, "index.html")
```
taratatach, line 84: nit: we might as well create a constant for `"index.html"`. `WebappIndexName` maybe?

model/app/registry_fs.go
```diff
@@ -0,0 +1,156 @@
...
+
+	case consts.KonnectorType:
+		manifestPath = path.Join(absDir, KonnectorManifestName)
+
+	default:
+		return fmt.Errorf("%q: %w", res.Type, ErrUnexpectedResourceType)
```
taratatach, line 90: Isn't the error type usually the first element of the message?

model/app/registry_fs.go
```diff
@@ -0,0 +1,156 @@
...
+func (r *LocalRegistry) GetWebAppManifest(slug string) (*WebappManifest, error) {
+	res, ok := r.resources[slug]
+	if !ok {
+		return nil, ErrNotFound
+	}
+
+	manFile, err := r.fs.Open(path.Join(res.Dir, WebappManifestName))
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, fmt.Errorf("could not find the manifest in your app directory %s", res.Dir)
```
taratatach, line 119: Why not wrapping `err` here?

model/app/registry_fs.go
```diff
@@ -0,0 +1,156 @@
...
+func (r *LocalRegistry) GetKonnManifest(slug string) (*KonnManifest, error) {
+	res, ok := r.resources[slug]
+	if !ok {
+		return nil, ErrNotFound
+	}
+
+	manFile, err := r.fs.Open(path.Join(res.Dir, KonnectorManifestName))
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, fmt.Errorf("could not find the manifest in your konnector directory %s", res.Dir)
```
taratatach, line 143: Why not wrapping `err` here?

model/app/webapp.go
```diff
@@ -642,9 +623,18 @@ func ListWebappsWithPagination(db prefixer.Prefixer, limit int, startKey string)
...
 	// - There are no docs at all
 	// We can load extra apps and append them safely to the list
-	for slug := range appsdir {
-		if man, err := loadManifestFromDir(slug); err == nil {
-			docs = append(docs, man)
+	if RegistryFS != nil {
+		for _, slug := range RegistryFS.List() {
+			appMan, err := RegistryFS.GetWebAppManifest(slug)
+			if errors.Is(err, ErrNotFound) {
+				continue
```
taratatach, line 630: This case should never happen since we're looping over the list of apps in the registry.
So I think we should remove this block for 2 reasons:
- it adds seemingly unnecessary logic
- if it ever happens, we have a problem so we should return the error

taratatach (changes requested):

model/app/registry_fs.go
Peltoche: For me the "default" is always at the end because this is the catch all if not other `case` have matched. When you read the code you check all the `case` and if you don't see a match you fallback on the `default`. 

After this is a question of taste. Maybe we could check if there is a linter for this kind of rule.

model/app/registry_fs.go
Peltoche: This is some code I have only move and not written so i didn't change the logic but I'm agree with you. I will change

model/app/registry_fs.go
Peltoche: same as above

model/app/registry_fs.go
Peltoche: the afero package also use the `filepath` packe to clean it's path and there is no path related method inside the `afero.FS`.

nono:
I think we can close this PR. Is it OK?

PR #3915, opened 2023-05-10
Add the `--konndir` flag support

Peltoche:
This flag allow to expose a directory as a konnector installed inside
the registry


Peltoche:
Replaces by #3917

PR #3916, opened 2023-05-11
feat: Add `icon-cozy-home-inverted.svg` icon

Merkur39:
This icon does not need to be reversed, but must exist because an icon of the same name is used in a client context.

(I don't know if it's relevant to add it [here](https://github.com/cozy/cozy-stack/blob/d76231c64e512f976a2b0f73f723362655a113b4/docs/config.md?plain=1#L255))


Crash--:
>(I don't know if it's relevant to add it [here](https://github.com/cozy/cozy-stack/blob/d76231c64e512f976a2b0f73f723362655a113b4/docs/config.md?plain=1#L255))

I think so, yes! 


The idea here is to have the possibility to have 2 "home" icons for the same context so that an application can choose one or the other depending on its background (dark or light). By putting this 2nd icon on the stack and via the assets, all the apps share the same URL to access it and thus we gain in http cache. 

Using an inline SVG in order to use a fill in the SVG in the cozy-bar was not easy and not very efficient.

Merkur39:
> > (I don't know if it's relevant to add it [here](https://github.com/cozy/cozy-stack/blob/d76231c64e512f976a2b0f73f723362655a113b4/docs/config.md?plain=1#L255))
> 
> I think so, yes!

Check

PR #3917, opened 2023-05-11
feat: Allow to dev in local with the `install` cmd

Peltoche:
This forces the konnectors/apps refresh from their source after each reload for apps or each run for konnectors.

This should improve the dev experience by allowing them to install a local application using the following command:

`cozy-stack konnectors install foo file://my/project/folder/build`

### DIFF
--- a/model/app/installer.go
+++ b/model/app/installer.go
@@ -14,6 +14,7 @@ import (
 	"github.com/cozy/cozy-stack/model/instance"
 	"github.com/cozy/cozy-stack/model/permission"
 	"github.com/cozy/cozy-stack/pkg/appfs"
+	build "github.com/cozy/cozy-stack/pkg/config"
 	"github.com/cozy/cozy-stack/pkg/config/config"
 	"github.com/cozy/cozy-stack/pkg/consts"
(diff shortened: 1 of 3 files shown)


model/app/installer.go
```diff
@@ -409,7 +410,22 @@ func (i *Installer) update() error {
 	// verifyPermissions flag is activated (for non manual updates for example),
 	// we cancel out the update and mark the UpdateAvailable field of the
 	// application instead of actually updating.
-	if makeUpdate && !isPlatformApp(oldManifest, i.context) {
+	if build.IsDevRelease() {
+		// If we are in dev release we want to automatically accept any permissions set
```
nono, line 414: I'm not sure I agree with this change. It can be useful for developers to test an upgrade with a permissions change.

Peltoche: You still can if you don't use the `--dev` flag. All depends of the meaning you want to give to the flag `--dev`. For me I understood it as the flag setup when you want to dev locally and you need to remove some protections allowing hot-reload, removing some check or stuff like that.

nono (approved):

PR #3920, opened 2023-05-11
feat: Add a SVgo implementation of initials

Peltoche:


pkg/avatar/testdata/initials-svgo.svg
taratatach, line 1: The content seems too large for the image size. Is that a configuration issue?

pkg/avatar/initials_svgo.go
```diff
@@ -0,0 +1,49 @@
...
+
+	b, err := hex.DecodeString(strings.TrimPrefix(color, "#"))
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse the color: %w", err)
+	}
+
+	fontSize := 128
+
+	width := 128
+	height := 128
```
taratatach, line 41: I think these should be attributes of the struct and passed to `NewSVGoInitials`.

pkg/avatar/initials_svgo.go
```diff
@@ -0,0 +1,49 @@
...
+// ContentType return the generated avatar content-type.
+func (a *SVGoInitials) ContentType() string {
+	return "image/svg+xml"
+}
+
+// Generate will create a new avatar with the given initials and color.
+func (a *SVGoInitials) Generate(ctx context.Context, initials, color string) ([]byte, error) {
+	var buf bytes.Buffer
+
+	b, err := hex.DecodeString(strings.TrimPrefix(color, "#"))
```
taratatach, line 33: What does `b` stand for here?

Peltoche: `bytes` the type is a `[]byte{}`

taratatach: what do you think of `rgb` or `rgbBytes` instead?

pkg/avatar/testdata/initials-svgo.svg
Peltoche: This seems to be an issue with github. You can check locally it render correctly.

pkg/avatar/initials_svgo.go
Peltoche: At the moment we only need one size and I'm really not sure that the icon will behave correctly if we change those values.

pkg/avatar/testdata/initials-svgo.svg
taratatach: I opened the raw file and even downloaded it and the result is the same. Besides, a circle with a `100px` radius is wider than a `128px` image.
I think you should check the values you use to create the circle and the text.

taratatach, line 1: The initials don't have the same size and are not horizontally centered while this seems to be the case in production.

Also, could you add a test with only one letter to make sure this case is covered as well?

pkg/avatar/initials_svgo.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	fontSize := 64
+
+	width := 128
+	height := 128
+	canvas := svg.New(&buf)
+	canvas.Start(width, height)
+	canvas.Circle(width/2, height/2, 64, canvas.RGB(int(rgbBytes[0]), int(rgbBytes[1]), int(rgbBytes[2])))
+	canvas.Text(width/2, height/2+fontSize/3, initials, fmt.Sprintf("text-anchor:middle;font-family:Roboto;font-size:%dpx;fill:white", fontSize))
```
nono, line 45: What happens if the user doesn't have the Roboto font installed on their computer?

Peltoche: It fallback to the an another font and this can be the browser default font or the default systeme font.

Peltoche:
After a second check by the frontend devs it appears that the migration from PNG to SVG will be more complicated than expected. The React Native components can't easily move from PNG to SVG so we need to keep the PNG generation by default. 

This PR main goal was to remove the use of `imagemagick` but if we are forced to keep it for the default behaviour it doesn't make as much sens to merge it. We would gain some perf for web applications but it would add complexity on server side. As the performance was not the original goad I close the PR for now.

PR #3928, opened 2023-05-15
refactor: Refactor CheckPreconditions and add tests

Peltoche:


pkg/utils/serve_content_test.go
```diff
@@ -0,0 +1,95 @@
...
+			IfNoneMatch: `"first-etag","second-etag"`,
+			Etag:        `"second-etag"`,
+			Match:       true,
+		},
+		// {
+		// 	Name:        "multiple inm values are trimmed",
+		// 	IfNoneMatch: `"first-etag" , "second-etag"`,
+		// 	Etag:        `"second-etag"`,
+		// 	Match:       true,
+		// },
```
taratatach, line 45: Is this supposed to be removed?

Peltoche: It is supposed to pass but it doesn't with the current code. I leaved commented in order to not forget. I would like to keep this PR only for the tests so this comment will disappear in the next PR, the one with the refactoring.

I will leave a comment

taratatach (approved):

PR #3932, opened 2023-05-16
Move the `serveContent` method

Peltoche:


pkg/appfs/server.go
```diff
@@ -178,8 +178,8 @@ func (s *swiftServer) ServeFileContent(w http.ResponseWriter, req *http.Request,
 	}
 
 	size, _ := strconv.ParseInt(contentLength, 10, 64)
-	web_utils.ServeContent(w, req, contentType, size, r)
-	return nil
+
+	return serveContent(w, req, contentType, size, r)
```
nono, line 182: It is a bit useless to return the error on the copy of the body, as we won't do anything of it (it is too late to send a 5xx error as the 200 code has already been sent). But I don't see any drawback, so why not.

nono (approved):

pkg/appfs/server.go
Peltoche: This can be useful for monitoring. This will allow use to monitor the number on incomplete download.

nono (approved):

PR #3934, opened 2023-05-19
Fix/konnector install

Crash--:


model/app/installer.go
```diff
@@ -619,7 +619,7 @@ func (i *Installer) ManifestChannel() chan Manifest {
 }
 
 // DoLazyUpdate tries to update an application before using it
-func DoLazyUpdate(in *instance.Instance, man Manifest, copier appfs.Copier, registries []*url.URL) Manifest {
+func DoLazyUpdate(in *instance.Instance, man Manifest, copier appfs.Copier, registries []*url.URL, appType consts.AppType) Manifest {
```
Ldoppea, line 622: nit: now that appType is given as param, I'm wondering if the `copier` param is still needed. Every calls create it from `app.Copier(appType, instance)`, so we have everything to create it from here and simplify the `DoLazyUpdate` api.

Ldoppea (approved):

Peltoche (changes requested):

Peltoche:
I have made a proposition on #3935, I have cherry-picked your doc

Crash--:
Thanks 🙏

PR #3935, opened 2023-05-19
feat: Update konnectors when fetching their manifest

Peltoche:
This PR is an another proposition for the issues addressed by #3934. it also contains some fixes

### DIFF
--- a/web/apps/apps.go
+++ b/web/apps/apps.go
@@ -107,10 +107,13 @@ func getHandler(appType consts.AppType) echo.HandlerFunc {
 		if err := middlewares.Allow(c, permission.GET, man); err != nil {
 			return err
 		}
+
+		registries := instance.Registries()
+		copier := app.Copier(man.AppType(), instance)
+		man = app.DoLazyUpdate(instance, man, copier, registries)
+
 		if appType == consts.WebappType {
-			registries := instance.Registries()
-			copier := app.Copier(consts.WebappType, instance)
-			man = app.DoLazyUpdate(instance, man, copier, registries)
+			// TODO: Check is this line is really necessary
 			man.(*app.WebappManifest).Instance = instance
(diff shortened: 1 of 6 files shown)


web/apps/apps.go
```diff
@@ -107,10 +107,12 @@ func getHandler(appType consts.AppType) echo.HandlerFunc {
 		if err := middlewares.Allow(c, permission.GET, man); err != nil {
 			return err
 		}
+
+		registries := instance.Registries()
+		copier := app.Copier(consts.WebappType, instance)
```
taratatach, line 112: We'll need a konnector copier to update konnectors.

```suggestion
		copier := app.Copier(man.AppType(), instance)
```

web/apps/apps.go
```diff
@@ -107,10 +107,12 @@ func getHandler(appType consts.AppType) echo.HandlerFunc {
...
+		man = app.DoLazyUpdate(instance, man, copier, registries)
+
 		if appType == consts.WebappType {
-			registries := instance.Registries()
-			copier := app.Copier(consts.WebappType, instance)
-			man = app.DoLazyUpdate(instance, man, copier, registries)
 			man.(*app.WebappManifest).Instance = instance
```
taratatach, line 116: I believe this can be removed as this is taken care of by `WebappManifest.ReadManifest()` which is called during the webapp `install()` and `update()` steps.

See https://github.com/cozy/cozy-stack/blob/master/model/app/webapp.go#L301

taratatach (changes requested):

web/apps/apps.go
Peltoche: I can't be sure this will not break something so at the moment I would prefere keeping this line.

taratatach (approved):

PR #3937, opened 2023-05-21
Split the debugger in `/pkg/logger` into two new implementations

Peltoche:


taratatach (changes requested):

taratatach:
It looks like this is not used yet while the globals used in production are removed.

pkg/logger/logger.go
```diff
@@ -13,12 +13,6 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-const (
-	debugRedisAddChannel = "add:log-debug"
-	debugRedisRmvChannel = "rmv:log-debug"
-	debugRedisPrefix     = "debug:"
-)
-
```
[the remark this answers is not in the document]
Peltoche: Yes off course, I absolutly not intend to remove them. I don't know how this diff ends up here...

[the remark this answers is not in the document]
Peltoche: They have been move inside the `debugger_redis.go` file, next to the implementation using it. 

https://github.com/cozy/cozy-stack/pull/3937/files#diff-9679e63b7906e8496babbe529a6cbff2b67cd997027b099d40eb7e647f50b37dR14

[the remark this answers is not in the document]
taratatach: Ah right, this is still the same package. Sorry.

taratatach:
So, this should not break the current implementation as I first thought (the required globals are still declared in the same package).
However, this is still not used… What's missing to switch to the new implementation?

Peltoche:
> So, this should not break the current implementation as I first thought (the required globals are still declared in the same package). However, this is still not used… What's missing to switch to the new implementation?

Nothing. I like to separate the PR where I add an implementation and the PR where I switch to it. This ensure the code isolation and can facilitate the rollback. If you don't like this system I can add the code switching to this implementation in this PR.

pkg/logger/debugger_redis.go
```diff
@@ -0,0 +1,152 @@
...
+	debugRedisRmvChannel = "rmv:log-debug"
+	debugRedisPrefix     = "debug:"
+)
+
+var (
+	ErrInvalidDomainFormat = errors.New("invalid domain format")
+)
+
+// RedisDebugger is a redis based [Debugger] implementation.
+// It sends the logs directly into a redis.
```
nono, line 24: I don't think the debugger sends logs, it only manipulates a list of Cozy domains which are in debug mode, right?

Peltoche: Correct, I fix this.

nono (approved):

pkg/logger/logger.go
```diff
@@ -149,28 +120,11 @@ func (e *Entry) Log(level logrus.Level, msg string) {
...
-			// it from the list and do nothing to the log.
-			err := RemoveDebugDomain(domain.(string))
-			if err != nil {
-				e.entry.Logger.Errorf("failed to remove debug domain %q: %s", domain, err)
-			}
-
-		default:
-			// Log with the default logger
-		}
+	if haveDomain && level == logrus.DebugLevel && debugger.ExpiresAt(domain.(string)) != nil {
```
taratatach, line 123: Why not use `DebugExpiration(domain)` here?

taratatach (approved):

pkg/logger/logger.go
Peltoche: DebuExpiration only global function calling `debugger.ExpiresAt` and this method is already called inside the condition.

PR #3939, opened 2023-05-22
Remove the `imagemagick` folder on stack shutdown

Peltoche:


pkg/avatar/service.go
```diff
@@ -75,6 +75,10 @@ func (s *Service) GenerateInitials(publicName string, opts ...Options) ([]byte,
 	return bytes, s.initials.ContentType(), nil
 }
 
+func (s *Service) Shutdown(ctx context.Context) error {
```
nono, line 78: Don't we need to add a call to this function somewhere?

Peltoche: Yep, in `cmd/serve.go`. I was working on a deeper refacto but I decided to keep this PR simple.

It appear this part of the diff have been erased with the other code. You are took quick for me, I didn't had the chance to review my PR myself :)

pkg/avatar/initials_convert_test.go
```diff
@@ -39,3 +40,32 @@ func TestInitialsPNG(t *testing.T) {
...
+	go func() {
+		isStarted <- struct{}{}
+		rawRes, err = client.Generate(context.Background(), "JD", "#FF7F1B")
+		generateFinished = true
+	}()
+
+	// Ensure that the Generate is started and not finished before calling
+	// `Shutdown`
+	<-isStarted
+	require.False(t, generateFinished)
```
nono, line 64: I don't like this test: it can easily be flaky. On this line, we don't have any guarantee that `Generate` is started. The call to `client.Shutdown` may take the lock before the call to `client.Generate`.

Peltoche: I can add a little sleep to be sure that `Generate` have the time to start but except this I don't really see how to improve this.

nono: I'm still not happy with this test. It could be flaky even with this change, and `go test -race` is also red (for the same reason).

Peltoche: I don't really see how to test it so I will remove the test

nono (approved):

PR #3940, opened 2023-05-22
Fix listing assets

nono:
https://github.com/cozy/cozy-stack/pull/3785/commits/e1fe8a537387228750e4213144c1557a74a2466f#diff-e35f56a6ed6c1d2da87f0ee14ed6e74b729656eb08a795d95b913e5726341727R75 has introduced a regression. The code was trying to read the assets without nbeing bound to the afero object, which doesn't work.

### DIFF
--- a/pkg/assets/dynamic/impl_afero_test.go
+++ b/pkg/assets/dynamic/impl_afero_test.go
@@ -1,0 +1,88 @@
+package dynamic
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"sort"
+	"testing"
+
+	"github.com/cozy/cozy-stack/pkg/assets/model"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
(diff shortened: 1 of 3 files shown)


pkg/assets/dynamic/impl_afero_test.go
```diff
@@ -0,0 +1,88 @@
...
+		require.Equal(t, content, actual)
+	})
+
+	t.Run("Remove", func(t *testing.T) {
+		fs := NewInMemoryFS()
+		content := []byte("content of foo.js")
+		require.NoError(t, addAsset(t, fs, "my-context", "foo.js", content))
+		require.NoError(t, fs.Remove("my-context", "/foo.js"))
+		_, err := fs.Get("my-context", "/foo.js")
+		assert.Error(t, err)
```
Peltoche, line 59: you can use `assert.ErrorIs(t, err, fs.ErrNotFound)` here. This will ensure us that we follow the fs errors

Peltoche (approved):

PR #3950, opened 2023-05-30
Fix revoked sharing for a single file

nono:
When a single file is shared, and the sharing is revoked, the stack should not transform the file to a directory (with a "sharing revoked" suffix in its name).


taratatach (approved):
It would be appropriate to have a test for this case.

PR #3955, opened 2023-05-31
fix: Improve the user-agent parsing for the app decomissioning

Peltoche:


cozy.example.yaml
```diff
@@ -1,4 +1,4 @@
-# This file contain an example of cozy-stack configuration file.
+#j This file contain an example of cozy-stack configuration file.
```
nono, line 1: I think it is a typo

nono (approved):

cozy.example.yaml
sblaisot, line 1: wut?

Peltoche: This have been fixed a few minutes after: https://github.com/cozy/cozy-stack/pull/3956/commits/a2ac74edb2d11f64647223d16d687b168dd6e19a

I did make a fix on this PR but I think I merged before it was taken into account.

PR #3956, opened 2023-05-31
fix: Remove some invalid parenthesis from a template

Peltoche:


web/auth/deprecated_app_list.go
```diff
@@ -77,6 +78,8 @@ func (d *DeprecatedAppList) RenderArgs(client *oauth.Client, inst *instance.Inst
 		WithField("app", app.Name).
 		Info("Deprecated app detected, stop authentication")
 
+	fmt.Printf("platform: %q\n\n", platform)
```
nono, line 81: I think you forgot a line for debugging

nono (approved):

PR #3957, opened 2023-06-01
feat: Add settings permission to bitwarden client

Crash--:
In order to create a "premium link", the client needs to fetch the uuid of the instance from the io.cozy.settings/instance document.

This permission will also be used to request
/data/io.cozy.settings/io.cozy.settings.bitwarden in order to get the organizationId and collectionId.

### DIFF
--- a/model/bitwarden/oauth.go
+++ b/model/bitwarden/oauth.go
@@ -23,6 +23,7 @@ var BitwardenScope = strings.Join([]string{
 	consts.Konnectors,
 	consts.AppsSuggestion,
 	consts.Support,
+	consts.Settings,
 }, " ")
 
 // oldBitwardenScope is here to help the transition of bitwarden tokens, as the


nono:
For `/data/io.cozy.settings/io.cozy.settings.bitwarden`, we already have made a change: https://github.com/cozy/cozy-stack/pull/3942.

How do you see the workflow for existing bitwarden clients? They will disconnect and reconnect?

nono:
We have made one time a migration for the permissions list of bitwarden client, cf the `oldBitwardenScope` from the same file that you have modified: https://github.com/cozy/cozy-stack/blob/master/model/bitwarden/oauth.go#L28-L38. But, I prefer when we can avoid that.

> As I understand today, even for an unlock we do create a new client, right?

I'm not sure, I think it is the case for web and browser extensions, but not for mobile apps. At least, it should be a good idea to check if it is the case.

Crash--:
> I'm not sure, I think it is the case for web and browser extensions, but not for mobile apps.

Yeah, I was thinking about the browser extension only because we'll at first only display this message within the browser extension and I forgot about other clients. 

So if we don't want to migrate the old client and we don't want to disconnect / reconnect, then the only solution is to do the thing as you have done in #3942 but here https://github.com/cozy/cozy-stack/blob/master/web/settings/instance.go#L66 . I'm not confortable with that, but if you prefer avoiding a migration go for it :)

PR #3960, opened 2023-06-01
Dont init the logger inside config

Peltoche:


nono:
OK, bad example. Let's take `cozy-stack fix contact-emails`.

nono:
Or the `cozy-stack swift ...`

Peltoche:
Both seems to do only calls via the http client and use `fmt.Print*` for the outputs. This seems coherent, if this was not the case this would mean that we are running server code directly inside the client and we would have logrus outputs for the cli.

nono:
> if this was not the case this would mean that we are running server code directly inside the client

We do have server code running inside the client. I don't remember for which commands, it is a legacy things. When I have time, I try to rewrite them to move the code on the server, but I don't think we are done.

Peltoche:
Ok, good to know. 

And for those case the default logrus config is not enough? All we want we want is to print the logs on stdout and the `logrus.StandardLogger` does that by default. If we don't call `logger.Init` we will fallback on the default logrus logger config with a `MemDebugger`.

Peltoche:
Everyone seems to aggreed that if there is some remaining logs on cozy-stack we should send them on syslog and not stdout. 

This review could be resolved when we will be sure that there is no more logs inside cozy-stack

PR #3967, opened 2023-06-05
Add some tests on `web/errors` and clean some code

Peltoche:


go.mod
```diff
@@ -36,7 +36,7 @@ require (
 	github.com/oschwald/maxminddb-golang v1.10.0
 	github.com/pquerna/otp v1.4.0
 	github.com/prometheus/client_golang v1.15.1
-	github.com/redis/go-redis/v9 v9.0.4
+	github.com/redis/go-redis/v9 v9.0.5
```
nono, line 39: Why do we have a module update in this PR?

web/errors/errors.go
```diff
@@ -59,20 +35,49 @@ func ErrorHandler(err error, c echo.Context) {
...
+	case errors.As(err, &couchError):
+		jsonError = &jsonapi.Error{
+			Status: couchError.StatusCode,
+			Title:  couchError.Name,
+			Detail: couchError.Reason,
+		}
+
+	case errors.As(err, &jsonError):
+		// jsonError is already filled
+		break
```
nono, line 59: Why do we need a `break` here?

client/files.go
```diff
@@ -39,18 +39,19 @@ type File struct {
...
+		CreatedAt  time.Time              `json:"created_at"`
+		UpdatedAt  time.Time              `json:"updated_at"`
+		Size       int64                  `json:"size,string"`
+		MD5Sum     []byte                 `json:"md5sum"`
+		Mime       string                 `json:"mime"`
+		Class      string                 `json:"class"`
+		Executable bool                   `json:"executable"`
+		Encrypted  bool                   `json:"encrypted"`
+		Tags       []string               `json:"tags"`
+		Metadata   map[string]interface{} `json:"metadata"`
```
nono, line 54: Why do we have this commit in the PR?

web/errors/errors.go
```diff
@@ -59,20 +35,47 @@ func ErrorHandler(err error, c echo.Context) {
 		log.Errorf("%s %s %s", req.Method, req.URL.Path, err)
```
nono, line 35: I would have preferred to keep the old way to have full details of the error in the log message

nono (approved):

PR #3974, opened 2023-06-12
Refactor: Create the `InstanceService` struct

Peltoche:
This struct aims to bring all the business logic around the instance in one place, with an available mock.

The code have been only moved, not modified except the callers.


model/instance/init.go
```diff
@@ -0,0 +1,42 @@
+package instance
+
+import "github.com/cozy/cozy-stack/pkg/config/config"
+
+var service *InstanceService
+
+func Init() *InstanceService {
```
nono, line 7: Why do you return an `*InstanceService` if it is never used?

model/instance/service.go
```diff
@@ -4,40 +4,51 @@ import (
...
-	return cachePrefix + inst.Domain
+type InstanceService struct {
+	cache  cache.Cache
+	logger *logger.Entry
+}
+
+func NewService(cache cache.Cache, lock lock.Getter) *InstanceService {
+	return &InstanceService{
+		cache:  cache,
+		logger: logger.WithNamespace("instance"),
```
nono, line 25: I don't see where this logger is used

model/instance/init.go
nono: OK, let's keep it

nono (approved):

PR #3975, opened 2023-06-12
Fix a race condition on io.cozy.shared

nono:
When a file is dissociated, its io.cozy.shared is updated twice. In most cases, it doesn't matter as both updates does nearly the same thing, but it may happen that a race condition arise for a sharing with many members. I don't really understand the detail, and I didn't find a way to reproduce in local, but I think that some members see the first revision and then the second revision, while most members only see the second revision.


Peltoche (approved):
I can only trust you on this one...

PR #3979, opened 2023-06-13
Add a mock implementation to `instance.Service`

Peltoche:


model/instance/init.go
```diff
@@ -4,6 +4,18 @@ import "github.com/cozy/cozy-stack/pkg/config/config"
 
 var service *InstanceService
 
+// Service handle all the interactions with the "instance" domaine.
```
nono, line 7: [nit] s/domaine/domain/

model/instance/init.go
```diff
@@ -4,6 +4,18 @@ import "github.com/cozy/cozy-stack/pkg/config/config"
...
+//
+// This interface has several implementations:
+// - [InstanceService] with the business logic
+// - [Mock] with a mock implementation
+type Service interface {
+	Get(domain string) (*Instance, error)
+	GetFromCouch(domain string) (*Instance, error)
```
nono, line 14: It looks like this method has been renamed to `GetWithoutCache`

Peltoche: True, I fixed and add a test to check the implementation.

nono (approved):

PR #3984, opened 2023-06-13
Test: Add a E2E test on bank decommission

Peltoche:


nono (approved):
LGTM but I won't called this a E2E test. For me, E2E test means a real browser and probably a screenshot or a video.

PR #3988, opened 2023-06-15
Add the first step for the email update with validation

Peltoche:


assets/mails/update_email.text
```diff
@@ -0,0 +1,6 @@
+{{t "Mail Update Email Intro 1" .PublicName}}
+{t "Mail Update Email Intro 2"}}
```
nono, line 2: I think one `{` is missing here

model/settings/service.go
```diff
@@ -2,53 +2,115 @@ package settings
...
+	return &SettingsService{emailer, instance, token, storage}
 }
 
-func (s *SettingsService) GetInstanceSettings(inst prefixer.Prefixer) (*couchdb.JSONDoc, error) {
-	doc := &couchdb.JSONDoc{}
-
-	err := couchdb.GetDoc(inst, consts.Settings, consts.InstanceSettingsID, doc)
+// PublicName returns the settings' public name or a default one if missing
+func (s *SettingsService) PublicName(db prefixer.Prefixer) (string, error) {
+	doc, err := s.storage.getInstanceSettings(db)
```
nono, line 53: While we are doing refactoring, I think it would be useful to have a way to get the public name and the email while fetching only once the settings document from CouchDB.

For example, https://github.com/cozy/cozy-stack/blob/master/model/sharing/sharing.go#L164-L171 is probably making two CouchDB requests.

model/instance/lifecycle/magic_link.go
```diff
@@ -5,6 +5,7 @@ import (
 	"net/url"
 
 	"github.com/cozy/cozy-stack/model/instance"
+	csettings "github.com/cozy/cozy-stack/model/settings"
```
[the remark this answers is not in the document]
Peltoche: `csettings` for cozy settings. I can't use `settings` for some package due to the conflics with the import `cozy-stack/model/bitwarden/settings`.

model/settings/service.go
Peltoche: Totally agree! This is completely on my bucket list but it was not a requirement for this PR and I tried to stay as focus as possible and as I see it this is not a trivial change. 

My vision for this issue would be to modfify the method `GetInstanceSettings` and make it return a structured `InstanceSettings` struct with the method `PublicName(db prefixer.Prefixer)` attached to it.

model/token/service.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func generateKey(db prefixer.Prefixer, ns Namespace, code string) (string, error) {
+	var nsStr string
+
+	switch ns {
+	case EmailUpdate:
+		nsStr = string(EmailUpdate)
+	case MagicLink:
+		nsStr = string(EmailUpdate)
```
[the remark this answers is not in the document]
Peltoche: good catch

model/instance/lifecycle/store.go
```diff
@@ -11,6 +11,8 @@ import (
 )
 
 // Store is an object to store and retrieve magic link codes.
+//
+// TODO: Move to [token.Service] with [token.MagicLink] namespace.
```
nono, line 15: Is it something you want to do in this PR? Or just a reminder for later?

model/token/service.go
```diff
@@ -0,0 +1,94 @@
...
+	val, ok := s.cache.Get(key)
+	if !ok {
+		return ErrInvalidToken
+	}
+
+	if string(val) != resource {
+		return ErrInvalidToken
+	}
+
+	// TODO: Should we delete automatically the token valid once validated?
```
nono, line 78: TODO spotted

web/settings/settings.go
```diff
@@ -84,6 +88,37 @@ func (h *HTTPHandler) warnings(c echo.Context) error {
 	return jsonapi.DataErrorList(c, warnings...)
 }
 
+// postEmail handle POST /settings/email
```
nono, line 91: Can you document this endpoint in https://github.com/cozy/cozy-stack/blob/master/docs/settings.md?

web/settings/integration_test.go
```diff
@@ -14,6 +14,7 @@ import (
 	"github.com/cozy/cozy-stack/model/instance/lifecycle"
```
nono, line 14: I would prefer to not rename this file to `integration_tests.go`. Currently, what we are calling integration tests are in https://github.com/cozy/cozy-stack/tree/master/tests/integration, and I would like to avoid the confusion between the two.

model/settings/service.go
```diff
@@ -2,53 +2,125 @@ package settings
...
-	if doc.DocType() != consts.Settings {
-		return ErrInvalidType
+	token, err := s.token.GenerateAndSave(inst, token.EmailUpdate, cmd.Email, TokenExpiration)
+	if err != nil {
+		return fmt.Errorf("failed to generate and save the confirmation token: %w", err)
 	}
 
-	if doc.ID() != consts.InstanceSettingsID {
-		return fmt.Errorf("%w: have %q, expected %q", ErrInvalidID, doc.ID(), consts.InstanceSettingsID)
+	link := inst.PageURL("/settings/email/confirm", url.Values{
```
nono, line 110: I suppose it will be new route added later, right?

model/instance/lifecycle/store.go
Peltoche: This is for later. I try to keep the changes as little as possible and this is not a hard requirement.

model/token/service.go
Peltoche: Same, this is not done yet and this doesn't looks like a requirement but maybe this could be great to have. 

What do you think of this feature?

nono: Probably a good idea for most tokens, but there are some kinds of tokens where we may want to allow to the reused.

model/settings/service.go
Peltoche: Yep

model/token/service.go
Peltoche: Agree to keep is as a todo for now?

nono: OK

model/settings/service.go
```diff
@@ -2,53 +2,125 @@ package settings
...
-	return doc, nil
-}
+	publicName, err := s.PublicName(inst)
+	if err != nil {
+		return fmt.Errorf("failed to retrieve the instance settings: %w", err)
+	}
 
-func (s *SettingsService) SetInstanceSettings(inst prefixer.Prefixer, doc *couchdb.JSONDoc) error {
-	// TODO: Validate input
+	settings.M["pending_email"] = cmd.Email
```
nono, line 103: I don't see where this change is persisted to CouchDB

Peltoche: I just added now.

model/settings/service_test.go
```diff
@@ -3,10 +3,128 @@ package settings
...
+
+	cmd := &UpdateEmailCmd{
+		Passphrase: []byte("some-pass"),
+		Email:      "some@email.com",
+	}
+
+	instSvc.On("CheckPassphrase", &inst, cmd.Passphrase).Return(nil).Once()
+
+	storage.On("getInstanceSettings", &inst).Return(&couchdb.JSONDoc{
+		M: map[string]interface{}{"public_name": "Jane Doe"},
```
nono, line 38: It is an example of what I try to explain sooner. Here, the mock is not exactly the same thing as the real implementation. It doesn't set the doctype, the ID and the revision. And it doesn't check them on `setInstanceSettings`.

In `func (s *CouchdbStorage) getInstanceSettings(inst prefixer.Prefixer) (*couchdb.JSONDoc, error) {`, if we had forgotten the line `doc.Type = consts.Settings`, the test would be green but the code wouldn't work.

model/settings/init.go
```diff
@@ -8,6 +8,7 @@ import (
 var service *SettingsService
 
 type Service interface {
+	PublicName(db prefixer.Prefixer) (string, error)
```
taratatach, line 14: Maybe use the same naming convention for the prefixer as the other methods of the service (i.e. `inst`)?

model/settings/init.go
```diff
@@ -18,9 +19,16 @@ func Init() *SettingsService {
 	return service
 }
 
+// PublicName returns the settings' public name or a default one if missing
+//
+// Deprecated: Use [Service.PublicName] instead.
+func PublicName(db prefixer.Prefixer) (string, error) {
```
taratatach, line 35: Same here.

model/settings/init.go
taratatach: Could you explain why you decided to move this method here? I'm curious.

model/settings/service_test.go
Peltoche: Yep, mocks doesn't works well when with unstructured types like `couchdb.JSONDoc`. That's why I tend to works with structured types and in this kind of case I would take a structured struct as argument and handle all the unstructured logic inside `setInstanceSettings`. 

Something like:
```go
type Settings  struct {
  PublicName string 
  PendingEmail string 
}

func (s *Storage) setInstanceSettings(inst prefixer.Prefixer, settings *Settings) error {
  doc := couchdb.JSONDoc{
    M: map[string]interface{}{
      "pending_email": settings.PendingEmail,
      "public_name": settings.PublicName,
    }
  }

  return couchdb.UpdateDoc(inst, doc)}
}

This ensure a strong contract between the parts.

Comment (Peltoche) [reply model/settings/service_test.go:38]: This is why I was talking about [in this previous comment](https://github.com/cozy/cozy-stack/pull/3988#discussion_r1231365898)

File: assets/locales/en.po [hunk 14]
```diff
@@ -763,6 +763,23 @@ msgstr ""
...
+msgid "Mail Update Email Intro 2"
+msgstr ""
+"You asked to change your password for Cozy Pass. "
+"If you didn't initiate this request, please contact us by replying directly to this email."
+
+msgid "Mail Update Email Button instruction"
+msgstr "Click on the this button to confirm the new email."
+
+msgid "Mail Update Email Button text"
+msgstr "Confirm my email"
```
taratatach, line 781:
```suggestion
msgstr "Confirm my new address"
```

assets/locales/fr.po
```diff
@@ -841,6 +841,23 @@ msgstr ""
...
+msgid "Mail Update Email Subject"
+msgstr "Mise a jour de l'email"
+
+msgid "Mail Update Email Intro 1"
+msgstr "Bonjour %s,"
+
+msgid "Mail Update Email Intro 2"
+msgstr ""
+"Vous avez demandé à changer votre adresse email. Si ce n'est pas le cas, "
+"contactez-nous en réponse à ce mail."
```
taratatach, line 853: Let's stay consistent with other e-mail wordings
```suggestion
"contactez-nous en réponse à cet email."
```

assets/locales/fr.po
```diff
@@ -841,6 +841,23 @@ msgstr ""
...
+
+msgid "Mail Update Email Button instruction"
+msgstr "Cliquez sur ce bouton pour valider votre changement d'email."
```
taratatach, line 856:
```suggestion
msgstr "Cliquez sur le bouton suivant pour valider votre changement d'adresse email."
```

assets/locales/fr.po
```diff
@@ -841,6 +841,23 @@ msgstr ""
...
+
+msgid "Mail Update Email Button text"
+msgstr "Je confirme mon adresse email"
```
taratatach, line 859: > 
```suggestion
msgstr "Je confirme ma nouvelle adresse"
```

assets/locales/en.po
```diff
@@ -763,6 +763,23 @@ msgstr ""
...
 
+msgid "Mail Update Email Subject"
+msgstr "Confirm your email"
+
+msgid "Mail Update Email Intro 1"
+msgstr "Hello %s,"
+
+msgid "Mail Update Email Intro 2"
+msgstr ""
+"You asked to change your password for Cozy Pass. "
```
taratatach, line 774: I don't get the link with Cozy Pass.

And aren't we updating the e-mail address rather than the passphrase?

```suggestion
"You asked to change your email address. "
```

assets/locales/en.po
```diff
@@ -763,6 +763,23 @@ msgstr ""
 "As a last resort, you can reset your password in the “forgot your password” section of your Cozy. "
 "If you did not initiate this request, contact us by replying directly to this email."
 
+msgid "Mail Update Email Subject"
+msgstr "Confirm your email"
```
taratatach, line 767:
```suggestion
msgstr "Confirm your email address"
```

docs/settings.md
```diff
@@ -46,6 +47,36 @@ Content-Type: application/vnd.api+json
 }
 ```
 
+## Email update
+
+### POST /settings/email
+
+The email update process starts with this call. The password is required in order to make
+a strong authentication. This endpoint will send a confirmation email to the new adress
+with a link. Once clicked this link will redirect the user to the second endpoint.
```
taratatach, line 56:
```suggestion
The email adress update process starts with this call. The password is required
in order to make a strong authentication. This endpoint will send a confirmation
email to the new address with a link. Once clicked, this link will redirect the
user to the second endpoint.
```

model/settings/service_test.go
nono: > mocks doesn't works well when with unstructured types like couchdb.JSONDoc

Structured types help, but `couchdb.JSONDoc` has an explicit field `Type` for the doctype, and it is not filled here.

web/settings/controller_test.go
```diff
@@ -28,7 +28,7 @@ import (
 	_ "github.com/cozy/cozy-stack/worker/mails"
```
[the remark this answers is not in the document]
Peltoche: Context: https://github.com/cozy/cozy-stack/pull/3988#discussion_r1234043674

Context 2: I wanted to write the tests for the `HTTPHandler` (and only this) but the file `setting_test.go` was already used with tests tests on more elements. As we didn't agree (yet) on what is unit/integration/E2E tests and where we use what we decided to name it `controller_test.go`. 

End: After some checks it will be very hard at the moment to make http tests due to the way the session is handled by global functions calling directly the db so I will revert the commit doing the renaming.

model/settings/service_test.go
Peltoche: The field `Type` is filled by the method `CouchdbStorage.getInstanceSettings` and used by `CouchdbStorage.setInstanceSettings. This field is part of the couchdb logic and only the `CouchdbStorage` should interact with it. The Service should't even be aware of this kind of internal fields. 

The fact we use the `couchdb.JSONDoc` is clearly a temporary state for me in order to make to gain some time.

assets/locales/en.po
Peltoche: I have taken the Password confirmation as example. I didn't see the part.

model/settings/init.go
Peltoche: This logic is part of the Settings as the data is stored inside the instance `settings` database. 

>  Maybe use the same naming convention for the prefixer as the other methods of the service (i.e. inst)?

Yep, didn't made my mind on it yet. I have started with `inst` but it appears that the rest of the codebase use `db` so I will make the changes to follow the convention.

nono (approved):

PR #3993, opened 2023-06-17
Move all the redis init logic inside it's own package

Peltoche:


pkg/config/config/config.go
```diff
@@ -292,27 +292,8 @@ func NewRedisConfig(u string) (conf RedisConfig, err error) {
 
 // GetRedisConfig returns a
 func GetRedisConfig(v *viper.Viper, mainOpt *redis.UniversalOptions, key, ptr string) (conf RedisConfig, err error) {
-	var localOpt *redis.Options
-
-	localKey := fmt.Sprintf("%s.%s", key, ptr)
```
nono, line 297: We must keep the `localOpt`. It is used for the flags, like in `cozy-stack serve --lock-url=redis://localhost:6379/1`.

nono (changes requested):

nono:
The other commits look good

PR #3997, opened 2023-06-17
Fix some dataraces

Peltoche:


model/app/installer_test.go
```diff
@@ -108,11 +108,8 @@ git checkout -`
 	localGitCmd.Stdout = &out2
 	localGitCmd.Stderr = &out2
 
-	err = localGitCmd.Start()
-	require.NoError(t, err, "failed to start the git server (output: %q): %s", out2.String(), err)
-
 	go func() {
-		err := localGitCmd.Wait()
+		err = localGitCmd.Run()
```
nono, line 112: I don't understand this change. How does it fix a datarace?

Peltoche:
```
Read at 0x00c000303ef0 by goroutine 52:
  bytes.(*Buffer).String()
      /usr/lib/go/src/bytes/buffer.go:65 +0x1b7
  github.com/cozy/cozy-stack/model/app_test.serveGitRep.func1()
      /home/peltoche/Cozy/cozy-stack/model/app/installer_test.go:117 +0x6b
 ```
 
 The variable `out2` is accessed by two goroutines:
 - The main one by `localGitCmd.Start()`
 - The second one by `localGitCmd.Run()`
 
 If you move the `Start` and `Run` (by using `Wait`) into a single goroutine you are ok.

PR #4013, opened 2023-06-21
Handle the case were ErrorHandler is called with nil

Peltoche:


nono:
I have cherry-picked the tests

PR #4015, opened 2023-06-21
Add the datarace detection for go tests

Peltoche:


nono:
I didn't realize that `-race` flag makes the tests two times slower. I don't think it is a good idea to activate it on the CI.

PR #4017, opened 2023-06-22
Add the second step to the email update proces

Peltoche:


web/settings/settings.go
```diff
@@ -123,6 +123,24 @@ func (h *HTTPHandler) postEmail(c echo.Context) error {
...
+	token := c.QueryParam("token")
+	inst := middlewares.GetInstance(c)
+
+	err := h.svc.ConfirmEmailUpdate(inst, token)
+	switch {
+	case err == nil:
+		// Redirect to the setting page
+		c.Redirect(http.StatusTemporaryRedirect, inst.SubDomain("settings").String())
+	case errors.Is(err, csettings.ErrNoPendingEmail), errors.Is(err, instance.ErrInvalidToken):
+		return jsonapi.BadRequest(errors.New("invalid token"))
```
nono, line 136: We should show a proper error page here. Something like https://github.com/cozy/cozy-stack/blob/master/web/auth/magic_link.go#L33-L45

Peltoche: Yep, this would be great. I will add this

nono:
Can you properly rebase the PR? There are some commits not related that have been included

assets/locales/fr.po
```diff
@@ -497,6 +497,15 @@ msgstr ""
...
 "l'e-mail que nous vous avons envoyé au moment de sa création."
 
+msgid "Error InvalidToken Title"
+msgstr "Lien non valide"
+
+msgid "Error InvalidToken Message"
+msgstr "Le lien que vous avez suivi a expiré ou est erroné"
+
+msgid "Error InvalidToken Link"
+msgstr "Retourner a la page de configuration"
```
nono, line 507: s/a/à/

pkg/config/config/config.go
```diff
@@ -839,6 +846,11 @@ func UseViper(v *viper.Viper) error {
 		return fmt.Errorf(`failed to parse the config for "deprecated_apps": %w`, err)
 	}
 
+	err = v.UnmarshalKey("clouderies", &config.Clouderies)
+	if err != nil {
+		return fmt.Errorf(`failed to parse the config for "deprecated_apps": %w`, err)
```
nono, line 851: s/deprecated_apps/clouderies/

docs/settings.md
```diff
@@ -77,6 +78,50 @@ Content-Type: application/json
...
+
+```http
+HTTP/1.1 307 Temporary Redirect
+Location: http://alice-settings.cozy.localhost:8080 
+```
+
+
+### DELETE /settings/email
+
+This endpoints allows to cancel the ongoin email adress update process.
```
nono, line 108: s/ongoin/ongoing/

model/cloudery/service.go
```diff
@@ -0,0 +1,62 @@
...
+
+	if !ok {
+		return fmt.Errorf("%w: tried %q and %q", ErrInvalidContext, inst.ContextName, config.DefaultInstanceContext)
+	}
+
+	client := manager.NewAPIClient(cfg.API.URL, cfg.API.Token)
+
+	url := fmt.Sprintf("/api/v1/instances/%s?source=stack", url.PathEscape(inst.UUID))
+	if err := client.Put(url, map[string]interface{}{
+		"local":       cmd.Locale,
```
nono, line 54: s/local/locale/

model/cloudery/service_noop.go
```diff
@@ -0,0 +1,14 @@
+package cloudery
+
+import "github.com/cozy/cozy-stack/model/instance"
+
+// NoopService implements [Service].
+//
+// This implem does nothing. It is used when no config
+// is provided.
```
nono, line 8: [nit] the wrapping looks weird. I would have expected a single line, not two here

model/instance/lifecycle/patch.go
```diff
@@ -200,29 +205,9 @@ func Patch(i *instance.Instance, opts *Options) error {
...
 
-	if !opts.FromCloudery {
-		managerUpdateSettings(i, clouderyChanges)
-	}
-
 	return nil
 }
 
-func managerUpdateSettings(inst *instance.Instance, changes map[string]interface{}) {
-	if inst.UUID == "" || len(changes) == 0 {
```
nono, line 211: I think we lost this `if len(changes) == 0 { return }` optimization to avoid calling the cloudery when not necessary

model/settings/service.go
```diff
@@ -129,3 +133,67 @@ func (s *SettingsService) StartEmailUpdate(inst *instance.Instance, cmd *UpdateE
...
+
+	err = s.storage.setInstanceSettings(inst, settings)
+	if err != nil {
+		return fmt.Errorf("failed to save the settings changes: %w", err)
+	}
+
+	err = s.cloudery.SaveInstance(inst, &cloudery.SaveCmd{
+		Locale:     inst.Locale,
+		Email:      settings.M["email"].(string),
+		PublicName: settings.M["public_name"].(string),
```
nono, line 168: I would prefer to use a local variable with:

```go
publicName, _ := settings.M["public_name"].(string)
```

If an instance has no public_name, it would avoid a panic.

model/instance/lifecycle/patch.go
Peltoche: This is wanted. We have two consecutive writes not protected by any commit system and this can lead to data corruption between the two bases. In case of a corruption calling `cloudery.SaveInstance` would make the data converge and this corruption would be fixed.

So by removing this optimization I improve the data quality. I choose this tradeoff because we don't have perf issues on this call.

nono: @sblaisot is it OK for you?

sblaisot: without this optimization, how do you stop the cycle stack -> cloudery -> stack -> cloudery -> ... ?

sblaisot: and IIRC this was made to avoid one round trip that was too costly when synchronously attibuting an instance to a user

nono: The cycle is stopped by the stack : when the cloudery makes a request, a parameter is added to tell to the stack that this change is coming from the cloudery, and the stack no longer makes a request to the cloudery in that case (cf line 181).

nono (approved):

PR #4038, opened 2023-07-05
Allow to use file:// for remote assets in dev

nono:
When the stack is built for a dev release, it is now possible to declare a remote asset in a configuration file with an URL for a local file by using the file:// protocol.


paultranvan (approved):
Thanks!

PR #4039, opened 2023-07-05
Block password reset for not onboarded instance

nono:
If an instance has not been onboarded, the stack will show an error screen if the user tries to reset their password. It may happen when trying to connect the Pass extension, typing the instance URL, and then then the "password forgotten" button.

### DIFF
--- a/web/auth/passphrase.go
+++ b/web/auth/passphrase.go
@@ -22,6 +22,17 @@ import (
 
 func passphraseResetForm(c echo.Context) error {
 	instance := middlewares.GetInstance(c)
+	if !instance.OnboardingFinished {
+		return c.Render(http.StatusOK, "need_onboarding.html", echo.Map{
+			"Domain":       instance.ContextualDomain(),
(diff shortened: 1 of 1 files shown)


web/auth/passphrase.go
```diff
@@ -22,6 +22,17 @@ import (
...
 func passphraseResetForm(c echo.Context) error {
 	instance := middlewares.GetInstance(c)
+	if !instance.OnboardingFinished {
+		return c.Render(http.StatusOK, "need_onboarding.html", echo.Map{
+			"Domain":       instance.ContextualDomain(),
+			"ContextName":  instance.ContextName,
+			"Locale":       instance.Locale,
+			"Title":        instance.TemplateTitle(),
+			"Favicon":      middlewares.Favicon(instance),
+			"SupportEmail": "contact@cozycloud.cc",
```
taratatach, line 32: Shouldn't this be context dependant?

taratatach (dismissed):

PR #4040, opened 2023-07-05
Allow to go back to settings after password reset

nono:


cballevre (approved):
I can't validate the code itself, given my knowledge of go. However, I can validate the functional aspect. It works well with my front-end modifications

taratatach (commented):
Why aren't we using the `redirect` query-param here? Is it already used for something else?

web/auth/passphrase.go
```diff
@@ -30,23 +30,27 @@ func passphraseResetForm(c echo.Context) error {
 	if resp, err := couchdb.NormalDocs(instance, consts.BitwardenCiphers, 0, 1, "", false); err == nil {
 		hasCiphers = resp.Total > 0
 	}
-	showBackButton := true
-	if c.QueryParam("hideBackButton") == "true" {
-		showBackButton = false
+	backButton := ""
+	from := c.QueryParam("from")
+	if from != "" {
+		backButton = instance.SubDomain(from).String()
```
taratatach, line 36: I'm wondering why we don't build the URL here rather than passing the app's subdomain.

taratatach (approved):

PR #4049, opened 2023-07-10
fix: Set correct bitwarden error (JSON) when wrong 2FA code

Ldoppea:
On Cozy Pass mobile when the login requires 2FA and when the user enters a wrong password, then the API should return a dedicated error message

With previous implementation, the same error would be returned on first login attempt and on second one containing a 2FA token. This would result on the app completing the login process but displaying an empty vault. This behaviour is wrong but is from Bitwarden's implementation. However with correct server implementation the correct behaviour happens

### DIFF
--- a/web/bitwarden/bitwarden.go
+++ b/web/bitwarden/bitwarden.go
@@ -360,6 +360,16 @@ func checkTwoFactor(c echo.Context, inst *instance.Instance) bool {
 		if token, ok := cache.Get(key); ok {
 			if inst.ValidateTwoFactorPasscode(token, passcode) {
 				return true
+			} else {
+				_ = c.JSON(http.StatusBadRequest, echo.Map{
+					"error":             "invalid_grant",
+					"error_description": "invalid_username_or_password",
+					"ErrorModel": map[string]string{
+						"Message": "Two-step token is invalid. Try again.",
+						"Object":  "error",
+					},
+				})
+				return false
 			}
 		}
 	}


Ldoppea:
Note that the current Bitwarden's implementation return this result on first login attempt:
```json
{
  "error": "invalid_grant",
  "error_description": "Two factor required.",
  "TwoFactorProviders": [
    "1"
  ],
  "TwoFactorProviders2": {
    "1": {
      "Email": "te************@gmail.com"
    }
  },
  "MasterPasswordPolicy": {
    "Object": "masterPass…"
  },
  "SsoEmail2faSessionToken": "BwSsoEmail2FaSessionToken_CfDJ8PNCK...",
  "Email": "testmail@gmail.com",
  "CaptchaBypassToken": "BWCaptchaBypass_CfDJ8PNCKA..."
}
```

On our side we return on first attempt:
```json
{
  "error": "invalid_grant",
  "error_description": "Two factor required.",
  "TwoFactorProviders": [
    1
  ],
  "TwoFactorProviders2": {
    "1": {
      "Email": "te************@gmail.com"
    }
  }
}
```

Some items are missing, is that expected?

And then this on 2nd attempt with a wrong 2FA code:
```json
{
  "error": "invalid_grant",
  "error_description": "invalid_username_or_password",
  "ErrorModel": {
    "Message": "Two-step token is invalid. Try again.",
    "Object": "error"
  },
  "CaptchaBypassToken": "BWCaptchaBypass_CfDJ8PNCKA..."
}
```

On our side:
```json
{
  "error": "invalid_grant",
  "error_description": "invalid_username_or_password",
  "ErrorModel": {
    "Message": "Two-step token is invalid. Try again.",
    "Object": "error"
  }
}
```

nono (approved):

nono:
For the missing fields, they were not here when I had added this endpoint, but the bitwarden API evolves and I don't follow the new changes. So, yes, it's expected that we have only a subset of what bitwarden does.

PR #4068, opened 2023-07-25
fix: Preserve other metadata when updating note

taratatach:
When updating a Note, we would take the new Document's metadata and
store it in stead of the old file's Metadata since the Note's title,
version, schema and content are stored there.

However, other apps can store data in a file's Metadata, such as a
qualification, and we don't want to lose this.

From now on, we'll merge the 2 Metadata objects, overriding any
existing Note metadata value but preserving metadata from other apps.


nono (approved):
Maybe we should remove certified metadata (https://github.com/cozy/cozy-stack/blob/master/model/vfs/metadata.go#L66), but I don't see why we would certify notes in the first place, so, let's go

PR #4069, opened 2023-07-25
feat: Add error for files larger than FileMaxSize

taratatach:
When uploading a file we can receive a `413` error response in return
in 2 cases:
1. uploading the file would exceed the quota of the Cozy
2. the file size is larger than the server's filesystem's maximum file size

We want clients to be able to make the difference between the 2
situations as they otherwise need to make an extra request to the
server to fetch the disk quota or hardcode the server's filesystem's
maximum file size.

Thus we introduce a new `ErrMaxFileSize` error which will still be
returned with a `413` HTTP status but with a different body.

### DIFF
--- a/web/files/files.go
+++ b/web/files/files.go
@@ -1946,7 +1946,7 @@ func wrapVfsError(err error) *jsonapi.Error {
 	case vfs.ErrFileInTrash, vfs.ErrNonAbsolutePath,
 		vfs.ErrDirNotEmpty:
 		return jsonapi.BadRequest(err)
-	case vfs.ErrFileTooBig:
+	case vfs.ErrFileTooBig, vfs.ErrMaxFileSize:
 		return jsonapi.Errorf(http.StatusRequestEntityTooLarge, "%s", err)
 	case vfs.ErrWrongToken:
 		return jsonapi.BadRequest(err)
--- a/web/notes/notes.go
+++ b/web/notes/notes.go
@@ -389,7 +389,7 @@ func UploadImage(c echo.Context) error {
 			return jsonapi.InvalidParameter(echo.HeaderContentLength, err)
 		}
 		if size > note.MaxImageWeight {
-			return jsonapi.Errorf(http.StatusRequestEntityTooLarge, "%s", vfs.ErrFileTooBig)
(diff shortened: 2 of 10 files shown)


web/files/files.go
```diff
@@ -1947,6 +1947,7 @@ func wrapVfsError(err error) *jsonapi.Error {
 		vfs.ErrDirNotEmpty:
 		return jsonapi.BadRequest(err)
 	case vfs.ErrFileTooBig:
+	case vfs.ErrMaxFileSize:
```
nono, line 1950: It should be `case vfs.ErrFileTooBig, vfs.ErrMaxFileSize:`!

web/notes/notes.go
```diff
@@ -473,6 +473,7 @@ func wrapError(err error) *jsonapi.Error {
 	case os.ErrNotExist, vfs.ErrParentDoesNotExist, vfs.ErrParentInTrash:
 		return jsonapi.NotFound(err)
 	case vfs.ErrFileTooBig:
+	case vfs.ErrMaxFileSize:
```
nono, line 476: Idem

web/sharings/sharings.go
```diff
@@ -905,6 +905,7 @@ func wrapErrors(err error) error {
 	case vfs.ErrConflict:
 		return jsonapi.Conflict(err)
 	case vfs.ErrFileTooBig:
+	case vfs.ErrMaxFileSize:
```
nono, line 908: Idem

web/shortcuts/shortcuts.go
```diff
@@ -198,6 +198,7 @@ func wrapError(err error) *jsonapi.Error {
 	case os.ErrNotExist, vfs.ErrParentDoesNotExist, vfs.ErrParentInTrash:
 		return jsonapi.NotFound(err)
 	case vfs.ErrFileTooBig:
+	case vfs.ErrMaxFileSize:
```
nono, line 201: Idem

nono (changes requested):

nono (approved):

nono (approved):

PR #4072, opened 2023-07-26
Don't try to upload deleted files

nono:
It should not happen, but if the stack try to upload a deleted file, it should detect it and skip this file instead of looping on this file with an error.

### DIFF
--- a/model/sharing/upload.go
+++ b/model/sharing/upload.go
@@ -224,6 +224,11 @@ func (s *Sharing) findNextFileToUpload(inst *instance.Instance, since string) (m
 				Warnf("missing results for bulk get %v", query)
 			return nil, 0, since, ErrInternalServerError
 		}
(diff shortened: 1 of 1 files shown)


model/sharing/upload.go
```diff
@@ -224,6 +224,11 @@ func (s *Sharing) findNextFileToUpload(inst *instance.Instance, since string) (m
 				Warnf("missing results for bulk get %v", query)
 			return nil, 0, since, ErrInternalServerError
 		}
+		if results[0]["_deleted"] == true {
+			inst.Logger().WithNamespace("upload").
+				Warnf("cannot upload file %v", results[0])
```
taratatach, line 229:
```suggestion
				Warnf("cannot upload _deleted file %v", results[0])
```

taratatach (approved):

PR #4080, opened 2023-08-03
feat: Allow limiting oauth clients

taratatach:
A given Cozy hosting service might want to limit the number of OAuth
  clients that a user can connect at any given time.

  We now allow defining such a limit via a feature flag,
  `cozy.oauthclients.max` with an integer value from -1 to MAX_INT where
  -1 means unlimited.

  When a limit is set and the user has reached it by connecting multiple
  OAuth clients, a new web page will be presented when trying to connect
  yet another client in stead of the OAuth permissions page.
  This page gives options to the user, based on the Cozy context, to
  either revoke existing clients or, in Cozy's case, subscribe to a plan
  that increases this limit.
  Once the number of connected clients is brought below the limit, the
  page will automatically refresh itself (a button allows doing a manual
  refresh) to present the permissions page and complete the OAuth
  connection flow.


assets/templates/too_many_oauth_clients.html
```diff
@@ -0,0 +1,46 @@
+<!DOCTYPE html>
```
nono, line 1: Don't forget to add this template to https://github.com/cozy/cozy-stack/blob/master/web/statik/handler.go#L29-L54

model/oauth/client.go
```diff
@@ -194,6 +194,27 @@ func GetNotifiables(i *instance.Instance) ([]*Client, error) {
 	return clients, nil
 }
 
+func GetConnectedUserClients(i *instance.Instance, limit int, bookmark string) ([]*Client, string, error) {
+	// Return clients with client_kind mobile, browser and desktop
+	var clients []*Client
+	req := &couchdb.FindRequest{
+		UseIndex: "connected-user-clients",
+		Selector: mango.And(mango.Gt("client_kind", ""), mango.Gt("client_name", "")),
```
nono, line 202: I think it is still WIP

web/statik/statik.go
nono, line 1: It's weird to have so many changes on this file

model/oauth/client.go
taratatach: that's the intended query if that's what you're referring to.

assets/templates/too_many_oauth_clients.html
taratatach: Ah OK. What's the purpose of this list?

pkg/couchdb/index.go
```diff
@@ -14,51 +14,58 @@ import (
...
-	mango.IndexOnFields(consts.Triggers, "by-worker-and-type", []string{"worker", "type"}),
+	mango.MakeIndex(consts.Triggers, "by-worker-and-type", mango.IndexDef{Fields: []string{"worker", "type"}}),
 
 	// Used to lookup oauth clients by name
-	mango.IndexOnFields(consts.OAuthClients, "by-client-name", []string{"client_name"}),
-	mango.IndexOnFields(consts.OAuthClients, "by-notification-platform", []string{"notification_platform"}),
+	mango.MakeIndex(consts.OAuthClients, "by-client-name", mango.IndexDef{Fields: []string{"client_name"}}),
+	mango.MakeIndex(consts.OAuthClients, "by-notification-platform", mango.IndexDef{Fields: []string{"notification_platform"}}),
+	mango.MakeIndex(consts.OAuthClients, "connected-user-clients", mango.IndexDef{
+		Fields: []string{"client_kind", "client_name"},
```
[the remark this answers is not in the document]
taratatach: My intention was for the Settings app to use this index as well and it apparently needs an index on `client_name` as well.

web/statik/statik.go
taratatach: I ran `make assets-fast`. Could this explain the number of changes?

model/oauth/client.go
```diff
@@ -793,4 +814,20 @@ func BuildLinkedAppScope(slug string) string {
 	return fmt.Sprintf("@%s/%s", consts.Apps, slug)
 }
 
+func HasReachedUserClientsLimit(inst *instance.Instance, limit int) (bool, error) {
+	if limit == -1 {
+		inst.Logger().Info("Limit is -1; bailing")
+		return false, nil
+	}
+
+	clients, _, err := GetConnectedUserClients(inst, 100, "")
```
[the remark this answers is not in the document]
taratatach: We could limit to `limit` indeed.

web/auth/oauth.go
```diff
@@ -238,9 +239,57 @@ func (a *AuthorizeHTTPHandler) authorizeForm(c echo.Context) error {
...
+		if clientsLimit, ok := flags.M["cozy.oauthclients.max"].(float64); ok && clientsLimit >= 0 {
+			limit := int(clientsLimit)
+
+			clients, _, err := oauth.GetConnectedUserClients(inst, 100, "")
+			if err != nil {
+				return fmt.Errorf("Could not get user OAuth clients: %w", err)
+			}
+			count := len(clients)
+
+			if count >= limit {
```
nono, line 257: From what I understand of the documentation, it should be `count > limit`

web/auth/oauth.go
```diff
@@ -238,9 +239,57 @@ func (a *AuthorizeHTTPHandler) authorizeForm(c echo.Context) error {
...
+				var manageDevicesURL, premiumURL string
+
+				connectedDevicesURL := inst.SubDomain(consts.SettingsSlug)
+				connectedDevicesURL.Fragment = "/connectedDevices"
+				manageDevicesURL = connectedDevicesURL.String()
+
+				if enablePremiumLinks, ok := flags.M["enable_premium_links"].(bool); ok && enablePremiumLinks {
+					var err error
```
nono, line 265: [nit] I think you can omit this line, as `err` is already declared before

taratatach: I remember getting a linter error without it but I don't anymore.

web/auth/oauth.go
```diff
@@ -238,9 +239,57 @@ func (a *AuthorizeHTTPHandler) authorizeForm(c echo.Context) error {
...
+	if !params.client.Flagship {
+		flags, err := feature.GetFlags(inst)
+		if err != nil {
+			return err
+		}
+
+		if clientsLimit, ok := flags.M["cozy.oauthclients.max"].(float64); ok && clientsLimit >= 0 {
+			limit := int(clientsLimit)
+
+			clients, _, err := oauth.GetConnectedUserClients(inst, 100, "")
```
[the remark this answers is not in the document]
taratatach: If we use `limit + 1` and we have say, 5 OAuth clients for a limit of 3, the modal will display `4/3 clients connected` instead of `5/3 clients connected`.

web/auth/oauth.go
nono: I'm confused. We are on the `authorize` page, the new OAuth client should have been created before we display it.

And the added part to the documentation says:

> In case a limit has been set on the Cozy to the number of user-connected OAuth
clients, and this limit has been exceeded already, the user will be presented a
screen requesting to either remove some existing clients or, if enabled,
increase the limit (e.g. by subscribing to a plan with a greater limit). Once
the number of connected clients is brought back to the limit or under, the
OAuth flow will resume and the permissions screen will be displayed.

nono:
By the way, I didn't say it, but I would really like to have some tests in this PR.

taratatach:
> By the way, I didn't say it, but I would really like to have some tests in this PR.

Yes, that's planned. I just wanted to have the feature working before my vacations so the front could use it and to get your feedback.

web/statik/statik.go
nono, line 1: It looks like something went wrong with this file: there are too many changed lines for just a few additions to assets

nono (approved):

PR #4081, opened 2023-08-03
Improve the way we get the sharing dir

nono:
It happens that the stack needs the main directory for a sharing. For example, when it receives a new file that is inside the shared directory. The stack was using a referenced_by to find it, but we have found in production that the referenced_by can have been put on the wrong directories. We will now use the ID when we have it, and only use the referenced_by if not.


model/sharing/files.go
```diff
@@ -335,6 +335,17 @@ func (s *Sharing) AddReferenceForSharingDir(inst *instance.Instance, rule *Rule)
 // GetSharingDir returns the directory used by this sharing for putting files
 // and folders that have no dir_id.
 func (s *Sharing) GetSharingDir(inst *instance.Instance) (*vfs.DirDoc, error) {
+	// When we can, find the sharding dir by its ID
```
taratatach, line 338:
```suggestion
	// When we can, find the sharing dir by its ID
```

PR #4082, opened 2023-08-04
feat: Handle Powens Uuids

doubleface:
Next versions of harvest, banking konnectors will handle powens uuids
which will allow the have the same identifiers for banks in every
environments (int, prod, local, etc)

### DIFF
--- a/model/account/type.go
+++ b/model/account/type.go
@@ -189,6 +189,9 @@ func (at *AccountType) MakeOauthStartURL(i *instance.Instance, state string, par
 		if id := params.Get("id_connector"); id != "" {
 			vv.Add("connector_ids", id)
 		}
+		if id := params.Get("connector_uuids"); id != "" {
+			vv.Add("connector_uuids", id)
+		}
 	case BIWebauth, BIWebauthAndSecret:
 		vv.Add("client_id", at.ClientID)
 		vv.Add("token", params.Get("token"))


taratatach (approved):

model/account/type.go
```diff
@@ -189,6 +189,9 @@ func (at *AccountType) MakeOauthStartURL(i *instance.Instance, state string, par
 		if id := params.Get("id_connector"); id != "" {
 			vv.Add("connector_ids", id)
 		}
+		if id := params.Get("connector_uuids"); id != "" {
+			vv.Add("connector_uuids", id)
+		}
```
Crash--, line 194: @doubleface what is the difference between `id_connector`and `connector_uuids` ? 

Can we remove `id_connector` later once Harvest is up to date everywhere? 

Don't we have documentation to update in order to reflect that change?

Crash--: Also, don't we need the same stuff for the webauth ? https://github.com/cozy/cozy-stack/blob/a097022fb87c8ed22c8d259f23e1444800b58ade/model/account/type.go#L195-L205

doubleface: id_connector expects to have bank ids.
Yes we can remove this once we have updated harvest everywhere. 
I will check the diagrams and docs one more time but I don't this any of theme talked about this param

About webauth, we don't use it directly anymore since webviews, webauth is integrated inside powens webviews. We could remove it since webviews are considered fully deployed.

PR #4086, opened 2023-08-10
feat: Block users with exceeded clients limit

taratatach:
If the user manages to connect more OAuth clients than is allowed
(e.g. by trying to connect a Flagship app while the limit was already
reached), we want to redirect them to the Settings app to manage their
clients or increase their limit if they can.

If the request is asking for a JSON response, we'll return the list of
warnings (including a new warning for the reached OAuth clients limit)
instead of redirecting.

### DIFF
--- a/model/oauth/client.go
+++ b/model/oauth/client.go
@@ -814,4 +814,21 @@ func BuildLinkedAppScope(slug string) string {
 	return fmt.Sprintf("@%s/%s", consts.Apps, slug)
 }
 
+func CheckOAuthClientsLimitReached(i *instance.Instance, limit int) (reached, exceeded bool) {
+	if limit == -1 {
+		return
+	}
+
+	clients, _, err := GetConnectedUserClients(i, 100, "")
+	if err != nil {
+		i.Logger().Errorf("Could not fetch connected OAuth clients: %s", err)
+		return
+	}
+	count := len(clients)
+
+	reached = count >= limit
+	exceeded = count > limit
+	return
+}
+
 var _ couchdb.Doc = &Client{}
--- a/tests/testutils/test_utils.go
+++ b/tests/testutils/test_utils.go
@@ -320,7 +320,7 @@ func (c *TestSetup) InstallMiniApp() (string, error) {
 	if err != nil {
 		return "", err
 	}
-	err = createFile(instance, appdir, "index.html", `this is index.html. <a lang="{{.Locale}}" href="https://{{.Domain}}/status/">Status</a> {{.Favicon}}`)
+	err = createFile(instance, appdir, "index.html", `<html><body>this is index.html. <a lang="{{.Locale}}" href="https://{{.Domain}}/status/">Status</a> {{.Favicon}}</body></html>`)
 	if err != nil {
 		return "", err
 	}
--- a/tests/testutils/test_utils.go
+++ b/tests/testutils/test_utils.go
@@ -26,6 +26,7 @@ import (
 	"github.com/cozy/cozy-stack/pkg/couchdb"
 	"github.com/cozy/cozy-stack/pkg/utils"
 	"github.com/gavv/httpexpect/v2"
+	"github.com/gofrs/uuid"
 	"github.com/labstack/echo/v4"
 	"github.com/ncw/swift/v2/swifttest"
 	"github.com/spf13/viper"
@@ -403,4 +404,48 @@ func compress(content string) []byte {
 	_, _ = bw.Write([]byte(content))
 	_ = bw.Close()
 	return buf.Bytes()
+}
+
+func WithManager(t *testing.T, inst *instance.Instance) (shouldRemoveUUID bool) {
+	if inst.UUID == "" {
+		uuid, err := uuid.NewV4()
+		require.NoError(t, err, "Could not enable test instance manager")
+		inst.UUID = uuid.String()
+		shouldRemoveUUID = true
+	}
+
+	config, ok := inst.SettingsContext()
+	require.True(t, ok, "Could not enable test instance manager: could not fetch test instance settings context")
+
+	managerURL, ok := config["manager_url"].(string)
+	require.True(t, ok, "Could not enable test instance manager: manager_url config is required")
+	require.NotEmpty(t, managerURL, "Could not enable test instance manager: manager_url config is required")
+
+	if inst.FeatureFlags == nil {
+		inst.FeatureFlags = map[string]interface{}{}
+	}
+	inst.FeatureFlags["enable_premium_links"] = true
+
+	t.Cleanup(func() {
+		err := DisableManager(inst, shouldRemoveUUID)
+		require.NoError(t, err)
+	})
+
+	err := instance.Update(inst)
+	require.NoError(t, err, "Could not enable test instance manager")
+
+	return shouldRemoveUUID
+}
+
+func DisableManager(inst *instance.Instance, shouldRemoveUUID bool) error {
+	if inst.FeatureFlags == nil {
+		inst.FeatureFlags = map[string]interface{}{}
+	}
+	inst.FeatureFlags["enable_premium_links"] = false
+
+	if shouldRemoveUUID {
+		inst.UUID = ""
+	}
+
+	return instance.Update(inst)
 }
--- a/web/apps/apps_test.go
+++ b/web/apps/apps_test.go
@@ -133,6 +133,41 @@ func TestApps(t *testing.T) {
 		assertInternalServerError(e, slug, testInstance.Domain, "/invalid")
 	})
 
+	t.Run("ServeWithClientsLimitExceeded", func(t *testing.T) {
+		e := testutils.CreateTestClient(t, ts.URL)
+
+		// Create the OAuth client for the flagship app
+		flagship := oauth.Client{
+			RedirectURIs: []string{"cozy://flagship"},
+			ClientName:   "flagship-app",
+			ClientKind:   "mobile",
+			SoftwareID:   "github.com/cozy/cozy-stack/testing/flagship",
+			Flagship:     true,
+		}
+		require.Nil(t, flagship.Create(testInstance, oauth.NotPending))
+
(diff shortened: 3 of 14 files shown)


model/oauth/client.go
```diff
@@ -814,4 +814,17 @@ func BuildLinkedAppScope(slug string) string {
 	return fmt.Sprintf("@%s/%s", consts.Apps, slug)
 }
 
+func CheckOAuthClientsLimitReached(i *instance.Instance, limit int) (reached, exceeded bool) {
+	clients, _, err := GetConnectedUserClients(i, 100, "")
```
nono, line 818: s/100/limit+1/

web/routing.go
```diff
@@ -210,14 +212,16 @@ func SetupRoutes(router *echo.Echo, services *stack.Services) error {
 	{
 		mwsNotBlocked := []echo.MiddlewareFunc{
 			middlewares.NeedInstance,
+			middlewares.NeedFlags,
```
nono, line 215: NO!!!

It would slow down every request to the stack for fetching the flags, while only very few needs them.

web/settings/settings.go
```diff
@@ -233,6 +233,7 @@ func (h *HTTPHandler) Register(router *echo.Group) {
 	router.GET("/clients", h.listClients)
 	router.DELETE("/clients/:id", h.revokeClient)
 	router.POST("/synchronized", h.synchronized)
+	router.GET("/clients/limit-exceeded", h.limitExceeded)
```
nono, line 236: Don't forget the documentation for this new endpoint

web/routing.go
```diff
@@ -79,9 +79,11 @@ func SetupAppsHandler(appsHandler echo.HandlerFunc) echo.HandlerFunc {
 		middlewares.Accept(middlewares.AcceptOptions{
 			DefaultContentTypeOffer: echo.MIMETextHTML,
 		}),
+		middlewares.NeedFlags,
 		middlewares.CheckInstanceBlocked,
 		middlewares.CheckInstanceDeleting,
 		middlewares.CheckTOSDeadlineExpired,
+		middlewares.CheckOAuthClientsLimitExceeded,
```
nono, line 86: I wouldn't have done that with a middleware:

1. I would prefer to do the check/redirection only for the HTML pages, not for the assets (JS, CSS, images)
2. I don't think it is a good idea to do that for people going to a shared by link page, as it can confuse them by saying that they have reached the limit of the number of their devices, but it's not their devices, but the devices of the owner of the instance with the sharing that is counted.

nono (commented):
> Warnings were previously included in webapps through the `CozyBar`
  template but less and less apps use this template and we want to make
  sure that warnings are always included and not leave this choice to
  the app developper.
  
Note that an app developer can still avoid that by skipping the `</head>` (it is authorized by html5). But, even if the `<meta>` for the warnings are injected in the html page, they won't do anyting without the cozy-bar or some other JS that read them and show them to the user. So, I'm not really convinced by the rationale to force the inclusion.

web/warnings/warnings.go
```diff
@@ -0,0 +1,27 @@
+package warnings
```
nono, line 1: I would prefer to avoid creating a package in `web` with no HTTP endpoint (and a single function). Maybe we can put it in `web/settings` (or `model/warnings` if you really prefer to have a new package).

web/settings/settings.go
```diff
@@ -64,20 +65,20 @@ func (h *HTTPHandler) getSessions(c echo.Context) error {
...
 		return err
 	}
 
-	warnings := inst.Warnings()
-	if warnings == nil {
-		warnings = []*jsonapi.Error{}
+	w := warnings.List(inst)
+	if w == nil {
+		w = []*jsonapi.Error{}
 	}
```
nono, line 79: [nit] I don't think we need to assign something to `w` when it is empty. These 3 lines may be removed.

model/oauth/client.go
taratatach: I did that in the first place but we figured out the clients count wouldn't be accurate if there are more clients than `limit + 1`.
I used `100` because we shouldn't have a limit this high ever.

web/routing.go
taratatach: Hum I tried to copy what was done for the ToS blocker as it seemed very similar.
I guess this check should be moved to the `Serve` method of the `apps` package?!

web/warnings/warnings.go
taratatach: I'll try `web/settings` since this is web related code. I don't specifically want to create a package but I had issues with circular dependencies and that was the way I found to fix the situation.

web/apps/apps_test.go
```diff
@@ -428,6 +428,7 @@ func TestApps(t *testing.T) {
 		attrs.Value("Cookie").String().Contains("HttpOnly")
 		attrs.Value("Token").String().NotEmpty()
 		attrs.ValueEqual("Flags", "{}")
+		attrs.Value("Warnings")
```
nono, line 431: hu, what are you testing here?

web/apps/serve.go
```diff
@@ -69,6 +69,12 @@ func Serve(c echo.Context) error {
 	}
 
 	if file == "" || file == route.Index {
+		if !route.Public {
+			if handled := middlewares.CheckOAuthClientsLimitExceeded(c); handled != nil {
```
nono, line 73: Are you sure about that? I think `middlewares.CheckOAuthClientsLimitExceeded` returns `nil` when it does the redirection

nono (commented):
Tests would be nice

web/apps/serve.go
taratatach: Hum, you're right. This should probably be something like `handled, err := middleware…`

web/apps/apps_test.go
taratatach: I thought `attrs.Value("Warnings").String().NotEmpty()` would throw since there are no warnings but it doesn't and probably does the actual presence check.

web/settings/settings_test.go
```diff
@@ -60,8 +63,14 @@ func TestSettings(t *testing.T) {
 	var oauthClientID string
 
 	config.UseTestFile(t)
+	conf := config.GetConfig()
+	conf.Assets = "../../assets"
+	_ = web.LoadSupportedLocales()
 	testutils.NeedCouchdb(t)
 	setup := testutils.NewSetup(t, t.Name())
+	render, _ := statik.NewDirRenderer("../../assets")
+	middlewares.BuildTemplates()
```
nono, line 72: I have fixed the test by copy-pasting from `web/sharings` tests a few lines

tests/testutils/test_utils.go
```diff
@@ -404,3 +406,45 @@ func compress(content string) []byte {
...
+
+	return instance.Update(inst), shouldRemoveUUID
+}
+
+func DisableManager(inst *instance.Instance, shouldRemoveUUID bool) error {
+	flags := inst.FeatureFlags
+	if flags == nil {
+		flags = map[string]interface{}{}
+	}
+	flags["enable_premium_links"] = true
```
nono, line 443: s/true/false/ ?

tests/testutils/test_utils.go
```diff
@@ -404,3 +406,45 @@ func compress(content string) []byte {
...
+		return fmt.Errorf("Could not enable test instance manager: could not fetch test instance settings context"), shouldRemoveUUID
+	}
+	manager_url, ok := config["manager_url"].(string)
+	if !ok || manager_url == "" {
+		return fmt.Errorf("Could not enable test instance manager: manager_url config is required"), shouldRemoveUUID
+	}
+
+	flags := inst.FeatureFlags
+	if flags == nil {
+		flags = map[string]interface{}{}
```
nono, line 431: We also need `inst.FeatureFlags = flags`, no?

tests/testutils/test_utils.go
```diff
@@ -404,3 +406,45 @@ func compress(content string) []byte {
...
+	}
+	flags["enable_premium_links"] = true
+
+	return instance.Update(inst), shouldRemoveUUID
+}
+
+func DisableManager(inst *instance.Instance, shouldRemoveUUID bool) error {
+	flags := inst.FeatureFlags
+	if flags == nil {
+		flags = map[string]interface{}{}
```
nono, line 441: idem

tests/testutils/test_utils.go
taratatach: Hum it seems to be working this way. Aren't maps actually pointers?

tests/testutils/test_utils.go
taratatach: Right. Copy/paste error.

tests/testutils/test_utils.go
```diff
@@ -404,3 +406,48 @@ func compress(content string) []byte {
 	_ = bw.Close()
 	return buf.Bytes()
 }
+
+func WithManager(t *testing.T, inst *instance.Instance) (err error, shouldRemoveUUID bool) {
```
nono, line 410: I don't see why you still return `shouldRemoveUUID`

tests/testutils/test_utils.go
```diff
@@ -404,3 +406,48 @@ func compress(content string) []byte {
...
+	if inst.UUID == "" {
+		uuid, err := uuid.NewV4()
+		if err != nil {
+			return fmt.Errorf("Could not enable test instance manager: %w", err), false
```
nono, line 414: You can use `require.NoError(t, err, "Could not enable test instance manager")` here instead of returning an error

tests/testutils/test_utils.go
taratatach: Because we might want to disable the Manager within a test and thus need it to call `DisableManager`.

tests/testutils/test_utils.go
taratatach: Ah yes, I didn't think about changing the behavior after passing `t`.

nono (approved):

PR #4087, opened 2023-08-10
feat: Add route to fetch OAuth clients usage

taratatach:
This will be useful to know how many OAuth clients a user has already
connected to their Cozy and how many they're allowed to create,
especially for the flagship app to decide whether it should complete
its onboarding flow or present the user with a modal requesting to
either remove some clients or increase the limit if they're able to.

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -36,7 +36,6 @@ Content-Type: application/vnd.api+json
         "type": "io.cozy.settings",
         "id": "io.cozy.settings.disk-usage",
         "attributes": {
-            "is_limited": true,
             "quota": "123456789",
             "used": "12345678",
             "files": "10305070",
@@ -46,6 +45,46 @@ Content-Type: application/vnd.api+json
     }
 }
 ```
+
+## OAuth clients usage
+
+### GET /settings/clients-usage
+
+This endpoint returns the number of user-connected OAuth clients, the limit set
+on the Cozy and if this limit has been reached or even exceeded.
+If there is no limit, the `limit` attribute won't be present in the response.
+
+#### Request
+
+```
+GET /settings/clients-usage HTTP/1.1
+Host: alice.example.com
+Accept: application/vnd.api+json
+Authorization: Bearer ...
+```
+
+#### Response
+
+```http
+HTTP/1.1 200 OK
+Content-Type: application/vnd.api+json
+```
+
+```json
+{
+    "data": {
+        "type": "io.cozy.settings",
+        "id": "io.cozy.settings.clients-usage",
+        "attributes": {
+          "limit": 3,
+          "count": 3,
+          "limitReached": true,
+          "limitExceeded": false
+        }
+    }
+}
+```
+
 
 ## Email update
 
--- a/web/settings/clients_usage.go
+++ b/web/settings/clients_usage.go
@@ -1,0 +1,75 @@
+package settings
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/cozy/cozy-stack/model/feature"
+	"github.com/cozy/cozy-stack/model/oauth"
+	"github.com/cozy/cozy-stack/model/permission"
+	"github.com/cozy/cozy-stack/pkg/consts"
+	"github.com/cozy/cozy-stack/pkg/couchdb"
+	"github.com/cozy/cozy-stack/pkg/jsonapi"
+	"github.com/cozy/cozy-stack/web/middlewares"
+	"github.com/labstack/echo/v4"
+)
+
+type apiClientsUsage struct {
+	Limit         *int `json:"limit,omitempty"`
+	Count         int  `json:"count"`
+	LimitReached  bool `json:"limitReached"`
+	LimitExceeded bool `json:"limitExceeded"`
+}
+
+func (j *apiClientsUsage) ID() string                             { return consts.ClientsUsageID }
+func (j *apiClientsUsage) Rev() string                            { return "" }
+func (j *apiClientsUsage) DocType() string                        { return consts.Settings }
+func (j *apiClientsUsage) Clone() couchdb.Doc                     { return j }
+func (j *apiClientsUsage) SetID(_ string)                         {}
+func (j *apiClientsUsage) SetRev(_ string)                        {}
+func (j *apiClientsUsage) Relationships() jsonapi.RelationshipMap { return nil }
+func (j *apiClientsUsage) Included() []jsonapi.Object             { return nil }
+func (j *apiClientsUsage) Links() *jsonapi.LinksList {
+	return &jsonapi.LinksList{Self: "/settings/clients-usage"}
+}
+
+// Settings objects permissions are only on ID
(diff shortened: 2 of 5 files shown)


docs/admin.md
```diff
@@ -426,6 +426,35 @@ Content-Type: application/zip
 Content-Disposition: attachment; filename="alice.cozy.localhost - part001.zip"
 ```
 
+## OAuth clients usage
+
+### GET /instances/:domain/clients-usage
```
nono, line 431: I don't understand why we have two endpoints that do the same thing.

docs/settings.md
```diff
@@ -47,6 +47,40 @@ Content-Type: application/vnd.api+json
...
+```
+
+```json
+{
+  "limit": 3,
+  "count": 3,
+  "reached": true,
+  "exceeded": false
+}
+```
```
nono, line 81: I think the response should in the JSON-API format

web/settings/clients_usage.go
```diff
@@ -0,0 +1,83 @@
...
+	if limit != -1 {
+		result.Limit = limit
+	}
+	result.Count = count
+	if count >= limit {
+		result.LimitReached = true
+	}
+	if count > limit {
+		result.LimitExceeded = true
+	}
```
nono, line 81: If `limit` is  set to `-1`, we will have `reached: true` and `exceeded: true`, but the documentation says that those attributes won't be present.

web/settings/clients_usage.go
```diff
@@ -0,0 +1,83 @@
...
+
+// checkAccessToClientsUsage validates the access control for the clients-usage
+// route. It checks if there is an explicit permission on this document, but
+// also allow every request from the flagship app as this route is used by it to
+// decide whether it can finalize its onboarding or not.
+func checkAccessToClientsUsage(c echo.Context, result *apiClientsUsage) error {
+	if err := middlewares.Allow(c, permission.GET, result); err == nil {
+		return nil
+	}
+	if client, ok := middlewares.GetOAuthClient(c); ok && client.Flagship {
```
nono, line 46: Do we really need this additional check? The flaship app has a wildcard permission, but maybe I am missing something.

docs/admin.md
taratatach: This route is an admin one. I used the same pattern as the disk usage one.

docs/settings.md
taratatach: Same here, I followed the same pattern as the disk usage one which I believe does not use the JSON-API format.

web/settings/clients_usage.go
taratatach: Good catch.

web/settings/clients_usage.go
taratatach: It could be that I don't understand correctly how the permission checks are done.
I thought the flagship app wouldn't have this specific permission and we needed to check the client type.

But if we have a wildcard permission then this is not needed indeed.

docs/admin.md
nono: hmmm, I don't see why we have two routes for disk usage too...

docs/settings.md
nono: https://docs.cozy.io/en/cozy-stack/settings/#get-settingsdisk-usage : it is in JSON-API format

docs/admin.md
```diff
@@ -426,6 +426,35 @@ Content-Type: application/zip
...
+
+This endpoint returns the number of user-connected OAuth clients, the limit set
+on the Cozy and if this limit has been reached or even exceeded.
+If there is no limit, the attribute won't be present in the response.
```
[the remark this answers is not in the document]
nono: You are right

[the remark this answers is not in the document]
taratatach: Now I understand why the disk usage int attributes are marshalled into strings. I'll do the same for the clients usage limit and count.

cc @Ldoppea this might change how you use the data in the flagship app

docs/admin.md
taratatach: To have different permissions checks? Or to avoid handling JSON-API data in the command line?

docs/admin.md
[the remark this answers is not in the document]
nono: They are marshaled into strings to avoid losing precision for big numbers. But I think you are looking at a pointer to a number for `limit` type. It allows to `omitempty` the `nil` value, and serializes it to a number even if it is 0.

docs/admin.md
nono: Maybe, but it looks like bad reasons for both cases.

taratatach: OK. Since this admin route for clients usage is not used yet, I can remove it. We can figure out what to do with the disk usage one later.

docs/admin.md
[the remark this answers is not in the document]
taratatach: Ah perfect! I'll go that route.

web/settings/clients_usage.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+type apiClientsUsage struct {
+	Limit         *int `json:"limit,omitempty"`
+	Count         int  `json:"count"`
+	LimitReached  bool `json:"limitReached"`
+	LimitExceeded bool `json:"limitExceeded"`
+}
+
+func (j *apiClientsUsage) ID() string                             { return consts.DiskUsageID }
```
nono, line 23: s/DiskUsageID/ClientsUsageID/ ?

web/settings/clients_usage.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// Settings objects permissions are only on ID
+func (j *apiClientsUsage) Fetch(field string) []string { return nil }
+
+// checkAccessToClientsUsage validates the access control for the clients-usage
+// route. It checks if there is an explicit permission on this document, but
+// also allow every request from the flagship app as this route is used by it to
+// decide whether it can finalize its onboarding or not.
+func checkAccessToClientsUsage(c echo.Context, result *apiClientsUsage) error {
```
nono, line 42: I think `return middlewares.Allow(c, permission.GET, result)` is enough

docs/settings.md
```diff
@@ -47,6 +46,46 @@ Content-Type: application/vnd.api+json
...
+```json
+{
+    "data": {
+        "type": "io.cozy.settings",
+        "id": "io.cozy.settings.clients-usage",
+        "attributes": {
+          "limit": 3,
+          "count": 3,
+          "reached": true,
+          "exceeded": false
```
nono, line 82: In the code, it is `limitReached` and `limitExceeded`, not just `reached` and `exceeded`

web/settings/clients_usage.go
nono: Do you think the `checkAccessToClientsUsage` function is still useful? I would inline it in the `clientsUsage` function

taratatach:
> Do you think the `checkAccessToClientsUsage` function is still useful? I would inline it in the `clientsUsage` function

It's probably not necessary. I'll inline the call.

nono (approved):

PR #4089, opened 2023-08-11
fix: Run eslint without third-party action

taratatach:
The third-party `eslint` action we were using has changed its API
making it very hard for us to check files from folders outside the
`scripts` folder.

Since it's pretty easy to run `eslint`, we'll stop using the action
altogether.

### DIFF
--- a/.github/workflows/js-lint.yml
+++ b/.github/workflows/js-lint.yml
@@ -19,13 +19,9 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: 16
-      - name: Yarn install 
+      - name: Yarn install
         run: |
           cd ./scripts
           yarn install
-      - uses: sibiraj-s/action-eslint@v3
-        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/js-lint.yml
```diff
@@ -19,13 +19,9 @@ jobs:
...
           cd ./scripts
           yarn install
-      - uses: sibiraj-s/action-eslint@v3
-        with:
-          eslint-args: '--config scripts/eslintrc.json --ignore-path=.gitignore --quiet'
-          extensions: 'js,jsx,ts,tsx'
-          bin-path: 'scripts/node_modules/eslint/bin'
-          annotations: true
+      - name: Run eslint
+        run: ./scripts/node_modules/.bin/eslint --config scripts/eslintrc.json --ignore-path=.gitignore --quiet
```
nono, line 27: I think you can use `make jslint`. It was what was used before: https://github.com/cozy/cozy-stack/pull/3810/commits/9b2d18f54cf27bd7243519b5966140fdb00201b6

PR #4101, opened 2023-08-22
Make the error for invalid/expired MetadataID more specific

nono:


Crash--:
Can we add a small hint here https://docs.cozy.io/en/cozy-stack/files/#post-filesuploadmetadata to say that the metadata will only have a 10 min lifetime ? 

Can we update also here https://docs.cozy.io/en/cozy-stack/files/#status-codes_2 to list the 422? and also here https://docs.cozy.io/en/cozy-stack/files/#post-filesdir-id ?

zatteo:
Works like a charm thanks!

PR #4102, opened 2023-08-22
Add support of task list and decision list for notes

nono:


model/note/markdown.go
```diff
@@ -36,11 +36,30 @@ func markdownSerializer(images []*Image) *markdown.Serializer {
...
+		"taskList": func(state *markdown.SerializerState, node, _parent *model.Node, _index int) {
+			state.RenderList(node, "  ", func(_ int) string { return "- " })
+		},
+		"taskItem": func(state *markdown.SerializerState, node, _parent *model.Node, _index int) {
+			if node.Attrs["state"] == "DONE" {
+				state.Write("[X] ")
+			} else {
+				state.Write("[ ] ")
+			}
+			state.RenderContent(node)
```
Crash--, line 56: It was for that!! I forgot to call `RenderContent` when I was trying to do that: 
```
"taskItem": func(state *markdown.SerializerState, node, _parent *model.Node, _index int) {
			var cross = " "
			value, _ := node.Attrs["state"].(string)
			if (value) == "DONE" {
				cross = "X"
			}
			state.Write(fmt.Sprintf("[%s]", cross))
			//state.Text(node.TextContent(), false)
		},
```

Is there a simple rule to know if we should call `RenderContent()` or `Text()`? 

`RenderInline` or `RenderList` seems to speak for themselves

model/note/note_test.go
```diff
@@ -46,7 +58,15 @@ hello world
...
-- baz`
+- baz
+
+This is a decision
+
+and another decision
+
+- A todo task
+
+- a done task`
```
Crash--, line 69: Where do we need this `Text` stuff where we have a markdown as input?

Crash-- (commented):
Nice! Well done

model/note/markdown.go
nono: If the thing that you want to render can have links or bold text, you should call `RenderContent()`, but if it is expected to be a raw text with no decoration (like in a block of code), you should call `Text()`.

model/note/note_test.go
nono: We don't. We want to test the `Text` output of a note, and it is too painful to create a note with its prosemirror tree manually to do that, so we import a markdown as a way to bootstrap a prosemirror tree in this test.

Crash--: And where do we need this `Text` output of a note? I'm asking in order to understand to not forget to update the test if I try to add a few plugins later.

nono: For the `MesPapiers` app, to copy the text of the notes and allow to paste it somewhere else

taratatach: I don't understand why we strip the `✍` character and `[ ]` or `[x]` strings in the text version.
These are perfectly understandable in their text form and removing them, especially the checkboxes, removes important information.

taratatach (commented):
No need to add a custom struct for `TaskList` nodes? Will we handle the state of sub-tasks correctly without it?

model/note/note_test.go
nono: The requirement is to remove all markup. The only exception is `-` for lists.

nono:
> No need to add a custom struct for TaskList nodes?

I really wanted to use the same format as GitHub Flavored Markdown for that (aka `- [ ]`). The markdown parser library we use can interpret that as a list of items, and inside the items, there is a checkbox. It would be really hard to implement a custom parser for task lists that doesn't mess the parsing of classical lists. So, it's parsed as a list of items, and it's when we transform the AST to Prosemirror tree that we do extra work to detect if it is a task list.

> Will we handle the state of sub-tasks correctly without it?

Yes, we are using the fact that the checkbox is checked or not for that.

model/note/note_test.go
taratatach: Ah OK.

taratatach (approved):

PR #4118, opened 2023-09-11
Fix a path inconsistency

nono:
When a conflict happens for two directories with the same path, one directory is renamed. Its path wasn't updated with the new name in this case.


taratatach (approved):
This crucial function would benefit from some tests (and this issue might have been caught earlier).

PR #4121, opened 2023-09-13
Avoid useless calls to imagemagick

nono:
It happens that the stack creates several jobs that will process a tiny thumbnail for the same image with imagemagick. We can avoid the useless calls to imagemagick with a lock + looking at Swift if the thumbnail exists.

### DIFF
--- a/model/vfs/vfsswift/thumbs_v2_v3.go
+++ b/model/vfs/vfsswift/thumbs_v2_v3.go
@@ -103,17 +103,15 @@ func (t *thumbsV2) CreateThumb(img *vfs.FileDoc, format string) (vfs.ThumbFiler,
 
 func (t *thumbsV2) ThumbExists(img *vfs.FileDoc, format string) (bool, error) {
 	name := t.makeName(img.ID(), format)
-	infos, headers, err := t.c.Object(t.ctx, t.container, name)
+	_, headers, err := t.c.Object(t.ctx, t.container, name)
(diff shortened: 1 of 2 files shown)


model/vfs/vfsswift/thumbs_v2_v3.go
```diff
@@ -103,17 +103,15 @@ func (t *thumbsV2) CreateThumb(img *vfs.FileDoc, format string) (vfs.ThumbFiler,
...
+	_, headers, err := t.c.Object(t.ctx, t.container, name)
 	if errors.Is(err, swift.ObjectNotFound) {
 		return false, nil
 	}
 	if err != nil {
 		return false, err
 	}
-	if infos.Bytes == 0 {
-		return false, nil
-	}
```
taratatach, line 115: Isn't there a risk to see empty thumbnails if something happens during their creation? Before this PR I suppose they would be overwritten at some point?!

nono: > Isn't there a risk to see empty thumbnails if something happens during their creation?

Nope, `ThumbExist` is not called to show a thumbnail. It is only called by the worker to known if it can avoid recomputing it.

> Without this PR I suppose they would be overwritten at some point?!

No, it was the case in the early days of cozy-stack, but we have observed lots of jobs for some bad images, and it was decided to not retry generating a thumbnail if it has failed.

taratatach (approved):

PR #4124, opened 2023-09-14
feat: Notify if exceeding clients limit with Pass

taratatach:
The Cozy Pass mobile app and web extension use OAuth clients to
connect to the Cozy so we want to include these in the OAuth clients
number limitation.
However, they do not go through our classic OAuth flow and use their
own so they would not see the blocking modal we added before the
permissions modal when connecting the client would exceed the limit
set on the Cozy.

For this reason, the connection of these app and extension will always
go through but, if this leads to exceeding the clients limit, the user
will see the blocking modal when trying to navigate on their Cozy.
This is why we'll send an e-mail notification to raise attention on
the limit being exceeded and ways to solve this issue in an effort to
anticipate future navigation that could be a bit annoying and
disorientating.


assets/locales/de.po
```diff
@@ -869,6 +869,30 @@ msgstr ""
 msgid "Notifications Disk Quota free text"
 msgstr "Speicherplatz freiräumen"
 
+msgid "Notifications OAuth Clients Subject"
+msgstr "You've exceeded the maximum of devices allowed in your plan"
```
nono, line 873: The maximum **number** of devices?

assets/mails/notifications_oauthclients.mjml
```diff
@@ -0,0 +1,26 @@
+{{define "content"}}
+<mj-text mj-class="title content-medium">
+	<img src="https://files.cozycloud.cc/email-assets/stack/icon-archive.png" width="16" height="16" style="vertical-align:sub;"/>&nbsp;
+	{{t "Notifications OAuth Clients Title"}}
+</mj-text>
+<mj-text mj-class="content-medium">
+	{{t "Notifications OAuth Clients Greeting"}}
+</mj-text>
+<mj-text mj-class="content-medium">
+	<p>
```
nono, line 10: Why the `<p>`? We haven't used that for the other emails, is there something special here?

taratatach: It's just that `<br/>` tags are supposed to have a parent such as `<p>`.

nono: The parent can be a `<div>`, which is what `mj-text` translates to

taratatach: This MDN documentation says otherwise but I can remove it if that's bothering you.

nono (approved):

PR #4127, opened 2023-09-15
fix: OAuth clients limit links open in current tab

taratatach:
When the connected OAuth clients limit is reached on a Cozy, we
display a blocking modal during the OAuth flow when trying to connect
yet another client.
In this situation we want the links displayed in the modal (i.e. to
manage the connected clients or subscribe to another offer) to open in
a new tab as this modal can be displayed in apps windows without any
navigation means.

However, when the modal is displayed when navigating in the Cozy in a
browser because the limit has been exceeded, we don't have this
limitation so the links should open in the same tab.


assets/templates/oauth_clients_limit_exceeded.html
```diff
@@ -26,14 +26,14 @@ <h1 class="h4 h2-md mb-4 text-center">{{t "Too Many OAuth Clients Title" .Client
               {{t "Too Many OAuth Clients Available Actions"}}
             </p>
             {{if .PremiumURL}}
-            <a href="{{.PremiumURL}}" target="_blank" class="btn btn-primary btn-md-lg w-100 mb-2">
+            <a href="{{.PremiumURL}}" {{with .OpenLinksInNewTab}}target="_blank"{{end}} class="btn btn-primary btn-md-lg w-100 mb-2">
```
nono, line 29: I would prefer to keep `if`, and not the fancier but more complicated `with` in templates

nono: It's also optional with `if`

nono (approved):

PR #4138, opened 2023-09-26
Fix onlyoffice behavior

nono:
When a new version of an office document is uploaded to a Cozy, the
stack will now open this new version, even if the previous version of
the document was already in the cache of OO. It was too confusing for
the user that has uploaded a new version to not see their changes if
they open the document in OO. If there was other users editing the
document in OO, they will be alerted of the conflict by cozy-drive and
can refresh their tab to see the new version. And if there were changes,
they are not lost, but will be saved in a conflict document.

### DIFF
--- a/docs/office.md
+++ b/docs/office.md
@@ -121,7 +121,7 @@ Content-Type: application/vnd.api+json
 }
 ```
 
-### GET /office/keys/:key
+### POST /office/keys/:key
 
 If a document is being edited while a new version is uploaded (via the desktop
 for example), the OO webapp should call this endpoint if the user chooses to
@@ -131,7 +131,7 @@ created, so that no work is lost.
 #### Request
 
 ```http
-GET /office/keys/7c7ccc2e7137ba774b7e44de HTTP/1.1
+POST /office/keys/7c7ccc2e7137ba774b7e44de HTTP/1.1
 Host: bob.cozy.example
 ```
 
--- a/docs/office.md
+++ b/docs/office.md
@@ -121,6 +121,66 @@ Content-Type: application/vnd.api+json
 }
 ```
 
+### GET /office/keys/:key
+
+If a document is being edited while a new version is uploaded (via the desktop
+for example), the OO webapp should call this endpoint if the user chooses to
+continue editing the version on which they were working. A conflict file is
+created, so that no work is lost.
+
+#### Request
+
+```http
+GET /office/keys/7c7ccc2e7137ba774b7e44de HTTP/1.1
+Host: bob.cozy.example
+```
+
+#### Response
+
+```http
+HTTP/1.1 200 OK
+Content-Type: application/vnd.api+json
+```
+
+```json
+{
+  "data": {
+    "type": "io.cozy.files",
+    "id": "32e07d806f9b0139c541543d7eb8149c",
+    "meta": {
+      "rev": "3-18c04daba326"
+    },
+    "attributes": {
+      "type": "file",
+      "name": "slideshow (2).pptx",
+      "trashed": false,
+      "md5sum": "ODZmYjI2OWQxOTBkMmM4NQo=",
+      "created_at": "2023-09-30T21:42:05Z",
+      "updated_at": "2023-09-30T22:38:04Z",
+      "tags": [],
+      "metadata": {},
+      "size": 12345,
+      "executable": false,
+      "class": "slide",
+      "mime": "application/vnd.ms-powerpoint",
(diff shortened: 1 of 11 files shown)


docs/office.md
```diff
@@ -121,6 +121,66 @@ Content-Type: application/vnd.api+json
 }
 ```
 
+### GET /office/keys/:key
```
taratatach, line 124: Shouldn't this endpoint rather be named something like `/office/files/:key` since it returns an `io.cozy.file` and not a key?

model/office/file_by_key.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"bytes"
+
+	"github.com/cozy/cozy-stack/model/instance"
+	"github.com/cozy/cozy-stack/model/vfs"
+)
+
+func GetFileByKey(inst *instance.Instance, key string) (*vfs.FileDoc, error) {
+	detector, err := GetStore().GetDoc(inst, key)
```
taratatach, line 11: What's the idea behind the `detector` name? I would expect a `doc` variable here (as is used in other parts of the office code).

docs/office.md
taratatach: I would expect a `POST` verb here since this route can create a new file.

model/office/file_by_key.go
```diff
@@ -0,0 +1,44 @@
+package office
+
+import (
+	"bytes"
+
+	"github.com/cozy/cozy-stack/model/instance"
+	"github.com/cozy/cozy-stack/model/vfs"
+)
+
+func GetFileByKey(inst *instance.Instance, key string) (*vfs.FileDoc, error) {
```
taratatach, line 10: I find this function name a bit misleading as a conflict can be generated.

model/office/file_by_key.go
nono: It's an id+rev+md5sum used to detect if a conflict has happened on this file. `doc` is used when we have a full document with the other fields (like name and size).

model/office/file_by_key.go
nono: Technically, the conflict has already happened, it was just not yet materialized on the Cozy. But, a file has been changed in OO and a new version has been uploaded, the stack will need to save the OO version in a new file. We just want to make it now, not wait for the OO callback.

taratatach: I understand the goal. I'm just saying that I would not expect an `io.cozy.file` to be created by calling a function named `GetFile…`.

docs/office.md
nono: I would prefer to keep `/office/keys/:key`, but you are right that it should be a `POST`

model/office/file_by_key.go
nono: OK, I have renamed it to `EnsureFileForKey`

taratatach (dismissed):

docs/office.md
nono: @cballevre I have changed the HTTP verb to `POST`

model/office/callback.go
```diff
@@ -148,6 +155,7 @@ func saveFile(inst *instance.Instance, detector conflictDetector, downloadURL st
 	newfile.UpdatedAt = time.Now()
 	newfile.CozyMetadata.UpdatedAt = newfile.UpdatedAt
 	newfile.CozyMetadata.UploadedAt = &newfile.UpdatedAt
+	newfile.CozyMetadata.UploadedBy = &vfs.UploadedByEntry{Slug: OOSlug}
```
cballevre, line 158: @nono I think we have a problem here. From what I understood, the uploadedBy is only changed when the file is created. So after this route is called, its value always remains at `{"slug": "onlyoffice-server" }`.  For the front, it means that the modals are no longer displayed from this point. A more effective way would be to add an entry in updatedByApps and based the front on it. What do you think?

taratatach: In the same way, `CozyMetadata.UploadedAt` should not be modified either. We don't modify these two attributes when uploading a new version via Drive or Desktop.

nono: We do modify these two attributes when a file is uploaded (creation or content modification). It's the stack that does the job. IS it OK for you?

cballevre: I must have a bug in local, the stack does not modify the fields and it always remains in onlyoffice-server after a modification by OnlyOffice 🙁

cballevre (approved):
Its working as expected for me 👌

PR #4147, opened 2023-10-03
Make the initial synchronization of a sharing faster

nono:


model/sharing/upload.go
```diff
@@ -52,25 +53,30 @@ func (s *Sharing) Upload(inst *instance.Instance, errors int) error {
...
-			errm = multierror.Append(errm, err)
-		}
-		if more {
-			members = append(members, m)
-		}
+	done := true
+	g, _ := errgroup.WithContext(context.Background())
+	for i := range members {
+		m := members[i]
+		g.Go(func() error {
```
taratatach, line 60: Don't we need to pass `m` as an argument to the goroutine to create a closure?

nono: `errgroup` doesn't allow to pass arguments to the goroutine. That's why I have written `for i := range members { m := members[i]` instead of the shorter `for _, m := range members`.  It forces `m` to have the correct range.

In the future, I hope that https://github.com/golang/go/wiki/LoopvarExperiment will make things easier.

taratatach: Thanks for the explanation.

taratatach (approved):

PR #4149, opened 2023-10-04
Make 2FA codes valid for 15 minutes

nono:


model/instance/auth.go
```diff
@@ -25,8 +25,8 @@ const (
 )
 
 var twoFactorTOTPOptions = totp.ValidateOpts{
-	Period:    30, // 30s
-	Skew:      10, // 30s +- 10*30s = [-5min; 5,5min]
+	Period:    60, // 60s
+	Skew:      14, // 60s +- 14*60s = [-14min; 15min]
```
taratatach, line 29: nit: I've not checked how `Skew` is really used but 60 s +- 14 min would actually be [-13 min; 15 min].

taratatach (dismissed):

PR #4150, opened 2023-10-06
fix: Preserve not_synchronized_on on shared dirs

taratatach:
When a member of a sharing modifies a directory within this sharing
then the `io.cozy.files` doc on the other members' Cozy will lose
their `not_synchronized_on` attribute.

This means the directory will get synchronized with their Desktop
client again while they haven't modified the configuration themselves.

This is because we forgot to keep the existing attribute when updating
the document in the sharing indexer.

### DIFF
--- a/model/sharing/files_test.go
+++ b/model/sharing/files_test.go
@@ -1,6 +1,7 @@
 package sharing
 
 import (
+	"strings"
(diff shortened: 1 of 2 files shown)


nono (approved):
Good catch

PR #4155, opened 2023-10-10
Fix a source of VFS inconsistencies

nono:
CouchDB is using ICU algorithm for views and mango requests. It can be surprizing, and it was the source of a bug in the stack. When a directory 'EDF' is trashed, and a directory 'EDF⚡️' was nearby, the path of this directory can be incorrectly modified to be put in the trash.

The system test 'filepath_operations.rb' was restored from a6a442b9075cef99667b36b7325b0a8b25eaf5b2, and updated to test this new case.

### DIFF
--- a/tests/system/tests/filepath_operations.rb
+++ b/tests/system/tests/filepath_operations.rb
@@ -1,0 +1,68 @@
+require_relative '../boot'
+require 'minitest/autorun'
+require 'pry-rescue/minitest' unless ENV['CI']
+
+describe "The VFS" do
+  it "is able to deal with special cases on filepaths (encoding, case sensitivity)" do
+    Helpers.scenario "filepath_operations"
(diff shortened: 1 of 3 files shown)


tests/system/tests/filepath_operations.rb
```diff
@@ -0,0 +1,68 @@
...
+    end
+
+    folder.remove inst
+    Folder.clear_trash inst
+
+    # Emojis can be tricky too
+    folder = Folder.create inst, name: "EDF"
+    Folder.create inst, name: "EDF⚡️"
+    folder.remove inst
+    assert_equal inst.fsck, []
```
taratatach, line 64: Maybe we could also check that `EDF⚡️` still exists?!

taratatach (approved):

PR #4157, opened 2023-10-10
Update JWT lib to v5

nono:


model/oauth/client.go
```diff
@@ -764,9 +764,9 @@ func validToken(i *instance.Instance, audience, token string) (permission.Claims
 		return claims, false
 	}
 	// Note: the refresh and registration tokens don't expire, no need to check its issue date
-	if claims.Audience != audience {
+	if len(claims.Audience) != 1 || claims.Audience[0] != audience {
```
taratatach, line 767: Adding an `Audience()` method to the `Claims` struct would help here. Something like this:

```go
func (c *Claims) Audience() String {
  if len(c.Audience) > 0 {
    return c.Audience[0]
  }
  return ""
}
```

model/office/open.go
```diff
@@ -94,7 +94,12 @@ func (o *apiOfficeURL) sign(cfg *config.Office) (string, error) {
 }
 
 // Valid is a method of the jwt.Claims interface
```
taratatach, line 96: Leftover?

nono: Yes

model/oauth/client.go
nono: This code does not compile

taratatach: The returned type needs to be `string`, not `String` and there might be other syntax errors but do you get my point?

nono: We could use another name, like `func (c *Claims) AudienceString() string {}`

nono: @taratatach what do you think?

taratatach: Sorry, I missed your answers. Your proposition is totally fine by me :).

taratatach (approved):

PR #4161, opened 2023-10-12
Show memberID in sharings check reports

sblaisot:
add member ID in sharings check reports


model/sharing/sharing.go
```diff
@@ -1568,6 +1568,7 @@ func (s *Sharing) checkSharingTreesConsistency(inst *instance.Instance, ownerDoc
 						"member":    m.Domain,
 						"ownerDoc":  ownerDoc,
 						"memberRev": memberDoc.Rev(),
+						"memberID":  memberDoc.M["_id"].(string),
```
taratatach, line 1571: There's an `ID()` method to do that :)
```suggestion
						"memberID":  memberDoc.ID(),
```

sblaisot: thanks, updated

taratatach (approved):

PR #4162, opened 2023-10-12
Add debian packaging

sblaisot:
This is the first step for the new selfhosting packaging. It produces packages for the last two debian and ubuntu versions as well as an upgrade package for the only currently supported debian version (Debian 10 buster)

It also upload deb packages as artifact to the release on each version tag

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,7 +2,7 @@ name: Create releases
 on:
   push:
     tags:
-    - '*.*.*' # Matching a version number like 1.4.19
+      - "*.*.*" # Matching a version number like 1.4.19
 jobs:
   publish_on_release:
     runs-on: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


debian/migrate-collect-to-home.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
```
nono, line 1: I think we can remove this script. The collect app has been removed 5 or 6 years ago.

scripts/packaging/installrequirements.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+set -euo pipefail
+
+GOVERSION="${GOVERSION:-1.20.4}"
```
nono, line 5: We can use a more recent version of go, like 1.21.3, or 1.20.10

debian/migrate-collect-to-home.sh
sblaisot: Removed

scripts/packaging/installrequirements.sh
sblaisot: Switched to `actions/setup-go@v4` using go version '1.21.x'

nono (approved):

sblaisot:
ping @ArchangeGabriel for information. We are improving our debian package build. maybe there are some inspiration for arch packaging.

PR #4172, opened 2023-10-17
Unify konnector execution scripts across node versions

sblaisot:
Merge various konnector-node-run scripts into one general-purpose script working with all node versions.

This will ease package creation and compatibility.


scripts/konnector-node-run.sh
```diff
@@ -1,11 +1,28 @@
...
+fi
+
+if ! [ -x "${NODE_BIN}" ]; then
+  >&2 echo "Unable to find nodejs binary, exiting..."
+  exit 1
+fi
+
+NODE_VERSION="$(${NODE_BIN} --version)"
+if [ "${NODE_VERSION%%.*}" = "v12" ]; then
+  NODE_OPTS="--max-http-header-size=16384 --tls-min-v1.0 --http-parser=legacy"
```
nono, line 16: Maybe it's time to no longer support node12

nono (approved):

scripts/konnector-node-run.sh
sblaisot: too soon, we still only have debian10/node12 selfhosted in the wild so we cannot get rid of node12 before we release selfhosted packages for debian11/node16+ and debian12/node16+

We need an upgrade path for existing selfhosted isntallations

sblaisot:
ping @ArchangeGabriel as this could affect arch linux packaging

PR #4178, opened 2023-10-19
Add cache for the Swift AppFS

nono:
When loading html, js, css, and images for a webapp or a konnector, the stack makes a call to the AppFS server. With many users and a Swift cluster, we can improve latency by using a cache.

### DIFF
--- a/pkg/appfs/server.go
+++ b/pkg/appfs/server.go
@@ -16,11 +16,13 @@ import (
 	"path"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/andybalholm/brotli"
 	"github.com/cozy/cozy-stack/pkg/consts"
 	web_utils "github.com/cozy/cozy-stack/pkg/utils"
+	lru "github.com/hashicorp/golang-lru/v2"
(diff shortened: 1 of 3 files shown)


pkg/appfs/server.go
```diff
@@ -99,19 +101,52 @@ func (g gzipReadCloser) Close() error {
...
+}
+
+var cache *lru.Cache[string, cacheEntry]
+var initCacheOnce sync.Once
+
 // NewSwiftFileServer returns provides the apps.FileServer implementation
 // using the swift backend as file server.
 func NewSwiftFileServer(conn *swift.Connection, appsType consts.AppType) FileServer {
+	initCacheOnce.Do(func() {
+		c, err := lru.New[string, cacheEntry](128)
```
taratatach, line 116: Why such a small cache? Don't we have more konnectors and applications already?

taratatach (approved):

pkg/appfs/server.go
nono: Well, it can take a lot of RAM. I prefer to start small and see later to increase.

PR #4189, opened 2023-10-23
Add completion configuration in deb package

sblaisot:


debian/rules
```diff
@@ -11,7 +11,10 @@ override_dh_auto_build:
 	go build -o $(CURDIR)/bin/cozy-stack -ldflags "\
 		-X github.com/cozy/cozy-stack/pkg/config.Version=$(UPSTREAM_VERSION) \
 		-X github.com/cozy/cozy-stack/pkg/config.BuildTime=$(BUILD_TIME) \
-		-X github.com/cozy/cozy-stack/pkg/config.BuildMode=production"
+		-X github.com/cozy/cozy-stack/pkg/config.BuildMode=production" &&\
+	$(CURDIR)/bin/cozy-stack completion bash > $(CURDIR)/bin/cozy-stack-completion.bash &&\
+	$(CURDIR)/bin/cozy-stack completion fish > $(CURDIR)/bin/cozy-stack-completion.fish &&\
+	$(CURDIR)/bin/cozy-stack completion zsh > $(CURDIR)/bin/cozy-stack-completion.zsh
```
nono, line 17: By curiosity, why do we rebuild the completion scripts instead of using those in the git repository (https://github.com/cozy/cozy-stack/tree/master/scripts/completion) ?

sblaisot: because "6 months ago"

image

sblaisot: (maybe there has been no change in cli commands for the last 6 monthes, but I'm not sure. Rebuilding on each package ensure they always be up to date.)

nono (approved):

PR #4191, opened 2023-10-24
Remove hooks scripts

nono:
They were not used, and removing them can make the code a bit cleaner.


taratatach:
Are we sure nobody uses this? Including outside Cozy.
We should probably deprecate it first.

nono:
> Are we sure nobody uses this? Including outside Cozy.

I really don't think someone is using it. This feature was meant for automations with many Cozy instances. And if someone is using without a support contract from Cozy Cloud, well, it's their issue.

taratatach (approved):

PR #4223, opened 2023-11-14
Add the notion of sponsorships and their logos

nono:


taratatach (commented):
Do we want to store Sponsors or Sponsorships? If we store Sponsors this means we won't be able to have multiple sponsorships from the same sponsor, right?

taratatach (approved):

PR #4235, opened 2023-11-23
Avoid double admin password request on instances add

sblaisot:
When adding a new instance and cozy-stack admin password has not been set in the environment variable, then it asks twice for admin password, one time to effectively create the instance and then another time to check if all required applications have been installed.

This pull request get rid of code duplication between `adminClient.NewInstanceClient()` and `newClientSafe()` and create a new client from adminClient on instance creation to avoid asking for the admin password more than once.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -12,7 +12,6 @@ import (
 	"github.com/cozy/cozy-stack/client/request"
 	build "github.com/cozy/cozy-stack/pkg/config"
 	"github.com/cozy/cozy-stack/pkg/config/config"
-	"github.com/cozy/cozy-stack/pkg/consts"
 	"github.com/cozy/cozy-stack/pkg/tlsclient"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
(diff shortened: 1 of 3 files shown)


taratatach (approved):

cmd/root.go
```diff
@@ -57,32 +56,7 @@ func newClientSafe(domain string, scopes ...string) (*client.Client, error) {
...
-		Scope:    scopes,
-	})
-	if err != nil {
-		return nil, err
-	}
-
-	httpClient, clientURL, err := tlsclient.NewHTTPClient(tlsclient.HTTPEndpoint{
-		Host:      config.GetConfig().Host,
-		Port:      config.GetConfig().Port,
-		Timeout:   15 * time.Minute,
```
taratatach, line 73: The timeout here is different than the one in `AdminClient.NewInstanceClient`. I don't know if this is important or not.

sblaisot: Oh, thanks, I didn't notice.

The difference comes from an earlier timeout increase that missed one part.

I updated it

PR #4238, opened 2023-11-27
Revert "Remove an old hack for migration path on references"

nono:
This reverts commit 21e94d28be2cc65cc23110328e82b115820cb2d1.

The hack is still needed, as some Cozy instances have references in the bogus format in production.

### DIFF
--- a/web/files/references.go
+++ b/web/files/references.go
@@ -77,6 +77,17 @@ func ListReferencesHandler(c echo.Context) error {
 	if len(resCount.Rows) > 0 {
 		count = int(resCount.Rows[0].Value.(float64))
 	}
+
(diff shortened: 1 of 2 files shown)


web/files/references.go
```diff
@@ -77,6 +77,17 @@ func ListReferencesHandler(c echo.Context) error {
 	if len(resCount.Rows) > 0 {
 		count = int(resCount.Rows[0].Value.(float64))
 	}
+
+	// XXX Some references can contains %2f instead of in the id (legacy),
```
taratatach, line 81:
```suggestion
	// XXX Some references can contain `%2f` instead of `/` in the id (legacy),
```

taratatach (approved):

Issue #4242, opened 2023-11-30
Docker feature request: moving user/group creation and ownership change to Dockerfile


jessebot:
Many users have security requirements to run applications in a way that they cannot modify the root file system or run commands as root.

These commands in the production docker image all require elevated privileges:
https://github.com/cozy/cozy-stack/blob/0fe78134b2d09c73813be48274c66ed8582328e6/scripts/docker/production/docker-entrypoint.sh#L11-L18

Since those commands are all in the entrypoint.sh, they run everytime the Docker container is restarted, which in some environments, such as Kubernetes, can happy more often due to the pod being moved to another node or scaled up or down in some way.

I'd like to see those privileged commands moved to the Dockerfile, similar to how nginx does their unprivileged docker container:
https://github.com/nginxinc/docker-nginx-unprivileged/blob/main/Dockerfile-debian.template

The downside is that you have to set pick a UID/GID and stick with it, but the up-side is that users can use this docker container in secure environments and know ahead of time the UID/GID the user runs as without having to allow the container root privileges.

Currently, if I try to run this docker container unprivileged, I get:

```
==========================================================================

Starting /usr/local/bin/docker-entrypoint.sh, Thu Nov 30 10:22:52 UTC 2023

==========================================================================

groupadd: Permission denied.

groupadd: cannot lock /etc/group; try again later.
```

I'm happy to submit a PR for this work if you'd like.

sblaisot:
Hi @jessebot 

Thank you for your feedback. I will take a look at it in the next few days.

jessebot:
Thanks for your help here and in the other issue! 🙏 

If it is helpful, this is what I ended up doing to unblock the helm chart I was writing, since I do everything kubernetes native:
- [Dockerfile](https://github.com/jessebot/cozy-stack/blob/non-root-docker-image/scripts/docker/production/Dockerfile)
- [entrypoint.sh](https://github.com/jessebot/cozy-stack/blob/non-root-docker-image/scripts/docker/production/docker-entrypoint.sh)

This is the helm chart I wrote that I'm happy to donate if this project would like it:
https://github.com/small-hack/cozy-helm-chart/

jessebot:
> The kubernetes use case is far beyond our simple selfhosting target.

Kubernetes is more and more the default use case for self-hosting these days. If you ever need help with that, please feel free to reach out.

[...]

PR #4243, opened 2023-11-30
create docker user and group outside of docker-entrypoint.sh

jessebot:
This moves user and group creation outside of the entrypoint script and into the Dockerfile. This would allow the docker image to run unprivileged as it won't need to run `useradd` or `groupadd` every time it restarts. It also sets the run user to be cozy.

This would help fix https://github.com/cozy/cozy-stack/issues/4242


nono:
As explained on https://github.com/cozy/cozy-stack/issues/4242, this is not something we are ready to accept and maintain.

PR #4249, opened 2023-12-04
Fix the empty trash handler

nono:
There was a panic when trying to get the ThumbFS, because the swift layout was missing from the given prefixer.

### DIFF
--- a/model/note/delete.go
+++ b/model/note/delete.go
@@ -1,10 +1,14 @@
 package note
 
 import (
+	"fmt"
+	"runtime"
+
 	"github.com/cozy/cozy-stack/model/instance"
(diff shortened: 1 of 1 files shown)


taratatach:
Does this PR fix the underlying issue or is it meant to properly catch and log the panic instead?

nono:
It fixes the issue (and also properly catch and log the panic).

taratatach:
So it all resides in https://github.com/cozy/cozy-stack/pull/4249/files#diff-ece0a6012820764c17eff9f5b9a9f56ef50d250fc3bc632bdf2236e59b5c5e3dR43?

taratatach (approved):

PR #4260, opened 2023-12-13
Optimize sharing for instances on the same stack

nono:
When two instances are on the same stack and there is a sharing between for files, the stack can optimize the share-upload worker. Before this change, the content of the file was fetched from the VFS of the source instance, then uploaded to the destination instance, which copies it to its VFS. Now, the stack just makes a copy from one VFS to the other. It's a lot lighter on I/O, and should help making the share-upload worker faster.

### DIFF
--- a/model/sharing/upload.go
+++ b/model/sharing/upload.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/cozy/cozy-stack/client/request"
 	"github.com/cozy/cozy-stack/model/instance"
+	"github.com/cozy/cozy-stack/model/instance/lifecycle"
 	"github.com/cozy/cozy-stack/model/vfs"
 	"github.com/cozy/cozy-stack/pkg/config/config"
 	"github.com/cozy/cozy-stack/pkg/consts"
@@ -28,6 +29,10 @@ type UploadMsg struct {
 	SharingID string `json:"sharing_id"`
 	Errors    int    `json:"errors"`
 }
+
+// fileCreatorWithContent is a function that can be used to create a file in
+// the given VFS. The content comes from the function closure.
+type fileCreatorWithContent func(fs vfs.VFS, newdoc, olddoc *vfs.FileDoc) error
 
 // Upload starts uploading files for this sharing
 func (s *Sharing) Upload(inst *instance.Instance, errors int) error {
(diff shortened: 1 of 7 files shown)


model/sharing/upload.go
```diff
@@ -350,6 +366,43 @@ func (s *Sharing) uploadFile(inst *instance.Instance, m *Member, file map[string
...
+	parts := strings.SplitN(src.Domain, ":", 2)
+	if len(parts) > 1 {
+		srcPort = parts[1]
+	}
+	parts = strings.SplitN(dst.Domain, ":", 2)
+	if len(parts) > 1 {
+		dstPort = parts[1]
+	}
+	return srcPort == dstPort
+}
```
taratatach, line 380: Cannot this be replaced with `src.InstanceHost() == dst.InstanceHost()`?

model/vfs/vfsafero/impl.go
```diff
@@ -572,6 +575,30 @@ func (afs *aferoVFS) RevertFileVersion(doc *vfs.FileDoc, version *vfs.Version) e
...
+		return err
+	}
+
+	_, err = io.Copy(fd, content)
+	_ = content.Close()
+	if err != nil {
+		return err
+	}
+
+	return fd.Close()
```
taratatach, line 599: Shouldn't this be defered in case we run into an error?

model/vfs/vfsswift/impl_v3.go
```diff
@@ -591,6 +594,101 @@ func (sfs *swiftVFSV3) RevertFileVersion(doc *vfs.FileDoc, version *vfs.Version)
...
+	if newsize > maxsize {
+		return vfs.ErrFileTooBig
+	}
+
+	newpath, err := sfs.Indexer.FilePath(newdoc)
+	if err != nil {
+		return err
+	}
+	if strings.HasPrefix(newpath, vfs.TrashDirName+"/") {
+		return vfs.ErrParentInTrash
```
taratatach, line 620: Why don't we check the value of `vfs.OptionsAllowCreationInTrash(opts))` here?

model/vfs/vfsswift/impl_v3.go
```diff
@@ -591,6 +594,101 @@ func (sfs *swiftVFSV3) RevertFileVersion(doc *vfs.FileDoc, version *vfs.Version)
 	return sfs.Indexer.DeleteVersion(version)
 }
 
+func (sfs *swiftVFSV3) CopyFileFromOtherFS(
```
taratatach, line 597: Looks like a great part of this function could be shared with `CreateFile` (as a guard method maybe?).

taratatach: Seems also that `CopyFile` could benefit from some of these checks.

taratatach: Looks like the end of the function could also be shared (versions management, modifications when uploading to the trash…).

nono: The exception is maybe the version management, which is a bigger block, I will look for it.

model/vfs/vfsswift/impl_v3.go
taratatach: OK. Thanks for the explanation.

model/vfs/vfsswift/impl_v3.go
taratatach: I probably overlooked things but I wanted to bring it up to you. Great if this can lead to some code sharing :).

model/sharing/upload.go
taratatach: You're right. I read that function too quickly (I read "host" as machine host rather than instance host).

taratatach (approved):
LGTM once the last comment is addressed.

PR #4262, opened 2023-12-15
feat: Log worker_id client

Crash--:
When sending logs from client side konnector, let's set the worker_id to client. Like that, it's easier to monitor those konnectors.

Cozy-Stack already add this information when updating the job status.

### DIFF
--- a/web/apps/apps.go
+++ b/web/apps/apps.go
@@ -253,6 +253,15 @@ func logsHandler(appType consts.AppType) echo.HandlerFunc {
 			}
 		}
 
+		clientSide := false
+		if appType == consts.KonnectorType {
+			man, err := app.GetKonnectorBySlug(inst, slug)
+			if err != nil {
+				return wrapAppsError(err)
+			}
+			clientSide = man.ClientSide()
+		}
+
 		var logs []AppLog
 		if err := json.NewDecoder(c.Request().Body).Decode(&logs); err != nil {
 			return jsonapi.BadJSON()
@@ -262,6 +271,9 @@ func logsHandler(appType consts.AppType) echo.HandlerFunc {
 			WithField("slug", slug).
 			WithField("job_id", c.QueryParam("job_id"))
 
+		if clientSide {
+			l = l.WithField("worker_id", "client")
+		}
 		for _, log := range logs {
 			level, err := logger.ParseLevel(log.Level)
 			if err != nil {
--- a/web/apps/apps_test.go
+++ b/web/apps/apps_test.go
@@ -85,8 +85,11 @@ func TestApps(t *testing.T) {
 	slug, err := setup.InstallMiniApp()
 	require.NoError(t, err, "Could not install mini app")
 
-	_, err = setup.InstallMiniKonnector()
+	konnectorSlug, err := setup.InstallMiniKonnector()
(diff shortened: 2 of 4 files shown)


Crash--:
Woa, thanks to the tests I see that this route can be used by a webapp and not only for a konnector. Maybe, in that case, the apps that send the logs should also pass the worker_id ? Something like : `WithField("worker_id",c.QueryParam("worker_id"))`? Or maybe we should make a call to couch to fetch the job type and if it's a clientSide job, then we add the client worker_id?

nono:
Ping @Crash-- : what should we do for this PR?

nono:
Ping @Crash--

Crash--:
@nono can you check the latest commit. I changed my mind and I want to get the manifest from the slug if this is a konnector. And then check if the konnector has the `clientSide` attribute from the manifest, but I don't how to do that.

nono:
@Crash-- I have added https://github.com/cozy/cozy-stack/pull/4262/commits/012811167c1642a66b820a8d3b935f35e7db0861 to the PR with the `clientSide` attribute

Crash--:
Oh nice, thanks! I'll try to run & fix the tests.

web/apps/apps.go
```diff
@@ -273,7 +271,7 @@ func logsHandler(appType consts.AppType) echo.HandlerFunc {
 			WithField("slug", slug).
 			WithField("job_id", c.QueryParam("job_id"))
 
-		if clientSide == true {
+		if clientSide {
```
taratatach, line 274: We could simply do `if man.ClientSide() {` now.

nono: No, we can't, `man` is not in the scope here

taratatach: Ah sorry, I missed the folded lines.
The assignment can be simplified though with `clientSide = man.ClientSide()`

nono:
@Crash-- I have fixed the code&tests. If it OK for you, I let you rebase the commits.

nono:
Ping @Crash--

web/apps/apps_test.go
```diff
@@ -653,14 +656,24 @@ func TestApps(t *testing.T) {
...
+			WithHost(testInstance.Domain).
+			WithHeader("Authorization", "Bearer "+token).
+			WithBytes([]byte(`[ { "timestamp": "2022-10-27T17:13:38.382Z", "level": "error", "msg": "This is an error message" } ]`)).
+			Expect().Status(204)
+
+		assert.Equal(t, `time="2022-10-27T17:13:38.382Z" level=error msg="This is an error message" domain=`+domain+" job_id= nspace=jobs slug="+clientSideSlug+" worker_id=client\n", testOutput.String())
+
 		// Send logs for a konnector
-		e.POST("/konnectors/"+slug+"/logs").
+		testOutput.Reset()
```
Crash--, line 669: @nono I'm not sure if this one is still needed?

nono: It is needed, we are using `testOutput` in the previous block of code, and if we don't `Reset` it, we will still have its log message in `testOutput` making the test fail.

Crash--:
Commits are cleaned and rebased. I let you merge :)

nono (approved):

PR #4263, opened 2023-12-15
Make it work with node 20

sblaisot:
Add `--no-experimental-fetch` to node20. This is needed for konnectors and service compatibility


Crash--:
I'm not a big fan of that. If we want to run our services & konnectors with this flag on our production servers, then we should develop & test with this flag. 

`--no-experimental-fetch` disable the new builtin `fetch` method. Since `services` seems to work with this new `fetch` we may want to remove the need of `node-fetch` (for instance) in order to reduce our bundle size / bundle time. But if we do that, it'll not work anymore on our remote server. 

I'm not saying we should not do that, what I'm saying is that: if we have constraints on our remote server, let's add them to our local env / CI in order to test the same program ;).

sblaisot:
I agree `--no-experimental-fetch` can only be a temporary solution.

I removed it from the plain nodejs script which doesn't require it.

The problem is only when running inside nsjail. built-in `fetch` run inside a wasm engine that allocate 10GB of memory address space. nsjail limit memory usage by reducing address space so we need to increase nsjail memory limit by 10GB.

I updated this PR

sblaisot:
ping @nono @Crash--

nono (approved):

PR #4271, opened 2023-12-20
feat: Notify users of plan changes

taratatach:
When we receive a plan change from the cloudery (i.e. a new feature
set that's not a sponsorship), we send a notification (preferably a
push one) to the user to invite them to reload their mobile app and/or
their browser page so flags can be refreshed.

This is especially useful when the user is expecting to gain new
features.


assets/mails/notifications_planchanged.mjml
```diff
@@ -0,0 +1,18 @@
+{{define "content"}}
```
nono, line 1: [nit] I would have put an underscore in the filename between `plan` and `changed`

nono (approved):

assets/mails/notifications_planchanged.mjml
taratatach: I decided to copy the existing pattern for notification e-mails templates but I agree with you.
Let's change that.

assets/mails/notifications_plan_changed.mjml
```diff
@@ -0,0 +1,18 @@
+{{define "content"}}
```
nono, line 1: Can you also update https://github.com/cozy/cozy-stack/blob/master/docs/assets.md#emails ?

taratatach: Yes I saw that it was missing so I added it.

PR #4286, opened 2024-01-10
Fix last-activity for Cozy not yet onboarded

nono:


web/instances/instances.go
```diff
@@ -498,7 +498,9 @@ func lastActivity(c echo.Context) error {
 		}
 		return nil
 	})
-	if err != nil {
+	// If the instance has not yet been onboarded, the io.cozy.sessions
+	// database will not exist.
+	if err != nil && !couchdb.IsNoDatabaseError(err) {
```
taratatach, line 503: Does the `io.cozy.oauth.clients` database exist though?

nono: Yes, we are creating this database and some indexes on it when the instance is created.

taratatach (approved):

PR #4287, opened 2024-01-11
Fix echo for deleted file

nono:


taratatach (approved):
I suppose this fix would not work if Alice's trash was emptied before it receives the creation "echo" from Charlie's instance?!

PR #4298, opened 2024-01-18
Add a route to copy an image from a note

nono:


model/note/image.go
```diff
@@ -178,6 +178,37 @@ func (u *ImageUpload) Close() error {
 	return nil
 }
 
+// CopyImageToAnotherNote makes a copy of an image from one note to be used in
+// another note.
+func CopyImageToAnotherNote(inst *instance.Instance, srcDoc *vfs.FileDoc, imageID string, dstDoc *vfs.FileDoc) (*Image, error) {
```
taratatach, line 183: What's the `srcDoc` argument used for here?

nono: Good question, and the response is nothing. I can remove it.

taratatach (approved):

PR #4302, opened 2024-01-23
feat: Add external ties route

taratatach:
An instance can be linked to third-party services that cannot be
cancelled/unsubscribed from programmatically and require customer
actions.
We call this an external tie as these connections prevent a gracious
instance deletion and users should be made aware of these when trying
to delete their instance.

The first external tie than can exist is a blocking premium
subscription bought via the App Store or Play Store.

### DIFF
--- a/model/cloudery/service.go
+++ b/model/cloudery/service.go
@@ -38,16 +38,10 @@ type SaveCmd struct {
 
 // SaveInstance data into the cloudery matching the instance context.
 func (s *ClouderyService) SaveInstance(inst *instance.Instance, cmd *SaveCmd) error {
-	cfg, ok := s.contexts[inst.ContextName]
-	if !ok {
-		cfg, ok = s.contexts[config.DefaultInstanceContext]
+	client, err := s.getClient(inst)
+	if err != nil {
+		return err
 	}
-
-	if !ok {
-		return fmt.Errorf("%w: tried %q and %q", ErrInvalidContext, inst.ContextName, config.DefaultInstanceContext)
-	}
-
-	client := manager.NewAPIClient(cfg.API.URL, cfg.API.Token)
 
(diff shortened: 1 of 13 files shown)


web/settings/external_ties.go
```diff
@@ -0,0 +1,56 @@
...
+func NewExternalTies(ties *csettings.ExternalTies) jsonapi.Object {
+	return &apiExternalTies{
+		ExternalTies: ties,
+		DocID:        consts.ExternalTiesID,
+	}
+}
+
+func (h *HTTPHandler) getExternalTies(c echo.Context) error {
+	// Any request with a token can ask for the capabilities (no permissions
+	// are required)
```
nono, line 42: I don't think it is true with current code. A token from the CLI or an OAuth client won't give access to this route. It probably doesn't matter, but I would prefer if we can rewrite the comment to make it clear

model/cloudery/service.go
```diff
@@ -60,3 +54,37 @@ func (s *ClouderyService) SaveInstance(inst *instance.Instance, cmd *SaveCmd) er
...
+}
+
+func (s *ClouderyService) getClient(inst *instance.Instance) (*manager.APIClient, error) {
+	cfg, ok := s.contexts[inst.ContextName]
+	if !ok {
+		cfg, ok = s.contexts[config.DefaultInstanceContext]
+	}
+
+	if !ok {
+		return nil, fmt.Errorf("%w: tried %q and %q", ErrInvalidContext, inst.ContextName, config.DefaultInstanceContext)
```
nono, line 84: What does this mean for self-hosted users? Does returning an error for external ties will block instance deletions?

taratatach: I'd say no but that will be up to the Settings app implementation to decide.
However, I've made some manual tests without any cloudery config and no errors were raised.

nono (approved):

PR #4306, opened 2024-01-25
Fix upload trigger creation

nono:
When a sharing is setup and the first replication fails, the upload trigger wasn't created.

### DIFF
--- a/model/sharing/setup.go
+++ b/model/sharing/setup.go
@@ -119,6 +119,12 @@ func (s *Sharing) Setup(inst *instance.Instance, m *Member) {
 		inst.Logger().WithNamespace("sharing").
 			Warnf("Error on setup replicate trigger (%s): %s", s.SID, err)
 	}
+	if s.FirstFilesRule() != nil {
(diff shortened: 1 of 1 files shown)


model/sharing/setup.go
```diff
@@ -119,21 +119,20 @@ func (s *Sharing) Setup(inst *instance.Instance, m *Member) {
...
 		}
-	} else {
-		if s.FirstFilesRule() == nil {
-			return
-		}
-		if err := s.AddUploadTrigger(inst); err != nil {
-			inst.Logger().WithNamespace("sharing").
-				Warnf("Error on setup upload trigger (%s): %s", s.SID, err)
-		}
+	} else if s.FirstFilesRule() != nil {
```
taratatach, line 135: Since we don't return anymore when `s.FirstFileRule() == nil` we'll call `s.NotifyRecipients(inst, m)` anyway.
Is that a problem?

nono: No, it's not a problem.

taratatach (approved):

PR #4309, opened 2024-01-30
Fix sharing stuck because of a missing shared ref

nono:
It happens that the job for creating an io.cozy.shared has been lost (stack restart for example), and we need to do something to avoid having the sharing stuck. The most efficient way to do that is to check that the file is actually in the sharing directory, and if it is the case, to create the missing io.cozy.shared.

### DIFF
--- a/model/sharing/upload.go
+++ b/model/sharing/upload.go
@@ -472,10 +472,19 @@ func (s *Sharing) SyncFile(inst *instance.Instance, target *FileDocWithRevisions
 
 	err = couchdb.GetDoc(inst, consts.Shared, sid, &ref)
 	if err != nil {
-		if couchdb.IsNotFoundError(err) {
+		if !couchdb.IsNotFoundError(err) {
+			return nil, err
+		}
+		// XXX It happens that the job for creating the io.cozy.shared has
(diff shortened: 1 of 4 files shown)


model/sharing/upload.go
```diff
@@ -472,10 +472,19 @@ func (s *Sharing) SyncFile(inst *instance.Instance, target *FileDocWithRevisions
...
+		}
+		// XXX It happens that the job for creating the io.cozy.shared has
+		// been lost (stack restart for example), and we need to do
+		// something to avoid having the sharing stuck. The most efficient
+		// way to do that is to check that the file is actually in the
+		// sharing directory, and if it is the case, to create the missing
+		// io.cozy.shared.
+		ref, err = s.fixMissingShared(inst, current)
+		if err != nil {
 			return nil, ErrSafety
```
taratatach, line 486: Why don't we return the error from `fixMissingShared` as it appears some are `ErrSafety` and some are not?

nono: It is safer to return `ErrSafety` if the fix hasn't worked, as it ensured that the caller will retry.

PR #4320, opened 2024-02-07
feat: Allow app services to send marketing emails

taratatach:
Preserve mail server used for transactional emails from spam detection by
allowing webapps to send marketing emails via a dedicated mail server.

### DIFF
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -89,6 +89,9 @@ example), you can use the --appdir flag like this:
 			cfg.Mail.NativeTLS = false
 			cfg.Mail.DisableTLS = true
 			cfg.Mail.Port = 1025
+			cfg.CampaignMail.NativeTLS = false
+			cfg.CampaignMail.DisableTLS = true
+			cfg.CampaignMail.Port = 1025
 		}
 
 		processes, services, err := stack.Start()
--- a/cozy.example.yaml
+++ b/cozy.example.yaml
@@ -234,6 +234,45 @@ mail:
       username: {{.Env.COZY_BETA_MAIL_USERNAME}}
       password: {{.Env.COZY_BETA_MAIL_PASSWORD}}
 
+# campaign mail service parameters for sending campaign emails via SMTP
+# If campaign_mail.host is empty, the default mail config will be used.
+campaign_mail:
+  # SMTP server host
+  # Defaults to empty string
+  host: smtp.home
+  # SMTP server port
+  # Defaults to 25
+  port: 587
+  # SMTP server username
+  # Defaults to empty string
+  username: {{.Env.COZY_MAIL_USERNAME}}
+  # SMTP server password
+  # Defaults to empty string
+  password: {{.Env.COZY_MAIL_PASSWORD}}
+  # Use SSL connection (SMTPS)
+  # Means no STARTTLS
+  # Defaults to false
+  use_ssl: false
+  # Disable STARTTLS for SMTP server
+  # Means using plain unencrypted SMTP
+  # Defaults to true
+  disable_tls: false
+  # Skip the certificate validation (may be useful on localhost)
+  # Defaults to false
+  skip_certificate_validation: false
+  # Local Name
+  # The hostname sent to the SMTP server with the HELO command
+  # Defaults to empty string
+  local_name: cozy.domain.example
+  # It is also possible to override the campaign mail config per context.
+  contexts:
+    beta:
+      # If the host is set to "-", no mail will be sent on this context
+      host: smtp.cozy.beta
+      port: 587
+      username: {{.Env.COZY_BETA_MAIL_USERNAME}}
+      password: {{.Env.COZY_BETA_MAIL_PASSWORD}}
+
 # location of the database for IP -> City lookups - flags: --geodb
 # See https://dev.maxmind.com/geoip/geoip2/geolite2/
 geodb: ""
--- a/docs/jobs.md
+++ b/docs/jobs.md
@@ -458,6 +458,66 @@ Content-Type: application/vnd.api+json
 HTTP/1.1 204 No Content
 ```
 
+### POST /jobs/campaign-emails
+
+Send a non transactional (or campaign) email to the user via the dedicated
+campaign mail server (configured via `campaign_mail` attributes in the config
+file, or overwritten by context with `campaign_mail.contexts.<name>`
+attributes).
+
+Both the subject and at least one part are required.
+
+#### Request
+
+```http
+POST /jobs/campaign-emails HTTP/1.1
+Content-Type: application/vnd.api+json
+```
+
+```json
+{
+  "data": {
+    "attributes": {
+      "arguments": {
+        "subject": "Checkout the new cool stuff!",
+        "parts": [
+          {
+            "body": "So many new features to check out!",
+            "type": "text/plain"
+          }
+        ]
+      }
+    }
+  }
+}
+```
+
+#### Response
+
+```http
+HTTP/1.1 204 No Content
+```
+
+#### Permissions
+
+To use this endpoint, an application needs a permission on the type
+`io.cozy.jobs` for the verb `POST` and the `sendmail` worker.
+This can be defined like so:
+
+```json
+{
+  "permissions": {
+    "campaign-emails": {
+      "description": "Required to send campaign emails to the user",
+      "type": "io.cozy.jobs",
+      "verbs": ["POST"],
+      "selector": "worker",
+      "values": ["sendmail"]
+    }
+  }
+}
+```
+
 ### GET /jobs/queue/:worker-type
 
 List the jobs in the queue.
--- a/pkg/emailer/init.go
+++ b/pkg/emailer/init.go
@@ -13,8 +13,9 @@ var service *EmailerService
 // - [EmailerService] sending email via an async job
 // - [Mock] with a mock implementation
 type Emailer interface {
-	SendEmail(inst *instance.Instance, cmd *SendEmailCmd) error
-	SendPendingEmail(inst *instance.Instance, cmd *SendEmailCmd) error
+	SendEmail(inst *instance.Instance, cmd *TransactionalEmailCmd) error
+	SendPendingEmail(inst *instance.Instance, cmd *TransactionalEmailCmd) error
+	SendCampaignEmail(inst *instance.Instance, cmd *CampaignEmailCmd) error
 }
 
 // Init the emailer package by setting up a service based on the
@@ -28,6 +29,6 @@ func Init() *EmailerService {
 // SendEmail send a mail to the instance owner.
 //
 // Deprecated: use [EmailerService.SendEmail] instead.
-func SendEmail(inst *instance.Instance, cmd *SendEmailCmd) error {
+func SendEmail(inst *instance.Instance, cmd *TransactionalEmailCmd) error {
 	return service.SendEmail(inst, cmd)
 }
--- a/pkg/emailer/service.go
+++ b/pkg/emailer/service.go
@@ -3,6 +3,7 @@ package emailer
 import (
 	"github.com/cozy/cozy-stack/model/instance"
 	"github.com/cozy/cozy-stack/model/job"
+	"github.com/cozy/cozy-stack/pkg/mail"
 )
 
 // EmailerService allows to send emails.
@@ -17,14 +18,19 @@ func NewEmailerService(jobBroker job.Broker) *EmailerService {
 	return &EmailerService{jobBroker}
 }
 
-// SendEmailCmd contains the informations to send a mail for the instance owner.
-type SendEmailCmd struct {
+// TransactionalEmailCmd contains the information to send a transactional email
+// to the instance owner.
+type TransactionalEmailCmd struct {
 	TemplateName   string
 	TemplateValues map[string]interface{}
 }
 
 // SendEmail sends a mail to the instance owner.
-func (s *EmailerService) SendEmail(inst *instance.Instance, cmd *SendEmailCmd) error {
+func (s *EmailerService) SendEmail(inst *instance.Instance, cmd *TransactionalEmailCmd) error {
+	if cmd.TemplateName == "" || cmd.TemplateValues == nil {
+		return ErrMissingTemplate
+	}
+
 	msg, err := job.NewMessage(map[string]interface{}{
 		"mode":            "noreply",
 		"template_name":   cmd.TemplateName,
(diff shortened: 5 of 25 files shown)


cozy.example.yaml
```diff
@@ -233,6 +233,13 @@ mail:
       port: 587
       username: {{.Env.COZY_BETA_MAIL_USERNAME}}
       password: {{.Env.COZY_BETA_MAIL_PASSWORD}}
+      # Mail server used for marketing communication; if not present, no
+      # marketing emails will be sent on this context.
+      marketing_emailer:
+        host: smtp.cozy.beta
+        port: 587
+        username: {{.Env.COZY_BETA_MAIL_USERNAME}}
+        password: {{.Env.COZY_BETA_MAIL_PASSWORD}}
```
nono, line 242: We should have a default for all contextes (like the classical emailer).

pkg/emailer/init.go
```diff
@@ -14,6 +14,7 @@ var service *EmailerService
 // - [Mock] with a mock implementation
 type Emailer interface {
 	SendEmail(inst *instance.Instance, cmd *SendEmailCmd) error
+	SendMarketingEmail(inst *instance.Instance, cmd *SendEmailCmd) error
```
nono, line 17: You also need to update the mock with this new method

pkg/emailer/service.go
```diff
@@ -62,3 +75,28 @@ func (s *EmailerService) SendPendingEmail(inst *instance.Instance, cmd *SendEmai
...
+		return fmt.Errorf("Missing email subject")
+	}
+	if cmd.Parts == nil {
+		return fmt.Errorf("Missing email content")
+	}
+
+	msg, err := job.NewMessage(map[string]interface{}{
+		"mode":    mail.ModeMarketing,
+		"subject": cmd.Subject,
+		"parts":   cmd.Parts,
```
nono, line 90: Are we using the right layout? The stack has 2 layouts for emails, but I don't remember which one should be used in what cases

web/jobs/jobs.go
```diff
@@ -259,7 +273,38 @@ func contactSupport(c echo.Context) error {
...
+
+	req := apiMarketing{}
+	if _, err := jsonapi.Bind(c.Request().Body, &req); err != nil {
+		return wrapJobsError(err)
+	}
+	subject, _ := req.Arguments["subject"].(string)
+	if subject == "" {
+		return echo.NewHTTPError(http.StatusBadRequest, "Invalid subject")
+	}
+	parts, _ := req.Arguments["parts"].([]mail.Part)
```
nono, line 291: It won't work. Currently, `req.Arguments` is defined as `map[string]interface{}`. When Go unmarshals JSON to it, it will only uses the basic types (nil, bool, float64, string, slices and maps). I think you can use a struct for `req.Arguments` to fix this issue.

pkg/emailer/service.go
taratatach: The layout is not specified when creating these jobs in Ecolyo so I believe we use the default one (and so do we here).

cozy.example.yaml
```diff
@@ -234,6 +234,40 @@ mail:
...
+  disable_tls: false
+  # skip the certificate validation (may be useful on localhost)
+  skip_certificate_validation: false
+  # Local Name
+  # The hostname sent to the SMTP server with the HELO command
+  # Defaults to localhost
+  # flags: --mail-local-name
+  local_name: cozy.domain.example
+# It is also possible to override the mail config per context.
+contexts:
```
nono, line 263: I think it is not correctly indented

web/jobs/jobs.go
```diff
@@ -259,7 +277,30 @@ func contactSupport(c echo.Context) error {
...
+	if _, err := jsonapi.Bind(c.Request().Body, &req); err != nil {
+		return wrapJobsError(err)
+	}
+
+	err := h.emailer.SendCampaignEmail(inst, &emailer.SendEmailCmd{
+		Subject: req.Arguments.Subject,
+		Parts:   req.Arguments.Parts,
+	})
+	if err != nil {
+		return err
```
nono, line 297: I think the response will be a 500 when the subject or the parts are missing, instead of a 4xx. We should fix that by wrapping the error.

worker/mails/mail.go
```diff
@@ -53,17 +54,26 @@ func SendMail(ctx *job.TaskContext) error {
...
+		cfgPerContext = config.GetConfig().CampaignMailPerContext
+	} else {
+		cfgPerContext = config.GetConfig().MailPerContext
+	}
+
 	ctxName := ctx.Instance.ContextName
-	cfgPerContext := config.GetConfig().MailPerContext
 	if ctxConfig, ok := cfgPerContext[ctxName].(map[string]interface{}); ok {
 		if host, ok := ctxConfig["host"].(string); ok && host != "" {
+			host, _ := ctxConfig["host"].(string)
```
nono, line 68: I don't understand this line. Is it a copy-paste error?

taratatach: ah yes, I changed my mind on where to store the campaign mailer config and I left this line.

web/jobs/jobs.go
taratatach: Won't we get an error when unmarshalling the request since the fields don't have the `omitempty` option?

worker/mails/mail.go
```diff
@@ -88,6 +98,17 @@ func SendMail(ctx *job.TaskContext) error {
 			opts.ReplyTo = &mail.Address{Name: name, Email: replyTo}
 		}
 		opts.RecipientName = toAddr.Name
+	case mail.ModeCampaign:
```
sblaisot, line 101: ModeFromStack case and ModeCampaign cas seems to be exactly the same (unless I miss something).

Can't golang group similar cases like we can do in C?

probably something like 

```golang
case mail.ModeFromStack, mail.ModeCampaign:
```

This would avoid code duplication.

sblaisot (commented):
It would be nice, if `campaign_mail` is not defined in config to use `mail` config as default.

This would ease usage on environments where specific non-transactional mail relays are not required, like on selfhosted environments.

cozy.example.yaml
```diff
@@ -234,6 +234,40 @@ mail:
       username: {{.Env.COZY_BETA_MAIL_USERNAME}}
       password: {{.Env.COZY_BETA_MAIL_PASSWORD}}
 
+# campaign mail service parameters for sending campaign emails via SMTP
+campaign_mail:
+  # mail smtp host - flags: --mail-host
```
sblaisot, line 239: flag conflict with transactional mail flag

sblaisot, line 239: flag conflict with transactional email flag

cozy.example.yaml
```diff
@@ -234,6 +234,40 @@ mail:
...
+  host: smtp.home
+  # mail smtp port - flags: --mail-port
```
sblaisot, line 241: flag conflict with transactional email flag

cozy.example.yaml
```diff
@@ -234,6 +234,40 @@ mail:
...
+  port: 587
+  # mail smtp username - flags: --mail-username
```
sblaisot, line 243: flag conflict with transactional email flag

cozy.example.yaml
```diff
@@ -234,6 +234,40 @@ mail:
...
+  username: {{.Env.COZY_MAIL_USERNAME}}
+  # mail smtp password - flags: --mail-password
```
sblaisot, line 245: flag conflict with transactional email flag

cozy.example.yaml
```diff
@@ -234,6 +234,40 @@ mail:
...
+  password: {{.Env.COZY_MAIL_PASSWORD}}
+  # Use SSL connection (SMTPS)
+  # Means no STARTTLS
+  # flags: --mail-use-ssl
```
sblaisot, line 249: flag conflict with transactional email flag

cozy.example.yaml
```diff
@@ -234,6 +234,40 @@ mail:
...
+  use_ssl: false
+  # disable mail STARTTLS
+  # Means using plain unencrypted SMTP
+  # flags: --mail-disable-tls
```
sblaisot, line 253: flag conflict with transactional email flag

cozy.example.yaml
```diff
@@ -234,6 +234,40 @@ mail:
...
+  disable_tls: false
+  # skip the certificate validation (may be useful on localhost)
+  skip_certificate_validation: false
+  # Local Name
+  # The hostname sent to the SMTP server with the HELO command
+  # Defaults to localhost
+  # flags: --mail-local-name
```
sblaisot, line 260: flag conflict with transactional email flag

sblaisot:
> It would be nice, if `campaign_mail` is not defined in config to use `mail` config as default.

said differently, what happen when sending campaign email if campaign_mail is not configured in cozy.yml?

web/jobs/jobs.go
```diff
@@ -807,28 +848,29 @@ func purgeJobs(c echo.Context) error {
...
-	router.POST("/clean", cleanJobs)
-	router.DELETE("/purge", purgeJobs)
-	router.GET("/:job-id", getJob)
-	router.PATCH("/:job-id", patchJob)
+// Register all the `/jobs` routes to the given router
+func (h *HTTPHandler) Register(router *echo.Group) {
+	router.GET("/queue/:worker-type", h.getQueue)
+	router.POST("/queue/:worker-type", h.pushJob)
+	router.POST("/support", h.contactSupport)
+	router.POST("/campaign-emails", h.sendCampaignEmail)
```
sblaisot, line 856: This endpoint allow to send ONE email per call, so I think it should be `/campaign-email` without S

worker/mails/mail.go
```diff
@@ -178,7 +199,11 @@ func doSendMail(ctx *job.TaskContext, opts *mail.Options, domain string) error {
 	email := gomail.NewMessage()
 	dialerOptions := opts.Dialer
 	if dialerOptions == nil {
-		dialerOptions = config.GetConfig().Mail
+		if opts.Mode == mail.ModeCampaign {
+			dialerOptions = config.GetConfig().CampaignMail
```
sblaisot, line 203: IIUC CampaignMail config, unlike Mail, have no defaults (because no flags with default value have been defined) config so we cannot directly use it here without verifying it has been configured.

If it is not defines, I think (correct me if I'm wrong) that Host will have value "" (empty string) which should fail to deliver mail

cozy.example.yaml
```diff
@@ -234,6 +234,40 @@ mail:
...
+  use_ssl: false
+  # disable mail STARTTLS
+  # Means using plain unencrypted SMTP
+  # flags: --mail-disable-tls
+  disable_tls: false
+  # skip the certificate validation (may be useful on localhost)
+  skip_certificate_validation: false
+  # Local Name
+  # The hostname sent to the SMTP server with the HELO command
+  # Defaults to localhost
```
sblaisot, line 259: Are you sure? where is this default defined? I can't find it anywhere

docs/jobs.md
```diff
@@ -458,6 +458,66 @@ Content-Type: application/vnd.api+json
...
 ```
 
+### POST /jobs/campaign-email
+
+Send a non transactional (or campaign) email to the user via the dedicated
+campaign mail server (configured via `campaign_mail` attributes in the config
+file, or overwritten by context with `campaign_mail.contexts.<name>`
+attributes).
+
+Both the subject and at leat one part are required.
```
Crash--, line 468:
```suggestion
Both the subject and at least one part are required.
```

web/jobs/jobs.go
sblaisot: and the documentation has no final S for this endpoint

cmd/serve.go
nono, line 1: The command line flags should be added to this file for compaign mail

pkg/emailer/service_test.go
```diff
@@ -66,3 +67,49 @@ func Test_Email_job_push_error(t *testing.T) {
...
+		})
+		assert.ErrorIs(t, err, ErrMissingContent)
+	})
+
+	t.Run("WithCompleteCmd", func(t *testing.T) {
+		brokerMock.On("PushJob", &inst, mock.MatchedBy(func(req *job.JobRequest) bool {
+			assert.Equal(t, "sendmail", req.WorkerType)
+			assert.JSONEq(t, `{
+    "mode": "campaign",
+	"subject": "Some subject",
```
nono, line 95: The indentation looks weird here

cmd/serve.go
sblaisot: I'm not sure we really need cmdline flags for campaign mail config.
config file and env variables should be sufficient if campaign mail server defaults to transactional mail server if unconfigured

worker/mails/mail.go
taratatach: Good point. They were not the same at first and I didn't think about grouping them afterwards.

web/jobs/jobs.go
taratatach: The documentation is wrong, I'll fix that.

The naming follows the REST conventions (POST request to the resources route to create a resource) same as the `/jobs` route.

pkg/emailer/service_test.go
taratatach: I let the tools deal with this usually, even if the result is not to my taste.
But the tests pass with a more readable indentation so let's go with that :).

cozy.example.yaml
taratatach: I forgot to define the flags with default values. That's done now.

pkg/emailer/service_test.go
nono: If you are speaking of `gofmt`, it doesn't format what is inside a string

cozy.example.yaml
```diff
@@ -235,31 +235,36 @@ mail:
...
 # campaign mail service parameters for sending campaign emails via SMTP
+# If campaign_mail.host is empty, the default mail config will be used.
 campaign_mail:
-  # mail smtp host - flags: --mail-host
+  # SMTP server host
+  # Defaults to empty string
   host: smtp.home
-  # mail smtp port - flags: --mail-port
+  # SMTP server port
+  # Defaults to 0
```
sblaisot, line 244: 25 should be the default.

https://github.com/spf13/viper?tab=readme-ov-file#establishing-defaults

cozy.example.yaml
```diff
@@ -235,31 +235,36 @@ mail:
...
   # Use SSL connection (SMTPS)
   # Means no STARTTLS
-  # flags: --mail-use-ssl
+  # Defaults to false
   use_ssl: false
-  # disable mail STARTTLS
+  # Disable STARTTLS for SMTP server
   # Means using plain unencrypted SMTP
-  # flags: --mail-disable-tls
+  # Defaults to false
```
sblaisot, line 258: Default value should be true

cozy.example.yaml
```diff
@@ -235,31 +235,36 @@ mail:
...
   disable_tls: false
-  # skip the certificate validation (may be useful on localhost)
+  # Skip the certificate validation (may be useful on localhost)
+  # Defaults to false
   skip_certificate_validation: false
   # Local Name
   # The hostname sent to the SMTP server with the HELO command
-  # Defaults to localhost
-  # flags: --mail-local-name
+  # Defaults to empty string
```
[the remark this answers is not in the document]
taratatach: The empty string is the way we detect we should use the default mail config. We'll never have an empty host in the final config.

nono (approved):

sblaisot (approved):

PR #4327, opened 2024-02-12
Fix cloning a note with images

nono:


taratatach (commented):
We should probably check the available disk space before creating the copy (especially if we have images attached)

taratatach (approved):

PR #4333, opened 2024-02-15
feat: Konnector worker honors skip-maintenance-for

taratatach:
The `harvest.skip-maintenance-for` flag can be set to a list of
konnector slugs to unlock these in Harvest and allow requesting an
execution even if it's maintenance mode has been activated either in
the registry or cozy-stack.

However, the flag was not honored in the konnector worker thus
preventing its execution.
We'll now check if the konnector to be executed is part of the list
(if set) and skip the maintenance mode when appropriate.

### DIFF
--- a/model/feature/flag.go
+++ b/model/feature/flag.go
@@ -71,6 +71,38 @@ func (f *Flags) UnmarshalJSON(bytes []byte) error {
 	return nil
 }
 
+func (f *Flags) GetList(name string) ([]interface{}, error) {
+	if f.M[name] == nil {
+		return []interface{}{}, nil
+	}
+
+	value, ok := f.M[name].(map[string]interface{})
+	if !ok {
+		return nil, fmt.Errorf("Flag %s is not a list flag", name)
+	}
+
+	list, ok := value["list"].([]interface{})
+	if !ok {
+		return nil, fmt.Errorf("Flag %s is not a list flag", name)
+	}
+
+	return list, nil
+}
+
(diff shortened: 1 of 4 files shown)


model/feature/flag.go
```diff
@@ -71,6 +71,38 @@ func (f *Flags) UnmarshalJSON(bytes []byte) error {
...
+
+	list, ok := value["list"].([]interface{})
+	if !ok {
+		return nil, fmt.Errorf("Flag %s is not a list flag", name)
+	}
+
+	return list, nil
+}
+
+func (f *Flags) HasListItem(name string, item string) (bool, error) {
```
nono, line 92: `name string, item string` can be shortened to `name, item string`

model/feature/flag.go
```diff
@@ -71,6 +71,38 @@ func (f *Flags) UnmarshalJSON(bytes []byte) error {
...
+	if f.M[name] == nil {
+		return []interface{}{}, nil
+	}
+
+	value, ok := f.M[name].(map[string]interface{})
+	if !ok {
+		return nil, fmt.Errorf("Flag %s is not a list flag", name)
+	}
+
+	list, ok := value["list"].([]interface{})
```
nono, line 84: I'm surprised by the format here. I would have expected `f.M[name]` to be `[]interface{}`, not having to look at a `list` thing. Are you sure about the format?

taratatach: As discussed, this is the format used for list flags.

nono (approved):

PR #4336, opened 2024-02-22
Make the bitwarden OAuth clients linked to cozy-pass web

nono:
It requires a new version of cozy-pass web, with some permissions added to the manifest (like com.bitwarden.folders).


nono:
> Are we sure the previous scope transition is complete and we can remove it?

If someone hasn't used the extension in the last 3 years, I really think they will login again if they want to use it again.

PR #4339, opened 2024-02-26
fix: Import notes' images from revoked sharings 

taratatach:
The `io.cozy.notes.images` documents associated with a Cozy Note are
  not shared with the note itself.
  Therefore, when the sharing is revoked and the note opened by the
  sharing recipient on their own Cozy (i.e. shared notes are opened on
  their creator's Cozy while the sharing is active), included images are
  shown as missing because we can't find their associated
  `io.cozy.notes.images` documents.

  However, the image binaries themselves are stored wihtin the note's
  binary (which is an archive in this case) and thus can be imported on
  the recipient's Cozy as `io.cozy.notes.images` documents and the note
  content can be updated to reference these documents.
  After that, the images will be displayed as expected when opening the
  note.

### DIFF
--- a/model/note/import.go
+++ b/model/note/import.go
@@ -4,6 +4,7 @@ import (
 	"archive/tar"
 	"bytes"
 	"errors"
+	"fmt"
 	"io"
 	"path"
 	"strconv"
@@ -43,7 +44,7 @@ func ImportFile(inst *instance.Instance, newdoc, olddoc *vfs.FileDoc, body io.Re
 	}
 
 	reader := io.TeeReader(body, file)
-	content, err := importReader(inst, newdoc, reader, schema)
+	content, _, err := importReader(inst, newdoc, reader, schema)
 
 	if content != nil {
 		fillMetadata(newdoc, olddoc, schemaSpecs, content)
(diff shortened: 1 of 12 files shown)


model/note/import.go
```diff
@@ -67,60 +69,128 @@ func ImportFile(inst *instance.Instance, newdoc, olddoc *vfs.FileDoc, body io.Re
...
+	}
+
+	content, images, err := importReader(inst, olddoc, file, schema)
+	if err := file.Close(); err != nil {
+		return fmt.Errorf("error while closing note file: %w", err)
+	}
+	if err != nil {
+		return fmt.Errorf("failed to read note content with images: %w", err)
+	}
+	if content == nil || len(images) == 0 {
```
nono, line 95: In which case can we have `content == nil` here? I would have expected `content` to be nil only if `err` wasn't nil, but we would have already returned from this function if that was the case.

model/note/import.go
```diff
@@ -67,60 +69,128 @@ func ImportFile(inst *instance.Instance, newdoc, olddoc *vfs.FileDoc, body io.Re
...
+		inst.Logger().WithNamespace("notes").
+			Infof("No images to import")
+		return nil
+	}
+
+	md := markdownSerializer(images).Serialize(content)
+	body := []byte(md)
+	if hasImages(images) {
```
nono, line 103: Why this condition? We already know that the note has some images, no?

model/note/import.go
```diff
@@ -67,60 +69,128 @@ func ImportFile(inst *instance.Instance, newdoc, olddoc *vfs.FileDoc, body io.Re
...
+	}
+
+	fs := inst.VFS()
+	file, err := fs.OpenFile(olddoc)
+	if err != nil {
+		return fmt.Errorf("failed to open file for note images import: %w", err)
+	}
+
+	content, images, err := importReader(inst, olddoc, file, schema)
+	if err := file.Close(); err != nil {
```
nono, line 89: [nit] maybe rename `err` in `cerr` to avoid confusion with `err` from the previous line

model/note/import.go
```diff
@@ -67,60 +69,128 @@ func ImportFile(inst *instance.Instance, newdoc, olddoc *vfs.FileDoc, body io.Re
...
+			return nil, nil, fmt.Errorf("failed to buffer note content: %w", err)
 		}
 		hasImages = false
 	} else {
 		hasImages = isTar(buf.Bytes())
 	}
 
 	if !hasImages {
+		inst.Logger().WithNamespace("notes").
+			Warnf("no images to import in note: %s", doc.Name())
```
nono, line 144: I don't think we should keep this log message

model/note/import.go
```diff
@@ -67,60 +69,128 @@ func ImportFile(inst *instance.Instance, newdoc, olddoc *vfs.FileDoc, body io.Re
...
 	}
 
 	var content *model.Node
 	var err error
 	var images []*Image
 	defer func() {
 		if err == nil && images != nil {
-			fixURLForProsemirrorImages(content, images)
+			inst.Logger().WithNamespace("notes").
+				Infof("fixing images in note %s: %v", doc.Name(), images)
```
nono, line 159: same

model/note/import.go
```diff
@@ -67,60 +69,128 @@ func ImportFile(inst *instance.Instance, newdoc, olddoc *vfs.FileDoc, body io.Re
...
 			if err != nil {
-				return nil, err
+				return nil, nil, fmt.Errorf("failed to parse note markdown: %w", err)
 			}
 		} else {
 			ext := path.Ext(header.Name)
 			contentType := filetype.ByExtension(ext)
 			upload, erru := NewImageUpload(inst, doc, header.Name, contentType)
+			inst.Logger().WithNamespace("notes").
+				Infof("importing image %s from note %s", header.Name, doc.Name())
```
nono, line 183: same

model/note/import.go
```diff
@@ -130,18 +200,22 @@ func importReader(inst *instance.Instance, doc *vfs.FileDoc, reader io.Reader, s
...
-func fixURLForProsemirrorImages(node *model.Node, images []*Image) {
+func fixURLForProsemirrorImages(node *model.Node, images []*Image, logger *logger.Entry) {
 	if node.Type.Name == "media" {
 		name, _ := node.Attrs["alt"].(string)
+		url, _ := node.Attrs["url"].(string)
+		logger.WithNamespace("note").Infof("fixing url of node %s (%s)", name, url)
 		for _, img := range images {
+			logger.WithNamespace("note").Infof("candidate image: %v", *img)
 			if img.originalName == name {
+				logger.WithNamespace("note").Infof("same name! changing node url")
```
nono, line 211: It looks like you have kept some logs for debugging

model/sharing/open_note.go
```diff
@@ -34,22 +34,22 @@ func (n *apiNoteURL) Fetch(field string) []string            { return nil }
 // Opener can be used to find the parameters for creating the URL where the
 // note can be opened.
 type Opener struct {
```
nono, line 36: We should probably rename `Opener` to `NoteOpener`.

model/sharing/open_note.go
```diff
@@ -34,22 +34,22 @@ func (n *apiNoteURL) Fetch(field string) []string            { return nil }
...
-	*sharing.FileOpener
+	*FileOpener
 }
 
 // Open will return an Opener for the given file.
-func Open(inst *instance.Instance, fileID string) (*Opener, error) {
+func OpenNote(inst *instance.Instance, fileID string) (*Opener, error) {
```
nono, line 41: We also have an `Opener` for `model/office`. What do you think of moving it to `model/sharing` to have a more symmetrical approach?

model/sharing/open_note.go
```diff
@@ -79,10 +79,6 @@ func (o *Opener) GetResult(memberIndex int, readOnly bool) (jsonapi.Object, erro
 }
 
 func (o *Opener) openLocalNote(memberIndex int, readOnly bool) (*apiNoteURL, error) {
-	// If the note came from another cozy via a sharing that is now revoked, we
-	// may need to recreate the trigger.
-	_ = setupTrigger(o.Inst, o.File.ID())
```
nono, line 84: I think we should keep this line for existing notes from a revoked sharing without a trigger.

model/sharing/open_note.go
taratatach: I was thinking about the same thing when I stumbled upon it while changing `Opener` into `NoteOpener` :).
I'll add a commit with this.

model/sharing/open_note.go
taratatach: Sounds like a good idea indeed.

model/note/import.go
taratatach: I forgot to remove this condition after the same other a bit before.

model/note/import.go
```diff
@@ -67,12 +68,66 @@ func ImportFile(inst *instance.Instance, newdoc, olddoc *vfs.FileDoc, body io.Re
...
+	}
+	if content == nil || !hasImages(images) {
+		inst.Logger().WithNamespace("notes").
+			Infof("No images to import")
+		return nil
+	}
+
+	md := markdownSerializer(images).Serialize(content)
+	body := []byte(md)
+	body, err = buildArchive(inst, []byte(md), images)
```
nono, line 100: you can compact the two lines in one `body, err := buildArchive(inst, []byte(md), images)`

PR #4354, opened 2024-03-19
Allow to configure some hosts authorized for confirm auth

nono:


web/auth/confirm.go
```diff
@@ -127,32 +127,41 @@ func ConfirmSuccess(c echo.Context, inst *instance.Instance, state string) error
...
+	if err != nil {
+		return nil, echo.NewHTTPError(http.StatusBadRequest, "bad url: could not parse")
+	}
+
+	if ok := checkRedirectToManager(inst, u); ok {
+		return u, nil
+	}
+
+	for _, host := range config.GetConfig().AuthorizedForConfirm {
+		if host == u.Host {
```
sblaisot, line 143: I wonder if this should be Host only or Host + Port (like in https://tool.domain.example:8080)

sblaisot (approved):
One small question, otherwise LGTM

web/auth/confirm.go
nono: It can be both. https://pkg.go.dev/net/url#URL says:

> Host        string    // host or host:port (see Hostname and Port methods)

PR #4355, opened 2024-03-20
Add cozyMetadata.favorite for io.cozy.files

nono:


zatteo:
Works fine 👍

web/files/files_test.go
```diff
@@ -1025,7 +1025,10 @@ func TestFiles(t *testing.T) {
             "tags": ["bar", "bar", "baz"],
             "name": "moved",
             "dir_id": "` + dirID + `",
-            "executable": true
+            "executable": true,
+			"cozyMetadata": {
+			  "favorite": true
+			}
```
paultranvan, line 1031: The indent looks weird

paultranvan (approved):
:+1: 
@zatteo There is no need to be able to directly create a favorite file?

taratatach (approved):

web/files/files_test.go
nono: Good catch, there is a mix of spaces and tabs

zatteo:
> 👍 @zatteo There is no need to be able to directly create a favorite file?

Did not think about it. But not for the moment.

PR #4360, opened 2024-03-25
Add a reference to sourceAccountIdentifier

nono:
When the stack executes a konnector that imports files, it checks that a directory exists for those files. And if the directory doesn't exist, it creates it. We have added a referencedBy on the directory for the sourceAccountIdentifier to be able to deal with several accounts for the same konnector, as each account should have its own directory.

### DIFF
--- a/worker/exec/konnector.go
+++ b/worker/exec/konnector.go
@@ -326,6 +326,11 @@ func (w *konnectorWorker) ensureFolderToSave(ctx *job.TaskContext, inst *instanc
 		return err
 	}
 
+	var sourceAccountIdentifier string
+	if acc != nil && acc.Metadata != nil {
+		sourceAccountIdentifier = acc.Metadata.SourceIdentifier
+	}
+
 	// 2. Check if the konnector has a reference to a folder
 	start := []string{consts.Konnectors, consts.Konnectors + "/" + w.slug}
 	end := []string{start[0], start[1], couchdb.MaxString}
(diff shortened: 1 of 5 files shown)


worker/exec/konnector.go
```diff
@@ -383,6 +392,12 @@ func (w *konnectorWorker) ensureFolderToSave(ctx *job.TaskContext, inst *instanc
 			Type: consts.Konnectors,
```
taratatach, line 392: Since we can now have 2 references to add to the folder, we should check for their presence on the folder instead of simply checking if none exist.
This way we can recreate a single missing reference.

worker/exec/konnector.go
```diff
@@ -341,10 +346,14 @@ func (w *konnectorWorker) ensureFolderToSave(ctx *job.TaskContext, inst *instanc
 		for _, row := range res.Rows {
 			dir := &vfs.DirDoc{}
 			if err := couchdb.GetDoc(inst, consts.Files, row.ID, dir); err == nil {
-				if !strings.HasPrefix(dir.Fullpath, vfs.TrashDirName) {
-					count++
-					dirID = row.ID
+				if strings.HasPrefix(dir.Fullpath, vfs.TrashDirName) {
+					continue
+				}
+				if !hasCompatibleSourceAccountIdentifier(dir, sourceAccountIdentifier) {
```
taratatach, line 352: Since we require the folder to have a referenced `sourceAccountIdentifier` now, we could move the [early return when `acc` is `nil`](https://github.com/cozy/cozy-stack/pull/4360/files#diff-4719b9e6012f031281e0cc6d5914f4771b18c611845a260472d2024546835f24R365-R368) towards the top of the function.

taratatach (commented):
It seems that we'll need a data migration to add the `SourceIdentifier` in the `Metadata` of all accounts, no?
Otherwise we'll never match any folder with its konnector?!

EDIT: answered by https://github.com/cozy/cozy-stack/pull/4360/files#r1538868420 (we don't need a migration).

worker/exec/konnector.go
nono: We don't require the folder to have a referenced `sourceAccountIdentifier`, it's optional.

nono:
> It seems that we'll need a data migration to add the SourceIdentifier in the Metadata of all accounts, no?

Most of the accounts already have a `sourceAccountIdentifier`.

> Otherwise we'll never match any folder with its konnector?!

No, we can match even without it. Most often, we are using the folder ID from the trigger, but even without it, we can match the folder.

worker/exec/konnector.go
taratatach: Ah yes, I missed the part where `hasCompatibleSourceAccountIdentifier` returns `true` if the folder does not have a `sourceAccountIdentifier`.
This answers my question about the data migration too.

doubleface (approved):

PR #4377, opened 2024-04-17
Avoid using startkey_docid in _all_docs

nono:
It was working, but relying on a bug of CouchDB, which may be fixed in later versions. Let's use a safer version with `startkey` and `endkey`.


taratatach (approved):
Why the need for quotes?

nono:
Well, `pkg/couchdb` already transforms the string in JSON, so there is no need to manually add the quotes (I didn't remember that when starting the PR).

PR #4384, opened 2024-04-22
Forbid IP like 169.254.169.254 in safehttp

### LINKED ISSUE #4380: SafeControl Controller is not checking all Internal URLs
Hi Team,

I found SSRF Vulnerability on Cozy-Stack Package. Also you can use ::1 IPV6 version to bypass it.

- https://github.com/cozy/cozy-stack/blob/26505bbe29a926aa46ed0251c4c9a6721c7135c4/pkg/safehttp/client.go#L80

You can find the POC on the following Playground link. The IsPrivate function did not check all of the internal URLs. So you can use IsLinkLocalUnicast, IsLinkLocalMulticast, IsLoopback functions to do that.

- https://go.dev/play/p/bi3OWV9sOih

Impact

169.254.169.254 is one of the AWS Metadata Server IP address and it should not be accessed by the external users. But If a web server uses the isPrivate function it can be bypassed.

nono:
Safehttp package is used to prevent Server-Side Requests Forgery. It prevents making HTTP requests to private IP addresses for things controlled by the user, like Bitwarden icons. This commit adds some more rules for link-local unicast and multicast.

Fix #4380

### DIFF
--- a/pkg/safehttp/client.go
+++ b/pkg/safehttp/client.go
@@ -77,6 +77,10 @@ func safeControl(network string, address string, conn syscall.RawConn) error {
 		return fmt.Errorf("%s is not a valid IP address", host)
 	}
 
+	if ipaddress.IsUnspecified() || ipaddress.IsLinkLocalUnicast() || ipaddress.IsLinkLocalMulticast() {
+		return fmt.Errorf("%s is not a valid IP address", host)
+	}
+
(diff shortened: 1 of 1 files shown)


sblaisot (approved):

sblaisot:
LGTM

PR #4386, opened 2024-04-23
fix: Redirect to clients paywall only if logged in

taratatach:
When an instance has more connected OAuth clients than their current
  plan and options allows them to, we redirect them to a paywall from
  any web app except the Settings app (because it allows disconnecting
  clients).

  However, if the user is not logged in, the paywall will return an
  unauthorized error.

  Therefore, we'll make sure the redirection to the paywall only occurs
  when the user is logged in (we don't duplicate the logic to redirect
  them to the login page in the paywall middleware).

### DIFF
--- a/web/apps/serve.go
+++ b/web/apps/serve.go
@@ -69,12 +69,6 @@ func Serve(c echo.Context) error {
 	}
 
 	if file == "" || file == route.Index {
-		if !route.Public {
-			if handled, err := middlewares.CheckOAuthClientsLimitExceeded(c); handled {
-				return err
-			}
-		}
-
 		webapp = app.DoLazyUpdate(i, webapp, app.Copier(consts.WebappType, i), i.Registries()).(*app.WebappManifest)
 	}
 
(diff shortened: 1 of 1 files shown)


web/apps/serve.go
```diff
@@ -313,6 +307,12 @@ func ServeAppFile(c echo.Context, i *instance.Instance, fs appfs.FileServer, web
 	sessID := ""
 	if isLoggedIn {
 		sessID = sess.ID()
+	} else if file == "" || file == route.Index {
+		if !route.Public {
+			if handled, err := middlewares.CheckOAuthClientsLimitExceeded(c); handled {
```
nono, line 312: I don't understand. The PR description says:

> Therefore, we'll make sure the redirection to the paywall only occurs when the user is logged in

But, here, we check for the OAuth clients limit only if the user is NOT logged in. What am I missing?

taratatach: You're not missing anything. That's my mistake :man_facepalming:.
I don't know how my manual test worked. I'll fix this and run some more tests.

nono (approved):

PR #4407, opened 2024-05-23
Fix weird behaviours for NextCloud/WebDAV

nono:
When the userId contains a space, the base path for webdav can be incorrectly matched in the xml response for PROPFIND, and the parent directory is no longer excluded. The fix is to encode sooner the base path.


taratatach (approved):

taratatach:
This PR does not fix the issue though.

nono:
@taratatach can you try again?

nono:
It looks like the bug has been fixed without this PR.

PR #4420, opened 2024-06-06
Fix uploading files to NextCloud via WebDAV

nono:
The HTTP request to NextCloud was made with Transfer-Encoding: chunked, because of the way the Content-Length was set. But it looks like NextCloud doesn't support that, and files were created empty. We can avoid that by setting request.ContentLength.

### DIFF
--- a/web/remote/nextcloud.go
+++ b/web/remote/nextcloud.go
@@ -199,8 +199,13 @@ func nextcloudDownstream(c echo.Context) error {
 		return jsonapi.BadRequest(errors.New("missing To parameter"))
 	}
 
+	kind := nextcloud.MoveOperation
+	if c.QueryParam("Copy") != "" {
+		kind = nextcloud.CopyOperation
+	}
+
 	cozyMetadata, _ := files.CozyMetadataFromClaims(c, true)
-	f, err := nc.Downstream(path, to, cozyMetadata)
+	f, err := nc.Downstream(path, to, kind, cozyMetadata)
 	if err != nil {
 		return wrapNextcloudErrors(err)
 	}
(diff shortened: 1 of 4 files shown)


web/remote/nextcloud.go
```diff
@@ -226,7 +226,12 @@ func nextcloudUpstream(c echo.Context) error {
 		return jsonapi.BadRequest(errors.New("missing From parameter"))
 	}
 
-	if err := nc.Upstream(path, from); err != nil {
+	kind := nextcloud.MoveOperation
+	if c.QueryParam("Copy") != "" {
```
taratatach, line 237: Checking the actual content of the param would allow requesters to always pass it with values depending on the calling context.

web/remote/nextcloud.go
```diff
@@ -199,8 +199,13 @@ func nextcloudDownstream(c echo.Context) error {
 		return jsonapi.BadRequest(errors.New("missing To parameter"))
 	}
 
+	kind := nextcloud.MoveOperation
+	if c.QueryParam("Copy") != "" {
```
taratatach, line 205: Same here.

web/remote/nextcloud.go
nono: I don't really see what you would want to do

taratatach: Well, if I'm not mistaken, making a request with `Copy=false` as parameter would still be considered as a copy request.
My suggestion is to check the value of the parameter so only requests with `Copy=true` would make copies.

nono: Instead of `false`, you can use an empty string, if you really want to give `Copy=` in the query-string.

taratatach: Sure, but it's less obvious that's all. And makes calls using a variable a little bit more complex.
It's not a big deal but would make the API a bit nicer to use.

taratatach (approved):

PR #4432, opened 2024-06-27
Manage the Nextcloud trash

nono:


docs/nextcloud.md
```diff
@@ -430,3 +434,127 @@ HTTP/1.1 204 No Content
...
+```json
+{
+  "data": [
+    {
+      "type": "io.cozy.remote.nextcloud.files",
+      "id": "613281",
+      "attributes": {
+        "type": "directory",
+        "name": "Old",
+        "path": "/trash/Old.d93571568",
```
nono, line 467: @cballevre be careful, the path for things inside the trash is a bit weird and cannot be computed with just the parent and the name, so it was added to the response

docs/nextcloud.md
```diff
@@ -430,3 +430,44 @@ HTTP/1.1 204 No Content
...
+      "id": "613281",
+      "attributes": {
+        "type": "directory",
+        "name": "Old",
+        "updated_at": "Tue, 25 Jun 2024 14:31:44 GMT",
+        "etag": "1719326384"
+      },
+      "meta": {},
+      "links": {
+        "self": "https://nextcloud.example.net/apps/files/trashbin/613281?dir=/Old"
```
taratatach, line 473: I'm confused about this URL. Isn't the `dir` param supposed to be the parent directory of the document, the `Old` directory in this case ?

model/nextcloud/nextcloud.go
```diff
@@ -99,48 +100,54 @@ func New(inst *instance.Instance, accountID string) (*NextCloud, error) {
...
-	if err := nc.fillBasePath(&doc); err != nil {
+	if err := nc.fillUserID(&doc); err != nil {
 		return nil, err
 	}
 	return nc, nil
 }
 
 func (nc *NextCloud) Download(path string) (*webdav.Download, error) {
-	return nc.webdav.Get(path)
+	return nc.webdav.Get("/files/" + nc.userID + "/" + path)
```
taratatach, line 119: Can `path` ever be `/`?

model/nextcloud/nextcloud.go
```diff
@@ -99,48 +100,54 @@ func New(inst *instance.Instance, accountID string) (*NextCloud, error) {
...
 }
 
 func (nc *NextCloud) Upload(path, mime string, contentLength int64, body io.Reader) error {
 	headers := map[string]string{
 		echo.HeaderContentType: mime,
 	}
+	path = "/files/" + nc.userID + "/" + path
```
taratatach, line 126: Looks like it could be the right time to add a `filePath(path string)` and a `trashedPath(path string)` methods?!

docs/nextcloud.md
nono: I'm confused by NextCloud too. The correct answer is `https://nextcloud.example.net/apps/files/trashbin/613281?dir=/Old.d93571568` (which is not the path of the parent directory, but the path of the directory its-self, without `/trash` at the beginning).

taratatach: OK, that's kind of the restore path (although slightly modified with some sort of id)?!

model/nextcloud/nextcloud.go
nono: Nope, you can only download a file, and `/` is not a correct path for a file.

docs/nextcloud.md
nono: No, there is another property `trashbin-original-location` for that

taratatach: OK. It's even weirder then :sweat_smile:

model/nextcloud/nextcloud.go
taratatach: Sure but this line is repeated throughout the file. My point is what happens if any of these methods is called with `path == "/"`?
Won't there be an error?

docs/nextcloud.md
cballevre: Interesting clarification. I saw that you added the `path` everywhere. In cozy-client, I add a second attribute named `parentPath`. It corresponds to the folder path. This is to make it easier to replay requests with sift.js ([commit with some explanation](https://github.com/cozy/cozy-client/pull/1488/commits/05293215bef289f0170b617bc278057916b5e778)). Wouldn't it make sense to add it to the back as well, so we wouldn't have to calculate anything on the front?

model/nextcloud/nextcloud.go
nono: It's not called with `path == "/"`. Even if it would have been the case, it's not an error, it just makes the URL looks worse, but it still works.

taratatach (approved):

PR #4444, opened 2024-07-11
feat: Remove default wallpaper as we use now an svg

JF-Cozy:
relative to https://github.com/cozy/cozy-home/pull/2153


nono:
We should merge this PR after a stable version of the home app has been published?

nono (approved):

PR #4445, opened 2024-07-16
Make the stack runnable from default config

paultranvan:
When running the stack after copying the default configuration file, the stack won't start because of errors due to fake paths defined for firebase and encryption.
Let's comment them to ease incoming developers' life.

### DIFF
--- a/cozy.example.yaml
+++ b/cozy.example.yaml
@@ -56,9 +56,9 @@ admin:
 # to generate the keys
 vault:
   # the path to the key used to encrypt credentials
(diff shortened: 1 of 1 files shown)


nono (approved):
I really think that developers should start from a blank file and just add what they need to their configuration file, but, well, in practice, it doesn't seem that developers do that...

paultranvan:
> in practice, it doesn't seem that developers do that...

I can confirm, I witnessed it many times. People tend to prefer a ready-to-work conf file and eventually adapt if required

PR #4450, opened 2024-08-01
Fix downloading a webapp code from a git branch

nono:


pkg/appfs/server.go
```diff
@@ -238,15 +238,19 @@ func (s *swiftServer) ServeCodeTarball(w http.ResponseWriter, req *http.Request,
...
 
 	file, err := s.c.ObjectCreate(s.ctx, s.container, objName, true, "", contentType, nil)
 	if err == nil {
-		_, _ = io.Copy(file, buf)
+		_, _ = io.Copy(file, bytes.NewReader(content))
 		_ = file.Close()
 	}
 
-	return serveContent(w, req, contentType, int64(buf.Len()), buf)
+	return serveContent(w, req, contentType, int64(buf.Len()), bytes.NewReader(content))
```
taratatach, line 253: Shouldn't we look at the size of `content` rather than `buf` now?

taratatach (commented):
Since the buffer returned by `prepareTarball` isn't used anymore in `ServeCodeTarball`, shouldn't we remove it?

PR #4457, opened 2024-09-02
Add external indexers for RAG

nono:


worker/index/index.go
```diff
@@ -0,0 +1,247 @@
...
+				}
+				needIndexation = metadata["md5sum"] != md5sum
+			case 404:
+				needIndexation = true
+			default:
+				return fmt.Errorf("GET status code: %d", res.StatusCode)
+			}
+			if !needIndexation {
+				// TODO we should patch the metadata in the vector db when a
+				// file has been moved/renamed.
```
paultranvan, line 143: There are also additional metadata that might be eventually be useful, e.g. the `metadata.referenceDate`, the `metadata.qualification.label`, ... At some point we will need to list which metadata to index

worker/index/index.go
```diff
@@ -0,0 +1,247 @@
...
+			}.Encode()
+			fs := inst.VFS()
+			fileDoc := &vfs.FileDoc{
+				Type:       consts.FileType,
+				DocID:      change.DocID,
+				DirID:      dirID,
+				DocName:    name,
+				InternalID: internalID,
+			}
+			// TODO notes with images
```
paultranvan, line 164: For now, we don't want to deal with images, right?

cozy.example.yaml
```diff
@@ -140,6 +140,7 @@ jobs:
   # List of available workers:
   #
   #   - "clean-clients":     delete unused OAuth clients
+  #   - "index":             send data to a RAG for being indexed
```
paultranvan, line 143: I'm wondering if it's not worth explicitly name it RAG, since the generic "index" terminology might be confusing with "traditional" index.

paultranvan (approved):
So it begins... :tada:

worker/index/index.go
```diff
@@ -0,0 +1,247 @@
...
+		return "", err
+	}
+	seq, _ := result["last_seq"].(string)
+	return seq, nil
+}
+
+// updateLastSequenceNumber updates the last sequence number for this
+// indexation if it's superior to the number in CouchDB.
+func updateLastSequenceNumber(inst *instance.Instance, doctype, seq string) error {
+	result, err := couchdb.GetLocal(inst, doctype, "index")
```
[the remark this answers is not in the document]
nono: It's lighter, we don't have to give it a name, create a database per user for it, etc.

worker/index/index.go
nono: Right, but currently, the file is not indexed at all (not even the text part). The stack sends the tarball, not the markdown, and the RAG just skip this file.

cozy.example.yaml
nono: I have hesitated, but I think the worker could easily evolve to allow other kinds of indexation (like a full-text index in meilisearch), so I kept a generic name

paultranvan: Ok, fair enough

worker/index/index.go
paultranvan: Oh I see. And for notes without images, is it seen as raw text or actual markdown?

nono: Just raw text for the moment

docs/workers.md
```diff
@@ -412,3 +412,9 @@ It can be launched from command-line with:
 ```sh
 $ cozy-stack jobs run migrations --domain example.mycozy.cloud --json '{"type": "to-swift-v3"}'
 ```
+
+## index
+
+This worker is used for sending data to a RAG. It looks at the changes feed for
+the given doctype, send the changes to an external indexer that will generate
+embeddings for the data and put them in a vector database.
```
taratatach, line 420: Since we're using the generic terms `index` and `indexer` shouldn't we stick to that here as well?
This would give something like:

This worker looks for changes to the given doctype in CouchDB's changes feed and send the associated documents to an external indexer.
The index is stored in the indexer's own database.
For `io.cozy.files` documents, the file content itself is sent to the indexer as well.

cozy.example.yaml
taratatach: Maybe we should stick to these generic terms and not talk avec RAGs at all except in the `ai` module documentation then.

worker/index/index.go
```diff
@@ -0,0 +1,242 @@
...
+	"github.com/cozy/cozy-stack/pkg/couchdb/revision"
+)
+
+func init() {
+	job.AddWorker(&job.WorkerConfig{
+		WorkerType:   "index",
+		Concurrency:  runtime.NumCPU(),
+		MaxExecCount: 1,
+		Reserved:     true,
+		Timeout:      5 * time.Minute,
```
taratatach, line 28: Is that enough?

worker/index/index.go
taratatach: Aren't we only sending the file's content? In which case metadata or path changes seem irrelevant.

docs/ai.md
```diff
@@ -46,3 +46,47 @@ can use the vector database to find relevant documents (technically, only some
...
+
+```http
+HTTP/1.1 200 OK
+Content-Type: application/vnd.api+json
+```
+
+```json
+{
+  "data": {
+    "type": "io.cozy.ai.url",
```
taratatach, line 72: Isn't it a bit too generic to use `url` as the last part here since this is only the frontend part and we might have other AI URLs ?
We kinda already do with the external indexers even if they don't have their own doctype for now.

Something like `io.cozy.ai.chatbot` or `io.cozy.ai.open` would be more appropriate in my opinion.

model/rag/chat.go
```diff
@@ -0,0 +1,200 @@
...
+
+type ChatCompletion struct {
+	DocID    string        `json:"_id"`
+	DocRev   string        `json:"_rev,omitempty"`
+	Messages []ChatMessage `json:"messages"`
+}
+
+type ChatMessage struct {
+	Role    string `json:"role"`
+	Content string `json:"content"`
```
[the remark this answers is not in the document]
paultranvan: We might also eventually need a dedicated id for each message : if we enable the possibility for a user to edit a message and re-execute the LLM response on it, we'll need a way to identify the message

pkg/consts/doctype.go
```diff
@@ -142,4 +142,8 @@ const (
 	// NextCloudFiles doc type is used when listing files from a NextCloud via
 	// WebDAV.
 	NextCloudFiles = "io.cozy.remote.nextcloud.files"
+	// ChatCompletions doc type is used for a chat between the user and a chatbot.
+	ChatCompletions = "io.cozy.ai.chat.completions"
```
paultranvan, line 146: I feel like it's more "conversation" than "completion" : completion sounds more like the API action, while the conversation is more the actual data type

model/rag/chat.go
```diff
@@ -0,0 +1,200 @@
...
+func callRAGQuery(inst *instance.Instance, payload map[string]interface{}) (*http.Response, error) {
+	ragServer := inst.RAGServer()
+	if ragServer.URL == "" {
+		return nil, errors.New("no RAG server configured")
+	}
+	u, err := url.Parse(ragServer.URL)
+	if err != nil {
+		return nil, err
+	}
+	u.Path = fmt.Sprintf("/query/%s", inst.Domain)
```
paultranvan, line 158: I wonder if `/query` is not too generic, as we might have other endpoints in the future. Maybe `/chat` would be more precise?

model/rag/chat.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+type ChatMessage struct {
+	Role    string `json:"role"`
+	Content string `json:"content"`
+}
+
+const (
+	HumanRole = "human"
+	AIRole    = "ai"
```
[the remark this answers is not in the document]
nono: Yes, I have taken the "human" and "ai" from LangChain, but OpenAI it is "user" and "assistant" for OpenAI and most other places. Let's switch!

worker/rag/rag.go
```diff
@@ -0,0 +1,49 @@
...
+		MaxExecCount: 1,
+		Reserved:     true,
+		Timeout:      15 * time.Minute,
+		WorkerFunc:   WorkerIndex,
+	})
+
+	job.AddWorker(&job.WorkerConfig{
+		WorkerType:   "rag-query",
+		Concurrency:  runtime.NumCPU(),
+		MaxExecCount: 1,
```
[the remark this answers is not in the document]
nono: No, `MaxExecCount: 1` means that we don't retry if a job fails. What you are looking for is the concurrency. By default, it is equal to the number of CPU, but in practice, in our environnement, it is defined in the configuration.

model/rag/chat.go
```diff
@@ -0,0 +1,200 @@
...
+	return &chat, nil
+}
+
+func Query(inst *instance.Instance, logger logger.Logger, query QueryMessage) error {
+	var chat ChatCompletion
+	err := couchdb.GetDoc(inst, consts.ChatCompletions, query.DocID, &chat)
+	if err != nil {
+		return err
+	}
+	msg := chat.Messages[len(chat.Messages)-1]
```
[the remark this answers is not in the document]
nono: - for the history, it is on my todo-list, I'm just not really clear on how to do that
- for system prompt, we already have one, but it is defined in the RAG component and I don't think we should persist it in CouchDB

model/rag/chat.go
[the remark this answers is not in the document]
nono: I agree. I also need to add CozyMetadata.

model/rag/chat.go
nono: I was also hesitating between a generic endpoint that accepts a parameter in the JSON (like `task: chat`), or having an endpoint per task.

PR #4474, opened 2024-10-10
Fix the myself contact at instance creation

nono:
- add the displayName
- add the index.byFamilyNameGivenNameEmailCozyUrl

In theory, there is a service from the contacts webapp that adds them, launched via an `@vevent` trigger on the io.cozy.contacts. But, the stack creates the myself contact before installing the contacts webapp at instance creation. Thus, the service is not launched. The easy fix is to just add the missing field when creating the myself contact.

### DIFF
--- a/model/contact/contacts.go
+++ b/model/contact/contacts.go
@@ -260,16 +260,33 @@ func FindByEmail(db prefixer.Prefixer, email string) (*Contact, error) {
 func CreateMyself(inst *instance.Instance, settings *couchdb.JSONDoc) (*Contact, error) {
 	doc := New()
 	doc.JSONDoc.M["me"] = true
-	if name, ok := settings.M["public_name"]; ok {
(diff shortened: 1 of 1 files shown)


model/contact/contacts.go
```diff
@@ -260,16 +260,33 @@ func FindByEmail(db prefixer.Prefixer, email string) (*Contact, error) {
...
+	cozyURL := inst.PageURL("", nil)
 	doc.JSONDoc.M["cozy"] = []map[string]interface{}{
-		{"url": inst.PageURL("", nil), "primary": true},
+		{"url": cozyURL, "primary": true},
+	}
+	index := email
+	if index == "" {
+		index = cozyURL
+	}
+	doc.JSONDoc.M["index"] = map[string]interface{}{
```
zatteo, line 288: Careful, it is stored as `indexes` in the database.

PR #4497, opened 2024-11-28
Fix HTTP keep-alive for CouchDB connections

nono:
Golang can reuse HTTP connections with keep alive, but the response body must have been fully read and closed. In the couchdb package, we were reading the body for sucessful response with a json.Decoder. It only reads a JSON value, and most responses from CouchDB have an additional \n after that JSON value. This last byte was not read from the response body, and it prevents Golang from reusing the connection. So, we are doing a new HTTP connection (with DNS resolution and TLS handshake) for each request to CouchDB.


sblaisot (approved):
Nice catch, thx @nono

PR #4499, opened 2024-12-10
Add a route for uploading avatars

nono:


web/settings/settings.go
```diff
@@ -253,6 +254,31 @@ func isMovedError(err error) bool {
...
 
+func (h *HTTPHandler) UploadAvatar(c echo.Context) error {
+	inst := middlewares.GetInstance(c)
+	if err := middlewares.AllowWholeType(c, http.MethodPut, consts.Settings); err != nil {
+		return err
+	}
+	header := c.Request().Header
+	size := c.Request().ContentLength
+	if size > 20_000_000 {
+		return jsonapi.BadRequest(errors.New("Avatar is too big"))
```
taratatach, line 265: I think we should return a 413 `Entity Too Large` error here.

web/settings/settings.go
```diff
@@ -253,6 +254,31 @@ func isMovedError(err error) bool {
...
 }
 
+func (h *HTTPHandler) UploadAvatar(c echo.Context) error {
+	inst := middlewares.GetInstance(c)
+	if err := middlewares.AllowWholeType(c, http.MethodPut, consts.Settings); err != nil {
+		return err
+	}
+	header := c.Request().Header
+	size := c.Request().ContentLength
+	if size > 20_000_000 {
```
taratatach, line 264: We could have a `MaxFileSize()` method in the `AvatarFS` struct like the one in the `Fs` interface.

taratatach: Maybe this should be an error returned by `CreateAvatar()`?!

taratatach (commented):
No risks of name collision in the `afero` version?

nono:
> No risks of name collision in the afero version?

No, we are using the same hidden directory that for thumbnails.

web/settings/settings.go
nono: I don't think we should have a method `MaxFileSize()` on the `AvatarFS` struct. This limit doesn't not depend of where the file is persisted. It is more a functional guard to avoid some users uploading a very large avatar that will slow down the login page and the sharing pages.

PR #4500, opened 2024-12-10
tech: Cleanup example config from prod elements

taratatach:
The example config should not contain elements such as URLs from our
production environment as people might simply copy the file without
changing some parts and end up making unnecessary requests to our
servers.

### DIFF
--- a/cozy.example.yaml
+++ b/cozy.example.yaml
@@ -218,7 +218,7 @@ mail:
   # mail noreply address - flags: --mail-noreply-address
   noreply_address: noreply@localhost
   noreply_name: My Cozy
-  reply_to: support@cozycloud.cc
+  reply_to: support@example.org
   # mail smtp host - flags: --mail-host
   host: smtp.home
   # mail smtp port - flags: --mail-port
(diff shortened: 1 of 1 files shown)


cozy.example.yaml
```diff
@@ -345,21 +345,21 @@ registries:
 
 # Wizard used for moving a Cozy from one place/hoster to another
 move:
-  url: https://move.cozycloud.cc/
+  url: https://move.example.org/
```
nono, line 348: Well, in that case, `move.cozycloud.cc` is fine and should be kept

cozy.example.yaml
```diff
@@ -457,7 +457,7 @@ contexts:
     default_redirection: drive/#/folder
     # This domain will be used as a suggestion for the members of a sharing
     # when they are asked for the URL of their Cozy instance
-    sharing_domain: mycozy.cloud
+    sharing_domain: example.org
```
nono, line 460: I think we should also keep this one

cozy.example.yaml
taratatach: I was going to but the flagship config has a `move_url` attribute with a different URL (see https://github.com/cozy/cozy-stack/pull/4500/files#diff-accceeea297d6b4bddb571eaab50129c26e0485c18c259d7822b192f3bc3df65L494) so I figured the move wizard did not have to be https://move.cozycloud.cc.

nono: It can be overloaded by context, but it is not used

taratatach: OK.

nono (approved):

PR #4511, opened 2025-02-10
Add a route to list shared drives

nono:


docs/shared-drives.md
```diff
@@ -0,0 +1,91 @@
...
+playing a proxy role.
+
+To create a shared drive (typically on the organization Cozy), we need the
+following steps:
+
+1. Ensure that the `/Drive` folder exists in the cozy instance with the
+   [`POST /files/shared-drives`](https://docs.cozy.io/en/cozy-stack/files/#post-filesshared-drives)
+   route.
+2. Create a folder inside it, with the name of shared drive.
+3. Create a sharing with the `sharing: true` attribute, and one rule for
```
zatteo, line 16: `drive: true` instead of `sharing: true`

docs/shared-drives.md
```diff
@@ -0,0 +1,128 @@
...
+# Shared drives
+
+A shared drive is a folder that is shared between several cozy instances. A
+member doesn't have the files in their Cozy, but can access them via the stack
+playing a proxy role.
+
+To create a shared drive (typically on the organization Cozy), we need the
+following steps:
+
+1. Ensure that the `/Drive` folder exists in the cozy instance with the
```
taratatach:
```suggestion
1. Ensure that the `/Drives` folder exists in the cozy instance with the
```

docs/shared-drives.md
```diff
@@ -0,0 +1,128 @@
...
+   shared folder (with `none` for `add`, `update` and `remove` attributes).
+
+# Routes
+
+A permission on the whole `io.cozy.files` doctype is required to use the
+following routes.
+
+## GET /sharings/drives
+
+The `GET /sharings/drives` route returns the list of shared drives.
```
taratatach, line 23: Is the response the same whether you're a member or the owner of the shared drive?

docs/sharing-design.md
taratatach, line 1: The file could be renamed as well.

docs/sharing-design.md
```diff
@@ -1,11 +1,17 @@
...
-Here we detail the baseline of the Cozy sharing design and provide some core
-statements.
+Here we detail the baseline of the Cozy sharing design where files are
+synchronized between several cozy instances.
+
+**Note:** shared drives is a newer feature, and it does not require to
+synchronize the files between the cozy instances. It is still possible to share
+files between cozy instances. As such, it is more scalable but does not provide
+the same features as the synchronized sharing. In particular, when a shared drive
+is revoked, the member has no longer access to a copy of the files in it.
```
taratatach:
```suggestion
is revoked, members no longer have access to a copy of the files in it.
```

docs/sharing-design.md
```diff
@@ -1,11 +1,17 @@
...
 ## Baseline
 
-Here we detail the baseline of the Cozy sharing design and provide some core
-statements.
+Here we detail the baseline of the Cozy sharing design where files are
+synchronized between several cozy instances.
+
+**Note:** shared drives is a newer feature, and it does not require to
+synchronize the files between the cozy instances. It is still possible to share
+files between cozy instances. As such, it is more scalable but does not provide
```
taratatach: What do you mean here by « It is still possible to share files between cozy instances. »?

I don't understand why introducing shared drives would change that :thinking:.

model/sharing/invitation.go
```diff
@@ -257,6 +257,89 @@ func getDocumentType(inst *instance.Instance, s *Sharing) string {
...
 }
 
+// CreateDriveShortcut will create a shortcut for a shared Drive.
+func (s *Sharing) CreateDriveShortcut(inst *instance.Instance, seen bool) error {
+	dir, err := inst.EnsureSharedDrivesDir()
+	if err != nil {
+		return err
+	}
+
+	filename := s.Rules[0].Title + ".url"
```
taratatach: We should sanitize the file name to make sure it is valid on all filesystems.

web/sharings/drives.go
```diff
@@ -104,13 +105,46 @@ func CopyFile(c echo.Context, inst *instance.Instance, s *sharing.Sharing) error
 	return files.CopyFile(c, inst)
 }
 
+func ChangesFeed(c echo.Context, inst *instance.Instance, s *sharing.Sharing) error {
+	// TODO: if owner then fail, shouldn't be accessing their own stuff, risk recursion download kinda thing
+	// TODO: should this break if there ever is actually more than 1 directory ?
+	// TODO: consider nested sharings
```
nono: The first 2 TODOs are legitimate but can be done later. For the third TODO, I don't see why we should do something special in that case, and I think we can remove it.

web/sharings/drives.go
```diff
@@ -54,41 +55,66 @@ func HeadDirOrFile(c echo.Context, inst *instance.Instance, s *sharing.Sharing)
...
 	}
 	return files.FileData(c, http.StatusOK, file, true, nil)
 }
 
-func GetDirSize(c echo.Context, inst *instance.Instance, s *sharing.Sharing) error {
-	fs := inst.VFS()
-	fileID := c.Param("file-id")
-	dir, err := fs.DirByID(fileID)
+func ReadFileContentFromIDHandler(c echo.Context, inst *instance.Instance, s *sharing.Sharing) error {
+	//TODO: CSP ?
```
nono: Yes, I think we need to add CSP stuff

web/sharings/drives.go
```diff
@@ -54,41 +55,66 @@ func HeadDirOrFile(c echo.Context, inst *instance.Instance, s *sharing.Sharing)
...
+	file, err := loadFileFromParam(c, inst, s)
 	if err != nil {
-		return files.WrapVfsError(err)
+		return err
 	}
+	//TODO: 403 if check perms is true
```
nono: We check permissions in the `loadFileFromParam` function, I don't see what more we can do here. Do you have something specific in mind?

web/sharings/drives.go
```diff
@@ -0,0 +1,422 @@
...
+		if dir != nil {
+			return dir, nil
+		}
+	}
+
+	return nil, jsonapi.NotFound(errors.New("shared drive not found"))
+}
+
+// drivesRoutes sets the routing for the shared drives
+func drivesRoutes(router *echo.Group) {
```
nono, line 339: Currently, some routes are documented and some others are not, and it's not clear why. Maybe we should at least add a list of the supported routes for shared drives in the documentation.

docs/shared-drives.md
nono: Yes, it is the same.

model/sharing/invitation.go
nono: Do you remember what are the rules for that?

There are the forbidden chars for windows (`<>:"/\|?*`), but is it all?

web/sharings/drives.go
taratatach: `files.SendFileFromDoc()` already adds `form-action: none` and `frame-ancestors: *`. Do you see other rules that should be added here that are not necessary in the basic file content download case?

web/sharings/drives.go
taratatach: I'm not really convinced by the first one. It's a way of getting a sub changes feed, with only files from a given shared drive.

The second one looks indeed legitimate although I don't really see what the experience should be for the user.

I agree about the third one. Since we're filtering changes by parent directory there shouldn't be any problem with nested sharings.

web/sharings/drives.go
nono: We should test it, but there are also the default CSP rules applied with the secure middleware which may contain the organization instance host, instead of the host of the proxy instance. But, yes, with `frame-ancestors: *`, there are good chances that it works as is.

model/sharing/invitation.go
taratatach: There's also a limit to the size of the name (but it's 255 characters so we're probably fine) and there cannot be any `/` on linux and macOS.

web/sharings/drives.go
```diff
@@ -0,0 +1,428 @@
...
+// ModifyMetadataByIDHandler handles PATCH requests on /files/:file-id
+//
+// It can be used to modify the file or directory metadata, as well as
+// moving and renaming it in the filesystem.
+func ModifyMetadataByIDHandler(c echo.Context, inst *instance.Instance, s *sharing.Sharing) error {
+	patch, err := getPatch(c, c.Param("file-id"))
+	if err != nil {
+		return files.WrapVfsError(err)
+	}
+	if err = applyPatch(c, inst.VFS(), patch); err != nil {
```
taratatach, line 158: We need to pass the sharing here so that its id can be added to the response.

PR #4520, opened 2025-04-20
Retro-compatible switch for `public/avatar` to an SVG output by default

ericlinagora:


docs/public.md
```diff
@@ -13,9 +13,14 @@ chosen, a fallback will be used, depending of the `fallback` parameter in the
 query-string:
 
 - `default`: a default image that shows the Cozy Cloud logo, but it can be
-  overriden by dynamic assets per context
-- `initials`: a generated image with the initials of the owner's public name
-- `404`: just a 404 - Not found error.
+  overriden by dynamic assets per context (always a png)
+- `404`: just a HTTP 404 - Not found error.
+- `anonymous`: a generic user icon without initials visible
+- `initials` or no `fallback` parameter: a generated image with the initials of
```
nono, line 19: Previously, when no `fallback` parameter, we were using the `default` behavior. I think we should stick to that for compatibility purposes. For example, the login page for the current `.mycozy.cloud` instances is using the `/public/avatar` route with no parameter, and we want to keep the logo for the moment. But, more generally, we try really hard to keep compatibility on the Rest API exposed by the stack, as it is a good practice.

nono (commented):
LGTM

docs/public.md
ericlinagora: Indeed, my bad, fixed it in a53a2b3

nono (approved):

pkg/avatar/testdata/ww.svg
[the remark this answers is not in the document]
zatteo: For the record, it was because `alignement-baseline` is not supported by Firefox.

PR #4524, opened 2025-05-07
style(templates): Update instance templates layout and add new icons

lenhanphung:


taratatach:
The test fails because we reach a deprecation date. This is not related to your work.

lenhanphung:
@taratatach can you also help me review on this PR? thanks you :D

taratatach (approved):
Yes, sorry! I just wanted to check if the support e-mail was correct but I forgot to come back here and approve the PR.

PR #4528, opened 2025-05-21
Allow to patch expired share-by-link permissions

nono:


zatteo:
Tested succesfully locally

PR #4539, opened 2025-06-16
Configure RAG indexing through flags for images and videos

paultranvan:
The RAG is able to index both images and videos, which is nice, but this can be a costful treatment.
Therefore, we prefer to make it configurable through flags.

### DIFF
--- a/model/rag/index.go
+++ b/model/rag/index.go
@@ -11,6 +11,7 @@ import (
 	"net/url"
 	"strings"
 
+	"github.com/cozy/cozy-stack/model/feature"
 	"github.com/cozy/cozy-stack/model/instance"
 	"github.com/cozy/cozy-stack/model/job"
 	"github.com/cozy/cozy-stack/model/note"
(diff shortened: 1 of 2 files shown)


model/rag/index.go
```diff
@@ -151,13 +153,39 @@ func callRAGIndexer(inst *instance.Instance, doctype string, change couchdb.Chan
...
+		if class == consts.VideoClass {
+			// Index Videos only if flag allows it
+			allowVideo, ok := flags.M["rag.index.video.enabled"].(bool)
+			if !ok || !allowVideo {
+				return nil
+			}
+		}
+
+		if class == consts.AudioClass {
+			// Index Videos only if flag allows it
```
taratatach:
```suggestion
			// Index Audio files only if flag allows it
```

paultranvan:
> nit: this could probably be moved to the top of the method to avoid some requests (the same way we do an early return for folders).

that's right, I moved it

PR #4548, opened 2025-07-11
Add shared_docs param

paultranvan:
A `shared_docs` param is added to the GET `/sharings/doctype` route, in order for the client to tell whether or not the `shared_docs` relationships is required in the response, as this can be costful in case of sharings with a lot of documents.

### DIFF
--- a/web/sharings/sharings.go
+++ b/web/sharings/sharings.go
@@ -198,7 +198,10 @@ func CountNewShortcuts(c echo.Context) error {
 func GetSharingsInfoByDocType(c echo.Context) error {
 	inst := middlewares.GetInstance(c)
 	docType := c.Param("doctype")
-
+	withSharedDocs, err := strconv.ParseBool(c.QueryParam("shared_docs"))
+	if err != nil {
+		withSharedDocs = true
+	}
 	sharings, err := sharing.GetSharingsByDocType(inst, docType)
 	if err != nil {
 		inst.Logger().WithNamespace("sharing").Errorf("GetSharingsByDocType error: %s", err)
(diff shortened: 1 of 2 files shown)


web/sharings/sharings.go
```diff
@@ -214,10 +215,14 @@ func GetSharingsInfoByDocType(c echo.Context) error {
 	for sID := range sharings {
 		sharingIDs = append(sharingIDs, sID)
 	}
-	sDocs, err := sharing.GetSharedDocsBySharingIDs(inst, sharingIDs)
-	if err != nil {
-		inst.Logger().WithNamespace("sharing").Errorf("GetSharedDocsBySharingIDs error: %s", err)
-		return wrapErrors(err)
+
+	var sDocs map[string][]couchdb.DocReference
+	if sharedDocsParam != "false" {
```
nono: [nit] For boolean parameters, we often use `strconv.ParseBool` (ie the client can send `false`, `False`, `0`, etc.)

paultranvan: Nice tip, I updated the commit

web/sharings/sharings.go
```diff
@@ -214,10 +217,14 @@ func GetSharingsInfoByDocType(c echo.Context) error {
 	for sID := range sharings {
 		sharingIDs = append(sharingIDs, sID)
 	}
-	sDocs, err := sharing.GetSharedDocsBySharingIDs(inst, sharingIDs)
-	if err != nil {
-		inst.Logger().WithNamespace("sharing").Errorf("GetSharedDocsBySharingIDs error: %s", err)
-		return wrapErrors(err)
+
+	var sDocs map[string][]couchdb.DocReference
+	if withSharedDocs == true {
```
nono: [nit] you can do `if withSharedDocs {` instead of `if withSharedDocs == true {`

PR #4555, opened 2025-09-01
feat: Add route that redirects to premium manager

taratatach:
We add a convenience route to allow redirecting a user to their
  premium management page without the client knowing about the manager
  itself (neither its base URL nor the instance UUID).

  If there aren't any manager set up for the given instance, the route
  returns an error page with a 404 status code.


web/settings/external_ties.go
```diff
@@ -52,3 +53,22 @@ func (h *HTTPHandler) getExternalTies(c echo.Context) error {
...
+	inst := middlewares.GetInstance(c)
+
+	// TODO: move to a SettingsService method to avoid the dependency on the
+	// instance module.
+	url, err := inst.ManagerURL(instance.ManagerPremiumURL)
+	if err != nil {
+		return echo.NewHTTPError(http.StatusBadRequest, err.Error())
+	}
+	if url == "" {
+		return echo.NewHTTPError(http.StatusNotFound, "This instance does not have a premium subscription manager.")
```
nono, line 70: I think we should show an html page. You can grep `error.html` in the code to have some examples.

web/settings/external_ties.go
```diff
@@ -52,3 +53,22 @@ func (h *HTTPHandler) getExternalTies(c echo.Context) error {
...
 	return jsonapi.Data(c, http.StatusOK, doc, nil)
 }
+
+func (h *HTTPHandler) redirectToPremium(c echo.Context) error {
+	if !middlewares.IsLoggedIn(c) {
+		return middlewares.ErrForbidden
+	}
+	inst := middlewares.GetInstance(c)
+
+	// TODO: move to a SettingsService method to avoid the dependency on the
```
nono, line 63: TODO spotted

web/settings/external_ties.go
taratatach: Yes, we do. The error gets picked up and `error.html` is displayed with the given message.

web/settings/external_ties.go
taratatach: That's kind of a long-term TODO as in the current state of the codebase this would just move the dependency to the service.

nono (approved):

PR #4556, opened 2025-09-03
Add an old domain field to Cozy instances

nono:
For the migration from Cozy to Twake Workplace, we need a way to redirect from the old domains to the new ones. So, an old domain field was added to the Cozy instances, and the stack will do the redirections from the old domains to the new ones. It works for the API endpoint, but also for the apps pages (eg old.mycozy.cloud -> new.twake.app).

The old domain of a Cozy instance can be set with the `cozy-stack instances modify` command, or directly by modifying the `old_domain` field of the CouchDB document of the instance in the global database.


Crash-- (approved):
thanks cc @rezk2ll

PR #4560, opened 2025-09-10
RabbitMQ connector + password change message handler

shepilov:


rabbitmq/handlers.go
```diff
@@ -0,0 +1,93 @@
...
+// PasswordChangeMessage represents a password change message using the new schema.
+type PasswordChangeMessage struct {
+	TwakeID    string `json:"twakeId"`    // external user identifier
+	Iterations int    `json:"iterations"` // PBKDF2 iterations used client-side (when applicable)
+	Hash       string `json:"hash"`       // client-side hashed passphrase (base64)
+	PublicKey  string `json:"publicKey"`  // Bitwarden public key (base64)
+	PrivateKey string `json:"privateKey"` // Bitwarden private key (encrypted, CipherString)
+	Key        string `json:"key"`        // encrypted symmetric key (CipherString)
+	Timestamp  int64  `json:"timestamp"`  // unix timestamp of the event
+	Domain     string `json:"domain"`     // domain of the instance, e.g. "twake.app"
```
rezk2ll: This won't be sent from signup as we agreed to leave it for later

rabbitmq/connection.go
```diff
@@ -0,0 +1,126 @@
...
+		}
+
+		// Connection created successfully
+		cm.conn = conn
+		cm.connClose = make(chan *amqp.Error, 1)
+		conn.NotifyClose(cm.connClose)
+		return cm.conn, nil
+	}
+
+	return nil, fmt.Errorf("exceeded maximum connection attempts")
```
taratatach: In which case do we reach this return?
Looks like we'll always return from line 60 in case we can't establish a connection.

Maybe we could declare `err` outside the loop, add it to this message and remove line 60.

rabbitmq/connection.go
```diff
@@ -0,0 +1,126 @@
...
+	cm.mu.Lock()
+	defer cm.mu.Unlock()
+
+	if cm.conn != nil && cm.isConnectionAlive() {
+		return cm.conn, nil // Already connected and alive
+	}
+
+	// Close existing connection if any
+	if cm.conn != nil {
+		cm.conn.Close()
```
taratatach: Do we need to close the channel again?

rabbitmq/connection_integration_test.go
```diff
@@ -0,0 +1,171 @@
...
+	t.Helper()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
+	defer cancel()
+
+	username := "guest"
+	password := "guest"
+
+	req := tc.ContainerRequest{
+		Image:        "rabbitmq:3.13-management",
```
taratatach: Is this the version used in production? I used the latest `4-management` version in the cloudery dev and test config.

rabbitmq/queue_runner.go
```diff
@@ -0,0 +1,136 @@
...
+	if _, err := r.ch.QueueDeclare(q.cfg.Name, true, false, false, false, qArgs); err != nil {
+		_ = ch.Close()
+		return nil, fmt.Errorf("failed to declare queue %s: %w", q.cfg.Name, err)
+	}
+	for _, binding := range q.cfg.Bindings {
+		if err := r.ch.QueueBind(q.cfg.Name, binding, exchangeName, false, nil); err != nil {
+			_ = ch.Close()
+			return nil, fmt.Errorf("failed to bind queue %s with routing key %s: %w", q.cfg.Name, binding, err)
+		}
+	}
```
taratatach: I would expect queues to be defined once and for all outside the stack (either manually or via a deployment script) like exchanges.

Is this a usual RabbitMQ way of creating queues?

rabbitmq/queue_runner.go
```diff
@@ -0,0 +1,136 @@
...
+	"context"
+	"fmt"
+	"time"
+
+	amqp "github.com/rabbitmq/amqp091-go"
+)
+
+const (
+	defaultPrefetch = 32
+	defaultType     = "quorum"
```
taratatach: I think it would be easier to reason about with a `QorumType` const.

rabbitmq/queue_runner.go
```diff
@@ -0,0 +1,136 @@
...
+			}
+		}
+	}
+
+	// Declare queue with args and bind to exchange
+	qArgs := amqp.Table{
+		"x-queue-type":           defaultType,
+		"x-dead-letter-exchange": q.dlxName,
+		"x-delivery-limit":       q.cfg.DeliveryLimit,
+	}
```
taratatach: Looks like we're missing 2 properties from the ADR:
- `x-dead-letter-strategy: at-least-once`
- `x-overflow: reject-publish`

rabbitmq/rabbitmq.go
```diff
@@ -0,0 +1,276 @@
...
+func declareRabbitMQExchange(conn *amqp.Connection, spec ExchangeSpec) error {
+	ch, err := conn.Channel()
+	if err != nil {
+		return fmt.Errorf("failed to open channel for exchange %s: %w", spec.cfg.Name, err)
+	}
+	defer ch.Close()
+
+	if err := ch.ExchangeDeclare(spec.cfg.Name, spec.cfg.Kind, spec.cfg.Durable, false, false, false, nil); err != nil {
+		return fmt.Errorf("failed to declare exchange %s: %w", spec.cfg.Name, err)
+	}
```
taratatach: Same thing as with queues: should we declare exchanges here?

I've read that they could be declared via "policies" and would expect them to be declared ahead of time.

rabbitmq/handlers.go
```diff
@@ -0,0 +1,60 @@
...
+func NewUserSettingsUpdateHandler() *UserSettingsUpdateHandler {
+	return &UserSettingsUpdateHandler{}
+}
+
+// UserSettingsUpdateMessage represents a user settings update message.
+type UserSettingsUpdateMessage struct {
+	Domain   string                 `json:"domain"`
+	Settings map[string]interface{} `json:"settings"`
+	Version  int                    `json:"version"`
+}
```
taratatach: I don't see this message in the ADR. Am I missing something or should it be added to the ADR?

rabbitmq/handlers.go
```diff
@@ -0,0 +1,60 @@
...
+func NewPasswordChangeHandler() *PasswordChangeHandler {
+	return &PasswordChangeHandler{}
+}
+
+// PasswordChangeMessage represents a password change message.
+type PasswordChangeMessage struct {
+	Domain      string `json:"domain"`
+	NewPassword string `json:"new_password"`
+	Version     int    `json:"version"`
+}
```
taratatach: This does not completely look like the message defined in the ADR (no version in the ADR and missing fields here).

model/instance/lifecycle/passphrase.go
```diff
@@ -247,6 +247,26 @@ func ForceUpdatePassphrase(inst *instance.Instance, newPassword []byte, params P
 	return update(inst)
 }
 
+func ForceUpdatePassphraseWithSHash(inst *instance.Instance, hash []byte, params PassParameters) error {
```
taratatach, line 250: nit: is the `S` in `SHash` extraneous?

rabbitmq/handlers.go
shepilov: all the fields including timestamp?

rabbitmq/connection.go
shepilov: Yes, make sense, I'll remove it

rabbitmq/connection.go
shepilov: I've added it just in case, not sure how rabbitmq go client is working, so closing it to make sure  everything is closes

rabbitmq/connection_integration_test.go
shepilov: In production I think 3.02 is right now

rabbitmq/queue_runner.go
shepilov: Yes, we don't have any external infra for queue and exchange declaration, so exchanges are declared by owner as well as queues

rabbitmq/queue_runner.go
shepilov: you mean name of the variable?

rabbitmq/queue_runner.go
shepilov: `x-dead-letter-strategy` it make sense, for `x-overflow` we don't publish anything from stack right now

rabbitmq/rabbitmq.go
shepilov: yes, the same as queue, publisher declares everything

rabbitmq/handlers.go
shepilov: nope, it was just my test message, I'll update with message from ADR

rabbitmq/handlers.go
shepilov: mmm, I don't know where you get this version, but right now message looks like 
```
type PasswordChangeMessage struct {
	TwakeID    string `json:"twakeId"`    // external user identifier
	Iterations int    `json:"iterations"` // PBKDF2 iterations used client-side (when applicable)
	Hash       string `json:"hash"`       // client-side hashed passphrase (base64)
	PublicKey  string `json:"publicKey"`  // Bitwarden public key (base64)
	PrivateKey string `json:"privateKey"` // Bitwarden private key (encrypted, CipherString)
	Key        string `json:"key"`        // encrypted symmetric key (CipherString)
	Timestamp  int64  `json:"timestamp"`  // unix timestamp of the event
	Domain     string `json:"domain"`     // domain of the instance, e.g. "twake.app"
}
````

File: model/instance/lifecycle/passphrase.go (hunk 11 above)
Comment (shepilov) [reply model/instance/lifecycle/passphrase.go:250]: yes

File: rabbitmq/queue_runner.go (hunk 5 above)
Comment (shepilov) [reply rabbitmq/queue_runner.go]: I've made it configurable in case we will have this separate declaration

File: rabbitmq/connection_integration_test.go (hunk 4 above)
Comment (taratatach) [reply rabbitmq/connection_integration_test.go]: I'll change that in the cloudery then. Thanks.

File: rabbitmq/rabbitmq.go (hunk 8 above)
Comment (taratatach) [reply rabbitmq/rabbitmq.go]: Publisher or consumer?

File: rabbitmq/handlers.go (hunk 1 above)
Comment (taratatach) [reply rabbitmq/handlers.go]: I don't know about the timestamp, it seems like this could be part of the message properties rather than the payload.

File: rabbitmq/rabbitmq.go (hunk 8 above)
Comment (shepilov) [reply rabbitmq/rabbitmq.go]: publisher

File: pkg/rabbitmq/handlers.go [hunk 12]
```diff
@@ -0,0 +1,118 @@
...
+	if msg.Iterations <= 0 {
+		return fmt.Errorf("password change: missing iterations")
+	}
+
+	params := lifecycle.PassParameters{
+		Pass:       []byte(msg.Hash),
+		Iterations: msg.Iterations,
+	}
+
+	// if one of the keys is missing, do not update any of the keys
```
nono: We should be able to update the key, without changing the public key and private key.

pkg/rabbitmq/handlers.go
```diff
@@ -0,0 +1,118 @@
...
+	Domain     string `json:"domain"`     // [OPTIONAL] domain of the instance, e.g. "twake.app"
+}
+
+// Handle processes a password change message.
+func (h *PasswordChangeHandler) Handle(ctx context.Context, d amqp.Delivery) error {
+	log.Infof("password change: received password change message: %s", d.RoutingKey)
+
+	var msg PasswordChangeMessage
+	if err := json.Unmarshal(d.Body, &msg); err != nil {
+		return fmt.Errorf("fpassword change:  ailed to unmarshal password change message: %w", err)
```
nono: [nit] s/fpassword/password/

pkg/rabbitmq/handlers.go
```diff
@@ -56,9 +56,12 @@ func (h *PasswordChangeHandler) Handle(ctx context.Context, d amqp.Delivery) err
...
 	}
 
-	// if one of the keys is missing, do not update any of the keys
-	if msg.Key != "" && msg.PublicKey != "" || msg.PrivateKey != "" {
+	if msg.Key != "" {
 		params.Key = msg.Key
+	}
+
+	// if one of the keys is missing, do not update any of the keys
+	if msg.PublicKey != "" || msg.PrivateKey != "" {
```
nono: If the public key is present in the message, but not the private key, the comment suggests that they are not updated, but the code does the update. I think we need `&&` instead of `||`. Am I wrong?

shepilov: fixed

nono (approved):

rabbitmq/rabbitmq.go
taratatach: After working a bit more on the RabbitMQ communication from the cloudery, I think consumers should be responsible for creating the queues since in our case, publishers send messages to exchanges and not queues directly.

pkg/rabbitmq/handlers.go
```diff
@@ -0,0 +1,121 @@
...
+	if msg.TwakeID == "" {
+		return fmt.Errorf("user.created: missing twakeId")
+	}
+	if msg.InternalEmail == "" {
+		return fmt.Errorf("user.created: missing internalEmail")
+	}
+
+	// For now, we simply acknowledge the message after validation.
+	// Integrate business logic (e.g., instance/user provisioning) here as needed.
+	return nil
```
Crash--, line 120: @shepilov I'm a bit concerned about the fact we don't handle this message because it means that we're not creating the Vault during the onboarding process. 

And if we don't do that, it means that we'll not be able to have the same password for the vault & the TWP for our saas users. (this is not 100% accurate, because we can imagine a flow where we ask for user's password to generate again the PKBF but this is not straight forward because we'll have to create specific flow for that...)

I can understand that, when we receive the message for the first time, we don't have the instance ready yet. 

Can we imagine having this kind of scenario: 

- We consume the message 
- We try to apply it 
- It works, fine. 
- Acknowledge and byebye 

- We consume the message
- We try to apply 
- It fails 
- Then we don't ACK nor NACK the message, we just put it right back to the queue with I don't know, 10s of invisibility. 

Since the message will be see again in the queue 10s later: 
- We consume again the message
- We try to apply
- .... blablabla 

If we can retry 3 times before sending it to the dead letter queue, it'll be awesome. 


(I'm not a RabbitMQ expert, so I don't know if this is something we can do, but I know that I've used that with AWS SQS lately and it was working really great!)

shepilov: if you go to `queue_runner.go` it calls `d.Nack(false, true)` whick means message will be requeued. By default we can't configure timeout, but we can configure number of retries before it goes to DLX.

In case of any trouble we can manually put message from DLX back to queue.

Do you want me to configure "x-delivery-count"? By default it's 0, But actually  if it will be 10, it won't make any difference before the instance creation

PR #4564, opened 2025-09-17
feat: Added move function for files in shared drives

shepilov:


web/sharings/drives.go
```diff
@@ -438,6 +780,8 @@ func drivesRoutes(router *echo.Group) {
 	drive.PUT("/:file-id", proxy(OverwriteFileContentHandler, true))
 	drive.POST("/upload/metadata", proxy(UploadMetadataHandler, true))
 	drive.POST("/:file-id/copy", proxy(CopyFile, true))
+	drive.POST("/:dir-id/downstream", proxy(MoveDownstreamHandler, true))
+	drive.POST("/:file-id/upstream", proxy(MoveUpstreamHandler, true))
```
nono: Can you add documentation for those 2 routes?

nono: Why `:dir-id` for one route and `:file-id` for the other?

nono: I wouldn't have used the proxy for those routes. It's possible to do the work for the recipient's Cozy (Betty in the tests) instead of doing it on the organization Cozy (acme in the tests). It's a bit faster as we avoid one http request. But it's not a big deal.

nono: We will need a third route to copy one file from one shared drive to another shared drive. Do you want to do it in this PR or in another PR?

web/sharings/drives.go
```diff
@@ -264,8 +268,346 @@ func CopyFile(c echo.Context, inst *instance.Instance, s *sharing.Sharing) error
...
+	}
+	return nil
+}
+
+func MoveDownstreamSameStack(c echo.Context, inst *instance.Instance, sourceInstance *instance.Instance, targetDirId string, sourceFileID string) error {
+	// Check if the source file exists and get its metadata
+	srcFile, err := sourceInstance.VFS().FileByID(sourceFileID)
+	if err != nil {
+		return files.WrapVfsError(err)
+	}
```
nono: We need to check the permission on this file

web/sharings/drives.go
```diff
@@ -264,8 +268,346 @@ func CopyFile(c echo.Context, inst *instance.Instance, s *sharing.Sharing) error
...
+	if _, err := url.Parse(instanceURL); err != nil {
+		return jsonapi.BadRequest(errors.New("invalid " + paramName + " parameter"))
+	}
+	return nil
+}
+
+// validateFileID validates a file ID and returns an error if it's invalid
+func validateFileID(fileID string) error {
+	if fileID == "" {
+		return jsonapi.BadRequest(errors.New("missing file-id parameter"))
```
nono: You can use `jsonapi.InvalidParameter` to be a bit more precise on the returned error

web/sharings/drives_test.go
```diff
@@ -981,4 +1111,241 @@ func TestSharedDrives(t *testing.T) {
...
+			// Verify the file was moved to the destination
+			movedFileID := responseObj.Path("$.data.id").String().Raw()
+			verifyFileMove(t, acmeInstance, movedFileID, "file-to-move-upstream.txt", destDirID, "foo")
+
+			// Verify the original file was deleted from source
+			verifyFileDeleted(t, acmeInstance, fileToMoveUpstream)
+		})
+
+		t.Run("InvalidFileParameter", func(t *testing.T) {
+			res := httpexpect.Default(t, tsA.URL).POST("/sharings/drives/"+sharingID+"//upstream").
```
nono: [nit] We have two slashes in `//upstream`

web/sharings/drives_test.go
```diff
@@ -456,21 +540,74 @@ func TestSharedDrives(t *testing.T) {
...
+				Object()
+			sharingID2 := sharing2.Path("$.data.id").String().NotEmpty().Raw()
+			// Ensure we got a valid sharing id before proceeding
+			require.NotEmpty(t, sharingID2)
+
+			// Create a file inside the first shared drive that we'll move to the second
+			sourceFileID := createFile(t, eA, meetingsID, "to-move-between-drives.txt", acmeAppToken)
 
+			// Move downstream from the first drive to the second drive's root
+			res := moveFileDownstream(t, eA, sharingID2, secondRootDirID,
```
nono: We should test by making request to `eB` (Betty's Cozy), as it is the common case

web/sharings/drives.go
```diff
@@ -522,3 +866,61 @@ func proxy(fn func(c echo.Context, inst *instance.Instance, s *sharing.Sharing)
...
+	u, err := url.Parse(urlStr)
+	if err != nil {
+		return nil, err
+	}
+
+	// Remove port if present
+	host := u.Host
+	if strings.Contains(host, ":") {
+		host = strings.Split(host, ":")[0]
+	}
```
nono, line 556: IIRC, we should keep the port. In local, it should be `lifecycle.GetInstance("cozy.localhost:8080")`.

web/sharings/drives.go
```diff
@@ -264,8 +268,346 @@ func CopyFile(c echo.Context, inst *instance.Instance, s *sharing.Sharing) error
...
+	// Delete the file from the source instance
+	if err := deleteSourceFile(srcInst.VFS(), srcFile); err != nil {
+		return err
+	}
+
+	// Return success response mirroring created remote file
+	c.Response().Header().Set("Content-Type", "application/vnd.api+json")
+	return c.JSON(http.StatusCreated, map[string]interface{}{
+		"data": map[string]interface{}{
+			"type": "io.cozy.files",
```
nono: [nit] You can use `consts.Files` instead of `"io.cozy.files"`

web/sharings/drives.go
```diff
@@ -264,8 +268,346 @@ func CopyFile(c echo.Context, inst *instance.Instance, s *sharing.Sharing) error
...
+	if err != nil {
+		return files.WrapVfsError(err)
+	}
+
+	// Extract bearer token from the request
+	bearer := extractBearerToken(c)
+	dstClient := NewRemoteClient(u, bearer)
+
+	// Open the source file for reading
+	srcFileHandle, err := srcInst.VFS().OpenFile(srcFile)
```
nono: We need to check the permission on the source file

web/sharings/drives.go
```diff
@@ -264,8 +268,346 @@ func CopyFile(c echo.Context, inst *instance.Instance, s *sharing.Sharing) error
...
+		return files.WrapVfsError(err)
+	}
+
+	// Delete the file from the source instance
+	if err := deleteSourceFile(srcInst.VFS(), srcFile); err != nil {
+		return err
+	}
+
+	// Return success response mirroring created remote file
+	c.Response().Header().Set("Content-Type", "application/vnd.api+json")
```
nono: This line can be removed: `c.JSON` already do that

web/sharings/drives.go
```diff
@@ -264,8 +268,346 @@ func CopyFile(c echo.Context, inst *instance.Instance, s *sharing.Sharing) error
...
+	srcFile, err := sourceInstance.VFS().FileByID(sourceFileID)
+	if err != nil {
+		return files.WrapVfsError(err)
+	}
+
+	// Create a new file document based on the source file
+	newFileDoc, err := createFileDocFromSource(srcFile, targetDirId)
+	if err != nil {
+		return files.WrapVfsError(err)
+	}
```
nono: We need to check the permission to write in this directory

web/sharings/drives.go
```diff
@@ -264,8 +268,346 @@ func CopyFile(c echo.Context, inst *instance.Instance, s *sharing.Sharing) error
...
+func copyFileContent(destVFS vfs.VFS, newFileDoc *vfs.FileDoc, srcVFS vfs.VFS, srcFile *vfs.FileDoc) error {
+	if err := destVFS.CopyFileFromOtherFS(newFileDoc, nil, srcVFS, srcFile); err != nil {
+		return files.WrapVfsError(err)
+	}
+	return nil
+}
+
+// deleteSourceFile deletes a file from the source instance
+func deleteSourceFile(srcVFS vfs.VFS, srcFile *vfs.FileDoc) error {
+	if err := srcVFS.GetIndexer().DeleteFileDoc(srcFile); err != nil {
```
nono: We must call `srcVFS.DestroyFile()`. By using the indexer, we just delete the file in CouchDB but not in Swift.

web/sharings/drives.go
```diff
@@ -264,8 +268,346 @@ func CopyFile(c echo.Context, inst *instance.Instance, s *sharing.Sharing) error
...
+	// Create the file in the destination VFS
+	fd, err := inst.VFS().CreateFile(newFileDoc, nil)
+	if err != nil {
+		return files.WrapVfsError(err)
+	}
+
+	// Download content from source and write into destination
+	rc, err := srcClient.DownloadByID(sourceFileID)
+	if err != nil {
+		return files.WrapVfsError(err)
```
nono: If we return an error here, `fd` won't be closed

web/sharings/drives.go
```diff
@@ -264,8 +268,346 @@ func CopyFile(c echo.Context, inst *instance.Instance, s *sharing.Sharing) error
...
+	srcFile, err := srcClient.GetFileByID(sourceFileID)
+	if err != nil {
+		return files.WrapVfsError(err)
+	}
+
+	// Create destination file doc using source metadata
+	newFileDoc, err := createFileDocFromRemoteFile(srcFile, targetDirId)
+	if err != nil {
+		return files.WrapVfsError(err)
+	}
```
nono: We also need to check permissions here

web/sharings/drives.go
```diff
@@ -264,8 +268,346 @@ func CopyFile(c echo.Context, inst *instance.Instance, s *sharing.Sharing) error
...
+	// Create a new file document based on the source file
+	newFileDoc, err := createFileDocFromSource(srcFile, dirID)
+	if err != nil {
+		return files.WrapVfsError(err)
+	}
+
+	// Copy the file content to the destination instance
+	if err := copyFileContent(destInst.VFS(), newFileDoc, srcInst.VFS(), srcFile); err != nil {
+		return err
+	}
```
nono: We also need to check permissions here

web/sharings/drives.go
```diff
@@ -264,8 +268,346 @@ func CopyFile(c echo.Context, inst *instance.Instance, s *sharing.Sharing) error
...
+	// Extract bearer token from the request
+	bearer := extractBearerToken(c)
+	dstClient := NewRemoteClient(u, bearer)
+
+	// Open the source file for reading
+	srcFileHandle, err := srcInst.VFS().OpenFile(srcFile)
+	if err != nil {
+		return files.WrapVfsError(err)
+	}
+	defer srcFileHandle.Close()
```
nono: We also need to check permissions here

docs/shared-drives.md
```diff
@@ -281,6 +281,74 @@ Duplicates a file.
...
+
+Body (preferred):
+
+```json
+{
+  "source": {
+    "instance": "https://alice.localhost:8080",
+    "sharing_id": "share_src_id",    
+    "file_id": "file123",
+    "dir_id": ""
```
nono, line 306: What is this `dir_id` parameter?

docs/shared-drives.md
```diff
@@ -281,6 +281,74 @@ Duplicates a file.
 Identical call to [`POST /files/:file-id/copy`](files.md#post-filesfile-idcopy) but over a shared drive.
 See there for request and response examples, the only difference is the URL.
 
+### POST /sharings/drives/move
+
+Move a file between locations (personal drive and/or shared drives).
```
nono, line 286: Do we have a way to make a copy instead of a move?

docs/shared-drives.md
```diff
@@ -281,6 +281,74 @@ Duplicates a file.
...
+    "id": "new-file-id",
+    "attributes": {
+      "name": "example.txt",
+      "dir_id": "destDir456",
+      "type": "file",
+      "size": 123,
+      "mime": "text/plain",
+      "class": "document",
+      "executable": false,
+      "tags": []
```
nono, line 344: I think we should also put the `driveId` here

web/sharings/drives.go
```diff
@@ -533,3 +533,58 @@ func proxy(fn func(c echo.Context, inst *instance.Instance, s *sharing.Sharing)
...
+		memberHost := m.InstanceHost()
+		if memberHost == "" {
+			continue
+		}
+		// Check by domain
+		if memberHost == inst.Domain || memberHost == currDomain {
+			isMember = true
+			break
+		}
+		// Check by email
```
nono, line 578: I think we always have the domain when using a shared drive, and we don't need to check by email.

web/sharings/move.go
```diff
@@ -0,0 +1,468 @@
...
+func getInstanceIdentifierFromURL(urlStr string) (*instance.Instance, error) {
+	u, err := url.Parse(urlStr)
+	if err != nil {
+		return nil, err
+	}
+	host := u.Host
+	if strings.Contains(host, ":") {
+		host = strings.Split(host, ":")[0]
+	}
+	inst, err := lifecycle.GetInstance(host)
```
nono: [nit] You can just `return lifecycle.GetInstance(host)`

web/sharings/move.go
```diff
@@ -0,0 +1,468 @@
...
+
+	var req moveRequest
+	if err := c.Bind(&req); err != nil {
+		return jsonapi.BadRequest(errors.New("error unmarshalling request"))
+	}
+	if err := validateMoveRequest(req); err != nil {
+		return err
+	}
+
+	// Authorization: if at least one side is not a shared drive, require whole-type permission
```
nono: Why don't you check permissions when moving a file from one shared drive to another shared drive?

web/sharings/move.go
```diff
@@ -0,0 +1,468 @@
...
+	if req.Source.SharingID == "" || req.Dest.SharingID == "" {
+		if err := middlewares.AllowWholeType(c, permission.PATCH, consts.Files); err != nil {
```
nono: I would have used `permission.POST` instead of `permission.PATCH`

web/sharings/move.go
```diff
@@ -0,0 +1,468 @@
...
+			return err
+		}
+	}
+
+	sourceInstance, err := resolveInstanceOrCurrent(req.Source.Instance, inst)
```
nono, line 248: If the source instance is not on the same stack as the local instance, it will return an error here

web/sharings/move.go
```diff
@@ -0,0 +1,468 @@
...
+	if err != nil {
+		return err
+	}
+	destInstance, err := resolveInstanceOrCurrent(req.Dest.Instance, inst)
```
nono, line 252: Same for dest instance

web/sharings/move.go
```diff
@@ -0,0 +1,468 @@
...
+	rc, err := srcClient.DownloadByID(fileID)
+	if err != nil {
+		// Best-effort close to avoid leaking descriptors on error paths
+		_ = fd.Close()
+		return files.WrapVfsError(err)
+	}
+	defer rc.Close()
+
+	// Use a small buffer to reduce allocations during large copies
+	if _, err := io.CopyBuffer(fd, rc, make([]byte, copyBufferSize)); err != nil {
```
nono: IIRC, `io.Copy` is enough: it uses a temporary buffer

docs/shared-drives.md
shepilov: it's a teaser of the "directory move function"

docs/shared-drives.md
shepilov: nope, source file is deleted, but we can add a param to make a "copy" from this function or extract some code and reuse it in copy

docs/shared-drives.md
shepilov: in tags?

web/sharings/drives.go
shepilov: I was not sure, and it's relevant at least for tests because in instance we have domain name and in invitations urls like "localhost:port" to make invitation process work with redirects

web/sharings/move.go
shepilov: it's also the same as previous, doesn't work if we have multiple instances running on different ports

web/sharings/move.go
shepilov: in this case permissions check for the sharings, and we don't need to check permission for local file system

web/sharings/move.go
shepilov: fixed

web/sharings/move.go
shepilov: double checked, my bad, now there is now port s there, removed it

web/sharings/move.go
shepilov: fixed

web/sharings/move.go
shepilov: fixed

web/sharings/drives.go
shepilov: fixed

web/sharings/drives.go
shepilov: actually it adds "application/json", I'm not sure what exactly content type do we need

docs/shared-drives.md
nono: In `attributes`, like we do for other routes

web/sharings/move.go
nono: But the permissions check for the sharings are not tied to the incoming http request: it means that anybody can trigger a move between two shared drives if they know the parameters. Or am I missing something?

shepilov: yes, you are right, I relied too much on permission check of http calls, but we need to check that we have access just get information about those sharings

docs/shared-drives.md
shepilov: added

nono (approved):

PR #4566, opened 2025-09-19
fix: We need to check PassPhrase

Crash--:
In order to have the same behavior as the previous route `/connect/token` we need to check that the passphrase is OK.

### DIFF
--- a/web/oidc/oidc.go
+++ b/web/oidc/oidc.go
@@ -135,7 +135,7 @@ func BitwardenExchange(c echo.Context) error {
 			"error": "invalid code",
 		})
 	}
-    if err := instance.CheckPassphrase(inst, []byte(pass)); err != nil {
+	if err := instance.CheckPassphrase(inst, []byte(pass)); err != nil {
 		return c.JSON(http.StatusUnauthorized, echo.Map{
 			"error": "invalid password",
 		})
--- a/web/oidc/oidc.go
+++ b/web/oidc/oidc.go
@@ -116,6 +116,7 @@ func BitwardenExchange(c echo.Context) error {
 	inst := middlewares.GetInstance(c)
 
 	code := c.FormValue("code")
+	pass := c.FormValue("password")
 	if code == "" {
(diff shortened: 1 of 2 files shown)


web/oidc/oidc.go
```diff
@@ -134,6 +135,11 @@ func BitwardenExchange(c echo.Context) error {
 			"error": "invalid code",
 		})
 	}
+	if err := instance.CheckPassphrase(inst, []byte(pass)); err != nil {
```
zatteo, line 138: Will it check vault password hash against instance password hash 🤔 ?

Crash--: kind of, yes.

zatteo: So it will not work with SSO again no?

Crash--: We have 3 cases: 

### First one, without SSO 

When you create your TWP, we init the vault with the same passphrase as your instance. If you change your instance passphrase, we update the vault one and so on. 

### Second  one, with our SaaS SSO (aka signup)

When you create your TWP, signup will send to your instance the hash of your passphrase. Like that, you'll have the same passphrase on the SSO but also on your instance. If you decide to change your passphrase on signup, we'll update your instance passphrase too. Work done in https://github.com/cozy/cozy-stack/pull/4560

### Third one, with external SSO 

In that case, when you create your TWP, we init it with a random passphrase on your instance. But in that case, when you open Pass (on Web), you'll have to initialize your passphrase. So we ask you about it and so on. Once it's done, we save this passphrase on your instance. 

It means that... if you allow to login with password on your instance even if you're linked to an OIDC provider, you're able to login with this password on your instance. 



(That's all in the past, I hope I'm not saying anything stupid.)

zatteo: Agreeing with 1. and 2.

For 3., we will get the issue if the external SSO implement the autodiscovery by email, so login with SSO and not by URL. But if I understand well it is not an issue at short term?

PR #4578, opened 2025-09-26
feat: Added a routing key for DLQ in case of DLX kind is topic

shepilov:


pkg/rabbitmq/queue_runner.go
```diff
@@ -63,13 +63,18 @@ func newQueueRunner(conn *amqp.Connection, exchangeName string, q QueueSpec) (*q
 		}
 		log.Debugf("Successfully declared DLQ: %s", q.dlqName)
 		if q.dlxName != "" {
-			log.Debugf("Binding DLQ %s to DLX %s", q.dlqName, q.dlxName)
-			if err := r.ch.QueueBind(q.dlqName, "", q.dlxName, false, nil); err != nil {
+			// Use configured DL routing key if present; otherwise keep empty key
+			bindingKey := ""
+			if q.cfg.DLRoutingKey != "" {
+				bindingKey = q.cfg.DLRoutingKey
+			}
```
taratatach:
```suggestion
			bindingKey := q.cfg.DLRoutingKey
```

This would have the same result.

pkg/rabbitmq/queue_runner.go
```diff
@@ -46,7 +46,7 @@ func newQueueRunner(conn *amqp.Connection, exchangeName string, q QueueSpec) (*q
 	// Declare DLX/DLQ using resolved fields already set on QueueSpec
 	if q.cfg.DeclareDLX && q.dlxName != "" {
 		log.Debugf("Declaring DLX: %s for queue: %s", q.dlxName, q.cfg.Name)
-		if err := r.ch.ExchangeDeclare(q.dlxName, "fanout", true, false, false, false, nil); err != nil {
+		if err := r.ch.ExchangeDeclare(q.dlxName, "topic", true, false, false, false, nil); err != nil {
```
taratatach, line 49: Shouldn't the exchange type be configurable?

shepilov: yes, it's better to be, but as far for stack it's not relevant right now, exchanges are created by other applicaiton, and this code used only in tests, I kept it as is

pkg/rabbitmq/queue_runner.go
shepilov: yes, my bad, here supposed to be first queue binding, I'll change it

pkg/rabbitmq/queue_runner.go
taratatach: OK

taratatach (approved):

PR #4580, opened 2025-09-26
fix: Changed a format of a domain in incoming messages(only top level domain)

shepilov:
… domain)

### DIFF
--- a/pkg/rabbitmq/handlers.go
+++ b/pkg/rabbitmq/handlers.go
@@ -125,6 +125,7 @@ type UserCreatedMessage struct {
 	PrivateKey    string `json:"privateKey"`
 	Key           string `json:"key"`
 	Timestamp     int64  `json:"timestamp"`
+	WorkplaceFqdn string `json:"workplaceFqdn"` // The fully qualified workplace domain
 }
 
 // Handle processes a user created message.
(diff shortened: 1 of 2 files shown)


pkg/rabbitmq/handlers.go
```diff
@@ -87,8 +87,9 @@ func (h *PasswordChangeHandler) Handle(ctx context.Context, d amqp.Delivery) err
 		log.Debugf("password change: skipping key parameters (incomplete pair) for TwakeID: %s", msg.TwakeID)
 	}
 
-	log.Debugf("password change: retrieving instance for domain: %s", msg.Domain)
-	inst, err := lifecycle.GetInstance(msg.Domain)
+	userDomain := msg.TwakeID + "." + msg.Domain
+	log.Debugf("password change: retrieving instance for domain: %s", userDomain)
+	inst, err := lifecycle.GetInstance(userDomain)
```
taratatach: You need the fully qualified domain name here (i.e. FQDN) which might not be `<twake_id>.<domain>` as the cloudery can change that before creating the instance.

pkg/rabbitmq/rabbitmq_test.go
```diff
@@ -128,7 +129,7 @@ func TestPasswordHandler(t *testing.T) {
 		// Compose message
 		testHash := "testhash123"
 		msg := rabbitmq.PasswordChangeMessage{
-			TwakeID:    inst.Prefix,
+			TwakeID:    username,
```
taratatach: We usually use the word `slug` for this part of the FQDN.

pkg/rabbitmq/handlers.go
shepilov: maybe i need, but I don't have, why cloudery can change it?

pkg/rabbitmq/rabbitmq_test.go
shepilov: fixed

taratatach:
Maybe these commits could be squashed?

PR #4581, opened 2025-09-30
feat: Notify members of sharing description update

taratatach:


model/sharing/sharing.go
```diff
@@ -1047,7 +1050,210 @@ func (s *Sharing) GetPreviewURL(inst *instance.Instance, state string) (string,
...
+		if err != nil {
+			inst.Logger().WithNamespace("sharing").
+				Warnf("Can't serialize the updated description for %s: %s", s.SID, err)
+			errors = append(errors, fmt.Errorf("serialization failed for %s: %w", m.Instance, err))
+			continue
+		}
+
+		// Optimization: use direct database update for recipients on the same stack
+		if IsSameStack(inst, m.Instance) {
+			err = s.UpdateDescriptionSameStack(inst, m.Instance, s.Description)
```
nono, line 1154: I don't think I would have taken the time to do this optimization, but now it's done, let's keep it

web/sharings/drives_test.go
```diff
@@ -983,6 +1016,100 @@ func TestSharedDrives(t *testing.T) {
 		})
 	})
 
+	// TODO refactor after move merge to use util function of separate drive created
```
nono: TODO spotted

web/sharings/replicator.go
```diff
@@ -175,6 +177,65 @@ func EndInitial(c echo.Context) error {
 	return c.NoContent(http.StatusNoContent)
 }
 
+// UpdateSharingMetadata is part of the replicator. It allows the owner of a
+// sharing to update the sharing metadata (currently just the description) on
+// recipient instances.
+//
+// This endpoint is called by the owner when the shared directory is renamed,
+// to keep the sharing description synchronized across all instances.
```
nono: The documentation of the API endpoint goes to `docs/sharings.md`

web/sharings/replicator.go
```diff
@@ -183,6 +236,7 @@ func replicatorRoutes(router *echo.Group) {
 	group.GET("/:sharing-id/io.cozy.files/:id", GetFolder, checkSharingReadPermissions)
 	group.PUT("/:sharing-id/io.cozy.files/:id/metadata", SyncFile, checkSharingWritePermissions)
 	group.PUT("/:sharing-id/io.cozy.files/:id", FileHandler, checkSharingWritePermissions)
+	group.PUT("/:sharing-id/metadata", UpdateSharingMetadata, checkSharingWritePermissions)
```
taratatach, line 222: Why create a new route when PATCH /sharing/:sharing-id already handles description updates?

model/sharing/sharing.go
```diff
@@ -1052,115 +1051,127 @@ func (s *Sharing) PatchDescription(inst *instance.Instance, description string)
...
+				errors = append(errors, fmt.Errorf("no token available for %s", m.Instance))
 				continue
 			}
 		}
 
-		params := PatchDescriptionParams{
-			Description: s.Description,
+		// Use the replicator endpoint which accepts sharing tokens
+		params := map[string]string{
+			"description": s.Description,
```
taratatach, line 1142: We lose type checking by not using a dedicated type for the params.

web/sharings/replicator.go
```diff
@@ -220,7 +221,14 @@ func UpdateSharingMetadata(c echo.Context) error {
 		return echo.NewHTTPError(http.StatusBadRequest, "JSON input is malformed")
 	}
 
-	s.Description = payload.Description
+	// Validate description input
+	description := strings.TrimSpace(payload.Description)
+	if len(description) > 1000 {
```
taratatach, line 201: This limit should be stored in a constant somewhere.
Was it shared with the front-end team?

shepilov: I hope they already have this limit

web/sharings/replicator.go
shepilov: /sharing/:sharing-id works only with client authentication, it doesn't support server-server replicator tokens

model/sharing/sharing.go
shepilov: why do we need type checking here? this tyte be used anywhere except next line?

taratatach: I just prefer to have type checking everywhere possible. Here nothing prevents us from putting something that's not a string in the `description` attribute of our JSON.

web/sharings/replicator.go
taratatach: Well, we should check.

PR #4586, opened 2025-10-02
feat: Allow to accept a shared drive with a SSO

nono:
Adds support for accepting shared drives using Single Sign-On (OIDC) authentication, with dynamic company branding for private OIDC providers. Users can now authenticate via either their organization's OIDC provider or the public Twake OIDC service when accepting a sharing invitation.

### DIFF
--- a/web/oidc/oidc.go
+++ b/web/oidc/oidc.go
@@ -73,7 +73,7 @@ func StartFranceConnect(c echo.Context) error {
 	return c.Redirect(http.StatusSeeOther, u)
 }
 
-// Sharing is the route to use the SSO to accept a shared drive.
+// Sharing is the route to use the SSO to accept a sharing.
 func Sharing(c echo.Context) error {
 	inst := middlewares.GetInstance(c)
 	conf, err := getGenericConfig(inst.ContextName)
@@ -88,7 +88,7 @@ func Sharing(c echo.Context) error {
 	return c.Redirect(http.StatusSeeOther, u)
 }
 
-// SharingPublic is the route to use the public Twake SSO to accept a shared drive.
+// SharingPublic is the route to use the public Twake SSO to accept a sharing.
 func SharingPublic(c echo.Context) error {
 	inst := middlewares.GetInstance(c)
 	conf, contextName, err := getPublicOIDCConfig()


pkg/config/config/testdata/full_config.yaml
```diff
@@ -171,6 +171,9 @@ authentication:
       logout_url: https://some-logout-url
       userinfo_instance_field: instance-field
 
+# Public OIDC context name - references the authentication context to use for public Twake accounts
```
taratatach, line 176: What's a public Twake account?

pkg/consts/consts.go
```diff
@@ -96,6 +96,11 @@ var KnownFlatDomains = []string{
 	"cozy.rocks",
 }
 
+// PublicSaaSDomains is a list of public SaaS domains (like *.twake.app)
+var PublicSaaSDomains = []string{
+	"twake.app",
+}
+
```
taratatach: I think this would be better in a config file so it can evolve easily and self-hosters can modify the list without modifying the code.

assets/templates/sharing_discovery_v2.html
```diff
@@ -0,0 +1,871 @@
...
+            font-style: normal;
+            font-weight: 600;
+            src: url(https://fonts.gstatic.com/s/inter/v20/UcCO3FwrK3iLTeHuS_nVMrMxCp50SjIw2boKoduKmMEVuLyfAZ9hiJ-Ek-_EeA.woff2) format('woff2');
+            unicode-range: U+0000-00FF, U+0131, U+0152-0153, U+02BB-02BC, U+02C6, U+02DA, U+02DC, U+0304, U+0308, U+0329, U+2000-206F, U+20AC, U+2122, U+2191, U+2193, U+2212, U+2215, U+FEFF, U+FFFD;
+        }
+
+        .mt-auto {
+            margin-top: auto;
+        }
+    </style>
```
Crash--: Why all this stuff? Can't we use cozy-bs & else?

assets/templates/sharing_discovery_v2.html
```diff
@@ -0,0 +1,871 @@
...
+            font-weight: 500;
+            src: url(https://fonts.gstatic.com/s/inter/v20/UcCO3FwrK3iLTeHuS_nVMrMxCp50SjIw2boKoduKmMEVuI6fAZ9hiA.woff2) format('woff2');
+            unicode-range: U+0000-00FF, U+0131, U+0152-0153, U+02BB-02BC, U+02C6, U+02DA, U+02DC, U+0304, U+0308, U+0329, U+2000-206F, U+20AC, U+2122, U+2191, U+2193, U+2212, U+2215, U+FEFF, U+FFFD;
+        }
+
+        @font-face {
+            font-family: 'Inter';
+            font-style: normal;
+            font-weight: 600;
+            src: url(https://fonts.gstatic.com/s/inter/v20/UcCO3FwrK3iLTeHuS_nVMrMxCp50SjIw2boKoduKmMEVuLyfAZ9hiJ-Ek-_EeA.woff2) format('woff2');
```
Crash--: No way we rely on google. On prod, it'll be blocked by our CSP.

assets/templates/sharing_discovery.html
```diff
@@ -49,14 +49,20 @@ <h1 class="h4 h2-md mb-2 text-center">{{t "Sharing Connect to Cozy"}}</h1>
...
+          {{end}}
           <p class="text-center mb-3">
             {{t "Sharing No Cozy"}}
             <a href="{{t "Sharing Discover Cozy URL"}}">
               {{t "Sharing Discover Cozy"}}
             </a>
           </p>
           <button id="login-submit" class="btn btn-primary btn-md-lg w-100 mb-3" type="submit">
-            {{if .Shortcut}}
+            {{if .Shortcut}}/
```
Crash--: Do we really need this "/"?

Crash--:
I don't understand the need of the private or public OIDC provider. We can't have both private and public for one instance. Or if we can have both, can you explain me how it works? 

So I don't know how a user can authenticate itself from external provider and in the same time with the public OIDC provider. I don't get it.

assets/templates/sharing_discovery_v2.html
shepilov: I've just copy pasted sign-up page, and no idea what is cozy-bs

assets/templates/sharing_discovery_v2.html
shepilov: yes,  I'll move all the statics inside project after I'll test the whole flow

assets/templates/sharing_discovery.html
shepilov: missprint

assets/templates/sharing_discovery_v2.html
Crash--: https://github.com/cozy/cozy-bootstrap this is the "design system" (only the css part) for cozy-stack. 

I find it unfortunate to add that much HTML and CSS for this page.

web/oidc/oidc.go
```diff
@@ -72,6 +73,36 @@ func StartFranceConnect(c echo.Context) error {
 	return c.Redirect(http.StatusSeeOther, u)
 }
 
+// Sharing is the route to use the SSO to accept a shared drive.
```
taratatach: Does that work only with shared drives or with all sharings?

web/oidc/oidc.go
```diff
@@ -72,6 +73,36 @@ func StartFranceConnect(c echo.Context) error {
...
+		return renderError(c, nil, http.StatusNotFound, "Sorry, the context was not found.")
+	}
+	u, err := makeSharingStartURL(inst.Domain, c.QueryParam("sharingID"), c.QueryParam("state"), conf, inst.ContextName)
+	if err != nil {
+		return renderError(c, nil, http.StatusNotFound, "Sorry, the server is not configured for OpenID Connect.")
+	}
+	return c.Redirect(http.StatusSeeOther, u)
+}
+
+// SharingPublic is the route to use the public Twake SSO to accept a shared drive.
```
taratatach: Same question here

web/oidc/oidc.go
```diff
@@ -156,8 +187,8 @@ func Redirect(c echo.Context) error {
 		return renderError(c, nil, http.StatusNotFound, "Sorry, the session has expired.")
 	}
 
-	if state.BitwardenContext != "" {
-		conf, err := getGenericConfig(state.BitwardenContext)
+	if state.OIDCContext != "" {
+		conf, err := getGenericConfig(state.OIDCContext)
```
taratatach: Won't that break Pass?

assets/templates/sharing_discovery_v2.html
shepilov: moved to static inside project

assets/templates/sharing_discovery_sso.html
```diff
@@ -0,0 +1,83 @@
...
+                <div class="flex flex-col lg:flex-row w-full xl:space-x-24 h-full">
+                    <div class="hidden lg:flex flex-col w-full lg:max-w-[506px] px-10 lg:px-0 space-y-6 pt-52 justify-center items-start pb-9">
+                        <div class="flex flex-col h-full justify-center space-y-[47px]">
+                            <img src="{{asset .Domain "/icons/twake-id-logo.svg" .ContextName}}" alt="Twake ID" width="174" height="85">
+                            <div class="flex flex-col space-y-5 h-full 4xl:h-auto pr-10"><h1 dir="auto"
+                                                                                             class="text-[32px] leading-10 not-italic font-semibold lg:leading-[44px] text-center lg:text-left w-full text-disabled-text rtl:text-right">
+                                                                                             {{t "Sharing Twake ID Description"}}</h1></div>
+                        </div>
+                        <div class="pt-auto"><span
+                                class="overflow-hidden text-sm not-italic font-medium leading-5 tracking-[0.1px] text-blueGray px-2">© 2000-2025, LINAGORA</span>
```
rezk2ll: Can we make the 2025 dynamic?

pkg/config/config/config.go
```diff
@@ -532,6 +532,50 @@ func GetFranceConnect(contextName string) (map[string]interface{}, bool) {
...
+
+	return ""
+}
+
+// GetPublicOIDC returns the public Twake OIDC configuration from the configured context
+// If contextName is empty, it uses the default context
+func GetPublicOIDC(contextName string) (map[string]interface{}, bool) {
+	if contextName == "" {
+		contextName = DefaultInstanceContext
+	}
```
taratatach, line 566: We're already falling back to the `DefaultInstanceContext` in `GetPublicOIDCContext()`

web/oidc/oidc.go
```diff
@@ -72,6 +73,36 @@ func StartFranceConnect(c echo.Context) error {
 	return c.Redirect(http.StatusSeeOther, u)
 }
 
+// Sharing is the route to use the SSO to accept a shared drive.
+func Sharing(c echo.Context) error {
+	inst := middlewares.GetInstance(c)
+	conf, err := getGenericConfig(inst.ContextName)
+	if err != nil {
+		inst.Logger().WithNamespace("oidc").Infof("Start error: %s", err)
+		return renderError(c, nil, http.StatusNotFound, "Sorry, the context was not found.")
```
taratatach, line 82: I'm not sure this error message will be useful to end users.

web/oidc/oidc.go
```diff
@@ -239,10 +270,20 @@ func Redirect(c echo.Context) error {
...
+		state = getStorage().Find(stateID)
+	}
+
 	var conf *Config
 	var err error
 	if c.QueryParam("franceconnect") == "" {
-		conf, err = getGenericConfig(inst.ContextName)
+		contextName := inst.ContextName
+		if state != nil && state.OIDCContext != "" {
+			contextName = state.OIDCContext
```
taratatach, line 287: Does that mean we can login with a different context than the instance's?

web/oidc/oidc.go
```diff
@@ -308,6 +350,54 @@ func Login(c echo.Context) error {
...
+	if !ok {
+		logger.WithNamespace("oidc").Errorf("Missing sub")
+		return ErrAuthenticationFailed
+	}
+	var memberInst *instance.Instance
+	domain, err := extractDomain(conf, params)
+	if err == nil {
+		memberInst, err = lifecycle.GetInstance(domain)
+	} else if err == ErrAuthenticationFailed && conf.AllowCustomInstance {
+		memberInst, err = findInstanceBySub(sub, state.OIDCContext)
```
taratatach, line 370: Shouldn't we try this right away if `sub` and `state.OIDCContext` are present?

web/oidc/oidc.go
```diff
@@ -308,6 +350,54 @@ func Login(c echo.Context) error {
...
+	}
+	if err != nil {
+		return renderError(c, nil, http.StatusNotFound, "Sorry, the cozy was not found.")
```
taratatach: `cozy` -> `twake`?

web/oidc/oidc.go
```diff
@@ -308,6 +350,54 @@ func Login(c echo.Context) error {
...
+	}
+
+	s, err := sharing.FindSharing(ownerInst, state.SharingID)
```
taratatach: I'm not sure I really understand what's happening here. Why are we looking for the sharing on the owner's instance (which might not even be on the same server) to accept the sharing on the member's side?

web/statik/handler.go
```diff
@@ -51,7 +51,7 @@ var (
 		"passphrase_choose.html",
 		"passphrase_reset.html",
 		"share_by_link_password.html",
-		"sharing_discovery.html",
+		"sharing_discovery_v2.html",
```
taratatach: If we replace the existing template, why not keep the name?

web/sharings/sharings.go
```diff
@@ -668,7 +668,53 @@ func renderDiscoveryForm(c echo.Context, inst *instance.Instance, code int, shar
...
+	if publicOIDC, ok := config.GetPublicOIDC(); ok {
+		if clientID, _ := publicOIDC["client_id"].(string); clientID != "" {
+			q := url.Values{
+				"sharingID": {sharingID},
+				"state":     {state},
+			}
+			twakeOIDCLink = inst.PageURL("/oidc/sharing/public", q)
+		}
+	}
+
```
taratatach, line 709: I don't really get in which situation we could have 2 OIDC providers.

pkg/config/config/testdata/full_config.yaml
shepilov: It's our SaaS offer, even if the user shared a drive with a private OIDC we want to have button to login with our public Twake Workplace

pkg/consts/consts.go
shepilov: I don't make sense for self-hosted, only if your own public cloud offer, but this check also probably don't make sense since we have `inst.ContextName != publicOIDCContext`, so I'll remove it completely

web/oidc/oidc.go
shepilov: It's for all sharings

pkg/config/config/config.go
shepilov: Yes, it's just in case GetPublicOIDCContext will be changed independently

web/oidc/oidc.go
shepilov: This error has been there since 2019

web/oidc/oidc.go
shepilov: Yes, we get OIDC configuration from a different context; the sharing might specify which OIDC context to use for
  authenticatio

web/oidc/oidc.go
shepilov: mmm, maybe like this logic will be more straightforward for OIDC flow

web/sharings/sharings.go
shepilov: When you share a file with your LNG account, a recipient will have two buttons, to login with public SaaS or your specific OIDC

web/oidc/oidc.go
taratatach: So I think the comment should be updated because right now it says `shared drive`.

pkg/config/config/config.go
taratatach: OK

web/oidc/oidc.go
taratatach: Why would the sharing (owner?) decide which context I should use to login and accept the sharing?

And I don't understand how I can log into my instance which is using context A by using the OIDC provider of context B.

web/oidc/oidc.go
taratatach: I think so.

web/sharings/sharings.go
taratatach: If we know what the recipient instance is, then we know which OIDC provider was used to sign up, right? So why offer 2 buttons when only one provider will log the user in?

shepilov: We don't know a recipient instance; we are sending an invitation by email. We know the sender's OIDC, so we offer two buttons: the sender's OIDC and the public OIDC

web/oidc/oidc.go
shepilov: We don't decide, we let users choose, does they want to use public OIDC or the sender's OIDC, which can be the case for the B2B Twake.  
There is a fallback to none-oidc flow when the sender doesn't have OIDC. 

Actually, we discussed it with @Crash--  that this "choose account" logic is supposed to be in the sign-up/login/registration application, and I hope one day it will be migrated.

For the scenario, when you installed your cozy-twake-stack in your garage, and you want to accept a share from, for example, LNG SSO, it won't work.

Also, until we have a B2B sign-up page, it won't work to accept shares between stacks with different OIDC providers until it's a public OIDC

taratatach: OK. Let's discuss this on Monday so I can understand better since it's been discusses already, let's move on.

web/sharings/sharings.go
taratatach: I still don't understand why we offer the sender's OIDC rather than the recipient's.

PR #4587, opened 2025-10-02
fix: Fix a panic when revoking a member from a shared drive

nono:


web/sharings/revoke.go
```diff
@@ -93,7 +93,7 @@ func RevocationRecipientNotif(c echo.Context) error {
 	if s.Drive {
 		token := c.Request().Header.Get(echo.HeaderAuthorization)
 		token = strings.TrimPrefix(token, "Bearer ")
-		if token == "" || token != s.Credentials[0].DriveToken {
+		if token == "" || len(s.Credentials) == 0 || token != s.Credentials[0].DriveToken {
```
taratatach, line 96: Isn't this a server error rather than an access to a forbidden resource?

nono: It could be both

taratatach: OK

PR #4588, opened 2025-10-02
fix: OIDC username with dot

Crash--:
We can have UserInfoField with "dotted string". In that case, cloudery remove the dot to create a FQDN without it.

So with username "toto.tata", cloudery will create "tototata.mycozycloud.com".

But then, when the stack does the OIDC check, stack doesn't remove the dot from the username. So we have inst.domain === tototata.mycozy.cloud but for the
constructed fqdn we have
toto.tata.mycozy.cloud.

result: error 500, Invalid domains

I don't know if this is the best way to fix the issue.

### DIFF
--- a/web/oidc/oidc.go
+++ b/web/oidc/oidc.go
@@ -425,9 +425,7 @@ func Logout(c echo.Context) error {
 				"error": "invalid claims",
 			})
 		}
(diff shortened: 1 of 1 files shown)


Crash--:
I didn't find where the cloudery remove this dot to be honest.

PR #4593, opened 2025-10-08
Common settings check remote version

shepilov:


model/instance/lifecycle/patch.go
```diff
@@ -205,11 +205,23 @@ func Patch(i *instance.Instance, opts *Options) error {
 		}
 	}
 
-	updated, err := common.UpdateCommonSettings(i, settings)
+	// Only propagate to common settings if relevant fields changed
+	shouldUpdateCommon := opts.Locale != "" ||
+		settings.M["email"] != nil ||
```
nono: Don't we always have an email here?

shepilov: Yes, we have all settings there, so this check is not relevant

PR #4597, opened 2025-10-16
fix: make sources' doctype generic

AlaeMghirbi:
Previously, the doc type was hardcoded as "io.cozy.files". This PR  makes it generic by using the value from src["doctype"].(string).

### DIFF
--- a/model/rag/chat.go
+++ b/model/rag/chat.go
@@ -155,13 +155,22 @@ func getSources(event map[string]interface{}) ([]Source, error) {
 		if !ok {
 			continue
 		}
+		doctype, _ := src["doctype"].(string)
+		file_id, _ := src["file_id"].(string)
+		file_name, _ := src["filename"].(string)
+		page := 0
+		if p, ok := src["page"].(float64); ok {
+			page = int(p)
+		}
+		file_url, _ := src["file_url"].(string)
+		chunk_url, _ := src["chunk_url"].(string)
 		sources = append(sources, Source{
-			ID:       src["file_id"].(string),
-			DocType:  src["doctype"].(string),
-			Filename: src["filename"].(string),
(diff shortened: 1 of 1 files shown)


model/rag/chat.go
```diff
@@ -157,7 +157,7 @@ func getSources(event map[string]interface{}) ([]Source, error) {
 		}
 		sources = append(sources, Source{
 			ID:       src["file_id"].(string),
-			DocType:  "io.cozy.files",
+			DocType:  src["doctype"].(string),
```
paultranvan: We should probably have more defensive code in case the doctype is not provided (not only for doctypes, btw)

model/rag/chat.go
```diff
@@ -155,13 +155,19 @@ func getSources(event map[string]interface{}) ([]Source, error) {
...
-			ID:       src["file_id"].(string),
-			DocType:  src["doctype"].(string),
-			Filename: src["filename"].(string),
-			Page:     int(src["page"].(float64)),
-			FileURL:  src["file_url"].(string),
-			ChunkURL: src["chunk_url"].(string),
+			ID:       file_id,
+			DocType:  doctype,
+			Filename: file_name,
+			Page:     int(page),
```
paultranvan: I think this can cause an error if the page is nil. You might need to do this instead
```go
page := 0
if p, ok := src["page"].(float64); ok {
    page = int(p)
}
```

paultranvan: Now that you made the change, there is no need to convert again `int(page`, `Page:     page` is enough

AlaeMghirbi: Abosolutely,   my bad i forgot to commit that line modifications'

nono (approved):

PR #4598, opened 2025-10-18
feat: Can patch metadata on folder

Crash--:
Today I think we can create `folder` with metadata by creating the metadata object with `POST /files/metadata` and then call `POST /files/...?metadataId=X`

Since we can create folder with metadata, we should be able to update them. 

I'm not super happy with the implementation because I had to edit the `DocPatch` structure. I didn't reuse `ModifyFileMetadata` but this method does work that is not useful for folder. 

Didn't tested yet and didn't launch tests neither (my local env is a big mess)

### DIFF
--- a/model/vfs/file.go
+++ b/model/vfs/file.go
@@ -342,6 +342,9 @@ func ModifyFileMetadata(fs VFS, olddoc *FileDoc, patch *DocPatch) (*FileDoc, err
 	newdoc.RestorePath = *patch.RestorePath
 	newdoc.UpdatedAt = *patch.UpdatedAt
 	newdoc.Metadata = olddoc.Metadata
+	if patch.Metadata != nil {
+		newdoc.Metadata = *patch.Metadata
+	}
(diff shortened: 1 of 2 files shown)


doubleface:
It works. Thank you @shepilov

model/vfs/file.go
```diff
@@ -342,6 +342,9 @@ func ModifyFileMetadata(fs VFS, olddoc *FileDoc, patch *DocPatch) (*FileDoc, err
 	newdoc.RestorePath = *patch.RestorePath
 	newdoc.UpdatedAt = *patch.UpdatedAt
 	newdoc.Metadata = olddoc.Metadata
+	if patch.Metadata != nil {
+		newdoc.Metadata = *patch.Metadata
+	}
```
nono, line 347: I think it was on purpose that we didn't allow to patch the metadata of a file without uploading a new version. At least, we have some specials metadata like electronic safe that must not be set with a patch.

PR #4600, opened 2025-10-20
Move folder to/from shared drive

shepilov:


web/sharings/move.go
```diff
@@ -288,162 +316,461 @@ func moveFileSameStack(c echo.Context, srcInst *instance.Instance, destInst *ins
...
 		}
-		if OnSameStackCheck(sourceInstance, destInstance) {
-			if moveDirectory {
-				return moveDirectorySameStack(c, sourceInstance, destInstance, req.Source.DirID, req.Dest.DirID, destSharing)
-			} else {
-				return moveFileSameStack(c, sourceInstance, destInstance, req.Source.FileID, req.Dest.DirID, destSharing)
-			}
+	}
+
+	// Delete source directories bottom-up (reverse order)
```
nono: [nit] I think this comment is a bit misleading

web/sharings/move.go
```diff
@@ -288,162 +316,461 @@ func moveFileSameStack(c echo.Context, srcInst *instance.Instance, destInst *ins
...
+			return files.WrapVfsError(errors.New("destination parent directory not found while moving file"))
+		}
+		// When moving a directory we copy files but do not delete sources here
+		if _, err := moveFileToSharedDriveCore(srcInst, destParentID, f.DocID, s, false); err != nil {
+			return files.WrapVfsError(err)
+		}
+	}
+
+	// Delete source directories bottom-up (reverse order)
+	_, _, err = srcInst.VFS().DeleteDirDocAndContent(srcRoot, false)
```
nono: IIRC, here, we put the dirs/files in the trash. But when moving/copying from a shared drive, we delete them (no trash). It's a bit weird to not have the same behavior.

PR #4602, opened 2025-10-29
SSO logout for flagship app

shepilov:


web/oidc/oidc.go
```diff
@@ -559,6 +559,25 @@ func AccessToken(c echo.Context) error {
 		}
 	}
 
+	// Store the session ID in the client for logout purposes
+	if client.Flagship && reqBody.IDToken != "" {
```
taratatach: There should be 2 nested conditional blocks otherwise lines 576 and 577 will apply to non-flagship clients.

model/oauth/client.go
```diff
@@ -891,3 +899,144 @@ func PushClientsLimitAlert(i *instance.Instance, clientName string, clientsLimit
 }
 
 var _ couchdb.Doc = &Client{}
+
+// OpenIDConfiguration represents the OpenID Connect configuration from the well-known endpoint
+type OpenIDConfiguration struct {
```
taratatach: What do you think about naming this type `OIDCConfiguration` instead? This way we use OIDC everywhere which makes it easier to search for.

model/oauth/client.go
```diff
@@ -891,3 +899,144 @@ func PushClientsLimitAlert(i *instance.Instance, clientName string, clientsLimit
...
+}
+
+var oidcClient = &http.Client{
+	Timeout: 15 * time.Second,
+}
+
+const cacheTTL = 24 * time.Hour
+
+// fetchOpenIDConfiguration fetches the OpenID Connect configuration from the well-known endpoint
+func fetchOpenIDConfiguration(contextName string) (*OpenIDConfiguration, error) {
```
taratatach: What do you think about naming this function `fetchOIDCConfiguration` instead? This way we use OIDC everywhere which makes it easier to search for.

model/oauth/client.go
```diff
@@ -891,3 +899,144 @@ func PushClientsLimitAlert(i *instance.Instance, clientName string, clientsLimit
...
+	oidc, ok := config.GetOIDC(contextName)
+	if !ok {
+		return nil, fmt.Errorf("no OIDC is configured for this context")
+	}
+
+	// Get the token URL to extract the base URL
+	tokenURL, ok := oidc["token_url"].(string)
+	if !ok {
+		return nil, fmt.Errorf("the token_url is missing for this context")
```
taratatach:
```suggestion
		return nil, fmt.Errorf("the OIDC token_url is missing for this context")
```

model/oauth/client.go
```diff
@@ -891,3 +899,144 @@ func PushClientsLimitAlert(i *instance.Instance, clientName string, clientsLimit
...
+	}
+
+	// Parse the token URL to extract the base URL
+	parsedURL, err := url.Parse(tokenURL)
+	if err != nil {
+		return nil, fmt.Errorf("invalid token URL: %w", err)
```
taratatach:
```suggestion
		return nil, fmt.Errorf("invalid OIDC token URL: %w", err)
```

model/oauth/client.go
```diff
@@ -891,3 +899,144 @@ func PushClientsLimitAlert(i *instance.Instance, clientName string, clientsLimit
...
+	cacheKey := "oidc-config:" + wellKnownURL.String()
+
+	data, ok := cache.Get(cacheKey)
+	if !ok {
+		// Fetch from the server
+		req, err := http.NewRequest(http.MethodGet, wellKnownURL.String(), nil)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create request: %w", err)
+		}
+		req.Header.Add("User-Agent", "cozy-stack "+build.Version+" ("+runtime.Version()+")")
```
taratatach: nit: we should extract this user agent to a constant since it's used in multiple places.

model/oauth/client.go
```diff
@@ -891,3 +899,144 @@ func PushClientsLimitAlert(i *instance.Instance, clientName string, clientsLimit
...
+	var oidcConfig OpenIDConfiguration
+	if err := json.Unmarshal(data, &oidcConfig); err != nil {
+		return nil, fmt.Errorf("failed to parse OpenID configuration: %w", err)
+	}
+
+	return &oidcConfig, nil
+}
+
+// PerformOIDCLogout calls the end_session_endpoint to terminate the SSO session
+func PerformOIDCLogout(contextName, sessionID string) error {
```
taratatach, line 987: This function always returns `nil`. We could return errors instead of logging them here and log the error in the caller instead.

This would still allow us to not fail the client deletion since we can ignore the error after logging it.

model/oauth/client.go
```diff
@@ -891,3 +899,144 @@ func PushClientsLimitAlert(i *instance.Instance, clientName string, clientsLimit
...
+		logger.WithNamespace("oidc").Warnf("Invalid end_session_endpoint URL: %s", err)
+		return nil
+	}
+
+	// Add the session_id parameter
+	q := endSessionURL.Query()
+	q.Add("session_id", sessionID)
+	endSessionURL.RawQuery = q.Encode()
+
+	req, err := http.NewRequest(http.MethodGet, endSessionURL.String(), nil)
```
taratatach, line 1018: Is it really ending the session on a `GET` request? I would expect a `POST` request here.

web/oidc/oidc_test.go
```diff
@@ -258,21 +266,341 @@ func TestOidc(t *testing.T) {
...
-          "scope": "*",
-          "code": "%s"
-        }`, oauthClient.ClientID, oauthClient.ClientSecret, code))).
+			WithJSON(map[string]string{
+				"client_id":     oauthClient.ClientID,
+				"client_secret": oauthClient.ClientSecret,
+				"scope":         "*",
+				"code":          code,
+				"id_token":      idTokenStr,
+			}).
```
taratatach, line 283: :+1:

model/oauth/client.go
```diff
@@ -20,10 +22,13 @@ import (
 	"github.com/cozy/cozy-stack/model/job"
 	"github.com/cozy/cozy-stack/model/notification"
 	"github.com/cozy/cozy-stack/model/permission"
+	build "github.com/cozy/cozy-stack/pkg/config"
```
Crash--, line 24: why do we have this "build" line here? (for my personal culture)

Crash-- (approved):
Can we add a bit of documentation for this new feature please? At least a word about this well-know file and so on.

Also, is it really a standard to have it here `/.well-known/openid-configuration` or should we make it configurable?

model/oauth/client.go
taratatach: It's a local package alias (otherwise the imported package name would be `config`).

model/oauth/client.go
shepilov: make sense, I just ton't like double C)

model/oauth/client.go
shepilov: not a constant maybe but function, it depends of the current stack version

model/oauth/client.go
shepilov: Both methods are supported in the OIDC spec

model/oauth/client.go
shepilov: but right now it won't help us, the behaviour will be the same

taratatach: It would mean a useful return value (I don't see the point of adding a return type that we'll have to deal with in every call when the returned value is always `nil`) and one call to the logger instead of many.

model/oauth/client.go
taratatach: Yes but the version does not change during execution time. So it could be set at startup.

model/oauth/client.go
taratatach: OK. I find it peculiar but :man_shrugging:

pkg/config/build.go
```diff
@@ -23,3 +28,14 @@ var (
...
+var userAgentOnce sync.Once
+var userAgent string
+
+// UserAgent returns the Cozy Stack User-Agent string and caches it.
+func UserAgent() string {
+	userAgentOnce.Do(func() {
+		userAgent = "cozy-stack " + Version + " (" + runtime.Version() + ")"
+	})
+	return userAgent
+}
```
taratatach, line 41: Nice!

PR #4604, opened 2025-10-29
feat: Use separate RabbitMQ Managers per contexts

taratatach:
We want to be able to isolate RabbitMQ messages per contexts to
  prevent conflicts and provide different QoS levels.

  Each context can be associated with a specific RabbitMQ node via its
  URL and be disabled separately.
  If a `default` context configuration is defined, messages meant for a
  context without configuration will be sent using this `default` node.

  Exchanges and queues definitions are shared across all contexts
  though.

### DIFF
--- a/model/stack/main.go
+++ b/model/stack/main.go
@@ -3,7 +3,6 @@ package stack
 import (
 	"context"
 	"fmt"
-	"github.com/cozy/cozy-stack/pkg/rabbitmq"
 	"os"
 
 	"github.com/cozy/cozy-stack/model/cloudery"
@@ -16,6 +15,7 @@ import (
 	"github.com/cozy/cozy-stack/pkg/config/config"
 	"github.com/cozy/cozy-stack/pkg/couchdb"
 	"github.com/cozy/cozy-stack/pkg/emailer"
+	"github.com/cozy/cozy-stack/pkg/rabbitmq"
 	"github.com/cozy/cozy-stack/pkg/utils"
 	"github.com/google/gops/agent"
 )
@@ -51,6 +51,7 @@ func (g gopAgent) Shutdown(ctx context.Context) error {
 type Services struct {
 	Emailer  emailer.Emailer
 	Settings settings.Service
+	RabbitMQ rabbitmq.Service
 }
 
 // Start is used to initialize all the
@@ -110,9 +111,15 @@ security features. Please do not use this binary as your production server.
 	instanceSvc := instance.Init()
 	clouderySvc := cloudery.Init(config.GetConfig().Clouderies)
 
+	rabbitmqSvc, err := rabbitmq.Init(config.GetConfig().RabbitMQ)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to init the rabbitmq service: %w", err)
+	}
+
 	services := Services{
 		Emailer:  emailerSvc,
 		Settings: settings.Init(emailerSvc, instanceSvc, tokenSvc, clouderySvc),
+		RabbitMQ: rabbitmqSvc,
 	}
 
 	// Initialize the dynamic assets FS. Can be OsFs, MemFs or Swift
@@ -123,13 +130,13 @@ security features. Please do not use this binary as your production server.
 		}
 	}
 
-	// Start RabbitMQ manager if enabled in configuration
-	if cfg := config.GetConfig().RabbitMQ; cfg.Enabled {
-		shutdowner, err := rabbitmq.Start(cfg)
-		if err != nil {
-			return nil, nil, fmt.Errorf("failed to start rabbitmq manager: %w", err)
-		}
-		shutdowners = append(shutdowners, shutdowner)
+	// Start RabbitMQ managers
+	rmqManagers, err := rabbitmqSvc.StartManagers()
+	for _, manager := range rmqManagers {
+		shutdowners = append(shutdowners, manager)
+	}
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to start rabbitmq manager: %w", err)
 	}
 
 	// Global shutdowner that composes all the running processes of the stack
--- a/pkg/config/config/config.go
+++ b/pkg/config/config/config.go
@@ -160,12 +160,18 @@ type Config struct {
 	AssetsPollingInterval time.Duration
 }
 
-// RabbitMQ contains configuration for the RabbitMQ consumer.
+// RabbitMQ contains configuration for the RabbitMQ consumers.
 type RabbitMQ struct {
-	Enabled   bool             `mapstructure:"enabled" yaml:"enabled"`
-	URL       string           `mapstructure:"url" yaml:"url"`
-	Exchanges []RabbitExchange `mapstructure:"exchanges" yaml:"exchanges"`
-	TLS       RabbitMQTLS      `mapstructure:"tls" yaml:"tls"`
+	Enabled   bool                    `mapstructure:"enabled" yaml:"enabled"`
+	Nodes     map[string]RabbitMQNode `mapstructure:"nodes" yaml:"nodes"`
+	Exchanges []RabbitExchange        `mapstructure:"exchanges" yaml:"exchanges"`
+}
+
+// RabbitMQNode defines the configuration for the connection with a RabbitMQ node.
+type RabbitMQNode struct {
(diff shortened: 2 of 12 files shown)


pkg/rabbitmq/service.go
```diff
@@ -0,0 +1,63 @@
...
+	Managers map[string]*RabbitMQManager
+}
+
+func NewService(cfg config.RabbitMQ) *RabbitMQService {
+	managers := map[string]*RabbitMQManager{}
+	for context, node := range cfg.Nodes {
+		manager, err := buildManager(node, cfg.Exchanges)
+		if err != nil {
+			log.Errorf("Error while building RabbitMQ manager: %v", err)
+			continue
```
shepilov: If we have rabbitmq configuraion ithen it supposed to work, i think we should not start stack if it's missconfigured.

pkg/config/config/config.go
```diff
@@ -160,12 +160,19 @@ type Config struct {
 	AssetsPollingInterval time.Duration
 }
 
-// RabbitMQ contains configuration for the RabbitMQ consumer.
+// RabbitMQ contains configuration for the RabbitMQ consumers.
 type RabbitMQ struct {
```
shepilov, line 164: could you please also update docs in docs/config.md with new configuration

pkg/rabbitmq/service_noop.go
```diff
@@ -0,0 +1,12 @@
+package rabbitmq
+
+// NoopService implements [Service].
+//
+// This implem does nothing. It is used when no config is provided.
+type NoopService struct{}
+
+// StartManagers does nothing.
+func (s *NoopService) StartManagers() ([]*RabbitMQManager, error) {
+	log.Warnf("No RabbitMQ managers to start")
```
shepilov, line 10: Consider Info to avoid noisy logs when RabbitMQ is intentionally disabled

pkg/config/config/config.go
```diff
@@ -160,12 +160,19 @@ type Config struct {
...
-	Enabled   bool             `mapstructure:"enabled" yaml:"enabled"`
-	URL       string           `mapstructure:"url" yaml:"url"`
-	Exchanges []RabbitExchange `mapstructure:"exchanges" yaml:"exchanges"`
-	TLS       RabbitMQTLS      `mapstructure:"tls" yaml:"tls"`
+	Enabled   bool                    `mapstructure:"enabled" yaml:"enabled"`
```
shepilov, line 165: You decide not to implement logic with the default context/node mentions in PR?

model/stack/main.go
```diff
@@ -123,13 +126,13 @@ security features. Please do not use this binary as your production server.
...
-	// Start RabbitMQ manager if enabled in configuration
-	if cfg := config.GetConfig().RabbitMQ; cfg.Enabled {
-		shutdowner, err := rabbitmq.Start(cfg)
-		if err != nil {
-			return nil, nil, fmt.Errorf("failed to start rabbitmq manager: %w", err)
-		}
-		shutdowners = append(shutdowners, shutdowner)
+	// Start RabbitMQ managers
+	rmqManagers, err := rabbitmqSvc.StartManagers()
+	for _, manager := range rmqManagers {
```
shepilov, line 135: We append shutdowners before checking error?

pkg/rabbitmq/rabbitmq_test.go
```diff
@@ -103,15 +106,15 @@ func TestRabbitMQManager(t *testing.T) {
 }
 
 func TestPasswordHandler(t *testing.T) {
```
shepilov, line 110: Could you please also add an integration test with multiple context and rabbitmqs?

pkg/rabbitmq/service_noop.go
taratatach: Well, I believe RabbitMQ will rarely be disabled and I want to avoid missing this crucial information.
Since we'll start the managers only once per stack there shouldn't be too many logs, right?

model/stack/main.go
taratatach: Yes because we could have successfully started some managers before we get an error. So I want them to be in the list of shutdowners so they can be stopped properly if the error leads to this.

pkg/rabbitmq/rabbitmq_test.go
taratatach: What would you want to test?

pkg/config/config/config.go
taratatach: I don't understand what you mean. I added the `Nodes` config attribute to deal with this.
This `Enabled` attribute is used to globally disable RabbitMQ.

pkg/rabbitmq/service.go
taratatach: The error could be that we failed to connect to the RabbitMQ node.
I wanted to avoid preventing the stack to start but maybe we wouldn't really catch that one manager was not started.
I'll return the error then.

shepilov: connection errors are handled with manager with reconnection logic

pkg/rabbitmq/rabbitmq_test.go
shepilov: that stack is started with multiple rabbitmq connection, ans consuming messages from all of them

pkg/config/config/config.go
shepilov: from the PR description:
"If a default context configuration is defined, messages meant for a
context without configuration will be sent using this default node."

taratatach: Yes, that's true. I still don't get what's the problem with this global `Enabled` flag though.

pkg/rabbitmq/rabbitmq_test.go
taratatach: There is a test for that: https://github.com/cozy/cozy-stack/pull/4604/files#diff-f0e7e66d6f49cc16d31d1404c3cf15ef1fd2997209e5d299fb789fc99181d75dR95-R147

pkg/rabbitmq/service.go
taratatach: The reconnection logic only works if we can connect in the first place. Otherwise we don't even create a manager.

pkg/config/config/config.go
taratatach: There was no docs for this actually.

pkg/rabbitmq/service.go
shepilov: then i think you should not create connection in initRabbitMQManager and keep in inside manager startup, you are ignoring this errors anyway, i don't see benefits, only degradation that we ignore all the errors during start up

pkg/config/config/config.go
shepilov: this one actually docs/rabbitmq.md

pkg/config/config/config.go
shepilov: could you change it in PR description then? confusing

shepilov: I thought that "Enabled" should be overriden, that's why on this line

pkg/rabbitmq/service_mock.go
```diff
@@ -0,0 +1,27 @@
+package rabbitmq
+
```
shepilov: also, why do we need mock service if we never use it?

pkg/config/config/config.go
taratatach: I'm sorry but  I still don't understand what you want me to change.

pkg/rabbitmq/service_mock.go
taratatach: I went ahead and created it as we did for other services but you're right, if not used, no need to create it now.

pkg/config/config/config.go
shepilov: description of the PR as far as we don't have "default" rabbitmq node

taratatach: But we do, if defined. That's what the PR description says: `If a default context configuration is defined [...]`

pkg/config/config/config.go
```diff
@@ -160,12 +160,19 @@ type Config struct {
...
+	URL       string                  `mapstructure:"url" yaml:"url"`
```
shepilov:
```suggestion
```

PR #4606, opened 2025-10-29
feat: RabbitMQ support for updating a user’s phone

shepilov:
Implemented RabbitMQ support for updating a user’s phone number from the new auth exchange event.

Documented the additional queue and JSON schema in docs/rabbitmq.md

```yaml
rabbitmq:
  exchanges:
    - name: auth
      queues:
        - name: stack.user.phone.updated
          declare: true
          declare_dlx: true
          declare_dlq: true
          dlq_name: stack.dead.letter.user.phone.updated
          dl_routing_key: user.phone.updated.dead
          prefetch: 8
          delivery_limit: 5
          bindings:
            - user.phone.updated
 ```

### DIFF
--- a/pkg/rabbitmq/handlers.go
+++ b/pkg/rabbitmq/handlers.go
@@ -262,7 +262,7 @@ func (h *UserPhoneUpdatedHandler) Handle(ctx context.Context, d amqp.Delivery) e
 
 	if err := lifecycle.Patch(inst, &lifecycle.Options{
 		Phone:        msg.Mobile,
-		FromCloudery: true,
+		FromCloudery: true, // XXX: do not update the instance's phone on the Cloudery as its API does not permit it and the Cloudery only uses it when requesting the instance creation from the stack
 	}); err != nil {
 		return fmt.Errorf("user.phone.updated: update settings: %w", err)
 	}

### CODE REVIEW

File: pkg/rabbitmq/handlers.go [hunk 1]
```diff
@@ -215,6 +215,62 @@ func (h *UserCreatedHandler) Handle(ctx context.Context, d amqp.Delivery) error
...
+	}
+
+	inst, err := lifecycle.GetInstance(msg.WorkplaceFqdn)
+	if err != nil {
+		return fmt.Errorf("user.phone.updated: get instance: %w", err)
+	}
+
+	if err := lifecycle.Patch(inst, &lifecycle.Options{
+		Phone:        msg.Mobile,
+		FromCloudery: true,
```
Crash--: Do we need this FromCloudery since this is fromSignup in fact

pkg/rabbitmq/handlers.go
```diff
@@ -215,6 +215,62 @@ func (h *UserCreatedHandler) Handle(ctx context.Context, d amqp.Delivery) error
...
+		return fmt.Errorf("user.phone.updated: missing twakeId")
+	}
+	if msg.Mobile == "" {
+		return fmt.Errorf("user.phone.updated: missing mobile")
+	}
+	if msg.WorkplaceFqdn == "" {
+		return fmt.Errorf("user.phone.updated: missing workplaceFqdn")
+	}
+
+	inst, err := lifecycle.GetInstance(msg.WorkplaceFqdn)
```
Crash--, line 258: this `GetInstance` can work both with domain & oldDomain? If the user tries to update phone number, it should have been migrated, so in fact no need to check oldDomain

rezk2ll: I'm afraid this won't work on migrated Cozy users.
Since the workplace FQDN comes from the `user.created` message, and this message is only sent after the instance is created.
However, in the case of migrated cozy users, the instance already exists, so no message is displayed.
Also, we can't hardcode this value in the migration script because we can't really tell if the instance will be `user.domain` and not `user2.domain`

shepilov: why we can't tell user.domain and not user2.domain, at what step we are starting to know what domain instance will have?

pkg/rabbitmq/handlers.go
shepilov: good question, is this case we won't send request with the updated setting to cloudery, but we I don't know should we or not.
@taratatach do you know, do we need updated phone in cloudery?

pkg/rabbitmq/handlers.go
rezk2ll: Currently, after instance creation:
- sign up -> onboarding
- login -> sso -> instance created

pkg/rabbitmq/handlers.go
```diff
@@ -215,6 +215,62 @@ func (h *UserCreatedHandler) Handle(ctx context.Context, d amqp.Delivery) error
...
+
+// NewUserPhoneUpdatedHandler creates a new user phone update handler.
+func NewUserPhoneUpdatedHandler() *UserPhoneUpdatedHandler {
+	return &UserPhoneUpdatedHandler{}
+}
+
+// UserPhoneUpdatedMessage represents a user phone update message.
+type UserPhoneUpdatedMessage struct {
+	TwakeID       string `json:"twakeId"`
+	Mobile        string `json:"mobile"`
```
taratatach, line 229: Let's use the `phone` wording everywhere.

```suggestion
	Phone        string `json:"phone"`
```

pkg/rabbitmq/handlers.go
taratatach: The cloudery's migration endpoint expects the `old_fqdn` and `new_fqdn` parameters so when the migration script calls it, it knows the instance's FQDN.

So we should be able to store it in LDAP at this point.

pkg/rabbitmq/handlers.go
taratatach: The phone is stored in the cloudery as well so we should update it there as well.

Therefore we should not use this parameter.

pkg/rabbitmq/handlers.go
rezk2ll: alright, i made a fix here: https://github.com/linagora/twake-workplace-private/pull/562

pkg/rabbitmq/handlers.go
shepilov: it's just a part of the spec https://www.notion.so/linagora/ADR-020-Communication-between-apps-using-RabbitMQ-26462718bad180ff9ee1f90f291f5cff

pkg/rabbitmq/handlers.go
shepilov: Does cloudery recieves messages from RabbitMQ about common settings updates?

shepilov: it looks like no

shepilov: @taratatach phone is stored in cloudery but it was never updated before during setting patch. 
Also it's used in coudery nowhere. If we need to have and updated phone number in cloudery, we should subscribe to this event there

taratatach: You're right, it's only used to create the instance by calling the stack.

@Crash-- if we tend towards removing instance creation from the Cloudery in favor of Signup, I think it's not /worth spending time to implement the message subscription in the Cloudery to update this attribute.
So unless you think we should make sure it stays up-to-date, I'll leave it as is.

pkg/rabbitmq/handlers.go
taratatach: OK. Too bad we didn't agree on a shared wording across apps :cry:

pkg/rabbitmq/handlers.go
taratatach: Maybe we could add a comment explaining why we're not syncing the data update with the Cloudery.
Something like this perhaps?

```suggestion
		FromCloudery: true, // XXX: do not update the instance's phone on the Cloudery as its API does not permit it and the Cloudery only uses it when requesting the instance creation from the stack
```

PR #4607, opened 2025-10-29
Accept shared drive on following invitation link

shepilov:
It will update member status for a shared drive owner and allow to call replication endpoint on shared drive member updates

### DIFF
--- a/web/auth/oauth.go
+++ b/web/auth/oauth.go
@@ -545,6 +545,15 @@ func (a *AuthorizeHTTPHandler) authorizeSharingForm(c echo.Context) error {
 		if err := s.AddShortcut(instance, params.state); err != nil {
 			return err
 		}
+
+		if s.Drive && !isMemberReady(s, params.state) {
+			if params.state == "" {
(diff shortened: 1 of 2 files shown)


web/auth/oauth.go
```diff
@@ -545,6 +545,15 @@ func (a *AuthorizeHTTPHandler) authorizeSharingForm(c echo.Context) error {
 		if err := s.AddShortcut(instance, params.state); err != nil {
 			return err
 		}
+
+		if s.Drive && !isMemberReady(s, params.state) {
```
taratatach, line 549: I think the second condition, `!isMemberReady(s, params.state)` should be nested in the `s.Drive` conditional block, after we check that `params.state` is not empty as it does not make much sense to check if the member is ready based on an empty state.

shepilov: if will just return false and if it's not a drive function won't be called

taratatach: Sure but I find it odd to test a member's state on a state that's empty is all.
Not a big deal though.

PR #4609, opened 2025-10-30
feat:  recognition of emailRessouce metadatas

AlaeMghirbi:
Adding email metadata required for later display in the "Resources" section

### DIFF
--- a/model/rag/chat.go
+++ b/model/rag/chat.go
@@ -78,12 +78,16 @@ type QueryMessage struct {
 }
 
 type Source struct {
-	ID       string `json:"id"`
-	DocType  string `json:"doctype"`
-	Filename string `json:"filename"`
-	FileURL  string `json:"fileUrl"`
-	ChunkURL string `json:"chunkUrl"`
-	Page     int    `json:"page"`
+	ID           string `json:"id"`
+	DocType      string `json:"doctype"`
+	Filename     string `json:"filename"`
+	FileURL      string `json:"fileUrl"`
+	ChunkURL     string `json:"chunkUrl"`
+	Page         int    `json:"page"`
+	EmailPreview string `json:"email.preview,omitempty"`
+	GroupID      string `json:"email.groupId,omitempty"`
+	Subject      string `json:"email.subject,omitempty"`
+	Date         string `json:"email.date,omitempty"`
 }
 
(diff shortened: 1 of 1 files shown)


shepilov:
@AlaeMghirbi do we need to bump DocTypeVersion in case of any changes?

model/rag/chat.go
```diff
@@ -165,12 +173,16 @@ func getSources(event map[string]interface{}) ([]Source, error) {
 		file_url, _ := src["file_url"].(string)
 		chunk_url, _ := src["chunk_url"].(string)
 		sources = append(sources, Source{
```
shepilov, line 175: or vice versa to avoid reconfiguration we should keep it the same?

model/rag/chat.go
```diff
@@ -78,12 +78,16 @@ type QueryMessage struct {
...
+	ID           string `json:"id"`
+	DocType      string `json:"doctype"`
+	Filename     string `json:"filename"`
+	FileURL      string `json:"fileUrl"`
+	ChunkURL     string `json:"chunkUrl"`
+	Page         int    `json:"page"`
+	EmailPreview string `json:"email.preview,omitempty"`
+	GroupID      string `json:"email.groupId,omitempty"`
+	Subject      string `json:"email.subject,omitempty"`
+	Date         string `json:"email.date,omitempty"`
```
paultranvan, line 90: This is not required, we should use `datetime` instead

model/rag/chat.go
```diff
@@ -78,12 +78,16 @@ type QueryMessage struct {
...
-	ChunkURL string `json:"chunkUrl"`
-	Page     int    `json:"page"`
+	ID           string `json:"id"`
+	DocType      string `json:"doctype"`
+	Filename     string `json:"filename"`
+	FileURL      string `json:"fileUrl"`
+	ChunkURL     string `json:"chunkUrl"`
+	Page         int    `json:"page"`
+	EmailPreview string `json:"email.preview,omitempty"`
+	GroupID      string `json:"email.groupId,omitempty"`
```
paultranvan, line 88: This is not specific to email, please use `group_id` instead

paultranvan: Also, `parent_id` is missing

PR #4610, opened 2025-10-31
fix :  making attributes coherent with the spec

AlaeMghirbi:
spec link : https://www.notion.so/linagora/Email-Metadata-Specification-for-RAG-Indexing-28e62718bad180efa1a7eadc08f66e9d#28e62718bad180b4979ddf199550f353

### DIFF
--- a/model/rag/chat.go
+++ b/model/rag/chat.go
@@ -78,16 +78,17 @@ type QueryMessage struct {
 }
 
 type Source struct {
-	ID           string `json:"id"`
-	DocType      string `json:"doctype"`
-	Filename     string `json:"filename"`
-	FileURL      string `json:"fileUrl"`
-	ChunkURL     string `json:"chunkUrl"`
-	Page         int    `json:"page"`
-	EmailPreview string `json:"email.preview,omitempty"`
-	GroupID      string `json:"email.groupId,omitempty"`
-	Subject      string `json:"email.subject,omitempty"`
-	Date         string `json:"email.date,omitempty"`
+	ID             string `json:"id"`
+	DocType        string `json:"doctype"`
+	Filename       string `json:"filename"`
+	FileURL        string `json:"fileUrl"`
+	ChunkURL       string `json:"chunkUrl"`
+	Page           int    `json:"page"`
+	EmailPreview   string `json:"email.preview,omitempty"`
+	RelationshipID string `json:"relationship_id,omitempty"`
+	ParentID       string `json:"parent_id,omitempty"`
+	Subject        string `json:"email.subject,omitempty"`
+	Datetime       string `json:"datetime,omitempty"`
 }
 
 func Chat(inst *instance.Instance, payload ChatPayload) (*ChatConversation, error) {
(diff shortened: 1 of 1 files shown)


model/rag/chat.go
```diff
@@ -85,9 +85,10 @@ type Source struct {
 	ChunkURL     string `json:"chunkUrl"`
 	Page         int    `json:"page"`
 	EmailPreview string `json:"email.preview,omitempty"`
-	GroupID      string `json:"email.groupId,omitempty"`
+	Group_id     string `json:"group_id,omitempty"`
```
shepilov: you don't need to rename fiels, just rename json mapping and keep go naming unchanged

model/rag/chat.go
```diff
@@ -85,9 +85,10 @@ type Source struct {
...
+	Parent_id    string `json:"parent_id,omitempty"`
```
shepilov: the same

model/rag/chat.go
AlaeMghirbi: Yes, you’re right. I just preferred to keep consistent names between the fields and the JSON mapping for easier manipulation.

model/rag/chat.go
```diff
@@ -85,9 +85,10 @@ type Source struct {
 	ChunkURL     string `json:"chunkUrl"`
 	Page         int    `json:"page"`
 	EmailPreview string `json:"email.preview,omitempty"`
-	GroupID      string `json:"email.groupId,omitempty"`
+	GroupId      string `json:"group_id,omitempty"`
+	ParentId     string `json:"parent_id,omitempty"`
```
taratatach: nit: We usually write id with uppercases in our variable and attribute names so it might be good to stay consistent with the rest of the code base and use `GroupID` and `ParentID` here.

model/rag/chat.go
```diff
@@ -155,13 +160,32 @@ func getSources(event map[string]interface{}) ([]Source, error) {
...
+		parentid, _ := src["parent_id"].(string)
+		doctype, _ := src["doctype"].(string)
+		file_id, _ := src["file_id"].(string)
+		file_name, _ := src["filename"].(string)
+		page := 0
+		if p, ok := src["page"].(float64); ok {
+			page = int(p)
+		}
+		file_url, _ := src["file_url"].(string)
+		chunk_url, _ := src["chunk_url"].(string)
```
paultranvan: in the stack, we are camelCase, not snake_case, for the var names. So `chunkUrl`, `emailPreview`, etc

model/rag/chat.go
```diff
@@ -78,16 +78,17 @@ type QueryMessage struct {
...
-	Date         string `json:"email.date,omitempty"`
+	ID             string `json:"id"`
+	DocType        string `json:"doctype"`
+	Filename       string `json:"filename"`
+	FileURL        string `json:"fileUrl"`
+	ChunkURL       string `json:"chunkUrl"`
+	Page           int    `json:"page"`
+	EmailPreview   string `json:"email.preview,omitempty"`
+	RelationshipId string `json:"relationship_id,omitempty"`
+	ParentId       string `json:"parent_id,omitempty"`
```
taratatach: Please use `ID` here to keep naming consistency.

```suggestion
	ParentID       string `json:"parent_id,omitempty"`
```

model/rag/chat.go
nono: `chunkURL` to be exact! Cf https://go.dev/wiki/CodeReviewComments#initialisms

paultranvan: Thank you, guardian of guidelines!

PR #4611, opened 2025-11-05
fix: Getting account refresh token from remote OIDC

shepilov:
Add more logs especially for better errors handling

Make AccountType.RefreshAccount respect the token_mode defined on each account type

For token_mode = "basic" send the refresh request as a POST with the Basic Authorization header, mirroring RequestAccessToken

For token_mode = "get" issue a query-parameter GET; default path stays a form POST
Keeps the response handling unchanged so providers that already worked continue to do so

### DIFF
--- a/model/account/type.go
+++ b/model/account/type.go
@@ -15,6 +15,7 @@ import (
 	"github.com/cozy/cozy-stack/pkg/consts"
 	"github.com/cozy/cozy-stack/pkg/couchdb"
 	"github.com/cozy/cozy-stack/pkg/couchdb/mango"
+	"github.com/cozy/cozy-stack/pkg/logger"
 	"github.com/cozy/cozy-stack/pkg/prefixer"
 	"github.com/labstack/echo/v4"
 )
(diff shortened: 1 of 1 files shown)


model/account/type.go
```diff
@@ -338,57 +339,116 @@ func (at *AccountType) RequestAccessToken(i *instance.Instance, accessCode, stat
...
+	log.Debugf("Account %s (type: %s): refreshing OAuth token using %s auth mode", a.ID(), at.Slug, authMode)
+
+	params := url.Values{
 		"grant_type":    []string{RefreshToken},
 		"refresh_token": []string{a.Oauth.RefreshToken},
-		"client_id":     []string{at.ClientID},
-		"client_secret": []string{at.ClientSecret},
-	})
+	}
+	if at.TokenAuthMode != BasicTokenAuthMode {
```
taratatach:
```suggestion
	if authMode != BasicTokenAuthMode {
```

model/account/type.go
```diff
@@ -338,57 +339,116 @@ func (at *AccountType) RequestAccessToken(i *instance.Instance, accessCode, stat
...
+		params.Add("client_id", at.ClientID)
+		params.Add("client_secret", at.ClientSecret)
+	}
+
+	var res *http.Response
+	var err error
+
+	switch at.TokenAuthMode {
```
taratatach:
```suggestion
	switch authMode {
```

PR #4612, opened 2025-11-06
fix: Update common settings on a user phone update

shepilov:


pkg/rabbitmq/rabbitmq_test.go
```diff
@@ -386,6 +415,9 @@ func TestPasswordHandler(t *testing.T) {
 		require.NoError(t, err)
 		phone, _ := settingsDoc.M["phone"].(string)
 		require.Equal(t, msg.Mobile, phone)
+
+		// Verify that common settings were called
+		require.True(t, commonSettingsCalled, "Common settings should have been called when updating phone")
```
Crash--, line 422: I would have expected to check the "body" send to the CS API

taratatach:
Could you give a little more context about this change please? I don't understand why it was done.

Crash--:
> Could you give a little more context about this change please? I don't understand why it was done.

Today the only application able to write to Common Settings is the stack. No one else write to CS. 

When you change your langage from the stack, the stack updates CS. 

Now, the stack listen phone changes, and update the phone number internally. But we forgot to update the CS when we updated the phone number internally.

pkg/rabbitmq/rabbitmq_test.go
shepilov: in general it should be done in common settings tests, but as far as we don't have them, I've added only phone verification for this flow

pkg/rabbitmq/handlers.go
```diff
@@ -260,8 +260,17 @@ func (h *UserPhoneUpdatedHandler) Handle(ctx context.Context, d amqp.Delivery) e
...
+	if err != nil {
+		return fmt.Errorf("user.phone.updated: get settings document: %w", err)
+	}
+
+	// Update phone in settings
+	settings.M["phone"] = msg.Mobile
+
 	if err := lifecycle.Patch(inst, &lifecycle.Options{
-		Phone:        msg.Mobile,
+		SettingsObj:  settings,
```
taratatach, line 273: Why pass a settings obj here rather than the phone like was done before?

shepilov: To propagate into common settings services, the patch is a mess function probably because of its history. We have some information in the instance, some inside the common settings DB, Cloudery.  It's a good candidate for refactoring during the next touch

PR #4614, opened 2025-11-14
Shared drive auto acceptance

shepilov:
Automatic acceptance of Drive sharings from trusted members, eliminating manual approval for trusted contacts.

  Features
  - Domain-based trust: Configure trusted domains (e.g., *.company.com) to auto-accept sharings within organizations
  - Contact-based trust: After manually accepting a sharing once, future sharings from that sender are auto-accepted

  Configuration
```yaml
  sharing:
    auto_accept_trusted: true
    auto_accept_trusted_contacts: true
    contexts:
      default:
        trusted_domains: ["linagora.com"]
```

### DIFF
--- a/model/contact/contacts.go
+++ b/model/contact/contacts.go
@@ -314,8 +314,8 @@ func CreateFromSharingMember(inst *instance.Instance, email, name, cozyURL strin
 	}
 	if name != "" {
 		doc.JSONDoc.M["fullname"] = name
-		doc.JSONDoc.M["displayName"] = displayName
 	}
+	doc.JSONDoc.M["displayName"] = displayName
 
 	if cozyURL != "" {
 		doc.JSONDoc.M["cozy"] = []map[string]interface{}{
--- a/model/sharing/oauth.go
+++ b/model/sharing/oauth.go
@@ -475,7 +475,7 @@ func (s *Sharing) SendAnswer(inst *instance.Instance, state string) error {
 
 	options := config.GetConfig().Sharing.OptionsForContext(inst.ContextName)
 	// Mark the sender's contact as trusted since we accepted their sharing
-	if len(s.Members) > 0 && s.Members[0].Email != "" && *options.AutoAcceptTrustedContacts {
+	if *options.AutoAcceptTrustedContacts && len(s.Members) > 0 && s.Members[0].Email != "" {
 		c, err := contact.FindByEmail(inst, s.Members[0].Email)
 		if err != nil {
 			// Contact doesn't exist, create it using the standardized method
--- a/model/sharing/trust.go
+++ b/model/sharing/trust.go
@@ -51,7 +51,7 @@ func IsTrustedMember(inst *instance.Instance, member *Member) bool {
 	if options.AutoAcceptTrustedContacts == nil || !*options.AutoAcceptTrustedContacts {
 		return false
 	}
-	if isContactTrusted(inst, member) {
+	if isTrustedContact(inst, member) {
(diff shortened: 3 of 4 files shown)


pkg/config/config/config.go
```diff
@@ -317,6 +319,49 @@ type Flagship struct {
...
+	// Apply global AutoAcceptTrusted if not set in context
+	if result.AutoAcceptTrusted == nil {
+		result.AutoAcceptTrusted = &c.AutoAcceptTrusted
+	}
+
+	// Ensure AutoAcceptTrusted is never nil (default to false)
+	if result.AutoAcceptTrusted == nil {
+		falseVal := false
+		result.AutoAcceptTrusted = &falseVal
+	}
```
taratatach: This block does not seem necessary since `&c.AutoAcceptTrusted` cannot be `nil`.

pkg/config/config/config.go
```diff
@@ -359,6 +361,14 @@ func (c SharingConfig) OptionsForContext(contextName string) SharingContext {
 		result.AutoAcceptTrusted = &falseVal
 	}
 
+	if result.AutoAcceptTrustedContacts == nil {
+		result.AutoAcceptTrustedContacts = &c.AutoAcceptTrustedContacts
+	}
+	if result.AutoAcceptTrustedContacts == nil {
+		falseVal := false
+		result.AutoAcceptTrustedContacts = &falseVal
+	}
```
taratatach: This block seems unnecessary since `&c.AutoAcceptTrustedContacts` cannot be `nil`.

model/sharing/trust.go
```diff
@@ -0,0 +1,75 @@
...
+		inst.Logger().WithNamespace("sharing-trust").
+			Infof("Member %s trusted (trusted contact)", member.Instance)
+		return true
+	}
+
+	return false
+}
+
+// isContactTrusted checks if a member is marked as a trusted contact
+func isContactTrusted(inst *instance.Instance, member *Member) bool {
```
taratatach: nit: let's use the same phrasing as `IsTrustedMember`:
```suggestion
func isTrustedContact(inst *instance.Instance, member *Member) bool {
```

model/contact/contacts.go
```diff
@@ -294,6 +294,66 @@ func CreateMyself(inst *instance.Instance, settings *couchdb.JSONDoc) (*Contact,
...
+
+	displayName := name
+	if name == "" {
+		parts := strings.SplitN(email, "@", 2)
+		name = parts[0]
+		displayName = email
+	}
+	if name != "" {
+		doc.JSONDoc.M["fullname"] = name
+		doc.JSONDoc.M["displayName"] = displayName
```
taratatach: Shouldn't we set the contact's display name even if `name` is empty? We always have a display name since we fallback to the email.

model/sharing/oauth.go
```diff
@@ -471,6 +472,43 @@ func (s *Sharing) SendAnswer(inst *instance.Instance, state string) error {
 	s.Credentials[0].Client = creds.Client
 	s.Active = true
 	s.Initial = s.NbFiles > 0
+
+	options := config.GetConfig().Sharing.OptionsForContext(inst.ContextName)
+	// Mark the sender's contact as trusted since we accepted their sharing
+	if len(s.Members) > 0 && s.Members[0].Email != "" && *options.AutoAcceptTrustedContacts {
```
taratatach: nit: putting `*options.AutoAcceptTrustedContacts` first makes it easier to see that this block will only be used in this situation.

PR #4616, opened 2025-11-20
Add openRAG routes

paultranvan:
This will be useful for apps to use LLM features

### DIFF
--- a/web/ai/ai.go
+++ b/web/ai/ai.go
@@ -1,6 +1,7 @@
 package ai
 
 import (
+	"io"
 	"net/http"
 
 	"github.com/cozy/cozy-stack/model/permission"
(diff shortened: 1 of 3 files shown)


web/ai/ai.go
```diff
@@ -29,7 +30,43 @@ func Chat(c echo.Context) error {
...
+	// TODO: handle streaming response
+	res, err := rag.CallRAGQuery(inst, body, path, contentType)
+	if err != nil {
+		return nil, err
+	}
+	return res, nil
+}
+
+func OpenAICompletion(c echo.Context) error {
+	res, err := callAI(c, "v1/chat/completions")
```
shepilov, line 56: Should we also add a permission check for these functions?

web/ai/ai.go
```diff
@@ -29,7 +30,43 @@ func Chat(c echo.Context) error {
 	return jsonapi.Data(c, http.StatusAccepted, chat, nil)
 }
 
+func callAI(c echo.Context, path string) (*http.Response, error) {
+	inst := middlewares.GetInstance(c)
```
shepilov, line 40: Just a proxy without any query validation?

paultranvan: Yes, for now, I just want to have this route available quickly, we can improve it later. It's also possible that we eventually mutualize some AI-service backend to avoid having to make the same implem in cozy-stack, tmail, etc

web/ai/ai.go
paultranvan: That's a good point, I wasn't sure it is relevant because we do not write/read any doctype here. However, if we don't check any permission, does that means there is no authentication check whatsoever on this route?
If yes, we definitely need permission check :smile:

web/ai/ai.go
```diff
@@ -29,7 +30,43 @@ func Chat(c echo.Context) error {
...
+
+	body, err := io.ReadAll(c.Request().Body)
+	if err != nil {
+		return nil, echo.NewHTTPError(http.StatusBadRequest, "Unable to read request body")
+	}
+	contentType := c.Request().Header.Get("Content-Type")
+	// TODO: handle streaming response
+	res, err := rag.CallRAGQuery(inst, body, path, contentType)
```
shepilov, line 48: Also, no validation on the path, we allow access to any data?

web/ai/ai.go
shepilov: Nope, we don't have any other transient middleware checks

paultranvan: Ok, I added a permission on the `io.cozy.ai.chat.conversations` then https://github.com/cozy/cozy-doctypes/blob/master/docs/io.cozy.ai.chat.conversations.md

web/ai/ai.go
paultranvan: Fair enough, I added a check on the 2 paths

PR #4621, opened 2025-12-03
feat: Notification about file creation inside sharing

shepilov:
Adds email notifications when files or folders are uploaded to a shared folder

image

### DIFF
--- a/model/sharing/shared.go
+++ b/model/sharing/shared.go
@@ -426,8 +426,8 @@ func UpdateShared(inst *instance.Instance, msg TrackMessage, evt TrackEvent) err
 		}
 	}
 
-	// Send notification to owner if a recipient modified a file
-	if !removed {
+	// Send notification to owner if a recipient created a file
+	if evt.Verb == "CREATED" {
 		MaybeNotifyFileCreated(inst, msg, evt)
 	}
 


model/sharing/notification.go
```diff
@@ -0,0 +1,132 @@
...
+			"FileURL":            itemURL.String(),
+			"IsFolder":           isFolder,
+		},
+		PreferredChannels: []string{"mail"},
+	}
+	return center.PushStack(inst.DomainName(), center.NotificationSharingFileChanged, n)
+}
+
+// MaybeNotifyFileChange checks if a file change notification should be sent.
+// Only sends notifications when a file or folder is CREATED, not updated or removed.
```
taratatach, line 37: nit: The name of the function leads to thinking it's the opposite.

model/sharing/notification.go
```diff
@@ -0,0 +1,132 @@
...
+	fileName, _ := evt.Doc.Get("name").(string)
+	if fileName == "" {
+		fileName = "unknown"
+	}
+
+	// Get file/folder ID from the event
+	fileID := evt.Doc.ID()
+	if fileID == "" {
+		return
+	}
```
taratatach, line 87: Maybe we should log here? This looks like an unexpected behavior.

model/sharing/notification.go
```diff
@@ -0,0 +1,132 @@
...
+
+	// Get parent folder ID from the event
+	dirID, _ := evt.Doc.Get("dir_id").(string)
+	if dirID == "" {
+		return
+	}
```
taratatach, line 95: Same thing here.

web/files/files.go
```diff
@@ -155,10 +155,24 @@ func createFileHandler(c echo.Context, fs vfs.VFS, sharedDrive *sharing.Sharing)
 	if err != nil {
 		return nil, wrapVfsError(err)
 	}
+	maybeNotifyShareByLinkUpload(c, inst, doc.DocName, doc.ID(), doc.DirID, false)
```
taratatach, line 158: We should call `sharing.MaybeNotifyFileChange()` if the file was created in a shared drive as this is not a link sharing.

taratatach: Also, we should not call the function if the file was not created in a shared drive as this would mean calling it every time a file is created on an instance.

web/files/files.go
```diff
@@ -172,6 +186,7 @@ func createDirHandler(c echo.Context, fs vfs.VFS, sharedDrive *sharing.Sharing)
 		if err != nil {
 			return nil, err
 		}
+		maybeNotifyShareByLinkUpload(c, inst, doc.DocName, doc.ID(), doc.DirID, true)
```
taratatach, line 189: Same thing here.

web/files/files.go
```diff
@@ -232,6 +246,7 @@ func createDirHandler(c echo.Context, fs vfs.VFS, sharedDrive *sharing.Sharing)
 		return nil, err
 	}
 
+	maybeNotifyShareByLinkUpload(c, inst, doc.DocName, doc.ID(), doc.DirID, true)
```
taratatach, line 249: Same thing here.

model/sharing/shared.go
```diff
@@ -426,6 +426,11 @@ func UpdateShared(inst *instance.Instance, msg TrackMessage, evt TrackEvent) err
 		}
 	}
 
+	// Send notification to owner if a recipient modified a file
+	if !removed {
+		MaybeNotifyFileChange(inst, msg, evt)
+	}
```
taratatach: We should probably not call the function on updates either since we know we won't send a notification in this case.

web/files/files.go
shepilov: It's checked inside the function that it's a link sharing

web/files/files.go
shepilov: same thing here

web/files/files.go
shepilov: and here as well

model/sharing/shared.go
shepilov: It's checked inside the function that it is a "CREATED" event

web/files/files.go
shepilov: and why we should call MaybeNotifyFileCreated? it's works with another data that we have from event

model/sharing/shared.go
taratatach: Then why the conditional block, calling `MaybeNotifyFileCreated()` only if it's not a `DELETED` event?

web/files/files.go
taratatach: Ah right, the two functions have similar names but very different signatures. And I wasn't expecting link sharings to use these routes but it makes sense.
Nevermind.

model/sharing/shared.go
shepilov: It's just cheap early-exit optimization. If it's confusing, we can remove it

taratatach: I'm just wondering why we're doing an early exit for `DELETED` events and not `UPDATED` ones :)

shepilov: For update events, I didn't see the quick check

taratatach: Actually, `removed` is set to true when the document is no longer shared, not when it's deleted.

We can check if it's a creation by looking at the event's verb:

```suggestion
	// Send notification to owner if a recipient created a file
	if evt.Verb == "CREATED" {
		MaybeNotifyFileCreated(inst, msg, evt)
	}
```

taratatach (approved):

PR #4653, opened 2026-01-28
docs: ADR for uploading large files

shepilov:


docs/adr/002-large-file-upload-chunking.md
```diff
@@ -0,0 +1,294 @@
...
+
+## Context
+
+Twake Workplace (Cozy-Stack) allows users to upload files via a single HTTP request that are stored in OpenStack Swift or local filesystem (Afero). 
+Currently, there are practical limitations:
+
+1. Swift has a ~5GB single object limit
+2. Very large file uploads can stress server resources
+3. Long uploads are more prone to network failures
+4. We plan to add S3 API support, which has similar chunking needs (multipart uploads)
```
taratatach, line 15: You mean as a storage backend or provide this API to our users?

docs/adr/002-large-file-upload-chunking.md
```diff
@@ -0,0 +1,294 @@
...
+- No CouchDB schema changes needed
+
+**For Local VFS (Afero):** Already handles large files natively (no chunking needed)
+
+**For Future S3:** Use S3 Multipart Upload API
+- S3/MinIO provide native multipart upload support similar to Swift SLO (streaming parts directly to disk without buffering entire files)
+- Same transparent approach can be used
+- Observe S3 protocol limits: a single object maxes out at 5TB, uploads can have at most 10,000 parts, and each part must be between 5MB and 5GB (last part can be smaller)
+- Pick part sizes small enough (and configurable) so the 10,000-part limit still covers the largest supported file; anything larger than 5TB must be chunked at the application level because the S3 API itself forbids it
+- **Note on checksums:** Like Swift SLO, S3 multipart ETags are not MD5 hashes of the content (they're a hash of part ETags). Application-side MD5 computation will be required for S3 multipart uploads, using the same pattern as Swift SLO
```
taratatach, line 51: Isn't there a small mistake here?
```suggestion
- **Note on checksums:** Like Swift SLO, S3 multipart ETags are not MD5 hashes of the content (they're a hash of part content). Application-side MD5 computation will be required for S3 multipart uploads, using the same pattern as Swift SLO
```
If not, I don't really understand.

docs/adr/002-large-file-upload-chunking.md
shepilov: mmm, yes, allow users to upload big files as well

Crash--: Today at least backend storage @taratatach. For some customers, we may want to use S3 backend instead of swift (because of secnumcloud certification or things like that) 

Having an S3 compatible API can be cool, but I don't know the cost of having the 2 API.

docs/adr/002-large-file-upload-chunking.md
```diff
@@ -0,0 +1,294 @@
...
+
+**Test override**: Tests will be able to set a tiny segment size (e.g., 1KB) via config overrides to exercise the SLO code path without uploading gigabytes.
+
+#### 2. Swift VFS Changes (`model/vfs/vfsswift/impl_v3.go`)
+
+**CreateFile method modifications:**
+
+The `CreateFile` method will be extended to:
+1. Read segment size and SLO threshold from configuration
+2. Determine whether to use SLO based on file size (files larger than threshold) or streaming mode (unknown size, indicated by negative `ByteSize`)
```
taratatach, line 78: Do we have to know the file size before hand to use SLO?

taratatach: Reading the "quota enforcement" paragraph it seems that we don't but then I don't really understand this sentence.

docs/adr/002-large-file-upload-chunking.md
```diff
@@ -0,0 +1,294 @@
...
+When `ByteSize < 0` (streaming/chunked transfer encoding), the total size is unknown upfront. Quota enforcement will work as follows:
+1. Before upload: check that instance has *some* available quota (reject if quota already exceeded)
+2. During upload: the `swiftLargeFileCreationV3.Write()` method will track cumulative bytes written
+3. On each write: compare cumulative bytes against instance quota; if exceeded, call `Abort()` to clean up segments and return a quota-exceeded error
+4. The existing `vfs.CheckAvailableDiskSpace` check runs at file creation time; for streaming uploads, we add a runtime check that aborts mid-stream if the limit is hit
+5. This mirrors the existing behavior for regular uploads where Swift/the VFS layer rejects writes that exceed quota
+
+#### 3. MD5/Checksum Handling
+
+**Important**: SLO manifests don't return a single MD5 hash like regular objects. The manifest's ETag is a hash of the segment ETags, not the content.
```
taratatach, line 94: Again you're saying here that the manifest's ETag is a hash of ETags. If that's not a mistake (i.e. we should not read "hash of segment") then what's the use of that manifest ETag?

docs/adr/002-large-file-upload-chunking.md
```diff
@@ -0,0 +1,294 @@
...
+
+We will implement application-side MD5 computation:
+
+1. Create a new `swiftLargeFileCreationV3` struct that holds an MD5 hasher alongside the Swift file writer
+2. On each `Write()` call, update the MD5 hash before passing data to Swift
```
taratatach, line 99: I'm curious about how we can update the hash without the full file content.

docs/adr/002-large-file-upload-chunking.md
```diff
@@ -0,0 +1,294 @@
...
+This ensures:
+- Antivirus scanning works (relies on MD5Sum)
+- File versioning works (compares MD5Sum to detect changes)
+- File integrity validation works
+
+#### 4. Failure Handling and Cleanup
+
+**Failure Scenarios:**
+
+1. **Client disconnects mid-upload**: The `swiftLargeFileCreationV3.Close()` is never called; segments remain orphaned
```
taratatach, line 112: Can't we call `Close()` in a `defer` function so it's called every time the handler finishes whether it's because the upload is done or the connection was closed?

docs/adr/002-large-file-upload-chunking.md
```diff
@@ -0,0 +1,294 @@
...
+    # ... existing config ...
+    # Age threshold for orphan cleanup (default: 24h)
+    orphan_segment_max_age: 24h
+```
+
+**Triggering cleanup:**
+- On upload error: immediate best-effort delete
+- On server startup: schedule GC job
+- Periodically: run GC worker (configurable interval)
+- Manual: `cozy-stack swift gc-segments` CLI command
```
taratatach, line 145: We have a `cozy-stack check fs` command that runs some consistency checks already. We might want to add the segments check to it.

docs/adr/002-large-file-upload-chunking.md
```diff
@@ -0,0 +1,294 @@
...
+| `cleanOldVersion` | `ObjectDelete` | Use `LargeObjectDelete` with fallback |
+| `EnsureErased` | `BulkDelete` / `ObjectDelete` | Use `LargeObjectDelete` for each |
+| `CopyFile` | `ObjectCopy` | Copy manifest + segments or re-upload |
+| `DissociateFile` | `ObjectCopy` + `ObjectDelete` | Handle SLO copy and delete |
+| `CopyFileFromOtherFS` | `ObjectCopy` | Handle SLO source objects |
+
+**Deletion pattern:**
+
+We will implement a `deleteObject` helper method that:
+1. First attempts `LargeObjectDelete` (which handles both SLO manifests with their segments and regular objects)
```
taratatach, line 163: Do you think this could add a lot of overhead when deleting many files? Would it be better to use SLO for all files even ones which are smaller that the threshold?

docs/adr/002-large-file-upload-chunking.md
```diff
@@ -0,0 +1,294 @@
...
+2. If Swift returns `NotLargeObject` error, falls back to regular `ObjectDelete`
+3. This unified approach ensures both SLO and regular objects are deleted correctly without needing to check the object type first
+
+**Copy pattern** (for SLO objects):
+
+Swift doesn't support copying SLO manifests directly. Available options:
+1. **Copy manifest content and update segment references** - Complex: requires parsing manifest JSON, copying each segment individually, updating references
+2. **Download and re-upload** - Simple but slow: streams entire file through the server
+3. **Copy segments individually then create new manifest** - Medium complexity: server-side segment copy + new manifest creation
```
taratatach, line 172: Does Swift provide helpers to create manifests from a list of segments?

docs/adr/002-large-file-upload-chunking.md
```diff
@@ -0,0 +1,326 @@
...
+**Pros:**
+- Client can resume uploads after failure
+- Better progress tracking per chunk
+- Works with any storage backend uniformly
+
+**Cons:**
+- Requires new HTTP API endpoints (`POST /files/chunks/start`, `PUT /files/chunks/{id}`, `POST /files/chunks/{id}/complete`)
+- Significant UI/client changes required
+- Server must track upload sessions and handle cleanup of incomplete uploads
+- Adds complexity to CouchDB (need to track chunk metadata)
```
taratatach, line 213: Couldn't we create a SLO manifest and create chunks by hand in this case? I'm just asking because I think adding this functionality later could be nice (for resuming uploads, parallel uploads…).

docs/adr/002-large-file-upload-chunking.md
shepilov: it's a  `MD5( segment1_etag + segment2_etag + ... )`

docs/adr/002-large-file-upload-chunking.md
shepilov: You don't need the full file content to calculate a hash with MD5; you do it in chunks

docs/adr/002-large-file-upload-chunking.md
shepilov: Yes, and we have to do it

docs/adr/002-large-file-upload-chunking.md
shepilov: yes, it can be part of `check fs`, but cleanup is destructive, and should be explicit

docs/adr/002-large-file-upload-chunking.md
shepilov: We can use also ObjectDelete for all object, for SLO it deletes only manifest and content will be deleted after with GC. But I'm a little but afraid of using GC for every file and prefer keep for extrodinay cases and even half manual.
There is also variants to store flag in file doc

docs/adr/002-large-file-upload-chunking.md
shepilov: Yep, there is and API to create manifest

docs/adr/002-large-file-upload-chunking.md
shepilov: Yes, when we are ready to do it, we can create segments separately and then create a manifest with uploaded chunks

docs/adr/002-large-file-upload-chunking.md
shepilov: At least we know it from the request header, but it's not 100%, we can rely on it at first to what method to use but not for quota, we need to calculate it on the fly as well

docs/adr/002-large-file-upload-chunking.md
taratatach: OK, so definitely a hash of ETags. Is the purpose of that hash to quickly know whether a segment was changed? I don't see much use for it besides this.

docs/adr/002-large-file-upload-chunking.md
taratatach: Hum, you're saying in the document that `Close()` can be never called though :thinking:

docs/adr/002-large-file-upload-chunking.md
taratatach: OK, I get your point.

shepilov:
Hopefully won't be implemented as we are migraing ot S3 step by step https://github.com/linagora/cozy-stack/pull/4716

PR #4654, opened 2026-01-30
docs: ADR for publishing a RabbitMQ settings update message

shepilov:


rezk2ll (approved):

docs/adr/003-rabbitmq-settings-publishing.md
```diff
@@ -0,0 +1,116 @@
...
+This integration is implemented in `model/settings/common/common.go` and performs:
+
+1. **HTTP POST** to `/api/admin/user/settings` when creating settings
+2. **HTTP GET** to `/api/admin/user/settings/{nickname}` to check remote version
+3. **HTTP PUT** to `/api/admin/user/settings/{nickname}` to update settings
+
+This approach has several issues:
+
+- **Synchronous blocking**: HTTP calls block the settings update operation
+- **Version conflict complexity**: The code checks remote versions before updates, creating complex conflict resolution logic
```
Crash--, line 19: This was done by purpose. Because we need to have a source of truth. 

I'm not sure we're dealing with that correctly from the common-settings app side right now. 

If we have 2 app sending a message to update CS, we'll have an issue on CS side. And this cozy-stack today is the entry point of CS, the check was done there. 

So if we remove this check, we need to be sure that CS can handle this stuff. cc @rezk2ll

docs/adr/003-rabbitmq-settings-publishing.md
```diff
@@ -0,0 +1,116 @@
...
+- **Tight coupling**: Direct HTTP dependency on the common settings app
+- **Failure propagation**: HTTP failures can affect the main settings update flow
+- **No longer needed**: The common settings app is no longer actively updating settings, making version conflict checking obsolete
```
Crash--: I don't understand this point, can you elaborate?

docs/adr/003-rabbitmq-settings-publishing.md
rezk2ll: Yes, the version control is done in CS for incoming requests.

but IMO, the CS app can retire and the stack will handle it. so in reality, we will have only one app sending the message updates.

docs/adr/003-rabbitmq-settings-publishing.md
shepilov: The common settings app is being deprecated, so the stack no longer needs to support conflict resolution

Crash--: What are you calling "common settings app"?

docs/adr/003-rabbitmq-settings-publishing.md
Crash--: What is CS app?

docs/adr/003-rabbitmq-settings-publishing.md
shepilov: linagora/twake-workplace-common-settings

docs/adr/003-rabbitmq-settings-publishing.md
shepilov: linagora/twake-workplace-common-settings

docs/adr/003-rabbitmq-settings-publishing.md
Crash--: Oh I didn't even know that there was a front app for this API. This front app is not used anywhere. 

The version's check on stack was not done because of that, it was done to be sure that we're sending the up to date data. And be sure that the user updated the latest version.

rezk2ll: Technically, it's a backend app

Crash--: So I still don't understand. We will alwaus need this backend app right?

rezk2ll: if the stack handles the message sending part, we no longer need this app.

The CS app was created because that stack could not send RabbitMQ messages and also connect to different RabbitMQ servers on a different infra.

Both of those issues are solved .

The other argument to keep the CS app was the on-prem installations. But I believe now it's mandatory to ship the stack on-prem if you want common settings. so really, no need to keep the app

Crash--: But we still need CS app to consume the message, no?

rezk2ll: no, the other apps will consume the message directly no need for CS to relay the message.

CS app was exposing an API to the stack -> the stack sends the settings -> the app stores the settings and checks the version etc -> then sends a rabbitmq message to everyone.

The stack is already a copy of the source of the truth and is now capable of sending the RabbitMQ messages. so we just need to add the version control check and replace the API call with a RabbitMQ message.

Crash--: But today we have a DB for CS, no? I liked the idea of having an external DB for that. Don't we have a "script" reading this DB to send event in order to fully align all the listeners (in case of something not working? Like what we had few weeks ago?) If we don't have this external DB anymore, and we only store in CouchDB, it'll be a mess to do. 

If we want to do that, then the cozy-stack still need to checks the version and else, so I don't understand how we can simplify or remove this check?

PR #4655, opened 2026-01-30
ADR for OIDC improvement to use TTL from oidc token

shepilov:


docs/adr/004-oidc-token_ttl_improvements.md
```diff
@@ -0,0 +1,177 @@
...
+    SID                string     `json:"sid,omitempty"`                   // Existing
+    OIDCTokenExpiresAt *time.Time `json:"oidc_token_expires_at,omitempty"` // New
+}
+```
+
+###  Session TTL Alignment
+
+#### Session Validity Middleware
+
+Add HTTP middleware that checks OIDC token expiry on each request:
```
Crash--, line 139: based on the `expires_in` attribute right? Not by doing an http call right?

docs/adr/004-oidc-token_ttl_improvements.md
```diff
@@ -0,0 +1,177 @@
...
+```mermaid
+sequenceDiagram
+    participant User
+    participant Cozy as Cozy Home
+    participant Twake Mail as Twake Mail (iframe)
+    participant Stack as Cozy Stack
+    participant OIDC 
+
+    User->>Cozy: Login via OIDC
+    Cozy->>OIDC: Authenticate
```
Crash--, line 33: Are you sure that the current flow is this one? 

Because Cozy Home will never communicate with the OIDC. It should be cozy-stack

docs/adr/004-oidc-token_ttl_improvements.md
```diff
@@ -0,0 +1,177 @@
...
+    OIDC-->>Cozy: Tokens (TTL: 1 hour)
+    Cozy->>Stack: Create session
```
Crash--, line 35: Create session is not done by Cozy Home. 

Cozy Stack will generate a token and pass this token to cozy home

docs/adr/004-oidc-token_ttl_improvements.md
```diff
@@ -0,0 +1,177 @@
...
+    Stack-->>Cozy: Session (TTL: 30 days)
+
+    Note over User,OIDC: Time passes... OIDC token expires
+
+    User->>Twake Mail: Open Twake Mail in iframe
```
Crash--, line 40: should be User->>Twake Mail Iframe ->> Twake Mail

docs/adr/004-oidc-token_ttl_improvements.md
```diff
@@ -0,0 +1,177 @@
...
+    OIDCTokenExpiresAt *time.Time `json:"oidc_token_expires_at,omitempty"` // New
+}
+```
+
+###  Session TTL Alignment
+
+#### Session Validity Middleware
+
+Add HTTP middleware that checks OIDC token expiry on each request: 
+if the token is expired, return a 401 Unauthorized response with an `oidc_token_expired` error indicating re-authentication is required.
```
Crash--, line 140: This http middleware will only work for "cozy apps" not for "iframed ones". 

But let's focus on theses cozy-apps first

I don't think cozy-stack should return a 401 Unauthorized. 

Because a cozy application will never know what to do with `oidc_token_expired`. 

Why the stack do not refresh the `oidc_token` based on the `oidc_refresh_token`. Like that, there is nothing to do for the client app in 95% of the time. 

We'll still have issue if the `oidc_token` is expired and the cozy stack is unable to refresh the token (because refresh token is not valid anymore). 

Then in this case, maybe cozy-stack should redirect to login page directly? 


Now let's talk about "iframed" app. 

I'm on an iframed app for a long time without any activities. I do not refresh my browser. So I've my cozy bar authenticated because I didn't do any http request throught the stack. I start navigating in my iframed app, and this iframed app is not able to refresh its own token, then this iframed app will redirect me to the sso. And we'll be in the use case we want to fix: not displaying the sso page in the iframe. 

Maybe we can do some magic from cozy-bridge (cc @zatteo) to detect the redirection to the SSO and do stuff at the cozy-app level (redirect to the SSO on top window). 

But maybe also, the cozy front end application can be aware of the expired time? Based on its own token with an expired_at attribute? In this case, cozy-bar can be able to make an http request and then run into this http middleware and so on?

docs/adr/004-oidc-token_ttl_improvements.md
Crash--: or based on introspect?

shepilov: yep, base on expires_in

docs/adr/004-oidc-token_ttl_improvements.md
shepilov: yes, it's a super-simplified one just to illustrate the problem.  It doesn't interracts with OIDC except redirect to sign-up to do the auth

docs/adr/004-oidc-token_ttl_improvements.md
shepilov: yes, I've mixed app with mobile app flow

docs/adr/004-oidc-token_ttl_improvements.md
shepilov: if Twake Mail Iframe is a cozy app, yes

docs/adr/004-oidc-token_ttl_improvements.md
shepilov: Because we don't have a problem when the token expires in the stack, we have problem when the token expires in twake male, and twake male doesn't refresh the session.

Yes, cozy stack I think even should do this, but it's just more changes to OIDC flow. But we don't know at what moment we should redirect to OIDC flow. we can't send 322 loging redirect to every post instead of 401. 

But the bigget problem is that you've mentioned, and it's not solved with this ADR

zatteo: > Maybe we can do some magic from cozy-bridge (cc @zatteo) to detect the redirection to the SSO and do stuff at the cozy-app level (redirect to the SSO on top window).

We won't be able to monitor from parent window URL change from twake mail if it redirects to SSO because we are not same origin so no access to url updates from iframe. But otherwise yes we can do stuff.

Crash--: > Because we don't have a problem when the token expires in the stack, we have problem when the token expires in twake male, and twake male doesn't refresh the session.

That's not totally true. TMail refresh the session, but sometimes your refresh token is expired. So you just can't refresh.

I still think this iframe stuff is not the right thing to do. We should let the app (stack / tmail / tchat) identify itself, and then inject the cozy-bar. Like that, no more issue like this one. 

I'll have a talk with @zatteo & @poupotte

shepilov: @Crash-- it's not 100% solution, since Twake Mail app has it's own auth oidc flow, but we can also can listen an event form stack in the bridge app when session is expired. We should be carefull at this step, to have the same oidc refresh flow and timeouts for both app(all settings shoudl be the same or mo restrictive on the stack side)

Crash--: Yes, we can do something: 
- cozy-bridge can expose a method which can be called by the embedded when refreshing 
- and then call a logout from the stack & start a full oidc flow 


I still think that, regarding this embedded/embedder approach, we might be going in the wrong direction.
We're considering adding a lot of methods and tools to work around something the application handle natively.
We should continue exploring alternatives without iframes.

Crash--: But let's fix at least the cozy's app issue. This ADR should fix that. Can you update it to target this objective please?

shepilov: What do you mean by "cozy's app issue"?

Crash--: - check token validity during the use of a cozy application. 
- check token validity when we open a new cozy app 

It'll fix these scenarios : 
- I've my cozy home page opened since a lot of time, and then I click on the mail application (iframed). => Cozy stack should detect that my current session is ended and then redirect me to the SSO. I will log again, and then, I'll come back to my mail app (iframed) and I'll be connected. 
- I'm writing things on my notes app, or sometimes on my only office document, then, if my session is expired, cozy stack should refresh it (without being disconnected & without losing my content). Then after, I can open my iframed app I'll be connected.

shepilov:
not relevant with https://github.com/linagora/cozy-stack/pull/4884

PR #4716, opened 2026-03-30
feat: add S3-compatible storage backend

Crash--:
## Summary

- Add S3-compatible object storage as a third VFS backend (alongside afero and Swift)
- Uses `minio-go/v7` — no AWS SDK dependency. Targets OVH S3, MinIO, Scaleway, and any S3-compatible provider
- Bucket strategy: one shared bucket per organization (`<prefix>-<orgId>`), key prefix per instance (`<DBPrefix>/`)
- Memory-efficient: single PUT when file size is known (streams like Swift ~32KB), multipart only for unknown size
- All 17 VFS integration tests pass for the S3 backend

### What's included

| Area | Status |
|------|--------|
| Main VFS (upload, download, delete, versions, fsck) | Done |
| Avatar storage | Done |
| Thumbnail storage | Done |
| App installation (Copier + FileServer) | Done |
| Dynamic assets | Done |
| Preview/icon cache | Done |
| Export archiver | Done |
| Documentation (`docs/s3.md`) | Done |
| Integration tests (testcontainers + MinIO) | Done |
| Security review fixes (MD5 verification, path traversal, error sanitization) | Done |

### Configuration

```yaml
fs:
  url: s3://s3.rbx.io.cloud.ovh.net?access_key=ACCESS&secret_key=SECRET&region=rbx&bucket_prefix=cozy&use_ssl=true
```

All buckets (apps, assets, previews, exports) are created automatically at startup.

### Bucket layout

| Bucket | Content |
|--------|---------|
| `<prefix>-<orgId>` | Main VFS data (files, versions, thumbs, avatars) |
| `<prefix>-apps-web` | Web application assets |
| `<prefix>-apps-konnectors` | Konnector assets |
| `<prefix>-assets` | Dynamic assets |
| `<prefix>-previews` | PDF preview and icon cache |
| `<prefix>-exports` | Instance export archives |

See `docs/s3.md` for full documentation including a local MinIO setup tutorial.

## Test plan

- [x] Unit tests: 15/15 naming tests pass
- [x] Integration tests: 17/17 VFS tests pass (afero, swift, s3)
- [x] Manual API test: upload, download, trash, permanent delete via `/files` API
- [x] Manual browser test: Drive UI works, file upload + thumbnail generation confirmed
- [x] Security review: all critical/high/medium findings fixed
- [ ] Test with OVH S3 in staging

🤖 Generated with [Claude Code](https://claude.com/claude-code)


model/vfs/vfss3/s3.go
```diff
@@ -0,0 +1,53 @@
+package vfss3
+
+import (
+	"context"
+
+	multierror "github.com/hashicorp/go-multierror"
```
shepilov: it's not necessary to name the package with the same name

model/vfs/vfss3/s3.go
```diff
@@ -0,0 +1,53 @@
+package vfss3
```
shepilov: and a little bit confusing to havein s3.go inly provate delete functions

model/vfs/vfss3/impl.go
```diff
@@ -0,0 +1,1155 @@
...
+
+type s3VFS struct {
+	vfs.Indexer
+	vfs.DiskThresholder
+	client      *minio.Client
+	cluster     int
+	domain      string
+	prefix      string // DBPrefix — used as key prefix in the bucket
+	contextName string
+	ctx         context.Context
```
shepilov, line 38: I see that we have context in all other vfs, but it's an execution context, not data. Let's do not spread anti-pattern and if not change interface, but at least just pass the context.Background as is

model/vfs/vfss3/impl.go
```diff
@@ -0,0 +1,1155 @@
...
+	bucket      string
+	keyPrefix   string // prefix + "/"
+	region      string
+	mu          lock.ErrorRWLocker
+	log         *logger.Entry
+}
+
+const maxFileSize = 5 << (3 * 10) // 5 GiB
```
shepilov: You decided not to implement a multipart upload of all files and keep this limit?

Crash--: To keep the same constraint as today we have on Swift. 

I prefer to merge S3 as is, and then add this new "feature" on S3 & Swift.

model/vfs/vfss3/s3.go
shepilov: or, maybe I understood, there is duplication with appfs, the same functions. 
Does it make sense then to move them to pkg/s3 or pkg/s3util package with 
```
  func EnsureBucket(ctx, client, bucket, region) error
  func DeleteObjects(ctx, client, bucket, names) error
  func WrapNotFound(err) error
```

model/vfs/vfs_test.go
```diff
@@ -17,6 +17,7 @@ import (
```
shepilov, line 17: Here, we don't test any error cases during upload. But we run all the PuObject in the background goroutine, so it would be gread to add test when miniio is down during upload to check that all errors are propagated to client

model/vfs/vfss3/impl.go
shepilov: But at the same time, you are setting uploadSize = -1, which means any size

model/vfs/vfss3/impl.go
Crash--: I understand the concern — storing a context in a struct is generally considered an anti-pattern in Go. However, the VFS interface (`model/vfs/vfs.go`) does not pass a `context.Context` in its method signatures, so all backends that need one store it as a field. The Swift implementation (`vfsswift/impl_v3.go`) uses the exact same pattern: `ctx: context.Background()` set at creation time. Changing this would require updating the VFS interface, which feels out of scope for this PR - and I'm not confortable with that - . Happy to discuss if you think it's worth a broader refactor though.

model/vfs/vfss3/impl.go
Crash--: You were right — the `maxFileSize` constant was not consistently enforced when `uploadSize = -1`. Since minio-go already handles multipart upload transparently (via the `PartSize` option), there is no real per-file size limit on S3 like there is on Swift. I've removed the constant entirely and `MaxFileSize()` now returns `0`. See commit 17859d7.

model/vfs/vfss3/s3.go
Crash--: Done — extracted the shared helpers into `pkg/s3util` with `DeleteObjects`, `DeletePrefixObjects`, `EnsureBucket`, `IsNotFound`, and `WrapNotFound`. Both `vfss3` and `appfs` now use this shared package. Also added integration tests for the new package. See commits 643cd8c and c14f0f1.

model/vfs/vfs_test.go
Crash--: https://github.com/cozy/cozy-stack/pull/4716/changes/10d471f44d5fbca223f53de516e53af1f1396875

Crash--:
@shepilov For your information we've tested this version of the stack and it seems to work. 

I'll redo my git history and also, maybe I'll try to be able to chose between s3 and swift not only on env level but also on instance. Like that, I'll be able to move some production instances to S3 in order to check & validate in real usage. WDYT?

shepilov:
image

Load consitency test results

shepilov (approved):

PR #4728, opened 2026-04-13
feat(nextcloud): trigger endpoint for bulk migrations

rezk2ll:
## Summary

Adds `POST /remote/nextcloud/migration` so the Settings UI can start a Nextcloud-to-Cozy bulk migration end-to-end. The endpoint probes the supplied credentials, upserts an `io.cozy.accounts` document with the resolved WebDAV user ID, creates an `io.cozy.nextcloud.migrations` tracking document in `pending` state, and publishes a `nextcloud.migration.requested` command to the `migration` RabbitMQ exchange. A separate service consumes the command and drives the existing `/remote/nextcloud/:account/*` routes, writing progress back to the tracking document so the UI can render a real-time progress bar.

Only one migration per instance can be in flight at a time; failed migrations do not block retries.

## API

`POST /remote/nextcloud/migration` (requires `POST io.cozy.nextcloud.migrations`)

```json
{
  "nextcloud_url": "https://nextcloud.example.com",
  "nextcloud_login": "alice",
  "nextcloud_app_password": "xxxxx-xxxxx-xxxxx-xxxxx-xxxxx",
  "source_path": "/",
  "target_dir": "/Nextcloud"
}
```

`source_path` and `target_dir` are optional. `source_path` defaults to `/`. `target_dir` defaults to `/Nextcloud` and must be a clean absolute path (validated with `path.Clean`, so `..`, double slashes, and relative segments are rejected). `nextcloud_app_password` should be a Nextcloud app password, not the account password.

### Error responses

| Code | When |
| ---- | ---- |
| 400 Bad Request | Required fields missing or `target_dir` is not a clean absolute path |
| 401 Unauthorized | The Nextcloud server rejected the supplied credentials |
| 409 Conflict | A `pending` or `running` migration already exists for this instance |
| 500 Internal Server Error | Account upsert or tracking document creation failed |
| 502 Bad Gateway | The Nextcloud instance is unreachable |
| 503 Service Unavailable | RabbitMQ publish failed; the tracking document is marked `failed` before returning |

## Companion route: recursive size

`GET /remote/nextcloud/:account/size/*path` returns the recursive byte total of a Nextcloud folder via a single PROPFIND on `oc:size`. The migration consumer uses it as a pre-flight quota check so it knows the true source size before starting; the previous shallow-sum estimate could be off by several orders of magnitude on deep trees.

## RabbitMQ contract

Exchange `migration`, routing key `nextcloud.migration.requested`. The consumer is responsible for declaring its queue and binding.

```json
{
  "migrationId": "d4e5f6a7b8c94d0ea1b2c3d4e5f6a7b8",
  "workplaceFqdn": "alice.cozy.example.com",
  "accountId": "a1b2c3d4e5f6",
  "sourcePath": "/",
  "timestamp": 1712563200
}
```

Credentials are never in the payload; they live in the `io.cozy.accounts` document referenced by `accountId`. `target_dir` is read by the consumer from the tracking document so legacy messages and new ones both flow through the same path. `MessageID` is set to the migration ID for cross-system tracing.

## Credentials probe

The OCS probe itself is not new. It already existed as a lazy fallback behind `(nc *NextCloud).fetchUserID()`, called from `fillUserID` only when an existing account document was missing its cached `webdav_user_id`. In practice that path was almost never exercised, so a latent bug sat there undetected: the probe targeted `apps/user_status`, and any non-200 (including a 404 from a managed Nextcloud host that strips the optional `user_status` app) was turned into `webdav.ErrInvalidAuth`.

What this PR adds is the synchronous, user-facing path: the migration trigger endpoint calls `FetchUserIDWithCredentials` before persisting anything, so the probe is now on the critical path of every trigger. That exposure turned the latent classification bug into a user-visible 401 for anyone pointing the feature at a managed Nextcloud like `thegood.cloud`.

A follow-up commit on this branch fixes it by switching the probe to OCS Core (`/ocs/v2.php/cloud/user`, which cannot be disabled by an admin) and narrowing the auth-failure classification to 401/403 only, so other non-2xx statuses bubble up with their real cause. Both code paths (the old lazy fallback and the new synchronous call) benefit from the fix.

Validated end-to-end against a real Nextcloud instance: the probe, the encrypted-at-rest account persistence, the tracking document, the `/size/*path` pre-flight, the RabbitMQ publish with no credentials in the payload, and the consumer-driven transfer all behave as documented.

### DIFF
--- a/model/nextcloud/nextcloud.go
+++ b/model/nextcloud/nextcloud.go
@@ -3,7 +3,9 @@
 package nextcloud
 
 import (
+	"context"
 	"encoding/json"
+	"fmt"
 	"io"
 	"net/http"
 	"net/url"
@@ -20,6 +22,7 @@ import (
 	"github.com/cozy/cozy-stack/pkg/consts"
 	"github.com/cozy/cozy-stack/pkg/couchdb"
 	"github.com/cozy/cozy-stack/pkg/jsonapi"
+	"github.com/cozy/cozy-stack/pkg/logger"
 	"github.com/cozy/cozy-stack/pkg/safehttp"
 	"github.com/cozy/cozy-stack/pkg/webdav"
 	"github.com/labstack/echo/v4"
@@ -63,10 +66,11 @@ func (f *File) Links() *jsonapi.LinksList {
 var _ jsonapi.Object = (*File)(nil)
 
 type NextCloud struct {
-	inst      *instance.Instance
-	accountID string
-	userID    string
-	webdav    *webdav.Client
+	inst        *instance.Instance
+	accountID   string
+	userID      string
+	installRoot string
+	webdav      *webdav.Client
 }
 
 func New(inst *instance.Instance, accountID string) (*NextCloud, error) {
@@ -97,19 +101,21 @@ func New(inst *instance.Instance, accountID string) (*NextCloud, error) {
 	}
 	username, _ := auth["login"].(string)
 	password, _ := auth["password"].(string)
+	installRoot := normalizeInstallRoot(u.Path)
 	logger := inst.Logger().WithNamespace("nextcloud")
 	webdav := &webdav.Client{
 		Scheme:   u.Scheme,
 		Host:     u.Host,
 		Username: username,
 		Password: password,
-		BasePath: "/remote.php/dav",
+		BasePath: installRoot + "/remote.php/dav",
 		Logger:   logger,
 	}
 	nc := &NextCloud{
-		inst:      inst,
-		accountID: accountID,
-		webdav:    webdav,
+		inst:        inst,
+		accountID:   accountID,
+		installRoot: installRoot,
+		webdav:      webdav,
 	}
 	if err := nc.fillUserID(&doc); err != nil {
 		return nil, err
@@ -119,6 +125,15 @@ func New(inst *instance.Instance, accountID string) (*NextCloud, error) {
 
 func (nc *NextCloud) Download(path string) (*webdav.Download, error) {
 	return nc.webdav.Get("/files/" + nc.userID + "/" + path)
+}
+
+// Size returns the recursive byte total of the resource at path, as
+// reported by Nextcloud's cached `oc:size` property. Works on the account
+// root (pass an empty string or "/") and on any sub-folder. Equivalent to
+// a single Depth:0 PROPFIND on the server, not a tree walk, so the cost
+// is constant regardless of how many files the folder contains.
+func (nc *NextCloud) Size(path string) (uint64, error) {
+	return nc.webdav.Size("/files/" + nc.userID + "/" + path)
 }
 
 func (nc *NextCloud) Upload(path, mime string, contentLength int64, body io.Reader) error {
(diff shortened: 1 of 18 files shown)


model/nextcloud/nextcloud.go
```diff
@@ -396,10 +428,14 @@ func (nc *NextCloud) fetchUserID() (string, error) {
...
 }
 
+func (nc *NextCloud) fetchUserID() (string, error) {
+	return fetchUserIDFromHost(nc.webdav.Scheme, nc.webdav.Host, nc.webdav.Username, nc.webdav.Password, nc.webdav.Logger)
+}
+
 type OCSPayload struct {
 	OCS struct {
 		Data struct {
-			UserID string `json:"userId"`
```
Crash--, line 402: sure about this change? old route & new route returns an ID === userID ?

web/remote/migration.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+	if err := body.normalize(); err != nil {
+		return jsonapi.BadRequest(err)
+	}
+	reqLogger := baseLogger.WithFields(logger.Fields{
+		"nextcloud_host":  utils.ExtractInstanceHost(body.NextcloudURL),
+		"nextcloud_login": body.NextcloudLogin,
+	})
+
+	active, err := nextcloud.FindActiveMigration(inst)
```
shepilov: there is no guarantee for two concurrent requests that two migration want be started.  You need to perform this query under a lock. what will be the consequences in case of UI bug if we have two parallel migrations?

model/nextcloud/nextcloud.go
```diff
@@ -360,16 +363,41 @@ func (nc *NextCloud) buildTrashedURL(item webdav.Item, path string) string {
...
+func FetchUserIDWithCredentials(nextcloudURL, username, password string, logger *logger.Entry) (string, error) {
+	u, err := url.Parse(nextcloudURL)
+	if err != nil {
+		return "", err
 	}
-	req, err := http.NewRequest(http.MethodGet, u.String(), nil)
+	if u.Scheme == "" || u.Host == "" {
+		return "", ErrInvalidAccount
+	}
+	return fetchUserIDFromHost(u.Scheme, u.Host, username, password, logger)
```
shepilov: why do you get only host from nexctloud url, what if it's deployed no in the root?

web/remote/migration.go
```diff
@@ -0,0 +1,237 @@
...
+	"github.com/labstack/echo/v4"
+)
+
+const nextcloudMigrationLogNamespace = "nextcloud-migration"
+
+// HTTPHandler owns the remote routes that need stack services. The
+// package-level [Routes] function is kept for handlers that don't depend on
+// any service; the migration endpoint lives on this struct because it
+// publishes commands to the migration broker.
+type HTTPHandler struct {
```
shepilov: Why is it in migration.go file if it's a package level handler?

Couldy you please move it to the remote.go, and also register all routes there or in routing file

web/remote/migration.go
```diff
@@ -0,0 +1,237 @@
...
+	if err != nil {
+		if errors.Is(err, webdav.ErrInvalidAuth) {
+			reqLogger.Infof("Nextcloud credentials probe rejected by remote host")
+			return jsonapi.Unauthorized(errors.New("nextcloud credentials are invalid"))
+		}
+		reqLogger.Warnf("Nextcloud credentials probe failed: %s", err)
+		return jsonapi.BadGateway(fmt.Errorf("nextcloud unreachable: %w", err))
+	}
+
+	accountID, err := ensureNextcloudAccount(inst, &body, userID)
```
shepilov: Every migraion start with different loging will create a new nextcloud account the we won't be able to view or delete

web/remote/migration.go
```diff
@@ -0,0 +1,237 @@
...
+	if err != nil {
+		reqLogger.Errorf("Failed to ensure nextcloud account: %s", err)
+		return jsonapi.InternalServerError(fmt.Errorf("ensure nextcloud account: %w", err))
+	}
+
+	doc := nextcloud.NewPendingMigration("")
+	if err := couchdb.CreateDoc(inst, doc); err != nil {
```
shepilov: to many logic in web nadler method, could you please consider moving it

web/remote/migration.go
shepilov: and also ensureNextcloudAccount it looks like model function, along with findNextcloudAccountDoc

model/nextcloud/nextcloud.go
shepilov: logger as a function param? why do we need  logic and infra?
If you really need a specific logger, consider passing it in the context

rezk2ll: thanks fixed

model/nextcloud/nextcloud.go
rezk2ll: Yes, both endpoints return the same thing, the user's Nextcloud login / WebDAV uid.  The field name on the wire is just different, `userId` in the old `user_status` response and `id` in the `cloud/user` response, which is why the JSON tag had to flip.

I checked it against two unrelated managed hosts (`thegood.cloud` and `webo.cloud`), and the value we store in `webdav_user_id` is the same as the one the old path would have given us, and subsequent WebDAV calls with it work.

On the "why switch now" part: the old probe only ever worked on hosts that ship the optional `user_status` app. It comes enabled on a vanilla Nextcloud, which is why local dev and the konnector tests never had a problem. `thegood.cloud` strips it (managed providers routinely drop optional bundled apps to cut their maintenance surface), so the old probe returns 404 there regardless of the password, and the "non-200 means bad credentials" classifier turned that into a false 401 for every `thegood.cloud` user.

The reason nobody noticed before is that this probe used to sit behind a lazy fallback in `fillUserID` that only fired when an existing account doc was missing its cached `webdav_user_id`. The konnector sets that field directly when it provisions the account, so the fallback basically never ran in practice, and the latent bug stayed latent. This PR is the first time the probe lives on the critical path of every migration trigger, which is how we ran into it on day one of real-world testing.

`cloud/user` is served by the `provisioning_api` Nextcloud app, which is in the core's always enabled list. An admin can't disable it. So it's strictly more reliable than `user_status`, which is just enabled by default. 

_It's not a 100% guarantee against a managed host that has hand-patched their Nextcloud fork to rip `provisioning_api`  out, and if it ever happens, the feature is fundamentally broken anyway because every other way to resolve the WebDAV uid from user-supplied credentials goes through the same app._
_The only escape hatch would be to assume `login === uid` and skip the probe entirely, which breaks LDAP-backed Nextclouds where those are two different strings._

model/nextcloud/nextcloud.go
rezk2ll: Agreed in principle, but the explicit `*logger.Entry` parameter is how this file already worked before this PR, the pre-existing `(nc *NextCloud).fetchUserID()` already took the webdav client's logger, and the new `FetchUserIDWithCredentials` wrapper followed the same convention so both callers could keep their per-migration `nspace=nextcloud-migration` namespace and request-scoped fields.

Switching to a context-bound logger is a reasonable cleanup but it would mean threading `context.Context` through every `model/nextcloud` function that currently takes `*logger.Entry` ( out of scope of this PR ).

Happy to do that as a follow-up refactor if you want it tracked, but I'd rather not expand the diff of the migration-trigger PR to cover it.

web/remote/migration.go
rezk2ll: The fix is a design decision I'd like to agree on before changing the behavior @shepilov @Crash--

1. **Key by `(account_type: "nextcloud", auth.url, auth.login)`** (current behavior). Safe when the user genuinely has two separate Nextcloud accounts they want to migrate (work + personal, different hosts). Downside: every typo or login change creates a ghost account that the Settings UI cannot clean up ( because we don't have the UI for it yet? )
2. **Key by `(account_type: "nextcloud")` only, one Nextcloud account per Cozy.** Calling the endpoint with a new login overwrites the previous account. No ghosts. Downside: breaks the multi-account story and destroys a previously valid account on any retry with a corrected password.
3. **Key by URL only, ignoring login.** A retry with a fixed password updates in place; a retry against a different host creates a new account. Prevents the orphan for the typical "I typed my password wrong" case, keeps the multi-account story.


The Figma design doesn't cover this use case and considers 1 user = one Nextcloud account. But it would be interesting for B2B to have multiple accounts connected to your account ( if we have the UI for it )

shepilov: why it's interesting for b2b? what is the difference with b2c?

model/nextcloud/nextcloud.go
shepilov: yes, it took WebDAV client logger, but not as a function parameter, there were no logger in fetchUserID signature

web/remote/migration.go
rezk2ll: PO reasoning, but we agreed to just one Nextcloud account per cozy for now and later we will introduce multi account if needed

model/nextcloud/nextcloud.go
rezk2ll: fixed

rezk2ll:
@shepilov i'll cleanup the commits once everything is settled with the PR logic

shepilov (approved):

PR #4729, opened 2026-04-17
fix(nextcloud): decode wildcard path before forwarding to WebDAV

rezk2ll:
## Summary

A Nextcloud migration against a real account was failing for a handful of files with a 404 from the Stack's downstream proxy. Every other file in the same folder transferred fine, so it wasn't a credentials or quota issue.

The common thread turned out to be the filename. The ones that failed all contained characters that Go's URL library leaves unescaped in paths (mostly `&`, but the same class of bug covers `+`, `,`, `;`, `=`):

- `Templates/Diagram & table.ods`
- `Templates/Mother's day.odt`

## Why it happened

The migration service sends `encodeURIComponent`-encoded filenames in the URL, so the Stack receives something like `Diagram%20%26%20table.ods`. Go's http parser sets `r.URL.RawPath` whenever the wire form differs from the default encoding of the decoded path, and Echo's wildcard param returns that raw (still-encoded) slice in that case. The handler then passed it straight into `url.URL.Path` when building the outgoing WebDAV request, and `u.String()` escaped every literal `%` a second time. Nextcloud saw `Diagram%2520%2526%2520table.ods` and looked for a file whose name literally contained `%20%26%20`, which of course doesn't exist, hence the 404.

## The fix

One `url.PathUnescape` at the handler boundary. The rest of the Stack then sees the literal filename and Go's encoder produces a correct single-encoded URL on the way out.

## Validation

Covered by a new table-driven test that asserts on `r.RequestURI` (not `r.URL.Path`) so the server-side decoding can't mask the bug. Also validated end-to-end against a real Nextcloud: after the fix, both files above transferred on the first try, full folder completed with zero errors.

### DIFF
--- a/web/remote/nextcloud.go
+++ b/web/remote/nextcloud.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -23,6 +24,22 @@ import (
 	"github.com/ncw/swift/v2"
 )
 
+// nextcloudPathParam returns the `*` wildcard param decoded to its literal
+// filename form. Echo's router stores the raw (percent-encoded) slice
+// whenever Go's http parser sets r.URL.RawPath — which happens as soon as a
+// character's default path encoding differs from what the client sent
+// (e.g. `&` is %26 on the wire but not escaped by Go's path encoder). If we
+// pass that encoded string straight into url.URL.Path when building the
+// WebDAV request, u.String() will escape every literal `%` again and the
+// wire-level URL becomes %2526 instead of %26 — Nextcloud then 404s looking
+// for a filename containing literal %XX sequences.
+//
+// Returns an error on malformed percent-encoding so callers can 400 rather
+// than forward garbage downstream.
+func nextcloudPathParam(c echo.Context) (string, error) {
+	return url.PathUnescape(c.Param("*"))
+}
+
 func nextcloudGetTrash(c echo.Context) error {
 	inst := middlewares.GetInstance(c)
 	if err := middlewares.AllowWholeType(c, permission.GET, consts.Files); err != nil {
@@ -35,7 +52,10 @@ func nextcloudGetTrash(c echo.Context) error {
 		return wrapNextcloudErrors(err)
 	}
 
-	path := c.Param("*")
+	path, err := nextcloudPathParam(c)
+	if err != nil {
+		return jsonapi.BadRequest(err)
+	}
 	files, err := nc.ListTrashed(path)
 	if err != nil {
 		return wrapNextcloudErrors(err)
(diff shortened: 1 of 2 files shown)


web/remote/nextcloud.go
```diff
@@ -23,6 +24,24 @@ import (
...
+// (e.g. `&` is %26 on the wire but not escaped by Go's path encoder). If we
+// pass that encoded string straight into url.URL.Path when building the
+// WebDAV request, u.String() will escape every literal `%` again and the
+// wire-level URL becomes %2526 instead of %26 — Nextcloud then 404s looking
+// for a filename containing literal %XX sequences.
+func nextcloudPathParam(c echo.Context) string {
+	raw := c.Param("*")
+	decoded, err := url.PathUnescape(raw)
+	if err != nil {
+		return raw
```
shepilov: why in case of error you return raw string?

web/remote/remote_test.go
```diff
@@ -428,3 +429,126 @@ func TestNextcloudDownstreamFailOnConflict(t *testing.T) {
...
+		require.Equal(t, http.StatusCreated, res.StatusCode)
+	}
+
+	cases := []struct {
+		name           string
+		encodedPath    string
+		wantDecodedURL string
+	}{
+		{"Ampersand", "Diagram%20%26%20table.ods", "/remote.php/dav/files/testuser/Diagram & table.ods"},
+		{"Apostrophe", "Mother's%20day.odt", "/remote.php/dav/files/testuser/Mother's day.odt"},
```
shepilov: Apostrophe is not actualy encoded

web/remote/remote_test.go
```diff
@@ -428,3 +429,126 @@ func TestNextcloudDownstreamFailOnConflict(t *testing.T) {
...
 }
+
+// TestNextcloudDownstreamPreservesSpecialChars pins the behaviour for
+// filenames whose encoded form sets RawPath on the incoming request — `&`,
+// `#`, apostrophes, and similar. Echo's wildcard param comes back encoded
+// in that case, and the older implementation let it through as url.URL.Path,
+// which double-encoded it on the way out to Nextcloud and produced a
+// phantom 404 ("Diagram%20%26%20table.ods" becoming
+// "Diagram%2520%2526%2520table.ods" on the wire).
+func TestNextcloudDownstreamPreservesSpecialChars(t *testing.T) {
```
shepilov: I think unit est for the function is enough, you don't need to test everythig passes t owebdav, and if you are testing this, then why only for one endpoint?

web/remote/nextcloud.go
rezk2ll: fixed

web/remote/remote_test.go
rezk2ll: I've replaced the integration test with a unit test on nextcloudPathParam itself

shepilov (approved):

PR #4733, opened 2026-04-20
Sync B2B contact on RabbitMQ `user.created` and `user.deleted` events

shepilov:
## Summary 
 For B2B users, the stack now keeps an external contact in every other instance of the same organization:
  - on `auth / user.created`, create the contact in all other org instances
  - on `b2b / user.deleted`, remove that external contact from all other org instances

The sync is idempotent, skips the user’s own instance, and fails loudly on inconsistent data so the message can be retried / dead-lettered.

  ### Implementation
  - filters to `metadata.external: true`
  - organization lookup is **domain-only**
  - contact lookup is email-based, then filtered to external contacts
  - contact creation uses the shared `contact.Create(...)` model API

### DIFF
--- a/model/job/globals.go
+++ b/model/job/globals.go
@@ -52,3 +52,8 @@ func System() JobSystem {
 	}
 	return globalJobSystem
 }
+
+// SystemUnsafe returns the global job system, even if it's not initialized
+func SystemUnsafe() JobSystem {
+	return globalJobSystem
+}
--- a/pkg/config/config/testdata/full_config.yaml
+++ b/pkg/config/config/testdata/full_config.yaml
@@ -129,10 +129,10 @@ rabbitmq:
       kind: topic
       durable: true
       declare_exchange: true
-      dlx_name: auth.dlx
-      dlq_name: auth.dlq
+      dlx_name: stack.dlx
+      dlq_name: stack.auth.dlq
       queues:
-        - name: user.password.updated
+        - name: stack.user.password.updated
           declare: true
           declare_dlx: true
           declare_dlq: true
@@ -140,7 +140,7 @@ rabbitmq:
           delivery_limit: 5
           bindings:
             - password.changed
-        - name: user.created
+        - name: stack.user.created
           declare: true
           declare_dlx: false
           declare_dlq: true
@@ -148,6 +148,21 @@ rabbitmq:
           delivery_limit: 5
           bindings:
             - user.created
+    - name: b2b
+      kind: topic
+      durable: true
+      declare_exchange: false
+      dlx_name: stack.dlx
+      dlq_name: stack.b2b.dlq
+      queues:
+        - name: stack.b2b.user.deleted
+          declare: true
+          declare_dlx: true
+          declare_dlq: true
+          prefetch: 8
+          delivery_limit: 5
+          bindings:
+            - user.deleted
 
 common_settings:
   default:
--- a/pkg/config/config/testdata/full_config.yaml
+++ b/pkg/config/config/testdata/full_config.yaml
@@ -129,7 +129,7 @@ rabbitmq:
       kind: topic
       durable: true
       declare_exchange: true
-      dlx_name: stack.dlx
+      dlx_name: stack.auth.dlx
       dlq_name: stack.auth.dlq
       queues:
         - name: stack.user.password.updated
@@ -152,7 +152,7 @@ rabbitmq:
       kind: topic
       durable: true
       declare_exchange: false
-      dlx_name: stack.dlx
+      dlx_name: stack.b2b.dlx
       dlq_name: stack.b2b.dlq
       queues:
         - name: stack.b2b.user.deleted
--- a/pkg/rabbitmq/handlers.go
+++ b/pkg/rabbitmq/handlers.go
@@ -138,7 +138,6 @@ type UserCreatedMessage struct {
 	WorkplaceFqdn      string `json:"workplaceFqdn"` // The fully qualified workplace domain
 	OrganizationID     string `json:"organizationId,omitempty"`
 	OrganizationDomain string `json:"organizationDomain,omitempty"`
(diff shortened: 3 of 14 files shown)


pkg/config/config/testdata/full_config.yaml
```diff
@@ -129,25 +129,40 @@ rabbitmq:
       kind: topic
       durable: true
       declare_exchange: true
-      dlx_name: auth.dlx
-      dlq_name: auth.dlq
+      dlx_name: stack.dlx
```
taratatach: This doesn't seem right: other apps are targeting the `auth` exchange and the contracts file also uses `auth`, not `stack` or `stack.dlx`.

Crash--:
Do we have a cli somewhere to run to check consistency ?

pkg/rabbitmq/handlers.go
```diff
@@ -125,17 +125,20 @@ func NewUserCreatedHandler() *UserCreatedHandler {
...
+	Iterations         int    `json:"iterations"`
+	Hash               string `json:"hash"`
+	PublicKey          string `json:"publicKey"`
+	PrivateKey         string `json:"privateKey"`
+	Key                string `json:"key"`
+	Timestamp          int64  `json:"timestamp"`
+	WorkplaceFqdn      string `json:"workplaceFqdn"` // The fully qualified workplace domain
+	OrganizationID     string `json:"organizationId,omitempty"`
+	OrganizationDomain string `json:"organizationDomain,omitempty"`
+	CanUpgrade         *bool  `json:"canUpgrade,omitempty"`
```
taratatach: Are you sure the `canUpgrade` attribute is part of the `user.created` message? I can't see it in the ADR and this is related to plans which is not handled by Signup.

shepilov:
> Do we have a cli somewhere to run to check consistency ?

nope, for what flow do you want cli?

Monitor errors during sync, and run check job manually after to check the inconsistency? Or insted or error monitoring, monitor this inconsistency job?

pkg/rabbitmq/org_contacts.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/cozy/cozy-stack/model/contact"
+	"github.com/cozy/cozy-stack/model/instance"
+	"github.com/cozy/cozy-stack/model/instance/lifecycle"
+	"github.com/cozy/cozy-stack/pkg/couchdb"
+	"github.com/cozy/cozy-stack/pkg/utils"
+)
+
+// SyncCreatedOrgContact syncs a newly created B2B user as an external contact
+// into every other instance of the organization.
+func SyncCreatedOrgContact(ctx context.Context, target *instance.Instance, msg UserCreatedMessage) error {
```
taratatach, line 18: Isn't everything stored in `msg` also stored in the instance itself?

pkg/rabbitmq/handlers.go
```diff
@@ -216,6 +219,70 @@ func (h *UserCreatedHandler) Handle(ctx context.Context, d amqp.Delivery) error
 		return fmt.Errorf("user.created: update passphrase: %w", err)
 	}
 	log.Infof("user.created: successfully updated passphrase for instance: %s (PasswordDefined: %v)", inst.Domain, inst.PasswordDefined)
+
+	if msg.OrganizationDomain != "" {
+		if err := SyncCreatedOrgContact(ctx, inst, msg); err != nil {
+			log.Errorf("user.created: failed to sync organization contacts for %s: %v", msg.WorkplaceFqdn, err)
+			return err
+		}
+	}
```
taratatach, line 227: Why wait until we receive this message to create the contact? We could create it as soon as the instance is created, couldn't we?

Crash--:
Something like cli fcsk, so "run check job manually after to check the inconsistency". So when I want to run this job, I should be able to do it. And the check means read LDAP & check if all the instances of the orga is OK.

model/job/globals.go
```diff
@@ -52,3 +52,8 @@ func System() JobSystem {
 	}
 	return globalJobSystem
 }
+
+// SystemIfStarted returns the global job system when it has been initialized.
+func SystemIfStarted() JobSystem {
+	return globalJobSystem
+}
```
taratatach, line 59: The only difference I see between this new function and `System()` is that this one does not actually make sure the job system has been initialized.
What's its purpose?

pkg/config/config/testdata/full_config.yaml
shepilov: it's a dlx exchange, and I think it's better to have it per service, and not one common, other apps don't care our deat letter

pkg/rabbitmq/handlers.go
shepilov: removed, I thing I've seen it in message example

pkg/config/config/testdata/full_config.yaml
taratatach: Ah yes, right. This doesn't match the actual config which uses `stack.auth.dlx` though.

pkg/rabbitmq/org_contacts.go
shepilov: InternalEmail and Mobile we get from message, and as far as we have it there,  we should probably have it in settings, but I'm not sure about internal email. We are changing mind about it too often

pkg/rabbitmq/handlers.go
shepilov: to be totally sure that it's fully created, yes, it could've been a job during instance creation, but not for deletion, and future blocking/unblocking. So I've decided to do everything in the same manner

model/job/globals.go
shepilov: for the tests, that doesn't start job system, to avoid panic on 'destory'

pkg/rabbitmq/org_contacts.go
taratatach: The email and phone are part of the settings document.

PR #4736, opened 2026-04-23
feat(nextcloud): endpoint to delete migrated folder from Nextcloud

rezk2ll:
## Summary

Adds `POST /remote/nextcloud/migration/:id/delete-source`, a synchronous cleanup endpoint the Settings UI calls after a migration completes to remove the Nextcloud-side data the user just imported into their Cozy. It reuses the credentials already stored on the instance's `io.cozy.accounts` document, so the caller never has to hand them back.

The product flow this enables: once the user has finished migrating their Nextcloud account into Cozy Drive, the Settings UI offers a button to wipe their Nextcloud. This endpoint is what that button calls.

## API

`POST /remote/nextcloud/migration/:id/delete-source` (no request body)

| Status | Meaning |
|---|---|
| 204 No Content | Source deleted, trash emptied, `source_deleted_at` stamped on the tracking doc. |
| 401 Unauthorized | Stored Nextcloud credentials were rejected. |
| 404 Not Found | Migration id does not exist, or the Nextcloud account was removed out of band. |
| 409 Conflict | Migration is not in the `completed` status, or the source has already been deleted. |
| 502 Bad Gateway | Nextcloud was reached but returned an unexpected error. The tracking doc is left untouched so the user can retry. |

## Behavior

The endpoint runs three steps end to end before stamping `source_deleted_at`:

1. **Remove the migrated content.** If the migration imported a sub-folder, one DELETE handles it. If the user migrated their whole Nextcloud (`source_path: "/"`, the dominant case), the endpoint lists the WebDAV home with PROPFIND and DELETEs each top-level child individually. See the limitation below for why.
2. **Empty the trash.** Nextcloud moves deleted files to a per-user trashbin where they keep counting against the user's quota until purged. The endpoint empties it so the migrated bytes are actually freed.
3. **Stamp `source_deleted_at`** on the tracking doc so the UI can hide the button and the endpoint refuses repeats.

A Nextcloud 404 at any step is treated as success: the goal is "the content is gone", which it is. The first hard failure aborts before stamping, so a partial cleanup leaves the doc in a state the user can retry from.

Only migrations in the `completed` status are eligible. Failed or canceled migrations may still hold files on Nextcloud that never reached Cozy, and removing them would lose data.

## Why we recurse on `source_path = "/"`

A direct `DELETE` on `/remote.php/dav/files/<user>/` returns 403 from Nextcloud: WebDAV does not let a regular user delete their own home folder. We could either reject `/` and tell the user "we cannot do this", or wipe the home in the way Nextcloud does support, by emptying its contents. We chose the latter because it matches the user's literal intent.

## Limitations

- **Not a full account deletion.** This wipes the user's *files* on Nextcloud, not the Nextcloud *account itself*. Removing the account would need admin OCS Provisioning credentials, which the Stack does not have. If the product later needs that, it is a separate flow.
- **Trash empty is total.** The endpoint clears the entire trashbin, not just items it created. If the user had unrelated items already in their trash, those are wiped too. This matches the "I'm done with Nextcloud" intent of the button but is worth flagging.
- **No parallel deletes.** Top-level children are deleted sequentially. A user with hundreds of folders would feel this; the typical home has well under 20.

## Tests performed

- [x] Unit tests cover the status gate, single-path delete, root recursion, trash empty, idempotent 404, 401 passthrough, and 409-on-repeat paths.
- [x] End-to-end against a real Nextcloud with `source_path: "/"` (account with 11 top-level entries plus one trash item): all entries gone, trash empty, `source_deleted_at` stamped, second call returns 409.

### DIFF
--- a/model/nextcloud/delete_source.go
+++ b/model/nextcloud/delete_source.go
@@ -1,0 +1,173 @@
+package nextcloud
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"path"
+	"time"
+
+	"github.com/cozy/cozy-stack/model/instance"
+	"github.com/cozy/cozy-stack/pkg/consts"
+	"github.com/cozy/cozy-stack/pkg/couchdb"
+	"github.com/cozy/cozy-stack/pkg/logger"
+	"github.com/cozy/cozy-stack/pkg/webdav"
+)
+
+// ErrMigrationNotDeletable gates delete-source to the completed status.
+// Failed and canceled migrations may still hold files on Nextcloud that
+// never reached Cozy, and removing the folder would lose them.
+var ErrMigrationNotDeletable = errors.New("nextcloud migration source can only be deleted after a completed migration")
+
+// ErrMigrationSourceAlreadyDeleted is surfaced rather than swallowed so
+// duplicate clicks from the UI show up as bugs instead of silent no-ops.
+var ErrMigrationSourceAlreadyDeleted = errors.New("nextcloud migration source already deleted")
+
+// ErrNextcloudAccountMissing means the single nextcloud account was
+// removed out-of-band between the migration and this cleanup call.
+var ErrNextcloudAccountMissing = errors.New("no nextcloud account configured for this instance")
+
+// DeleteMigrationSource removes the Nextcloud content ingested by the
+// given migration, then empties the user's trashbin so the migrated
+// bytes stop counting against quota.
+//
+// SourcePath == "/" enumerates the WebDAV home and deletes each
+// top-level child: Nextcloud refuses DELETE on the home itself with
+// 403, so a single call would fail. An empty SourcePath is treated as
+// "/" to match legacy docs predating source_path persistence.
+//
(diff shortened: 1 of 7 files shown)


model/nextcloud/delete_source.go
```diff
@@ -0,0 +1,173 @@
...
+		log.WithField("status", doc.Status).
+			Infof("Delete-source rejected: migration is not completed")
+		return ErrMigrationNotDeletable
+	}
+	if doc.SourceDeletedAt != nil {
+		log.Infof("Delete-source rejected: source already deleted")
+		return ErrMigrationSourceAlreadyDeleted
+	}
+	sourcePath := doc.SourcePath
+	if sourcePath == "" {
```
shepilov, line 70: are you sure you want to delete root of source path is not set?

rezk2ll: Yes, that's the intended behaviour, same as the migration. If the source path is not specified, that means it's the root.

while in theory we won't need to have this check but it's good to have

shepilov (approved):

PR #4737, opened 2026-04-23
fix: 409 conflict after auto accept of a shared drive

shepilov:
## Summary
 - AddGroupsAndContacts adds Bob, saves doc (rev 1)
 - SendInvitations starts → SendShortcut → sends sharing request to Bob's cozy
 - Bob's cozy auto-accepts (trusted sender) → calls back POST /sharings/:id/answer on your cozy
 - ProcessAnswer on your cozy updates the sharing doc (rev 1 → 2) — member status becomes Ready
 - Meanwhile, SendInvitations finishes the goroutine and calls couchdb.UpdateDoc(inst, s) with stale rev 1 → CouchDB 409  

The sharing actually worked, the member was added and accepted, but the response to the client is 409 because of the stale revision.

### DIFF
--- a/model/sharing/invitation.go
+++ b/model/sharing/invitation.go
@@ -72,10 +72,45 @@ func (s *Sharing) SendInvitations(inst *instance.Instance, perms *permission.Per
 		})
 	}
 	errg := g.Wait()
-	if err := couchdb.UpdateDoc(inst, s); err != nil {
+	if err := s.updateMemberStatusesWithRetry(inst); err != nil {
 		return err
 	}
 	return errg
+}
+
+// updateMemberStatusesWithRetry persists the current member statuses to
+// CouchDB, retrying on conflict. A conflict can happen when ProcessAnswer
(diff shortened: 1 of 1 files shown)


model/sharing/invitation.go
```diff
@@ -72,12 +72,47 @@ func (s *Sharing) SendInvitations(inst *instance.Instance, perms *permission.Per
...
+		if err == nil || !couchdb.IsConflictError(err) || attempt >= maxRetries {
+			return err
+		}
+		s, err = FindSharing(inst, s.SID)
+		if err != nil {
+			return err
+		}
+		for i, status := range desiredStatuses {
+			if i >= len(s.Members) {
+				continue
```
taratatach, line 106: nit: we could break here

taratatach (approved):
Maybe we could compare members' instances to be sure we're updating the right member instead of relying on indexes which can be invalid when retrying?

PR #4738, opened 2026-04-23
feat: Proxy get shortcut route for shared drives

zatteo:


shepilov (approved):

zatteo:
Thank you for the follow up @shepilov!

PR #4742, opened 2026-04-29
Support for shared drives whose root is a file, not only a folder.

shepilov:
## Summary 
 - shared drive creation from either a file or a folder
  - a new drive_root_type field on shared drives
  - support for file-root read and update routes on the root file
  - share-by-link support for file-root shared drives
  - _changes and realtime support for file-root shared drives
  - directory-only routes are rejected
  - move/copy rules for file-root drives:
      - copy out is allowed
      - move is rejected
      - file-root drives cannot be destinations


docs/shared-drives.md
```diff
@@ -290,6 +337,11 @@ Content-Type: application/vnd.api+json
 Unless stated otherwise, a permission on the whole `io.cozy.files` doctype is
 required to use the following routes.
 
+For file-root shared drives (`drive_root_type = file`), iteration 1 currently
+supports only file-shaped routes. Directory-only routes return
+`422 Unprocessable Entity`. `_changes` and realtime are available with exact
```
taratatach, line 342: I don't think it makes much sense to provide a changes feed for file-root shared drives as the only relevant information will be the latest version of the file which can be fetched more easily via other routes.

model/sharing/files.go
```diff
@@ -302,106 +302,135 @@ func (s *Sharing) CreateDirForSharing(inst *instance.Instance, rule *Rule, paren
...
+		return dir, nil, nil
 	}
-
-	// Check not in trash (path-based check for folders inside trash)
-	if strings.HasPrefix(dir.Fullpath, vfs.TrashDirName+"/") {
-		return nil, ErrSystemFolder
+	if file.Trashed {
+		return nil, nil, ErrFileInTrash
+	}
+	if err := checkRootForSharing(file.ReferencedBy, ErrFileAlreadyShared); err != nil {
```
taratatach, line 400: Will that detect a replicated sharing? Because `AddReferenceForSharing()` won't add a reference if the file sharing is not a federated one 🤔

web/sharings/drives.go
```diff
@@ -233,6 +252,9 @@ func HeadDirOrFile(c echo.Context, inst *instance.Instance, s *sharing.Sharing)
 
 // ReadMetadataFromPath allows to get file/dir information for a path.
 func ReadMetadataFromPath(c echo.Context, inst *instance.Instance, s *sharing.Sharing) error {
+	if err := ensureDirectoryBackedSharedDrive(s); err != nil {
+		return err
+	}
```
taratatach: Why this limitation here? `files.ReadMetadataFromPath` also works with files.

web/sharings/drives.go
```diff
@@ -292,6 +317,9 @@ func ModifyMetadataByIDHandler(c echo.Context, inst *instance.Instance, s *shari
 	if err != nil {
 		return files.WrapVfsError(err)
 	}
+	if s.HasFileDriveRoot() && patch.DirID != nil {
+		return jsonapi.NewError(http.StatusUnprocessableEntity, "cannot move the root file of a file-root shared drive")
+	}
```
taratatach: This should probably apply to directory roots as well.

web/sharings/drives.go
```diff
@@ -448,6 +496,8 @@ func ThumbnailHandler(c echo.Context, inst *instance.Instance, s *sharing.Sharin
 // FileDownloadCreateHandler stores the required path into a secret usable for
 // the download handler below.
 func FileDownloadCreateHandler(c echo.Context, inst *instance.Instance, s *sharing.Sharing) error {
+	// TODO: The route contract can be broader than the drive root, especially
+	// for owner requests. To fix it, we need explicit shared-drive scope check here.
```
taratatach, line 496: Not sure what you mean here. Sure the given `Path` or `Id` params can be outside the shared drive but we have permission checks for that already.
And this applies to other routes like `CopyFile`.

web/sharings/drives.go
```diff
@@ -384,12 +412,23 @@ func applyPatch(c echo.Context, fs vfs.VFS, patch *docPatch) (err error) {
...
+// use an exact root-file match and ignore unrelated file changes.
 func ChangesFeed(c echo.Context, inst *instance.Instance, s *sharing.Sharing) error {
 	// TODO: if owner then fail, shouldn't be accessing their own stuff, risk recursion download kinda thing
 	// TODO: should this break if there ever is actually more than 1 directory ?
+	if s.HasFileDriveRoot() {
+		rule := s.FirstFilesRule()
+		if rule == nil || rule.Selector == couchdb.SelectorReferencedBy || len(rule.Values) == 0 {
+			return jsonapi.NotFound(errors.New("shared drive not found"))
+		}
+		sharedFile, err := inst.VFS().FileByID(rule.Values[0])
```
taratatach: nit: Accessing the shared document or at least its ID could be extracted in a `Sharing` method as we're using pretty much these same lines in different places.

web/sharings/drives.go
```diff
@@ -682,20 +751,19 @@ func ListSharedDriveShareByLinkPermissions(c echo.Context, inst *instance.Instan
...
 
-	fs := inst.VFS()
 	ids := make([]string, 0)
 	for _, id := range strings.Split(rawIDs, ",") {
 		id = strings.TrimSpace(id)
 		if id == "" {
 			return nil, jsonapi.InvalidParameter("ids", errors.New("ids must contain file or folder IDs"))
 		}
-		if err := isWithinDirectory(fs, id, rootDir); err != nil {
+		if err := validateSharedDrivePermissionTargetID(c, inst, s, id); err != nil {
```
taratatach, line 762: I don't really understand this change as `extractSharedDrivePermissionTargetIDs` seems to only apply to dir-root federated sharings.

web/sharings/move.go
```diff
@@ -74,6 +74,25 @@ func validateMoveRequest(req moveRequest) error {
 	return nil
 }
 
+func validateFileBackedDriveMoveCopy(req moveRequest, sourceSharing, destSharing *sharing.Sharing) error {
+	if sourceSharing != nil && sourceSharing.HasFileDriveRoot() {
+		if !req.Copy {
+			return jsonapi.NewError(http.StatusUnprocessableEntity, "file-root shared drives only support copy as a source")
```
taratatach: nit:

```suggestion
			return jsonapi.NewError(http.StatusUnprocessableEntity, "file-root shared drives only support copy")
```

web/sharings/move.go
```diff
@@ -74,6 +74,25 @@ func validateMoveRequest(req moveRequest) error {
...
+		}
+		if req.Source.DirID != "" {
+			return jsonapi.NewError(http.StatusUnprocessableEntity, "file-root shared drives can only copy their root file")
+		}
+		rule := sourceSharing.FirstFilesRule()
+		if rule == nil || len(rule.Values) == 0 || req.Source.FileID != rule.Values[0] {
+			return jsonapi.NewError(http.StatusUnprocessableEntity, "file-root shared drives can only copy their root file")
```
taratatach: Was this message copy/pasted? It doesn't look right 🤔

docs/shared-drives.md
shepilov: But for the local files, we have this changes feed, why should we not implement it for shared drives?

model/sharing/files.go
shepilov: Nope, do we need it?

web/sharings/drives.go
shepilov: Actually, yes, the only thing we don't need it for is files; we don't have a file tree if it's only one file

web/sharings/drives.go
shepilov: Yes, that's why it's just TODO, if use use "shared drive" API to access files outside of a shared drive. Yes, you have permissions to this, but it's kind of out of API contract

web/sharings/move.go
shepilov: but they don't support copy as a destination, when we copy "to"

web/sharings/drives.go
shepilov: nope, it's applied for the files as well

web/sharings/move.go
shepilov: AI generated, my bad, missed it

docs/shared-drives.md
taratatach: We have a changes feed for the whole drive which makes sense because we have more than one file. In the case of federated sharing of files, we have only one file so it does not really make sense to return a changes feed since that feed will contain lots of `deleted` docs for only one interesting doc which can be fetched directly.

web/sharings/move.go
taratatach: OK, but we have a different error message when a file-root federated sharing is passed as destination.

taratatach (approved):

PR #4745, opened 2026-04-30
feat: Allow io.cozy.contacts and io.cozy.contacts.groups in token exc…

zatteo:
…hange

Replace the single allowed scope constant with an allowlist and a validateTokenExchangeScope function that accepts any space-separated combination of io.cozy.files, io.cozy.contacts, and io.cozy.contacts.groups.

### DIFF
--- a/web/auth/token_exchange.go
+++ b/web/auth/token_exchange.go
@@ -25,10 +25,15 @@ import (
 const (
 	tokenExchangeAdminRole             = "admin"
 	tokenExchangeOwnerRole             = "owner"
-	tokenExchangeAllowedScope          = "io.cozy.files"
 	tokenExchangeOAuthClientName       = "Twake Admin Panel"
 	tokenExchangeOAuthClientSoftwareID = "twake-admin-panel"
 )
+
+var tokenExchangeAllowedScopes = map[string]struct{}{
+	"io.cozy.files":           {},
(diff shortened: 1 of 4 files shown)


web/auth/token_exchange.go
```diff
@@ -47,8 +52,8 @@ func executeTokenExchange(c echo.Context, inst *instance.Instance, req tokenExch
 	if err != nil {
 		return nil, echo.NewHTTPError(http.StatusBadRequest, err.Error())
 	}
-	if req.Scope != tokenExchangeAllowedScope {
-		return nil, echo.NewHTTPError(http.StatusBadRequest, "invalid scope")
+	if err := validateTokenExchangeScope(req.Scope); err != nil {
```
shepilov, line 55: Could you please trim spaces before validation, and use `stings.Split` please, because later we use it, and with `stings.Fields` you allow tokens like `io.cozy.files\tio.cozy.contact`.

Or if you want a multiline support, you then need to use `strings.Join(fields, " ")` as scope, and return it from validation function, and then probably rename it

shepilov (approved):

web/auth/token_exchange.go
zatteo: Done. I let you check again and merge if it's okay 👍

PR #4751, opened 2026-05-05
feat: return explicit refresh-token invalid error for OAuth refresh failures

shepilov:
## Bug
When a konnector calls POST /accounts/:accountType/:accountID/refresh, stack refreshes OAuth token with provider.
If provider rejects refresh token, stack currently returns a generic internal error.
Connector cannot reliably detect “refresh token is invalid, user must reconnect account”, so it often falls back to generic VENDOR_DOWN.

This hides the real problem for users and support. And doesn't allow users to reinitialize the refresh token manually

  ## What changed

  - Added typed refresh error classification in account refresh flow.
  - On provider OAuth errors invalid_token or invalid_grant, stack now classifies as:
      - code: oauth_refresh_invalid_token
  - In /accounts/:accountType/:accountID/refresh, this classified error is returned as JSON:API:
      - status 401
      - code oauth_refresh_invalid_token
      - detail: reconnect needed message
  - Kept other refresh failures on generic error path.

### DIFF
--- a/model/account/type.go
+++ b/model/account/type.go
@@ -65,6 +65,56 @@ var RefreshToken = "refresh_token"
 // within an account does not allow refreshing it.
 var ErrUnrefreshable = errors.New("this account can not be refreshed")
 
+const (
+	// RefreshOAuthErrorCodeInvalidToken indicates the refresh token has been
+	// rejected by the OAuth provider and the account should be reconnected.
+	RefreshOAuthErrorCodeInvalidToken = "oauth_refresh_invalid_token"
+)
+
+// RefreshOAuthError represents a classified OAuth refresh failure.
+type RefreshOAuthError struct {
+	Code           string
+	ProviderStatus int
+	ProviderError  string
+}
+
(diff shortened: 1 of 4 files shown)


model/account/type.go
```diff
@@ -417,8 +467,32 @@ func (at *AccountType) RefreshAccount(a Account) error {
 	}
 
 	if res.StatusCode != 200 {
-		log.Errorf("Account %s: OAuth service error (status %d): %s", a.ID(), res.StatusCode, string(resBody))
-		return fmt.Errorf("oauth service responded with status %d: %s", res.StatusCode, string(resBody))
+		var out tokenEndpointResponse
+		if err = json.Unmarshal(resBody, &out); err != nil {
+			log.Errorf("Account %s: OAuth service error (status %d): non-JSON response", a.ID(), res.StatusCode)
```
taratatach: nit: we could log the response body like we used to to make debugging easier.

taratatach (approved):

PR #4759, opened 2026-05-18
fix: fix OIDC acceptance from preview sharing links

shepilov:
## Context
- alice.twake.app shares a folder with bob.twake.app
    - recipient instance: not known yet and Alice share by email [[bob@example.com](mailto:bob@example.com)](mailto:bob@example.com)
    - Bob member exists in Alice’s sharing document `email = bob@example.com, instance = "", status = pending`
    - Bob also has a sharing credential state, for example: `credentials[0].state = abc123`
    - And preview permission exists `sharecode = 123` mapped to Bob
- Bob receives/open this preview link [https://alice-drive.twake.app/preview?sharecode=](https://alice-drive.twake.app/preview?sharecode=lgGAegWLpvqk)123
    - Important: this URL contains sharecode, not state.
    - Bob Opens The Link
    - Bob lands on Alice’s Drive preview page.
    - The stack can identify Bob at this point because it has sharecode = 123
    - Internally this works: sharecode -> preview permission -> [[bob@example.com](mailto:bob@example.com)](mailto:bob@example.com) -> Bob member, so Bob is not missing yet.
- Bob Clicks “Synchronize With My Twake”
    - The UI sends Bob to Alice’s sharing discovery page, still based on the preview link/sharecode flow.
    - Sharecode -> Bob member -> Bob credential state abc123, but before the fix, the discovery page did not do that conversion, it renders OIDC button with https://alice.twake.app/oidc/sharing?sharingID=SHARING_ID&state=
    - Bob Clicks “Log In With OIDC” on the button without sharing state

## Fix
 When discovery receives a valid `sharecode`, resolve the member from that sharecode and derive the member credential `state` before rendering the page. This keeps the existing OIDC sharing flow unchanged while making preview links provide the state expected by the OIDC callback.

### DIFF
--- a/web/sharings/sharings.go
+++ b/web/sharings/sharings.go
@@ -686,6 +686,20 @@ func renderAlreadyAccepted(c echo.Context, inst *instance.Instance, cozyURL stri
 	})
 }
 
+func renderInvalidSharing(c echo.Context, inst *instance.Instance) error {
+	return c.Render(http.StatusBadRequest, "error.html", echo.Map{
+		"Domain":         inst.ContextualDomain(),
+		"ContextName":    inst.ContextName,
+		"Locale":         inst.Locale,
+		"Title":          inst.TemplateTitle(),
+		"Favicon":        middlewares.Favicon(inst),
+		"Illustration":   "/images/generic-error.svg",
+		"Error":          "Error Invalid sharing",
+		"SupportEmail":   inst.SupportEmailAddress(),
+		"SupportPageURL": inst.SupportPageURL(),
+	})
+}
+
 func renderDiscoveryForm(c echo.Context, inst *instance.Instance, code int, sharingID, state, sharecode, shortcut string, m *sharing.Member) error {
 	publicName, _ := settings.PublicName(inst)
 	fqdn := strings.TrimPrefix(m.Instance, "https://")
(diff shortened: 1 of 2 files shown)


web/sharings/sharings.go
```diff
@@ -886,6 +897,9 @@ func PostDiscovery(c echo.Context) error {
 				return wrapErrors(err)
 			}
 		}
+		if state, err = discoveryStateForMember(s, state, member); err != nil {
+			return wrapErrors(err)
+		}
```
taratatach, line 902: Shouldn't we rely solely on the state that was sent by the client at this point?

If it was missing, the state should now have been computed in `GetDiscovery()` and sent as parameter to `PostDiscovery()`.

shepilov: it was "just in case" fix, I can't imagine flow when we have post with empty state here, but in the same time I didn't find a clean separation between different flows with sharecode and state in this function, so I've decided add this "helper" to have PostDiscovery wtihout state.
It's not connected to the initial buf, so if you also don't have in mind such kind of flows and this lines confuse you, I can remove them.

taratatach: I don't have all the flows in mind. I'm even surprised to see the member can either be computed from sharecode or state at this point since having a state seems necessary…

If I had all the pieces together I'd say let's remove all sharecode use in this function but maybe there are weird cases.

So you're right, let's keep this "just in case" fix.

taratatach (approved):

PR #4760, opened 2026-05-18
Fix shared drive failures caused by multiple `share-interact` permission docs being created for the same sharing

shepilov:
## Summary

The issue could happen when drive sharing invitations were prepared concurrently: several members could trigger interact-code creation at the same time, and CouchDB generated different permission document IDs. Later reads could pick a partial permission doc, causing missing drive tokens and errors like “member was not found”

 ## Changes

  - Use a canonical deterministic ID for `share-interact` permission docs.
  - Repair existing duplicate `share-interact` permission docs by merging codes into the canonical doc and deleting legacy duplicates.
  - Make `GetForShareInteract` repair duplicates before returning permissions.
  - Serialize `GetInteractCode` with the existing instance lock to avoid local concurrent writes for the same sharing.

### DIFF
--- a/model/permission/permissions.go
+++ b/model/permission/permissions.go
@@ -3,6 +3,7 @@
 package permission
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"net/http"
@@ -10,6 +11,7 @@ import (
 	"time"
 
 	build "github.com/cozy/cozy-stack/pkg/config"
+	stackconfig "github.com/cozy/cozy-stack/pkg/config/config"
 	"github.com/cozy/cozy-stack/pkg/consts"
 	"github.com/cozy/cozy-stack/pkg/couchdb"
 	"github.com/cozy/cozy-stack/pkg/couchdb/mango"
@@ -17,6 +19,7 @@ import (
 	"github.com/cozy/cozy-stack/pkg/logger"
 	"github.com/cozy/cozy-stack/pkg/metadata"
 	"github.com/cozy/cozy-stack/pkg/prefixer"
+	"github.com/cozy/cozy-stack/pkg/utils"
 	"github.com/labstack/echo/v4"
 )
 
@@ -243,15 +246,27 @@ func GetForSharePreview(db prefixer.Prefixer, sharingID string) (*Permission, er
 }
 
 // GetForShareInteract retrieves the Permission doc for a given sharing to
-// read/write a note
+// read/write a note. It may repair legacy duplicate share-interact permission
+// docs as part of the read by creating/updating the canonical permission doc
+// and deleting duplicate legacy docs.
 func GetForShareInteract(db prefixer.Prefixer, sharingID string) (*Permission, error) {
-	return getFromSource(db, TypeShareInteract, consts.Sharings, sharingID)
+	mu := stackconfig.Lock().ReadWrite(db, shareInteractLockName(sharingID))
+	if err := mu.Lock(); err != nil {
+		return nil, err
+	}
+	defer mu.Unlock()
+
+	return getOrRepairShareInteractPermissions(db, sharingID)
 }
 
 // ShareInteractPermissionID returns the canonical permission document ID for a
 // share-interact permission set.
 func ShareInteractPermissionID(sharingID string) string {
 	return TypeShareInteract + "-" + sharingID
+}
+
+func shareInteractLockName(sharingID string) string {
+	return "permissions/share-interact/" + sharingID
 }
 
 func getShareInteractPermissions(db prefixer.Prefixer, sharingID string) ([]Permission, error) {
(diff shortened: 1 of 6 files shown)


model/permission/permissions.go
```diff
@@ -247,6 +248,143 @@ func GetForShareInteract(db prefixer.Prefixer, sharingID string) (*Permission, e
...
+		}
+
+		if len(merged.Permissions) == 0 && len(p.Permissions) > 0 {
+			merged.Permissions = p.Clone().(*Permission).Permissions
+		}
+		if merged.Metadata == nil && p.Metadata != nil {
+			merged.Metadata = p.Metadata.Clone()
+		}
+		if merged.ExpiresAt == nil && p.ExpiresAt != nil {
+			merged.ExpiresAt = p.ExpiresAt
```
taratatach, line 413: Should we keep the furthest expiration date? Or maybe there's a better selection criterion.
It seems weird to just keep the expiration date of the last permission doc we process.

model/permission/permissions.go
```diff
@@ -247,6 +248,143 @@ func GetForShareInteract(db prefixer.Prefixer, sharingID string) (*Permission, e
...
+		hasUsablePermission = true
+		if p.ID() == canonicalID {
+			hasCanonical = true
+			merged.SetRev(p.Rev())
+		} else {
+			duplicates = append(duplicates, p)
+		}
+
+		if len(merged.Permissions) == 0 && len(p.Permissions) > 0 {
+			merged.Permissions = p.Clone().(*Permission).Permissions
```
taratatach: Could we have different permission sets in different permission docs for the same sharing?

model/permission/permissions.go
```diff
@@ -247,6 +248,143 @@ func GetForShareInteract(db prefixer.Prefixer, sharingID string) (*Permission, e
...
+		SourceID: consts.Sharings + "/" + sharingID,
+		Codes:    make(map[string]string),
+	}
+	hasUsablePermission := false
+	hasCanonical := false
+	duplicates := make([]*Permission, 0, len(perms))
+	for i := range perms {
+		p := &perms[i]
+		if p.Expired() {
+			continue
```
taratatach: By continuing here we won't add `p` to the list of duplicates and it won't be deleted. Could that be an issue later when requesting permissions?

model/permission/permissions.go
```diff
@@ -318,15 +366,17 @@ func repairShareInteractPermissions(db prefixer.Prefixer, sharingID string) (*Pe
 	duplicates := make([]*Permission, 0, len(perms))
 	for i := range perms {
 		p := &perms[i]
+		if p.ID() == canonicalID {
+			hasCanonical = true
+			// If the canonical doc is expired, update that doc ID but rebuild its content from non-expired docs.
```
taratatach: Are we really updating the ID? 🤔

model/permission/permissions.go
```diff
@@ -743,21 +793,75 @@ func CreateSharePreviewSet(db prefixer.Prefixer, sharingID string, codes, shortc
...
 		Codes:       codes,
 		SourceID:    consts.Sharings + "/" + sharingID,
 		Metadata:    subdoc.Metadata,
 	}
-	err := couchdb.CreateDoc(db, doc)
-	if err != nil {
+
+	if doc.Codes == nil {
+		doc.Codes = make(map[string]string)
+	}
```
taratatach: I don't really get why we're doing this.

model/permission/permissions.go
```diff
@@ -743,21 +793,75 @@ func CreateSharePreviewSet(db prefixer.Prefixer, sharingID string, codes, shortc
...
+			}
+			return merged, nil
+		}
+
+		err = couchdb.CreateNamedDoc(db, doc)
+		if err == nil {
+			return doc, nil
+		}
+		return nil, err
+	})
```
taratatach, line 875: We're already fixing docs when fetching the permission doc so do we really need to do it at creation time as well?
I would expect a "fetch or create" situation here instead.

model/permission/permissions.go
shepilov: As far as all these docs were created with the bug, and were created at the same time, so strategy "first non-nil expiration win" is just minimal rational, do you think we need to add funciton to calculate ExpiresAt ?

model/permission/permissions.go
shepilov: in theory - yes if duplicate docs were created before/after a sharing rules update, so  is also arbitrary assuming that we didn't modify anything since the bug, for our specific race it should be the same

model/permission/permissions.go
shepilov: mmm, why do you think so? The merged doc is already initialized with canonicalID

model/permission/permissions.go
shepilov: yes, it's too defensive, if is nil Reading/ranging it is fine, but writing into it panics, so we can remove it

model/permission/permissions.go
shepilov: it was also "just in case" but I think it would be cleaner to remove it

model/permission/permissions.go
taratatach: Hum, I forgot about the "at the same time" part.
It seems OK to not care about this then.

model/permission/permissions.go
taratatach: OK seems fair.

model/permission/permissions.go
taratatach: I'm just reading the comment that says we update the doc ID. I was surprised since I didn't see where we were editing the ID.

taratatach (approved):

PR #4763, opened 2026-05-20
fix: add a file "shared drive" metadata to rules

shepilov:
## Context
 https://github.com/linagora/twake-drive/issues/3850

  Make rules[0].mime for file-root shared drives a derived value from the VFS file as far as UI use this field to check that the shared drive is a file and doesn't use specific field 

  ## Changes:
  - Add the root file MIME to shared-drive rules.
  - Add `drive_root_type` to recipient shortcut target metadata.
  - Add shortcut target MIME for file-root shared drives.
  - Canonicalize `drive_root_type` from the real VFS root instead of trusting legacy input.
  - Make invalid drive root inference fail instead of silently creating incomplete metadata.

  ## Assumptions
  - The VFS file document MIME is the trusted source of truth.
  - Client-provided rule MIME should never be preserved for file-root shared drive creation.
  - Directory-root shared drives should still not get a synthetic rule MIME.

### DIFF
--- a/model/sharing/sharing.go
+++ b/model/sharing/sharing.go
@@ -459,7 +459,9 @@ func (s *Sharing) Create(inst *instance.Instance) (*permission.Permission, error
 		if !IsValidDriveRootType(s.DriveRootType) {
 			return nil, ErrInvalidRule
 		}
-		s.inferDriveRootType(inst)
+		if err := s.inferDriveRootType(inst); err != nil {
+			return nil, err
+		}
 	}
 	if !s.Drive && len(s.Members) < 2 {
 		return nil, ErrNoRecipients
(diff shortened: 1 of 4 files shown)


model/sharing/sharing.go
```diff
@@ -481,25 +483,28 @@ func (s *Sharing) Create(inst *instance.Instance) (*permission.Permission, error
...
-	}
-	rootID, err := s.DriveRootID()
-	if err != nil {
-		return
+func (s *Sharing) inferDriveRootType(inst *instance.Instance) error {
+	rule := s.FirstFilesRule()
+	if rule == nil || rule.Selector == couchdb.SelectorReferencedBy || len(rule.Values) == 0 {
+		return ErrDriveRootNotFound
 	}
+	rootID := rule.Values[0]
```
taratatach, line 491: Too bad we have to duplicate this code to get access to the rule :/.
I find `rule.Values[0]` not to be clear it represents the drive ID and was happy to see it used only in a helper.

But I don't have a better version to offer at the moment.

model/sharing/sharing.go
```diff
@@ -518,6 +523,10 @@ func CreateDrive(inst *instance.Instance, rootID, description, appSlug string) (
 		rootType = DriveRootTypeFile
 		rootName = file.DocName
 	}
+	rootMime := ""
+	if file != nil {
+		rootMime = file.Mime
+	}
```
taratatach: nit: the declaration of `rootMime` could be moved with the declarations of `rootType` and `rootName` and the re-assignment of `rootMime` moved into the file block above.

taratatach (approved):

PR #4767, opened 2026-05-26
fear: allow token exchange CORS from SaaS admin panels

shepilov:
## Summary
  - Add `public_domain_suffix` context config with default-context fallback.
  - Allow `/auth/token_exchange` CORS from `https://<owner>-admin.<public_domain_suffix>`.
  - Resolve `<owner>.<public_domain_suffix>` and require it to belong to the same organization as the target instance.
  - Keep the existing `*.org_domain` CORS behavior unchanged.

### DIFF
--- a/pkg/config/config/config.go
+++ b/pkg/config/config/config.go
@@ -588,6 +588,31 @@ func GetPublicOIDCContext(contextName string) string {
 	return ""
 }
 
+// GetPublicDomainSuffix returns the public domain suffix from the specific context
+// or falls back to the default context if not set.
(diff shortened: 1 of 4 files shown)


zatteo:
Will it work for https://<**admin not only owner**>-admin.<public_domain_suffix> ?

shepilov:
> Will it work for https://<**admin not only owner**>-admin.<public_domain_suffix> ?

yep, it's checking right now only "the same org", we don't store org permissions in stack

zatteo (approved):

PR #4768, opened 2026-05-26
feat: forward LLM override for custom assistant providers

paultranvan:
When a chat conversation is bound to an assistant linked to a non-OpenRAG provider (OpenAI, Mistral, …), resolve the assistant's account and forward the model name, API key, and base URL to OpenRAG as `metadata.llm_override` so it can proxy completions to the external provider

### DIFF
--- a/model/rag/chat.go
+++ b/model/rag/chat.go
@@ -113,6 +113,33 @@ type Source struct {
 	Snippet string `json:"snippet,omitempty"`
 }
 
+type chatAssistant struct {
+	DocID         string                 `json:"_id,omitempty"`
+	DocRev        string                 `json:"_rev,omitempty"`
+	Relationships chatAssistantRelations `json:"relationships,omitempty"`
+}
+
+type chatAssistantRelations struct {
+	Provider struct {
+		Data struct {
+			// For LLM accounts, the assistant app uses the "relationships"
+			// (not "referenced_by") format, so the fields are "_id"/"_type".
+			ID       string `json:"_id"`
+			Type     string `json:"_type"`
+			Metadata struct {
+				ProviderID string `json:"providerId"`
+			} `json:"metadata"`
+		} `json:"data"`
+	} `json:"provider"`
+}
+
+func (a *chatAssistant) ID() string         { return a.DocID }
+func (a *chatAssistant) Rev() string        { return a.DocRev }
+func (a *chatAssistant) DocType() string    { return consts.ChatAssistants }
+func (a *chatAssistant) SetID(id string)    { a.DocID = id }
+func (a *chatAssistant) SetRev(rev string)  { a.DocRev = rev }
+func (a *chatAssistant) Clone() couchdb.Doc { c := *a; return &c }
+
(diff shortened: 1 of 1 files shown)


model/rag/chat.go
```diff
@@ -252,6 +252,72 @@ func getSources(event map[string]interface{}) ([]Source, error) {
...
+
+	override := map[string]interface{}{}
+	if auth, ok := accountDoc.M["auth"].(map[string]interface{}); ok {
+		// The account's "login" field stores the LLM model name (e.g.
+		// "gpt-4o"); "password" stores the API key. This repurposing
+		// matches the io.cozy.accounts schema used by the assistant app
+		if model, _ := auth["login"].(string); model != "" {
+			override["model"] = model
+		}
+		if apiKey, _ := auth["password"].(string); apiKey != "" {
```
shepilov: two questions:
 - it's supposed to be encrypted, isn't it?
 - Do we return the api_key back to the client? why?

model/rag/chat.go
```diff
@@ -252,6 +252,72 @@ func getSources(event map[string]interface{}) ([]Source, error) {
 	return sources, nil
 }
 
+// buildLLMOverride returns the `metadata.llm_override“ map forwarded to
+// OpenRAG when the conversation is bound to an assistant that uses an
+// external provider (OpenAI, Mistral, …). It returns nil to leave the
+// stack's default RAG configuration in place. Either no assistant is
+// attached, the provider is the default "openrag", or the linked account
+// could not be resolved.
+func buildLLMOverride(inst *instance.Instance, logger logger.Logger, chat *ChatConversation) map[string]interface{} {
```
shepilov: logger is never used

model/rag/chat.go
```diff
@@ -252,6 +252,72 @@ func getSources(event map[string]interface{}) ([]Source, error) {
...
+	rel, ok := chat.Rels["assistant"]
+	if !ok {
+		return nil
+	}
+	relData, _ := rel.Data.(map[string]interface{})
+	assistantID, _ := relData["_id"].(string)
```
shepilov, line 295: also, is it possible to avoid such kind of reads with a typed doc?

model/rag/chat.go
paultranvan: > it's supposed to be encrypted, isn't it?

yes, and I removed the decryption in a previous test, my very bad . It's back now.

> Do we return the api_key back to the client? why? 

no, we transmit the api_key to openRAG. It goes client -> cozy-stack -> openRAG

model/rag/chat.go
paultranvan: Good catch, removed

model/rag/chat.go
paultranvan: Like this ? [c637fde](https://github.com/cozy/cozy-stack/pull/4768/commits/c637fdef4931573b9e31442fca7c5db4e0559aac)

shepilov: yes, like this

shepilov (approved):

PR #4772, opened 2026-05-28
fix: revoke file-root drive sharing when root file is trashed

shepilov:
## Summary

  Fix revocation for file-root shared drives when the owner deletes the root file via `DELETE /files/:id`.

  Directory-root shared drives already revoked correctly because directory updates checked the transition from normal path to trash path. 

  ## Changes

  - Detect file trash transition using both old and new file state/path.
  - Call the existing sharing revocation hook when a referenced file enters trash.
  - Add/adjust shared-drive tests for file-root deletion behavior.

### DIFF
--- a/model/vfs/couchdb_indexer.go
+++ b/model/vfs/couchdb_indexer.go
@@ -223,11 +223,7 @@ func (c *couchdbIndexer) UpdateFileDoc(olddoc, newdoc *FileDoc) error {
 		return err
 	}
 
-	fullpath, err := olddoc.Path(c)
-	if err != nil {
-		return err
-	}
-	if strings.HasPrefix(fullpath, TrashDirName) {
+	if !olddoc.Trashed && newdoc.Trashed {
 		c.checkTrashedFileIsShared(newdoc)
 	}
 
--- a/web/sharings/drives_test.go
+++ b/web/sharings/drives_test.go
@@ -3064,12 +3064,14 @@ func TestFileRootSharedDriveMutationRoutes(t *testing.T) {
 			Object().
 			Path("$.data.attributes.trashed").Boolean().True()
 
+		require.Eventually(t, func() bool {
+			s, err := sharing.FindSharing(env.betty, sharingID)
+			return err == nil && !s.Active
+		}, 5*time.Second, 50*time.Millisecond)
+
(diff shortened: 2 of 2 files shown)


model/vfs/couchdb_indexer.go
```diff
@@ -223,11 +223,18 @@ func (c *couchdbIndexer) UpdateFileDoc(olddoc, newdoc *FileDoc) error {
...
 		return err
 	}
-	if strings.HasPrefix(fullpath, TrashDirName) {
+	newpath, err := newdoc.Path(c)
+	if err != nil {
+		return err
+	}
+
+	oldTrashed := olddoc.Trashed && strings.HasPrefix(oldpath, TrashDirName)
+	newTrashed := newdoc.Trashed && strings.HasPrefix(newpath, TrashDirName)
```
taratatach: Why check if the path is prefixed by the trash dir name when we can just compare the value of `Trashed`?

shepilov: no idea, but it was done like this before

shepilov: just to be "totally" sure that a file was deleted

shepilov: Actually I found this comment that it can be inconsistent:

shepilov: // If there is only the trashed attribute or the parent in trash, but not
// both, we can try again to move the file to the trash to fix the
// inconsistency.

taratatach: OK. Then we should probably have something like this to be consistent with the current situation:

```suggestion
	newTrashed := newdoc.Trashed || strings.HasPrefix(newpath, TrashDirName)
```

shepilov: nope, it's not "or", it's 100% "and" or only strings.HasPrefix(newpath, TrashDirName), as I understood `Trashed` it's just a denormaliztion

taratatach: Like the comment says, we can have inconsistencies with trashed files that have the `Trashed` attribute but are not in a trashed directory or vice-versa.
Using "or" instead of "and" would make this code resilient to such inconsistencies.

shepilov: I'll remove "Trashed" altogether

taratatach (approved):

PR #4774, opened 2026-05-29
fix: delete owner drive sharing after root deletion

shepilov:
## Summary
- delete the owner-side drive sharing document after automatic shared-drive root deletion revokes recipients
- keep non-drive and recipient-side revocation behavior unchanged
- assert owner sharing doc deletion for file-root and directory-root shared drives

### DIFF
--- a/model/sharing/revoke_trashed.go
+++ b/model/sharing/revoke_trashed.go
@@ -7,6 +7,7 @@ import (
 	"github.com/cozy/cozy-stack/model/instance"
 	"github.com/cozy/cozy-stack/model/vfs"
 	"github.com/cozy/cozy-stack/pkg/consts"
+	"github.com/cozy/cozy-stack/pkg/couchdb"
 	"github.com/cozy/cozy-stack/pkg/prefixer"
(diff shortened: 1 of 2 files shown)


model/sharing/revoke_trashed.go
```diff
@@ -38,6 +39,12 @@ func revokeTrashed(db prefixer.Prefixer, sharingID string) {
 	log.Infof("revokeTrashed called for sharing %s", sharingID)
 	if s.Owner {
 		err = s.Revoke(inst)
+		if err == nil && s.Drive {
+			err = couchdb.DeleteDoc(inst, s)
+			if couchdb.IsNotFoundError(err) {
+				err = nil
+			}
+		}
```
taratatach, line 47: Why remove the sharing only if it's a Drive? Are there good reasons to keep revoked sharings around?

taratatach: The idea is to avoid touching legacy sharings if possible.

taratatach (approved):

PR #4775, opened 2026-05-29
fix: delete drive sharing after last member removal

shepilov:
## Summary
- delete owner-side drive sharing documents once all recipients are revoked
- keep drive sharing documents when at least one recipient remains
- cover both last-recipient removal and bulk recipient revocation

### DIFF
--- a/model/sharing/group.go
+++ b/model/sharing/group.go
@@ -106,6 +106,16 @@ func (s *Sharing) RevokeGroup(inst *instance.Instance, index int) error {
 	}
 
 	s.Groups[index].Revoked = true
+	if s.Drive {
+		if errm != nil {
+			if err := couchdb.UpdateDoc(inst, s); err != nil {
+				errm = multierror.Append(errm, err)
+			}
+			return errm
+		}
+		return s.NoMoreRecipient(inst)
+	}
+
 	if err := couchdb.UpdateDoc(inst, s); err != nil {
 		errm = multierror.Append(errm, err)
 	}
--- a/model/sharing/sharing.go
+++ b/model/sharing/sharing.go
@@ -636,6 +636,9 @@ func (s *Sharing) Revoke(inst *instance.Instance) error {
 	if err := couchdb.UpdateDoc(inst, s); err != nil {
 		return err
 	}
+	if s.Drive && errm == nil {
+		return s.deleteRevokedDriveSharing(inst)
+	}
 	return errm
 }
 
@@ -892,7 +895,23 @@ func (s *Sharing) RevokeRecipientByNotification(inst *instance.Instance, m *Memb
 // NoMoreRecipient cleans up the sharing if there is no more active recipient
 func (s *Sharing) NoMoreRecipient(inst *instance.Instance) error {
 	if s.Drive {
-		return couchdb.UpdateDoc(inst, s)
+		if s.hasRemainingRecipients() {
+			return couchdb.UpdateDoc(inst, s)
+		}
+		if err := s.RemoveTriggers(inst); err != nil {
+			return err
+		}
+		if err := RemoveSharedRefs(inst, s.SID); err != nil {
+			return err
+		}
+		if err := s.RevokeInteractPermissions(inst); err != nil {
+			return err
+		}
+		s.Active = false
+		if err := couchdb.UpdateDoc(inst, s); err != nil {
+			return err
+		}
+		return s.deleteRevokedDriveSharing(inst)
 	}
 	for _, m := range s.Members {
 		if m.Status == MemberStatusReady {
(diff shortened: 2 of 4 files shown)


model/sharing/sharing.go
taratatach, line 918: I know it's not part of this PR but why are we updating the sharing doc once for each member while we haven't made any change between updates?

model/sharing/sharing.go
```diff
@@ -892,7 +895,23 @@ func (s *Sharing) RevokeRecipientByNotification(inst *instance.Instance, m *Memb
...
+			return err
+		}
+		if err := s.RevokeInteractPermissions(inst); err != nil {
+			return err
+		}
+		s.Active = false
+		if err := couchdb.UpdateDoc(inst, s); err != nil {
+			return err
+		}
+		return s.deleteRevokedDriveSharing(inst)
```
taratatach, line 914: It looks like a lot of this could be shared with the non-drive following "block".

model/sharing/sharing.go
```diff
@@ -892,7 +895,23 @@ func (s *Sharing) RevokeRecipientByNotification(inst *instance.Instance, m *Memb
 // NoMoreRecipient cleans up the sharing if there is no more active recipient
 func (s *Sharing) NoMoreRecipient(inst *instance.Instance) error {
 	if s.Drive {
-		return couchdb.UpdateDoc(inst, s)
+		if s.hasRemainingRecipients() {
+			return couchdb.UpdateDoc(inst, s)
```
taratatach, line 899: nit: it feels super weird to update a document here while changes were made outside this function. We don't know what we're saving and what for.
Like, why are we updating the document only if it has remaining recipients now?

model/sharing/sharing.go
```diff
@@ -909,6 +928,26 @@ func (s *Sharing) NoMoreRecipient(inst *instance.Instance) error {
...
 }
 
+func (s *Sharing) hasRemainingRecipients() bool {
+	for i, m := range s.Members {
+		if i > 0 && m.Status != MemberStatusRevoked {
+			return true
+		}
+	}
+	return false
+}
```
taratatach, line 943: Should this function check also for the presence of groups?

model/sharing/group.go
```diff
@@ -105,6 +105,10 @@ func (s *Sharing) RevokeGroup(inst *instance.Instance, index int) error {
 		}
 	}
 
+	if s.Drive && !s.hasRemainingRecipients() {
+		return errm
+	}
+
```
taratatach: Why does this only apply to shared drives and not all kinds of sharings? If there aren't any members left I see no reason to mark them as revoked.

model/sharing/sharing.go
shepilov: i had the same question, because we changed this doc before, it was disigned like we are updating this doc here, also cofusing for me, this function is supposed to be called "finalize recipient revocation"

model/sharing/sharing.go
shepilov: I don't want to touch non-drive block, and keep this branches in parallel where possible, especially here

model/sharing/group.go
shepilov: - because I don't want to touch legacy sharing
- for the "sync" sharing we detact active recipient difeerently, for this case everything that is not active is considered as "revoked" and can be easily removed what is not the case for drive

model/sharing/sharing.go
shepilov: i'll add it, but for this groups it's not super relevant, as far as we won't have groups for b2c and for b2b nobody will have this "sharings" tab with zero member sharings

model/sharing/sharing.go
shepilov: nope, we are returning  on this step, for another case update is a little bit later

model/sharing/sharing.go
taratatach: Won't these sharings be displayed in the admin panel for b2b?

model/sharing/sharing.go
taratatach: I don't understand your comment, sorry. I'll try rephrasing my question: why do we update the sharing only if it has remaining recipients when we were updating it in all cases?

The fact that we're saving changes made to `s` outside `NoMoreRecipient` is making it harder to understand.

model/sharing/sharing.go
shepilov: I'm not sure how it's done in b2b admin panel, If I were doing it, it would've been just a folder structure, but anyway, it's deleting for the groups as well now

model/sharing/group.go
```diff
@@ -106,6 +106,16 @@ func (s *Sharing) RevokeGroup(inst *instance.Instance, index int) error {
 	}
 
 	s.Groups[index].Revoked = true
+	if s.Drive {
+		if errm != nil {
+			if err := couchdb.UpdateDoc(inst, s); err != nil {
```
taratatach, line 111: Since we're already updating `s` in `NoMoreRecipient`, do we really need this call here?

model/sharing/group.go
```diff
@@ -106,6 +106,16 @@ func (s *Sharing) RevokeGroup(inst *instance.Instance, index int) error {
...
+				errm = multierror.Append(errm, err)
+			}
+			return errm
+		}
+		return s.NoMoreRecipient(inst)
+	}
```
taratatach, line 117: Since we have branches in `NoMoreRecipient` to handle both Drive sharings and non-Drive sharings, I feel like we should be calling `NoMoreRecipient` all the time here 🤔.

We could ditch the call to `UpdateDoc` below then.

model/sharing/group.go
shepilov: but it's skipped when err != nil

model/sharing/group.go
shepilov: We could've if we wanted to delete the share even if one of the recipients' revocations was faild.

If a member revocation failed, persist the revoked group state but skip final cleanup so we do not delete the sharing after a partial failure

model/sharing/sharing.go
taratatach: Actually, this is an early return to prevent cleaning up sharings that still have recipients.

taratatach (approved):

model/sharing/sharing.go
taratatach: Actually we're calling `UpdateDoc` only once, for the first member who's status is `MemberStatusReady`, since we return.

PR #4776, opened 2026-05-29
feat: redirect shared drive invites to target

shepilov:
## Context

Shared-drive invitation links could leave users in the generic shared-drives folder. If auto-accept had already run, the same authorization URL could also be rejected as an already-active sharing instead of opening the target.

## Summary

- Redirect shared-drive invite and shortcut opens to the specific shared drive root instead of the generic shared-drives folder.
- Use existing Cozy Drive routes for folder-root and file-root shared drives.
- Allow the sharing authorization route to act as an open link after auto-accept.

### DIFF
--- a/model/sharing/sharing.go
+++ b/model/sharing/sharing.go
@@ -156,6 +156,36 @@ func (s *Sharing) DriveRootID() (string, error) {
 	return rule.Values[0], nil
 }
 
+// DriveTargetURL returns the Drive URL that opens the shared drive root.
+func (s *Sharing) DriveTargetURL(inst *instance.Instance) *url.URL {
+	u := inst.SubDomain(consts.DriveSlug)
+	u.RawQuery = url.Values{"sharing": {s.SID}}.Encode()
+
+	rootID, err := s.DriveRootID()
+	if err != nil {
+		u.Fragment = "/folder/" + consts.SharedDrivesDirID
+		return u
+	}
+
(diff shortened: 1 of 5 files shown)


model/sharing/sharing.go
```diff
@@ -156,6 +156,36 @@ func (s *Sharing) DriveRootID() (string, error) {
...
+
+// AuthorizeSharingURL returns the recipient-side URL that opens the sharing
+// authorization flow.
+func (s *Sharing) AuthorizeSharingURL(inst *instance.Instance, state string) string {
+	return inst.PageURL("/auth/authorize/sharing", url.Values{
+		"sharing_id": {s.SID},
+		"state":      {state},
+	})
+}
+
```
taratatach: nit: shouldn't that be in the `invite.go` file?

web/auth/oauth.go
```diff
@@ -556,9 +562,9 @@ func (a *AuthorizeHTTPHandler) authorizeSharingForm(c echo.Context) error {
 			}
 		}
 		u := instance.SubDomain(consts.DriveSlug)
-		u.RawQuery = "sharing=" + s.SID
+		u.RawQuery = url.Values{"sharing": {s.SID}}.Encode()
```
taratatach: This could be moved in the following `else` block as this attribute is already set by `DriveTargetURL()`

taratatach (approved):

taratatach (approved):

PR #4780, opened 2026-06-02
fix: stale sharing shortcut counter after recipient removal

shepilov:
## Summary

  Fixes the Drive sharing counter after a recipient is removed from a pending shared drive.

  A recipient can have multiple shortcut files for the same sharing: for example, one pending `new` shortcut and another shortcut created when opening the sharing authorization flow. Previously, revocation only removed the shortcut
  pointed to by `ShortcutID`, leaving older referenced shortcuts counted by `/sharings/news`.

  ## Changes

  - Clean up the shortcut referenced by `ShortcutID`.
  - Also remove remaining shortcut files linked to the same sharing through `referenced_by`.
  - Keep referenced cleanup limited to shortcut files.

### DIFF
--- a/model/sharing/invitation.go
+++ b/model/sharing/invitation.go
@@ -353,24 +353,7 @@ func (s *Sharing) CreateDriveShortcut(inst *instance.Instance, seen bool) error 
 		Type: consts.Sharings,
 	})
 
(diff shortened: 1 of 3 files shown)


taratatach (approved):
I wonder if we can still have multiple shortcuts for the same sharing now that we reuse any existing shortcut.

PR #4786, opened 2026-06-04
fix: stabilize flaky unit tests in ci

doubleface:
## Summary

- Drop the redundant sharings database teardown in TestUpdateSelfMemberInstance that occasionally failed in CI with a CouchDB file_exists error.
- Re-enqueue the auto-accept job in TestSharedDriveGroupDynamicMembership after faking the owner URL so the sharing is accepted via the test server URL even when the original auto-accept job already ran with the unresolvable domain URL.


shepilov (changes requested):
It should be replace with EnsureDBExist or require.Eventually on db create

also we should NOT modify the auth flow and reenqueue the auto-accept job in test, in this case test no longer proves what it claims

PR #4787, opened 2026-06-04
fix: Automated partition creation

ewan102:
Function added to the chat.go file that creates a partition on RAG whenever there is no files (and partition) on RAG.


model/rag/chat.go
```diff
@@ -387,6 +387,16 @@ func Query(inst *instance.Instance, logger logger.Logger, query QueryMessage) er
 	if err != nil {
 		return err
 	}
+	if res.StatusCode == http.StatusNotFound {
+		res.Body.Close()
+		if partErr := createPartition(inst); partErr != nil {
```
paultranvan: it seems the `createPartition` does pretty much the same thing than `callRAGQuery`, i.e. requesting the RAG. I feel like we could remove the createPartition func and simplify the whole code by doing just this: 

```
  if res.StatusCode == http.StatusNotFound {
      // Partition does not exist, try create it
      partRes, partErr := CallRAGQuery(inst, nil, fmt.Sprintf("/partition/%s", inst.Domain), echo.MIMEApplicationJSON)
      if partErr == nil || partRes.StatusCode >= 400 {
              logger.Warnf("Failed to create RAG partition: %s", partErr)
          }
          partRes.Body.Close()
      }
      res, err = CallRAGQuery(inst, body, "v1/chat/completions", echo.MIMEApplicationJSON)
      if err != nil {
          return err
      }
  }
```

model/rag/chat.go
```diff
@@ -387,6 +387,16 @@ func Query(inst *instance.Instance, logger logger.Logger, query QueryMessage) er
 	if err != nil {
 		return err
 	}
+	if res.StatusCode == http.StatusNotFound {
```
shepilov, line 390: why do we think that in general 404 is missing partition?

How did we created partitions before? Just uploding files? And if yes, probably we are hiding error of the partion creation here or we can just skip it and do nothing

shepilov, line 390: Do we have also error payload to check that it's really partion missin and not something elease

model/rag/chat.go
```diff
@@ -387,6 +387,16 @@ func Query(inst *instance.Instance, logger logger.Logger, query QueryMessage) er
...
+		res.Body.Close()
```
shepilov, line 391: log here please status and response body to be able to understand why we are creting partition at this step

model/rag/chat.go
shepilov: what will happen if we have two conccurrent chat queries that create partintion at the same time? do we Have 409 for one of the queries?

model/rag/chat.go
```diff
@@ -561,6 +571,32 @@ func handleNonStreamResponse(inst *instance.Instance, msg ChatMessage, body io.R
...
+	if err != nil {
+		return err
+	}
+	req.Header.Add(echo.HeaderAuthorization, "Bearer "+ragServer.APIKey)
+	res, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return err
+	}
+	defer res.Body.Close()
+	if res.StatusCode >= 400 {
```
shepilov: http.StatusBadRequest for 400

shepilov: and also do we consider 302 for example as partion created?

ewan102:
Hello, I took into account the points you made and tried and new version with the advice of Paul.  (Yes, we have a 409 for one of the queries and you are right 404 might be something else than missing partition so a GET partition was added in case a 404 occurs)

model/rag/chat.go
paultranvan: Even better: when we have a 404, let's query the rag to check whether or not the partition exist. 
This way, we do not rely on fragile error payload check

model/rag/chat.go
paultranvan: yes, should be 409

model/rag/chat.go
```diff
@@ -387,6 +387,42 @@ func Query(inst *instance.Instance, logger logger.Logger, query QueryMessage) er
...
+		if err != nil {
+			return err
+		}
+		u.Path = fmt.Sprintf("/partition/%s", inst.Domain)
+		checkReq, err := http.NewRequest(http.MethodGet, u.String(), nil)
+		if err != nil {
+			return err
+		}
+		checkReq.Header.Add(echo.HeaderAuthorization, "Bearer "+ragServer.APIKey)
+		checkRes, err := http.DefaultClient.Do(checkReq)
```
paultranvan: all this part looks very close to what `CallRAGQuery` does, I think the only difference is the GET vs POST ?
What I suggest is to refactor a bit CallRAGQuery to accept a `method` parameter:
```
func CallRAGQuery(inst *instance.Instance, method string, payload []byte, path string, contentType string) (*http.Response, error) {
  ...
  req, err := http.NewRequest(method, u.String(), bytes.NewReader(payload))
  ...

```
And then you just call : 
`CallRAGQuery(inst, http.MethodGet, nil, fmt.Sprintf("/partition/%s", inst.Domain), echo.MIMEApplicationJSON)`

And of course adapt all calls to `CallRAGQuery`

ewan102:
Thank you for your feedback, I just modified Callragquery implementing what you explained

shepilov (approved):

model/rag/chat.go
ewan102: I forgot to add this before, I just did

paultranvan (approved):
Thank you!

shepilov:
@paultranvan do you still need this PR to be merged?

PR #4788, opened 2026-06-05
fix: stale shared drive references

shepilov:
## Summary
- remove stale root `referenced_by` entries when sharings are fully revoked
- clean both directory-root and file-root shared drive references
- add regressions proving roots can be shared again after final recipient removal

## Root cause
Deleting a drive sharing after the last recipient was removed deleted the `io.cozy.sharings` document, but did not remove its id from the owner root file or folder `referenced_by`. That stale reference made later `POST /sharings/drives` fail with 409.

### DIFF
--- a/model/sharing/files.go
+++ b/model/sharing/files.go
@@ -357,6 +357,58 @@ func (s *Sharing) AddReferenceForSharing(inst *instance.Instance, rule *Rule) er
 	return fs.UpdateFileDoc(olddoc, file)
 }
 
+// RemoveReferenceForSharing removes the reference to the sharing from its root
+// file or folder. It is a no-op when the root is already gone.
+func (s *Sharing) RemoveReferenceForSharing(inst *instance.Instance, rule *Rule) error {
+	if rule == nil || rule.Selector == couchdb.SelectorReferencedBy || len(rule.Values) == 0 {
+		return nil
+	}
+
+	indexer := vfs.NewCouchdbIndexer(inst)
+	parts := strings.Split(rule.Values[0], "/")
+	dir, file, err := indexer.DirOrFileByID(parts[len(parts)-1])
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) || couchdb.IsNotFoundError(err) {
+			return nil
+		}
+		inst.Logger().WithNamespace("sharing").
+			Warnf("RemoveReferenceForSharing failed to find root: %s", err)
+		return err
+	}
+	if dir == nil && file == nil {
+		return nil
+	}
+
+	ref := couchdb.DocReference{
+		ID:   s.SID,
+		Type: consts.Sharings,
+	}
+
+	if dir != nil {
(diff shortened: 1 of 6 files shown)


model/sharing/files.go
```diff
@@ -357,6 +357,58 @@ func (s *Sharing) AddReferenceForSharing(inst *instance.Instance, rule *Rule) er
...
 }
 
+// RemoveReferenceForSharing removes the reference to the sharing from its root
+// file or folder. It is a no-op when the root is already gone.
+func (s *Sharing) RemoveReferenceForSharing(inst *instance.Instance, rule *Rule) error {
+	if rule == nil || rule.Selector == couchdb.SelectorReferencedBy || len(rule.Values) == 0 {
+		return nil
+	}
+
+	indexer := vfs.NewCouchdbIndexer(inst)
```
taratatach, line 355: We usually use `fs := inst.VFS()` for this purpose. It returns the instance's indexer rather than creating a new one.

model/sharing/files.go
```diff
@@ -357,6 +357,58 @@ func (s *Sharing) AddReferenceForSharing(inst *instance.Instance, rule *Rule) er
...
+		return err
+	}
+	if dir == nil && file == nil {
+		return nil
+	}
+
+	ref := couchdb.DocReference{
+		ID:   s.SID,
+		Type: consts.Sharings,
+	}
```
taratatach: nit: I think it would be better to have a method on the sharing to return this reference to be sure it's always the same we use everywhere.

model/sharing/files.go
shepilov: yes, i did at first with VFS(),  in the "trash" handler, there is no VFS, only indexer, and not only in tests in prodction code as well

model/sharing/files.go
```diff
@@ -319,15 +313,12 @@ func (s *Sharing) AddReferenceForSharing(inst *instance.Instance, rule *Rule) er
 
 	if dir != nil {
 		for _, ref := range dir.ReferencedBy {
-			if ref.Type == consts.Sharings && ref.ID == s.SID {
+			if ref == s.DocReference() {
```
taratatach, line 316: Is this comparing values field by field in Go? Because if it does a reference comparison I fear we won't have the expected result.

model/sharing/files.go
taratatach: I don't understand what you mean. Since we have an instance, we have access to `inst.VFS()` 🤔

model/sharing/files.go
shepilov: yes, struct comparison if field by field, and it's a struct and not a pointer

model/sharing/files.go
shepilov: https://github.com/cozy/cozy-stack/actions/runs/27027417809/job/79770825616

in this test we had panic on calling VFS(), revoke_trashed.go creates a lightweight Instance with only prefix

taratatach (approved):

PR #4791, opened 2026-06-10
exchange OIDC token for app token

shepilov:
## Summary

  Adds OIDC app token exchange support to `/auth/token_exchange`.

  This lets a trusted frontend app, like Twake Mail, exchange an external OIDC `id_token` for a local Cozy OAuth client and access/refresh tokens, and initialize cozy-client

  ## Changes

  - Adds `exchange_type` to `/auth/token_exchange`.
    - Missing value keeps the existing admin exchange behavior.
    - `exchange_type: "app"` enables app token exchange.
  - Adds OIDC config for trusted app exchanges:
    - `allow_app_token_exchange`
    - `app_token_exchange`
  - Maps an OIDC token audience, like `twake-mail-web`, to a Cozy linked app, like `registry://mail`.
  - Derives the local OAuth scope from the linked app manifest instead of accepting caller-provided scopes.

### DIFF
--- a/model/oauth/client.go
+++ b/model/oauth/client.go
@@ -393,6 +393,11 @@ const (
 	// sharings by example, as a token is created just after the client
 	// creation.
 	NotPending CreateOptions = iota + 1
+	// SoftwareIDPrevalidated tells Create to skip the registry probe in
+	// CheckSoftwareID. The caller is responsible for proving that the
+	// registry://<slug> SoftwareID is trusted (e.g. by reading the locally
+	// installed manifest).
+	SoftwareIDPrevalidated
 )
 
 func hasOptions(needle CreateOptions, haystack []CreateOptions) bool {
@@ -456,8 +461,10 @@ func (c *Client) Create(i *instance.Instance, opts ...CreateOptions) *ClientRegi
 	if err := c.checkMandatoryFields(i); err != nil {
 		return err
 	}
-	if err := c.CheckSoftwareID(i); err != nil {
-		return err
+	if !hasOptions(SoftwareIDPrevalidated, opts) {
+		if err := c.CheckSoftwareID(i); err != nil {
+			return err
+		}
 	}
 
 	if err := c.ensureClientNameUnicity(i); err != nil {
--- a/model/oidc/provider/instance.go
+++ b/model/oidc/provider/instance.go
@@ -1,0 +1,144 @@
+package provider
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/cozy/cozy-stack/model/instance"
+	"github.com/cozy/cozy-stack/model/instance/lifecycle"
+	"github.com/cozy/cozy-stack/pkg/consts"
+	"github.com/cozy/cozy-stack/pkg/couchdb"
+	"github.com/cozy/cozy-stack/pkg/couchdb/mango"
+	"github.com/cozy/cozy-stack/pkg/logger"
+	"github.com/cozy/cozy-stack/pkg/prefixer"
+	"github.com/golang-jwt/jwt/v5"
+)
+
+// ErrInstanceAuthenticationFailed is returned when OIDC claims cannot be
+// interpreted (missing/malformed sub or user info field). It signals an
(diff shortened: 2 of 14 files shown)


pkg/config/config/config.go
```diff
@@ -553,6 +566,47 @@ func GetOIDC(contextName string) (map[string]interface{}, bool) {
...
+
+	enabled, _ := oidc["allow_app_token_exchange"].(bool)
+	cfg.Enabled = enabled
+	if !enabled {
+		return cfg, nil
+	}
+
+	rawApps, ok := oidc["app_token_exchange"]
+	if !ok {
+		cfg.Apps = map[string]OIDCAppTokenExchangeAppConfig{}
```
taratatach, line 620: Isn't this already done when declaring `cfg`?

pkg/config/config/config.go
```diff
@@ -553,6 +566,47 @@ func GetOIDC(contextName string) (map[string]interface{}, bool) {
...
+		return cfg, fmt.Errorf("invalid app token exchange configuration: %w", err)
+	}
+
+	for audience, appConfig := range apps {
+		appConfig.Audience = audience
+		appConfig.SoftwareID = strings.TrimSpace(appConfig.SoftwareID)
+		if appConfig.SoftwareID == "" {
+			return cfg, fmt.Errorf("software_id is required for app token exchange audience %q", audience)
+		}
+		if !strings.HasPrefix(appConfig.SoftwareID, "registry://") || strings.TrimPrefix(appConfig.SoftwareID, "registry://") == "" {
```
taratatach, line 635: We could use `oauth.GetLinkedAppSlug(appConfig.SoftwareID) == ""`. It's simpler, reused existing helpers and would help make clear why we enforce this here.

model/oauth/client.go
```diff
@@ -456,8 +461,10 @@ func (c *Client) Create(i *instance.Instance, opts ...CreateOptions) *ClientRegi
 	if err := c.checkMandatoryFields(i); err != nil {
 		return err
 	}
-	if err := c.CheckSoftwareID(i); err != nil {
-		return err
+	if !hasOptions(SoftwareIDPrevalidated, opts) {
+		if err := c.CheckSoftwareID(i); err != nil {
+			return err
+		}
```
taratatach, line 467: Why do we need this? `registry://mail` does not exist?

model/oidc/provider/instance.go
```diff
@@ -0,0 +1,144 @@
...
+	if conf.AllowCustomInstance {
+		sub, ok := claims["sub"].(string)
+		expected := inst.OIDCID
+		if conf.Provider == FranceConnectProvider {
+			expected = inst.FranceConnectID
+		}
+		if !ok || sub == "" {
+			inst.Logger().WithNamespace("oidc").
+				Errorf("Invalid sub claim: %v != %s", claims["sub"], expected)
+			return ErrInstanceAuthenticationFailed
```
taratatach, line 56: nit: it would be nice to have this block next to the `sub` and `ok` var assignment.
I would say, move the assignment right before this block.

model/oauth/client.go
taratatach: OK, I see this is to allow doing the check by reading the locally installed manifest. Shouldn't switch to a local check entirely instead of making requests to the registry? I don't really see a situation where we'd want to allow creating an OAuth client for a linked app that's not installed on the instance.

shepilov: I think yes, but I wanted to double check all the flows that we can have before making this change

pkg/config/config/config.go
shepilov: nope, id doesn't do map allocation

pkg/config/config/config.go
shepilov: config can't import oauth, the only variant it to move it to a separate helper on pkg level

shepilov: Maybe it makes sense, but in a separate pr, too many changes will be for one line extraction, helper in oauth package is used in too many places

taratatach (approved):

PR #4793, opened 2026-06-10
fix: keep sharing news counter for background drive realtime connections

rezk2ll:
## Problem
The recipient websocket `/sharings/drives/:id/realtime` marks the sharing as seen on every connection. The dataproxy indexer connects to every accepted drive sharing within seconds to index it, so on auto-accepted sharings the shortcut flips from new to seen before the user ever sees the sharings counter. The recipient never gets the red badge for a newly shared drive.

## Solution
Background consumers now pass `?background=true` to keep the sharing unseen. The default is unchanged, so the drive folder view (and any deployed client) still marks the sharing as seen when the user actually opens the drive. The dataproxy side is in cozy-libs (cozy-realtime grows a `background` option, the search indexer passes it).

### DIFF
--- a/web/sharings/realtime.go
+++ b/web/sharings/realtime.go
@@ -339,12 +339,17 @@ func Ws(c echo.Context) error {
 		return jsonapi.InternalServerError(err)
 	}
 
-	go func() {
-		if err := markSharingAsSeen(inst, s); err != nil {
-			inst.Logger().WithNamespace("sharing").
-				Warnf("markSharingAsSeen failed: %s", err)
-		}
-	}()
+	// Background consumers (like the dataproxy indexer) watch the drive
+	// without the user looking at it, so they must not consume the "new"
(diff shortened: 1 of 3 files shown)


web/sharings/realtime.go
```diff
@@ -339,12 +339,17 @@ func Ws(c echo.Context) error {
...
-	go func() {
-		if err := markSharingAsSeen(inst, s); err != nil {
-			inst.Logger().WithNamespace("sharing").
-				Warnf("markSharingAsSeen failed: %s", err)
-		}
-	}()
+	// Background consumers (like the dataproxy indexer) watch the drive
+	// without the user looking at it, so they must not consume the "new"
+	// status that feeds the sharings news counter.
+	if c.QueryParam("background") != "true" {
```
shepilov, line 345: strconv.ParseBool() maybe?

shepilov (approved):

PR #4795, opened 2026-06-11
fix: update shared drive target URLs

shepilov:
## Summary

  - Update `DriveTargetURL` to stop adding the `?sharing=<id>` query parameter.
  - Change file-root shared drive redirects to `#/sharings/shareddrive/<sharing_id>/file/<file_id>`.

### DIFF
--- a/model/sharing/sharing.go
+++ b/model/sharing/sharing.go
@@ -168,7 +168,6 @@ func (s *Sharing) DriveRootID() (string, error) {
 // DriveTargetURL returns the Drive URL that opens the shared drive root.
 func (s *Sharing) DriveTargetURL(inst *instance.Instance) *url.URL {
 	u := inst.SubDomain(consts.DriveSlug)
-	u.RawQuery = url.Values{"sharing": {s.SID}}.Encode()
 
 	rootID, err := s.DriveRootID()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


model/sharing/sharing.go
```diff
@@ -179,7 +178,7 @@ func (s *Sharing) DriveTargetURL(inst *instance.Instance) *url.URL {
 	sharingID := url.PathEscape(s.SID)
 	rootID = url.PathEscape(rootID)
 	if s.HasFileDriveRoot() {
-		u.Fragment = "/shareddrive/" + sharingID + "/file/" + rootID
+		u.Fragment = "/sharings/shareddrive/" + sharingID + "/file/" + rootID
```
taratatach, line 181: Why do we add this prefix only for file-root drives? 🤔

model/sharing/sharing.go
```diff
@@ -168,7 +168,6 @@ func (s *Sharing) DriveRootID() (string, error) {
 // DriveTargetURL returns the Drive URL that opens the shared drive root.
 func (s *Sharing) DriveTargetURL(inst *instance.Instance) *url.URL {
 	u := inst.SubDomain(consts.DriveSlug)
-	u.RawQuery = url.Values{"sharing": {s.SID}}.Encode()
```
taratatach, line 171: That was redundant indeed 👍

PR #4797, opened 2026-06-11
feat/resolve OIDC context by email

shepilov:
## Context

  Today, when a user opens a sharing invitation and the stack does not already know their Cozy URL, we show the discovery page. The user has to enter their server address, or use an available OIDC button if one is shown.

  For B2B SaaS users, we already know the invited user email. With this PR, the stack can use that email to find the company/org context and start the right OIDC sharing flow automatically. The expected UX is: user opens the sharing
  link, and if their email domain belongs to a known org with an OIDC context, they are redirected directly to the company SSO instead of filling the discovery form.

  ## Changes

  - Add OIDC context lookup by configured `oidc.domain`.
  - Add automatic sharing discovery redirect to OIDC when the recipient email domain matches a configured OIDC context.
  - Add B2B SaaS support by resolving the recipient email domain against existing instances with the same `org_domain`.
  - For org-domain resolution, use the first matched instance context directly.
  - If the matched org instance has no context, log a warning and fall back to the normal discovery page.

### DIFF
--- a/web/sharings/sharings.go
+++ b/web/sharings/sharings.go
@@ -8,6 +8,7 @@ package sharings
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 	"net/http"
 	"net/url"
 	"os"
@@ -30,11 +31,14 @@ import (
 	"github.com/cozy/cozy-stack/pkg/jsonapi"
 	"github.com/cozy/cozy-stack/pkg/logger"
 	"github.com/cozy/cozy-stack/pkg/safehttp"
+	"github.com/cozy/cozy-stack/pkg/utils"
 	"github.com/cozy/cozy-stack/web/middlewares"
 	weboidc "github.com/cozy/cozy-stack/web/oidc"
 	"github.com/hashicorp/go-multierror"
 	"github.com/labstack/echo/v4"
 )
+
+var errMissingOIDCContextForOrgDomain = errors.New("organization domain instance has no context")
 
 // CreateSharing initializes a new sharing (on the sharer)
 func CreateSharing(c echo.Context) error {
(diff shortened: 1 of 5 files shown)


web/sharings/sharings.go
```diff
@@ -30,12 +31,15 @@ import (
...
 	"github.com/cozy/cozy-stack/pkg/logger"
 	"github.com/cozy/cozy-stack/pkg/safehttp"
+	"github.com/cozy/cozy-stack/pkg/utils"
 	"github.com/cozy/cozy-stack/web/middlewares"
 	weboidc "github.com/cozy/cozy-stack/web/oidc"
 	"github.com/hashicorp/go-multierror"
 	"github.com/labstack/echo/v4"
 )
 
+var errMissingOIDCContextForOrgDomain = errors.New("organization domain instance has no context")
```
taratatach, line 41: nit: We're missing a context entirely here, not just for OIDC.

web/sharings/sharings.go
```diff
@@ -803,12 +807,38 @@ func discoveryStateForMember(s *sharing.Sharing, currentState string, m *sharing
...
+}
+
+func oidcContextForMemberEmail(email string) (string, error) {
+	domain := domainFromEmail(email)
+	if domain == "" {
+		return "", nil
+	}
+	if contextName, err := oidcContextForOrgDomain(domain); err != nil || contextName != "" {
+		return contextName, err
+	}
```
taratatach, line 840: I don't really get why we make this call here. We might end up with a context that has no OIDC configuration.

shepilov: in theory - yes, in practice - no

taratatach (approved):

PR #4799, opened 2026-06-15
fix: update federated sharing email translation

shepilov:


rezk2ll (approved):

assets/locales/en.po
```diff
@@ -1041,7 +1041,7 @@ msgid "Notification Sharing Title Note"
 msgstr "a note"
 
 msgid "Notification Sharing Type Shared Drive"
```
taratatach, line 1043: Maybe it's the right moment to start using "federated sharing"? Because this introduces even more confusion by using "shared drive" in locale keys with a different meaning user-wise.

shepilov: I'm waiting for the last urgent bugfix to refactor it. I don't want to have one nameing in one place and in another another

PR #4801, opened 2026-06-15
feat: allow shared drive writers to manage links

shepilov:
## Summary
- add contexts.<context>.sharing.allow_writers_to_manage_links with a safe default of false
- allow resolved non-read-only shared-drive members to patch or revoke links created by others when the flag is enabled
- preserve owner-or-creator default behavior plus public-token and read-only restrictions

### DIFF
--- a/web/sharings/drives.go
+++ b/web/sharings/drives.go
@@ -16,6 +16,7 @@ import (
 	"github.com/cozy/cozy-stack/model/permission"
 	"github.com/cozy/cozy-stack/model/sharing"
 	"github.com/cozy/cozy-stack/model/vfs"
+	"github.com/cozy/cozy-stack/pkg/config/config"
 	"github.com/cozy/cozy-stack/pkg/consts"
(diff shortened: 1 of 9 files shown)


web/sharings/drives.go
```diff
@@ -951,8 +953,13 @@ func CheckSharedDrivePermissionMutationAuthorization(
 		creatorDomain := existingPerm.Metadata.UpdatedByApps[0].Instance
 		isCreator = creatorDomain == callerDomain
 	}
+	isSharedDriveWriter := allowWritersToManageLinks &&
```
taratatach: nit: it's a local variable so it's not a big issue but it feels weird to define a role by active config switches pertaining to this role

taratatach (approved):

web/sharings/drives.go
shepilov: Rename to `allowToManageLinks`? or I didn't get what do you mean?

taratatach: You got the meaning :)

PR #4803, opened 2026-06-15
fix: allow delegated drive recipient removal

shepilov:
## Summary

- allow write-capable shared-drive recipients to remove another recipient through the existing recipient removal endpoint
- delegate recipient-side removal requests to the owner sharing using the existing sharing OAuth token refresh pattern
- keep direct delegated recipient removal forbidden for non-drive sharings, including open sharings

## Root cause

`DELETE /sharings/:sharing-id/recipients/:index` was owner-side only. A write-capable recipient could pass local permissions, but the local recipient-side sharing has `owner: false`, so the model returned `ErrInvalidSharing` and the API responded with 400.

### DIFF
--- a/model/sharing/member.go
+++ b/model/sharing/member.go
@@ -356,6 +356,9 @@ func (s *Sharing) SendDelegated(inst *instance.Instance, api *APIDelegateAddCont
 	if err != nil {
 		return err
 	}
+	if len(s.Credentials) == 0 {
+		return ErrInvalidSharing
+	}
 	c := &s.Credentials[0]
 	if c.AccessToken == nil {
 		return ErrInvalidSharing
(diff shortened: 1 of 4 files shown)


web/sharings/revoke.go
```diff
@@ -38,24 +38,48 @@ func RevokeRecipient(c echo.Context) error {
...
+	if s.Owner {
+		if err = s.RevokeRecipient(inst, index); err != nil {
+			return wrapErrors(err)
+		}
+		go s.NotifyRecipients(inst, nil)
+		return c.NoContent(http.StatusNoContent)
+	}
+	if !s.Drive {
+		return echo.NewHTTPError(http.StatusForbidden)
+	}
```
taratatach: We need to forbid revocation for organization shared drive editors as well.

model/sharing/member.go
```diff
@@ -989,6 +992,48 @@ func (s *Sharing) DelegateRemoveReadOnlyFlag(inst *instance.Instance, index int)
...
+		Method: http.MethodDelete,
+		Scheme: u.Scheme,
+		Domain: u.Host,
+		Path:   fmt.Sprintf("/sharings/%s/recipients/%d", s.SID, index),
+		Headers: request.Headers{
+			echo.HeaderAuthorization: "Bearer " + c.AccessToken.AccessToken,
+		},
+		ParseError: ParseRequestError,
+	}
+	res, err := request.Req(opts)
```
taratatach, line 1019: Can we add an optimization to avoid http requests for same stack sharings?

web/sharings/revoke.go
taratatach: Besides, this check is already done in `authorizeRevokeRecipient()`

taratatach (changes requested):

web/sharings/revoke.go
taratatach: We'll deal with organization shared drives later. The check is still done twice though so we could remove these lines.

taratatach (approved):

web/sharings/revoke.go
```diff
@@ -38,24 +38,59 @@ func RevokeRecipient(c echo.Context) error {
...
+func authorizeRevokeRecipient(c echo.Context, s *sharing.Sharing) error {
+	if _, err := checkCreatePermissions(c, s); err == nil {
+		if s.Owner || s.Drive {
+			return nil
+		}
+		return echo.NewHTTPError(http.StatusForbidden)
+	}
+	if !s.Owner || !s.Drive {
+		return echo.NewHTTPError(http.StatusForbidden)
+	}
```
taratatach, line 84:
```suggestion
	if !s.Owner || !s.Drive {
		return echo.NewHTTPError(http.StatusForbidden)
	}
	if _, err := checkCreatePermissions(c, s); err == nil {
			return nil
	}
```

PR #4807, opened 2026-06-17
fix: readonly sharing member shouldn't be able to self-promote

shepilov:
## Summary

  Prevent read-only shared drive recipients from promoting themselves to editor.

  ## Details

  A read-only shared drive recipient could call the role-change endpoint with a sharing protocol token and remove their own `read_only` flag. The token can legitimately include `io.cozy.sharings:ALL:<sharingID>` during the downgrade
  protocol, but that scope should not imply editor privileges on the shared drive.

### DIFF
--- a/web/sharings/readonly.go
+++ b/web/sharings/readonly.go
@@ -22,7 +22,7 @@ func AddReadOnly(c echo.Context) error {
 	_, err = checkCreatePermissions(c, s)
 	if err != nil {
 		// It can be a delegated call from a member on an open sharing to the owner
-		if err = hasSharingWritePermissions(c); err != nil {
+		if err = authorizeDelegatedReadOnlyChange(c, s); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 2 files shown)


web/sharings/readonly.go
```diff
@@ -100,6 +100,28 @@ func RemoveReadOnly(c echo.Context) error {
...
+	}
+
+	// A read-only drive member can temporarily hold io.cozy.sharings:ALL to
+	// finish the downgrade protocol, but that scope must not let them promote
+	// themselves or another member to editor.
+	member, err := requestMember(c, s)
+	if err != nil {
+		return wrapErrors(err)
+	}
+	if member.ReadOnly {
```
taratatach, line 119: Maybe this should be the actual implementation of `hasSharingWritePermissions()`?!

taratatach (approved):

web/sharings/readonly.go
shepilov: Yep, maybe it make sense, I'll create an issue for this

shepilov: https://github.com/cozy/cozy-stack/issues/4813

PR #4808, opened 2026-06-17
feat: add open route for excalidraw documents

Crash--:
## Summary

- Add `GET /excalidraw/:file-id/open` returning where to open a `.excalidraw` document (owner instance, subdomain, protocol, sharecode), typed as `io.cozy.files`.
- Add the shared drive variant `GET /sharings/drives/:id/excalidraw/:file-id/open`.
- Add `ExcalidrawOpener` in `model/sharing`, reusing `FileOpener` for the local/shared and drive/cozy-to-cozy routing, with no collaborative-edition infra (whole-file open/save).
- Validate the file by its `.excalidraw` extension via the new `ExcalidrawExtension` constant.

## Updated
  - Add `GET /editor/:file-id/open` returning where to open a file-backed editor document (owner instance, subdomain, protocol, target `file_id`, sharecode), typed as `io.cozy.files`.
  - Add the shared drive variant `GET /sharings/drives/:id/editor/:file-id/open`.
  - Use the generic editor opener for Excalidraw files instead of adding an Excalidraw-specific route.
  - Reuse shared `FileOpener` logic for local vs owner-instance routing across editor, notes, and office open flows.
  - Keep Notes and OnlyOffice public routes and response contracts unchanged because they need editor-specific params/setup.
  - Force shared editor opens to read-only unless the `io.cozy.files` permission grants write access.


Crash--:
@shepilov I don't like this approach of adding a new route every time we add a new collaborative editor. 

Maybe we can have something more generics like "editor/fileId/open"? And if we need more work (like notes) then we create a dedicated route? 

I need this to open excalidraw file on the owner instance. If I don't do that, today the websocket of the recipients & the websocket of the owner do not communicate. So even if we have the same file id, we do not share realtime events. So I need to open on only one instance.

shepilov:
> @shepilov I don't like this approach of adding a new route every time we add a new collaborative editor.
> 
> Maybe we can have something more generics like "editor/fileId/open"? And if we need more work (like notes) then we create a dedicated route?
> 
> I need this to open excalidraw file on the owner instance. If I don't do that, today the websocket of the recipients & the websocket of the owner do not communicate. So even if we have the same file id, we do not share realtime events. So I need to open on only one instance.

Yep, for this "simple" open we can have a generic route, also what I want to get rid of during "shared drives" refactoring, is ot get rid of memberIndex, and it would be great if we have it in one place for all editors

Right now OO and Notes has some specific behavior, validation, tokens, etc ... 

So we can make this route generic, and reuse some of the models in the OO and Notes handlers

I'll take a look into it

shepilov:
#4818 next

Crash--:
@shepilov great! There is no documentation? 

Since I've created the PR I can't approve it

shepilov:
> #4818
I'll add it

Crash--:
thanks! Let's merge & try <3

shepilov (approved):

PR #4810, opened 2026-06-18
feat: Resolve client URL via app clientURL or default

zatteo:
An app can declare a "client_url_flag" key in its manifest. Two paths now:

  Case A — client_url_flag resolves
  Look up the flag value from the instance's feature flags and use
  it as the client URL.

  Case B — subdomain (original behavior, also the fallback)
  Construct https://<user + slug>.<domain> via SubDomain().

This URL is used to specify valid origin when using postMessage in intents. It is required to create an intent from tmail (which is hosted at mail.twake.app and not <user>-mail.twake.app).

Todo:
- [x] validate manifest key name "client_url_flag"
- [x] add manifest key to cozy-twakemail https://github.com/cozy/cozy-twakemail/pull/39
- [x] add manifest key to manifest documentation https://github.com/cozy/cozy-apps-registry/pull/478

### DIFF
--- a/web/apps/apps_test.go
+++ b/web/apps/apps_test.go
@@ -25,6 +25,7 @@ import (
 	"github.com/cozy/cozy-stack/model/instance/lifecycle"
 	"github.com/cozy/cozy-stack/model/intent"
 	"github.com/cozy/cozy-stack/model/oauth"
+	"github.com/cozy/cozy-stack/model/permission"
 	"github.com/cozy/cozy-stack/model/session"
 	"github.com/cozy/cozy-stack/model/stack"
 	"github.com/cozy/cozy-stack/pkg/assets"
@@ -32,6 +33,7 @@ import (
 	"github.com/cozy/cozy-stack/pkg/assets/model"
 	"github.com/cozy/cozy-stack/pkg/config/config"
 	"github.com/cozy/cozy-stack/pkg/consts"
+	"github.com/cozy/cozy-stack/pkg/couchdb"
 	"github.com/cozy/cozy-stack/pkg/filetype"
 	"github.com/cozy/cozy-stack/tests/testutils"
 	"github.com/cozy/cozy-stack/web"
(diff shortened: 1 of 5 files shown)


web/intents/intents.go
```diff
@@ -52,15 +53,40 @@ func (i *apiIntent) MarshalJSON() ([]byte, error) {
...
+		i.doc.Client = i.resolveClientURL(parts[1])
 	}
 	res, err := json.Marshal(i.doc)
 	i.doc.Client = was
 	return res, err
 }
 
+func (i *apiIntent) resolveClientURL(slug string) string {
+	var appDoc couchdb.JSONDoc
+	err := couchdb.GetDoc(i.ins, consts.Apps, consts.Apps+"/"+slug, &appDoc)
```
shepilov: it should be app.GetWebappBySlug

web/intents/intents.go
```diff
@@ -52,15 +53,40 @@ func (i *apiIntent) MarshalJSON() ([]byte, error) {
...
+	if err != nil {
+		return i.defaultClientURL(slug)
+	}
+	flagKey, ok := appDoc.M["clientURL"].(string)
```
shepilov: and this can be moved to  func (m *WebappManifest) ClientURLFlag()

shepilov: also mabe it make sens to vlidate flagKey as a valid URL and fallback to default otherwise. But I'm not sure what will be better from the client side, to recieve invalid URL and handle errors there or deal with the default one that is not what expected

shepilov:
@zatteo should we also add it to CSP `frame-ancestors`? or wrapper app URL is enough? 
if iframe is opened from `mail.twake.app` I think we should add it as well, but I'm not sure

web/intents/intents.go
zatteo: Done 👍

web/intents/intents.go
zatteo: Done 👍

zatteo: I added a basic check to ensure we have a correctly formatted URL. But at the end, we can consider that it is the responsability of the client to deal with this kind of error (for example if flag is not set).

zatteo:
> @zatteo should we also add it to CSP `frame-ancestors`? or wrapper app URL is enough? if iframe is opened from `mail.twake.app` I think we should add it as well, but I'm not sure

Good catch I forgot about it. I added a second commit for this to homogenize the behavior.

zatteo:
Also, I renamed the manifest.webapp field to client_url_flag to be more clear and stick with usual naming of manifest.webapp attribute.

web/apps/apps_test.go
```diff
@@ -280,9 +282,68 @@ func TestApps(t *testing.T) {
 			Header(echo.HeaderContentSecurityPolicy)
 		csp.Contains("script-src")
 		csp.Contains("'wasm-unsafe-eval'")
-		csp.Contains("connect-src blob:")
 		csp.Contains("worker-src 'self' blob:;")
-		csp.Contains("frame-ancestors 'self' https://test-app.cozywithapps.example.net/;")
+		csp.Contains("frame-ancestors 'self' https://test-app.cozywithapps.example.net;")
```
zatteo, line 286: @shepilov is this trailing slash important?

shepilov: no, I don't think that we need a slash there

shepilov (approved):

PR #4814, opened 2026-06-22
feat: support instance_claim for OIDC app token exchange

shepilov:
## Summary

  Add optional `instance_claim` support to OIDC app token exchange configuration.

  When configured for an app token exchange audience, Stack uses that ID token claim as the target Cozy instance domain instead of the default OIDC instance mapping. If`instance_claim` is not configured, the existing behavior is unchanged.

  This allows apps like Twake Mail to keep an app-specific `sub` while still exchanging tokens against the correct Cozy instance.

  ## Changes

  - Add `instance_claim` to `app_token_exchange` config entries.
  - Validate configured instance claims against the target instance domain.
  - Keep existing fallback behavior when `instance_claim` is absent.

### DIFF
--- a/web/auth/token_exchange.go
+++ b/web/auth/token_exchange.go
@@ -315,10 +315,38 @@ func validateTokenExchangeAppToken(inst *instance.Instance, conf *oidcprovider.C
 	if !ok {
 		return nil, errors.New("invalid token audience")
 	}
-	if err := oidcprovider.CheckClaimsForInstance(conf, inst, claims); err != nil {
+	if err := tokenExchangeCheckAppInstance(conf, inst, claims, *appConfig); err != nil {
 		return nil, err
 	}
 	return appConfig, nil
+}
+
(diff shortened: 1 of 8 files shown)


web/auth/token_exchange.go
```diff
@@ -315,12 +315,40 @@ func validateTokenExchangeAppToken(inst *instance.Instance, conf *oidcprovider.C
...
+	if !ok || rawDomain == "" {
+		inst.Logger().WithNamespace("oidc").
+			Warnf("Cannot extract instance claim %s", appConfig.InstanceClaim)
+		return oidcprovider.ErrInstanceAuthenticationFailed
+	}
+
+	domain := utils.NormalizeDomain(rawDomain)
+	if domain == "" {
+		inst.Logger().WithNamespace("oidc").
+			Warnf("Cannot extract domain from instance claim %s", appConfig.InstanceClaim)
```
taratatach, line 339: Maybe we could log the raw domain here to help with debugging.

taratatach (approved):

web/auth/token_exchange.go
shepilov: mmm, if it's empty here, it means it was empty before, it won't help a lot if we log an empty string

PR #4816, opened 2026-06-23
New metadata structure/trigger worker

ewan102:
Adding a RAG indexation result into a typed FilesCozyMetadata.RAG sub-object. The webhook result is a boolean.
This PR changes nothing for the rest of Cozy while CallRAGindexer is not modified
- Add RAGMetadata struct (Indexed bool + LastSuccessDate/LastErrorDate) to FilesCozyMetadata
- Modification of the PR from last week of worker/rag/status.go: statusMessage{Indexed bool}, updateRAGStatus
  and timestamp atomically in one couchdb.UpdateDoc with a 3-retry conflict loop.
- Add status_test.go


worker/rag/status.go
```diff
@@ -0,0 +1,120 @@
...
+	log.Debugf("Updating RAG status for file %s: indexed=%v", msg.FileID, msg.Indexed)
+
+	return updateRAGStatus(inst, file, msg.Indexed)
+}
+
+// updateRAGStatus writes the RAG indexation result into cozyMetadata.RAG.
+// It retries on CouchDB conflict errors (409) up to maxRetries times,
+// re-fetching a fresh document revision each iteration to avoid stale-rev
+// conflicts. Both the boolean flag and the timestamp are written in a single
+// UpdateDoc call to stay atomic.
```
paultranvan: Please don't describe what the implem does except for non-trivial stuff. Implementation can change over time, and comments like this are easy to forget

worker/rag/status.go
```diff
@@ -0,0 +1,120 @@
...
+		}
+		if newdoc.CozyMetadata.RAG == nil {
+			newdoc.CozyMetadata.RAG = &vfs.RAGMetadata{}
+		}
+		newdoc.CozyMetadata.RAG.Indexed = indexed
+		now := time.Now()
+		if indexed {
+			newdoc.CozyMetadata.RAG.LastSuccessDate = &now
+		} else {
+			newdoc.CozyMetadata.RAG.LastErrorDate = &now
```
paultranvan: There is a confusion here: `indexed: false` is not an error, we can have this value when we delete a file. So for me, those `LastSuccessDate` / `LastErrorDate` should be provided in the webhook directly, alongside the `Indexed` info.

worker/rag/status.go
```diff
@@ -0,0 +1,120 @@
...
+	var err error
+	log := inst.Logger().WithNamespace("rag")
+	fs := inst.VFS()
+
+	for i := 0; i < maxRetries; i++ {
+		if i > 0 {
+			log.Infof("Retrying RAG status update for file %s (attempt %d/%d)", file.DocID, i+1, maxRetries)
+		}
+		var f *vfs.FileDoc
+		f, err = fs.FileByID(file.DocID)
```
paultranvan: you already have the `file` provided as func parameter

worker/rag/status.go
```diff
@@ -0,0 +1,120 @@
...
+// re-fetching a fresh document revision each iteration to avoid stale-rev
+// conflicts. Both the boolean flag and the timestamp are written in a single
+// UpdateDoc call to stay atomic.
+func updateRAGStatus(inst *instance.Instance, file *vfs.FileDoc, indexed bool) error {
+	const maxRetries = 3
+	var err error
+	log := inst.Logger().WithNamespace("rag")
+	fs := inst.VFS()
+
+	for i := 0; i < maxRetries; i++ {
```
paultranvan: I don't think this maxRetries is useful, there is already an automatic worker retry. I would simply remove the loop. For conflicts, we could have this logic: if a conflict occurs, it means a parallel webhook was called. So, compare the current file `LastSuccessDate` and ` LastErrorDate` with the received dates. The most recent date win: if it's the ones from file: drop it, no need to update the doc. If it's the new dates: do a new fetch to have the latest file and update

model/rag/webhook.go
```diff
@@ -0,0 +1,43 @@
...
+		return "", err
+	}
+	for _, t := range triggers {
+		info := t.Infos()
+		if info.Type == "@webhook" && info.WorkerType == "rag-index-status" {
+			webhookURL := inst.PageURL("/jobs/webhooks/"+t.ID(), nil)
+			log.Debugf("RAG webhook trigger already exists: %s", t.ID())
+			return webhookURL, nil
+		}
+	}
```
paultranvan: I think we can do better than that : currently, it means we fetch all the triggers from db, and check if the correct trigger exist. It's ok, but not very efficient.  We could do that with one db query.
When you create the trigger, provide a TID: https://github.com/cozy/cozy-stack/blob/master/model/job/scheduler.go#L63 ; for instance `TID=RAG_STATUS_TRIGGER_NAME`, with `RAG_STATUS_TRIGGER_NAME="rag-index-status"`
Then, you can query it directly with a `sched.GetTrigger(inst, RAG_TRIGGER_STATUS_ID)`

model/vfs/cozy_metadata.go
```diff
@@ -194,6 +216,21 @@ func (fcm *FilesCozyMetadata) ToJSONDoc() map[string]interface{} {
 		}
 		doc["trashedBy"] = trashed
 	}
+	if fcm.RAG != nil {
+		rag := map[string]interface{}{
+			"indexed": fcm.RAG.Indexed,
```
paultranvan: Why is indexed not treated like the other attributes?

worker/rag/status.go
```diff
@@ -0,0 +1,80 @@
...
+		WorkerFunc:   WorkerIndexStatus,
+	})
+}
+
+type statusMessage struct {
+	Partition string `json:"partition"`
+	FileID    string `json:"file_id"`
+	Status    string `json:"status"`    // "success" | "error" | "notsupported"
+	Md5Sum    string `json:"md5sum"`    // content hash (hex) the status refers to; echoed by rag-indexer
+	Timestamp string `json:"timestamp"` // RFC3339Nano; absent in older emitters
```
paultranvan: what is "older"? I don't think we should have time-relative comment

worker/rag/status.go
```diff
@@ -0,0 +1,80 @@
...
+		Timeout:      30 * time.Second,
+		WorkerFunc:   WorkerIndexStatus,
+	})
+}
+
+type statusMessage struct {
+	Partition string `json:"partition"`
+	FileID    string `json:"file_id"`
+	Status    string `json:"status"`    // "success" | "error" | "notsupported"
+	Md5Sum    string `json:"md5sum"`    // content hash (hex) the status refers to; echoed by rag-indexer
```
paultranvan: "echoed by rag-indexer" -> names can change , and we could have other emitters than rag-indexer, so no need to mention it

worker/rag/callback_status.go
```diff
@@ -0,0 +1,80 @@
+package rag
```
paultranvan: we might rename this file "callback_status", this way the purpose is clearer

model/rag/status.go
```diff
@@ -0,0 +1,133 @@
...
+		timestamp = time.Now()
+	}
+	fs := inst.VFS()
+	log := inst.Logger().WithNamespace("rag")
+
+	apply := func(file *vfs.FileDoc) error {
+		// Callback about an outdated version (a newer content is now on the file):
+		// it must not influence Status, but a past success still means a version
+		// was indexed.
+		if isOutdated(file, md5sum) {
```
paultranvan, line 36: if the md5 is outdated, I think we actually should drop it entirely. setting `indexed:true` could actually be false, since the newest status could be `indexed: false` if we removed the file

ewan102:
Just a precision, some PRs will follow to have a complete functioning : an authentification token will be added, a new doctype will be created and therefore the rag status will be referenced only in this new doctype io.cozy.ai.chat.rag

PR #4817, opened 2026-06-24
fix: stop unverified email changes via the settings API

rezk2ll:
## Problem

The instance email is the delivery address for password-reset, 2FA and magic-link mails, but `PUT /settings/instance` let any caller set it directly, no passphrase, no confirmation. A settings-scoped token could swap the email, trigger the public password reset and take over the account, which is exactly what the verified `POST /settings/email` flow exists to prevent.

## Solution

Non-CLI callers can no longer change `email`/`pending_email` through `PUT /settings/instance`; those go through `POST /settings/email` or the CLI, so other deployments keep working. Contexts wired to the signup app (`signup_url` set) also lock `phone`/`recovery_email` and disable `POST /settings/email`, since the signup flow owns the profile.

### DIFF
--- a/web/settings/settings_test.go
+++ b/web/settings/settings_test.go
@@ -324,7 +324,9 @@ func TestSettings(t *testing.T) {
 	t.Run("PatchWithBadRevAndChanges", func(t *testing.T) {
 		e := testutils.CreateTestClient(t, tsURL)
 
-		// We are defining a random rev, but make changes in the instance values
+		// We are defining a random rev, but make changes in the instance values.
+		// email is guarded, so the change has to be on an editable settings-doc
+		// field (tz) to still trigger the conflict.
 		rev := "6-2d9b7ef014d10549c2b4e206672d3e44"
 
 		e.PUT("/settings/instance").
@@ -340,7 +342,7 @@ func TestSettings(t *testing.T) {
             "rev": "%s"
           },
           "attributes": {
-            "tz": "Europe/London",
+            "tz": "Asia/Tokyo",
             "email": "alice@example.com",
             "locale": "en"
           }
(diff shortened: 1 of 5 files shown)


Crash--:
We should do that only if there is the signup flag no? Because if not then there is no reason to block email or phone number changes.

Crash--:
We should also block the post /settings/email in the case of an active signup.

web/settings/settings_test.go
```diff
@@ -324,7 +324,9 @@ func TestSettings(t *testing.T) {
 	t.Run("PatchWithBadRevAndChanges", func(t *testing.T) {
 		e := testutils.CreateTestClient(t, tsURL)
 
-		// We are defining a random rev, but make changes in the instance values
+		// We are defining a random rev, but make changes in the instance values.
+		// email is guarded, so the change has to be on an editable settings-doc
+		// field (tz) to still trigger the conflict.
```
shepilov, line 329: why do we need this changes it tests?

web/settings/settings_test.go
```diff
@@ -340,7 +342,7 @@ func TestSettings(t *testing.T) {
             "rev": "%s"
           },
           "attributes": {
-            "tz": "Europe/London",
+            "tz": "Asia/Tokyo",
```
shepilov, line 345: and this one as well?

shepilov (approved):

web/settings/settings_test.go
rezk2ll: Before this PR the conflicting change here was the email (body alice@example.com vs stored alice@example.org). Email is now guarded/ignored on PUT, so it no longer counts as a change, and this test needs a real change to hit the bad-rev 409.

web/settings/settings_test.go
rezk2ll: Same reason: stored tz is already Europe/London here, so it was a no-op once email stopped counting. Changed it to Asia/Tokyo to make a real settings change that triggers the 409

PR #4819, opened 2026-06-29
fix: avoid revoked note sharing panic

shepilov:
## Summary

- drop nil file docs returned for missing direct-note sharing keys before repairing revoked notes
- keep `FixRevokedNotes` defensive so it skips nil docs instead of calling methods on them

## Bug

Production panics happened on `DELETE /sharings/:sharing-id/recipients/self` while revoking a recipient-side note sharing.

For direct note sharings, `GetNotes` fetches `rule.Values` with CouchDB `_all_docs`. When one of those file ids is already missing or deleted locally, CouchDB still returns a row for the requested key instead of failing the whole request. The stack `GetAllDocs` helper preserves that missing row as `null`, which unmarshals into a nil `*vfs.FileDoc` inside the notes slice.

`FixRevokedNotes` then iterated over that slice and called `doc.ID()` to recreate the note trigger/import images. When `doc` was nil, this caused the nil pointer panic seen in production.

### DIFF
--- a/model/sharing/files.go
+++ b/model/sharing/files.go
@@ -828,7 +828,7 @@ func (s *Sharing) GetNotes(inst *instance.Instance) ([]*vfs.FileDoc, error) {
 					return nil, fmt.Errorf("failed to fetch notes shared by themselves: %w", err)
 				}
 
-				return notes, nil
+				return nonNilFileDocs(notes), nil
 			} else {
 				return nil, nil
 			}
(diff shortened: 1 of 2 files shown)


model/sharing/files.go
```diff
@@ -869,6 +879,9 @@ func (s *Sharing) FixRevokedNotes(inst *instance.Instance) error {
 
 	var errm error
 	for _, doc := range docs {
+		if doc == nil {
+			continue
+		}
```
taratatach, line 884: It's probably defensive but it's redundant with the call to `nonNilFileDocs()` in `GetNotes()`

taratatach (approved):

PR #4820, opened 2026-06-29
feat: Intent client empty when using a token from token_exchange

zatteo:
AI report

## The bug

When an app uses an OAuth access token obtained from `POST /auth/token_exchange` to
create an intent, the `client` field in the JSON-API response is `""` instead of
the resolved client URL.

## Root cause

The chain from token → intent response:

### 1. JWT subject is the OAuth client CouchID

`model/oauth/client.go:834`

```go
func (c *Client) CreateJWT(i *instance.Instance, audience, scope string) (string, error) {
    token, err := crypto.NewJWT(i.OAuthSecret, permission.Claims{
        RegisteredClaims: jwt.RegisteredClaims{
            Subject:  c.CouchID,   // ← UUID, e.g. "c1a2b3c4d5e67890abcdef0123456789"
            ...
        },
        Scope: scope,
    })
```

`CouchID` is a UUID — no `/` character.

### 2. SourceID is set to claims.Subject

`web/middlewares/permissions.go:128-133`

```go
func GetForOauth(instance *instance.Instance, claims *permission.Claims, client *oauth.Client) (*permission.Permission, error) {
    pdoc := &permission.Permission{
        Type:     permission.TypeOauth,
        SourceID: claims.Subject,  // ← the UUID, no "/"
        Client:   client,
    }
}
```

### 3. Intent.Client is set to SourceID

`web/intents/intents.go:83`

```go
intent.Client = pdoc.SourceID  // ← "c1a2b3c4d5e67890abcdef0123456789"
```

### 4. MarshalJSON splits on "/" — finds nothing

`web/intents/intents.go:50-57`

```go
func (i *apiIntent) MarshalJSON() ([]byte, error) {
    was := i.doc.Client
    parts := strings.SplitN(i.doc.Client, "/", 2)
    if len(parts) < 2 {
        i.doc.Client = ""   // ← always hit for OAuth tokens
    } else {
        i.doc.Client = i.resolveClientURL(parts[1])  // ← never reached
    }
    res, err := json.Marshal(i.doc)
    i.doc.Client = was
    return res, err
}
```

Since the UUID has no `/`, `strings.SplitN` returns a single-element slice,
`len(parts) < 2` is true, and `client` is set to `""`. The `resolveClientURL`
call (which would use `client_url_flag` from the manifest) is never reached.

## SourceID per token type

| Token audience | SourceID | Has `/`? | MarshalJSON result |
|---|---|---|---|
| `AppAudience` (webapp token) | `consts.Apps + "/" + slug` → `io.cozy.apps/mydrive` | yes | resolves correctly |
| `KonnectorAudience` | `consts.Konnectors + "/" + slug` | yes | resolves correctly |
| `AccessTokenAudience` (OAuth token) | `claims.Subject` = OAuth `CouchID` (UUID) | **no** | **`""`** |
| `CLIAudience` | unset/empty | no | `""` |
| Register token | unset/empty | no | `""` |

For webapp tokens, `SourceID` is built as `docType + "/" + slug` in
`createAppSet` (`model/permission/permissions.go:613`). For OAuth tokens,
`SourceID` is just `claims.Subject` — the OAuth client's CouchDB doc ID, which
is a UUID with no `/`.

## What needs to happen for OAuth tokens

`MarshalJSON` in `web/intents/intents.go` needs to handle the OAuth case. When
`SourceID` has no `/`, it should:

1. Retrieve the OAuth client from `claims.Subject` (the CouchID)
2. Find the app linked to that OAuth client (via the client's `SoftwareID` or
   linked app scope)
3. Use the app's slug to call `resolveClientURL(slug)`, which will pick up
   the `client_url_flag` from the manifest

The relevant helpers already exist:

- `oauth.FindClient(inst, clientID)` in `model/oauth/client.go`
- `oauth.GetLinkedAppSlug(softwareID)` to extract the slug from a software ID
- `app.ResolveClientURL(inst, slug)` to resolve the URL from the manifest's
   `client_url_flag` or fall back to subdomain

### DIFF
--- a/web/intents/intents.go
+++ b/web/intents/intents.go
@@ -65,6 +65,23 @@ func (i *apiIntent) resolveClientURL(slug string) string {
 	return app.ResolveClientURL(i.ins, slug)
 }
 
+func RequestAppSourceID(pdoc *permission.Permission) string {
+	if pdoc.Type != permission.TypeOauth {
+		return pdoc.SourceID
+	}
+
+	oc, ok := pdoc.Client.(*oauth.Client)
+	if !ok {
+		return pdoc.SourceID
+	}
+
+	if slug := oauth.GetLinkedAppSlug(oc.SoftwareID); slug != "" {
+		return consts.Apps + "/" + slug
+	}
+
+	return pdoc.SourceID
+}
+
 func createIntent(c echo.Context) error {
 	pdoc, err := middlewares.GetPermission(c)
 	if err != nil {
@@ -81,12 +98,7 @@ func createIntent(c echo.Context) error {
 	if intent.Type == "" {
 		return jsonapi.InvalidParameter("type", errors.New("Type is missing"))
 	}
-	intent.Client = pdoc.SourceID
-	if oc, ok := pdoc.Client.(*oauth.Client); ok {
-		if slug := oauth.GetLinkedAppSlug(oc.SoftwareID); slug != "" {
-			intent.Client = consts.Apps + "/" + slug
-		}
-	}
+	intent.Client = RequestAppSourceID(pdoc)
 	intent.SetID("")
 	intent.SetRev("")
 	intent.Services = nil
--- a/web/intents/intents.go
+++ b/web/intents/intents.go
@@ -9,6 +9,7 @@ import (
 	"github.com/cozy/cozy-stack/model/app"
 	"github.com/cozy/cozy-stack/model/instance"
 	"github.com/cozy/cozy-stack/model/intent"
+	"github.com/cozy/cozy-stack/model/oauth"
 	"github.com/cozy/cozy-stack/model/permission"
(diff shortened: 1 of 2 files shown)


web/intents/intents.go
```diff
@@ -81,6 +82,11 @@ func createIntent(c echo.Context) error {
 		return jsonapi.InvalidParameter("type", errors.New("Type is missing"))
 	}
 	intent.Client = pdoc.SourceID
+	if oc, ok := pdoc.Client.(*oauth.Client); ok {
+		if slug := oauth.GetLinkedAppSlug(oc.SoftwareID); slug != "" {
+			intent.Client = consts.Apps + "/" + slug
+		}
+	}
```
zatteo: I let you double check this one as I could not test locally 🙏

shepilov (approved):

PR #4824, opened 2026-07-01
Missing CSP for intent on a non-cozy domain 

zatteo:
When the client app is hosted on a non-cozy domain (via its client_url_flag), it can itself be framed by its cozy subdomain parent. frame-ancestors must list every ancestor in the iframe chain, so also allow the cozy subdomain.

### DIFF
--- a/model/app/webapp.go
+++ b/model/app/webapp.go
@@ -209,27 +209,27 @@ func (m *WebappManifest) ClientURLFlag() string { return m.val.ClientURLFlag }
 func ResolveClientURL(ins *instance.Instance, slug string) string {
 	manifest, err := GetWebappBySlug(ins, slug)
 	if err != nil {
-		return defaultClientURL(ins, slug)
+		return DefaultClientURL(ins, slug)
 	}
 	flagKey := manifest.ClientURLFlag()
 	if flagKey == "" {
-		return defaultClientURL(ins, slug)
+		return DefaultClientURL(ins, slug)
 	}
 	flags, err := feature.GetFlags(ins)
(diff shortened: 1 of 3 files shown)


taratatach:
I'm not sure I really get what the situation is. Could you give us an example please?

zatteo:
Context: We want to open a Twake Drive intent from Twake Mail (which is not a cozy application).

So we have:
- Drive app: user-drive.twake.app (per user)
- Todo app (just an example app): user-todo.twake.app (per user)
- Mail container app: user-mail.twake.app (per user)
- Mail app: mail.twake.app (shared for every users)

Previously, when an app requested an intent, it was adding <user>-<slug of the calling app>.twake.app to the requested app `frame-ancestors`. So if Todo app request Drive app, we added user-todo.twake.app to Drive `frame-ancestors`. But this did not work for mail because mail has the same domain for everyone. This was fixed here https://github.com/cozy/cozy-stack/pull/4810.

But actually, we need both, because we can open mail directly or from the Mail container app (actually this is more probable). So we have an iframe in an iframe and as `frame-ancestors` apply to _every_ parent we need to allow both. 

So we need to support these both cases:
- mail.twake.app that loads in iframe user-drive.twake.app ✅
- user-mail.twake.app loads in iframe mail.twake.app that loads in iframe user-drive.twake.app: ➡️ this PR because we missed in `frame-ancestors` user-mail.twake.app

taratatach:
Thank you for the explanation, that's really clear!

taratatach (approved):

PR #4828, opened 2026-07-02
feat: add B2B organization directory group synchronization through RabbitMQ.

### LINKED ISSUE #4821: B2B: RabbitMQ handler to sync organization groups to stack

shepilov:
## Summary

  This introduces support for `b2b.group.*` lifecycle events on the `b2b` exchange and replicates managed contact groups, group memberships, and related external contacts across all instances of an organization.

  ## What Changed

  - Added a new `stack.b2b.group.lifecycle` queue bound to:
    - `b2b.group.created`
    - `b2b.group.updated`
    - `b2b.group.deleted`
    - `b2b.group.member.added`
    - `b2b.group.member.removed`

  - Added organization directory group sync logic:
    - creates/updates/deletes managed `io.cozy.contacts.groups` documents
    - uses stable group IDs derived from `organizationId + external group id`
    - stores external metadata under `twakeDirectory`
    - refuses to overwrite/delete unmanaged groups

  - Added membership sync:
    - creates or reuses managed external contacts by email / Cozy URL
    - attaches/removes group relationships on contacts
    - supports members that only have an email, with workplace FQDN as optional data

  - Extended `user.created` organization contact sync:
    - supports `organizationId` or organization domain resolution
    - copies existing managed groups and contacts from the organization instance to newly created member instances

  - Added managed directory protections in the data API:
    - managed contacts/groups cannot be created, updated, or deleted through regular data routes
    - only the organization directory sync path owns those documents

  - Added organization instance lookup by `org_id`:
    - new global `by-orgid` index
    - org instance and managed directory queries now paginate with CouchDB bookmarks

Fixes #4821

### DIFF
--- a/model/instance/service.go
+++ b/model/instance/service.go
@@ -79,7 +79,11 @@ func (s *InstanceService) listByOrgField(indexName, fieldName, value string) ([]
 		if err != nil {
 			return nil, err
 		}
-		docs = append(docs, page...)
+		for _, inst := range page {
+			if inst != nil {
+				docs = append(docs, inst)
+			}
+		}
 
 		nextBookmark := ""
 		if res != nil {
--- a/model/instance/service.go
+++ b/model/instance/service.go
@@ -67,6 +67,19 @@ func (s *InstanceService) ListByOrgDomain(orgDomain string) ([]*Instance, error)
 	return docs, nil
 }
 
+func (s *InstanceService) ListByOrgID(orgID string) ([]*Instance, error) {
+	var docs []*Instance
+	req := &couchdb.FindRequest{
+		UseIndex: "by-orgid",
+		Selector: mango.Equal("org_id", orgID),
+	}
+	err := couchdb.FindDocs(prefixer.GlobalPrefixer, consts.Instances, req, &docs)
+	if err != nil {
+		return nil, err
+	}
+	return docs, nil
+}
+
 // Update saves the changes in CouchDB.
 func (s *InstanceService) Update(inst *Instance) error {
 	return couchdb.UpdateDoc(prefixer.GlobalPrefixer, inst)
--- a/model/instance/service.go
+++ b/model/instance/service.go
@@ -15,6 +15,8 @@ type InstanceService struct {
 	logger logger.Logger
 }
 
+const orgInstanceListPageSize = 1000
+
 func NewService(logger logger.Logger) *InstanceService {
 	return &InstanceService{
 		logger: logger,
@@ -55,29 +57,39 @@ func (s *InstanceService) Get(domain string) (*Instance, error) {
 }
 
 func (s *InstanceService) ListByOrgDomain(orgDomain string) ([]*Instance, error) {
-	var docs []*Instance
-	req := &couchdb.FindRequest{
-		UseIndex: "by-orgdomain",
-		Selector: mango.Equal("org_domain", orgDomain),
-	}
-	err := couchdb.FindDocs(prefixer.GlobalPrefixer, consts.Instances, req, &docs)
-	if err != nil {
-		return nil, err
-	}
-	return docs, nil
+	return s.listByOrgField("by-orgdomain", "org_domain", orgDomain)
 }
 
 func (s *InstanceService) ListByOrgID(orgID string) ([]*Instance, error) {
+	return s.listByOrgField("by-orgid", "org_id", orgID)
+}
+
+func (s *InstanceService) listByOrgField(indexName, fieldName, value string) ([]*Instance, error) {
 	var docs []*Instance
-	req := &couchdb.FindRequest{
-		UseIndex: "by-orgid",
-		Selector: mango.Equal("org_id", orgID),
+	var bookmark string
+	for {
+		var page []*Instance
+		req := &couchdb.FindRequest{
+			UseIndex: indexName,
+			Selector: mango.Equal(fieldName, value),
+			Limit:    orgInstanceListPageSize,
+			Bookmark: bookmark,
+		}
+		res, err := couchdb.FindDocsRaw(prefixer.GlobalPrefixer, consts.Instances, req, &page)
+		if err != nil {
+			return nil, err
+		}
+		docs = append(docs, page...)
+
+		nextBookmark := ""
+		if res != nil {
+			nextBookmark = res.Bookmark
+		}
+		if len(page) < orgInstanceListPageSize || nextBookmark == "" || nextBookmark == bookmark {
+			return docs, nil
+		}
+		bookmark = nextBookmark
 	}
-	err := couchdb.FindDocs(prefixer.GlobalPrefixer, consts.Instances, req, &docs)
-	if err != nil {
-		return nil, err
-	}
-	return docs, nil
 }
 
 // Update saves the changes in CouchDB.
--- a/model/orgdirectory/contact_patch.go
+++ b/model/orgdirectory/contact_patch.go
@@ -1,0 +1,110 @@
+package orgdirectory
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/cozy/cozy-stack/model/contact"
+	"github.com/cozy/cozy-stack/model/instance"
+	"github.com/cozy/cozy-stack/pkg/couchdb"
+	"github.com/cozy/cozy-stack/pkg/utils"
+)
+
+// ContactPatch describes a B2B contact replica to create or update.
+type ContactPatch struct {
+	OrganizationID string
+	Username       string
+	Email          string
+	FirstName      string
+	LastName       string
+	WorkplaceFQDN  string
+	Name           string
+	CozyURL        string
+	Phone          string
+}
+
+func (patch ContactPatch) validate() error {
+	if strings.TrimSpace(patch.Email) == "" {
+		return fmt.Errorf("contact missing email")
+	}
+	return nil
+}
+
+func (patch ContactPatch) shouldSkipOwn(inst *instance.Instance) bool {
+	if patch.WorkplaceFQDN != "" && inst.HasDomain(utils.ExtractInstanceHost(patch.WorkplaceFQDN)) {
+		return true
+	}
+	if patch.Email == "" {
+		return false
+	}
+	email, err := inst.SettingsEMail()
+	if err != nil {
+		return false
+	}
+	return strings.EqualFold(strings.TrimSpace(email), patch.Email)
+}
+
+func (patch ContactPatch) displayName() string {
+	if strings.TrimSpace(patch.Name) != "" {
+		return strings.TrimSpace(patch.Name)
+	}
+	name := strings.TrimSpace(strings.TrimSpace(patch.FirstName) + " " + strings.TrimSpace(patch.LastName))
+	if name != "" {
+		return name
+	}
+	if patch.Username != "" {
+		return patch.Username
+	}
+	if patch.Email != "" {
+		parts := strings.SplitN(patch.Email, "@", 2)
+		return parts[0]
+	}
+	return ""
+}
+
+func contactPatchFromMember(organizationID string, member GroupMember) ContactPatch {
+	workplaceFQDN := strings.TrimSpace(member.WorkplaceFQDN)
+	return ContactPatch{
+		OrganizationID: strings.TrimSpace(organizationID),
+		Username:       strings.TrimSpace(member.Username),
+		Email:          strings.TrimSpace(member.Email),
+		FirstName:      strings.TrimSpace(member.FirstName),
+		LastName:       strings.TrimSpace(member.LastName),
+		WorkplaceFQDN:  workplaceFQDN,
+		CozyURL:        cozyURLFromWorkplaceFQDN(workplaceFQDN),
+	}
+}
+
+func contactPatchFromDoc(organizationID string, doc *couchdb.JSONDoc) ContactPatch {
+	meta := directoryMetadata(doc)
+	contactDoc := &contact.Contact{JSONDoc: *doc}
+	patch := ContactPatch{OrganizationID: organizationID}
+	patch.Username, _ = meta["username"].(string)
+	patch.Email, _ = meta["email"].(string)
+	patch.WorkplaceFQDN, _ = meta["workplaceFqdn"].(string)
+	patch.CozyURL = contactDoc.PrimaryCozyURL()
+	if patch.CozyURL == "" {
+		patch.CozyURL = cozyURLFromWorkplaceFQDN(patch.WorkplaceFQDN)
+	}
+	patch.Name = primaryContactName(doc)
+	patch.Phone = contactDoc.PrimaryPhoneNumber()
+	return patch
+}
+
+func primaryContactName(doc *couchdb.JSONDoc) string {
+	c := &contact.Contact{JSONDoc: *doc}
+	name := c.PrimaryName()
+	if name != "" {
+		return name
+	}
+	displayName, _ := doc.M["displayName"].(string)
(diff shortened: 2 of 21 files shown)


model/instance/service.go
```diff
@@ -15,6 +15,8 @@ type InstanceService struct {
 	logger logger.Logger
 }
 
+const orgInstanceListPageSize = 1000
```
Crash--, line 18: ~~So we've an hardlimit to 1000 users in an org?~~~

It's just the default limit

pkg/couchdb/couchdb.go
```diff
@@ -776,6 +776,13 @@ func FindDocsUnoptimized(db prefixer.Prefixer, doctype string, req *FindRequest,
 	return err
 }
 
+// FindDocsUnoptimizedRaw allows search on non-indexed fields and returns the
+// raw find response, including the bookmark used for pagination.
+// /!\ Use with care
+func FindDocsUnoptimizedRaw(db prefixer.Prefixer, doctype string, req *FindRequest, results interface{}) (*FindResponse, error) {
+	return findDocsRaw(db, doctype, req, results, true)
+}
```
Crash--, line 784: why do we need to allow search on a non indexed fields?! It'll timeout in production. Specially on contacts... We have users with 10k + contacts

web/data/data.go
```diff
@@ -48,6 +49,27 @@ func fixErrorNoDatabaseIsWrongDoctype(err error) error {
 	return err
 }
 
+func managedDirectoryWriteError() error {
+	return jsonapi.Errorf(http.StatusForbidden, "managed organization directory documents are read-only")
```
Crash--: can you add a log? Because today it's a nightmare to have a 403 in graylog, but we are not being able to see the reason it was rejected. it's kind of complicated to debug

model/instance/service.go
shepilov: deafault one, we pginate them

pkg/couchdb/couchdb.go
shepilov: and to go into table with less then 1 million records without index supposed to be fine

web/data/data.go
shepilov: it will be logged upper in the handler

model/orgdirectory/metadata.go
```diff
@@ -0,0 +1,60 @@
...
+func setGroupDirectoryMetadata(doc *couchdb.JSONDoc, organizationID, externalID string) {
+	if doc.M == nil {
+		doc.M = make(map[string]interface{})
+	}
+	doc.M[DirectoryMetadataKey] = map[string]interface{}{
+		"managed":        true,
+		"kind":           metadataKindGroup,
+		"organizationId": organizationID,
+		"externalId":     externalID,
+	}
```
taratatach: We could create a new struct for this directory metadata.

web/data/data.go
```diff
@@ -280,6 +321,10 @@ func DeleteDoc(c echo.Context) error {
 	doc.Type = doctype
 	doc.SetRev(rev)
 
+	if orgdirectory.IsManagedDirectoryDoctype(doctype) && orgdirectory.IsManagedDirectoryDoc(&doc) {
+		return managedDirectoryWriteError()
+	}
```
taratatach: Create  a `rejectManagedDirectoryDelete` function?

model/orgdirectory/organizations.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	if len(list) == 0 {
+		return OrganizationInstances{}, fmt.Errorf("organization has no instances")
+	}
+
+	resolvedID := organizationID
+	if resolvedID == "" {
+		for _, inst := range list {
+			if inst == nil {
```
taratatach: Should we remove these from the list before returning it?

model/orgdirectory/contact_patch.go
```diff
@@ -0,0 +1,110 @@
...
+	return patch
+}
+
+func primaryContactName(doc *couchdb.JSONDoc) string {
+	c := &contact.Contact{JSONDoc: *doc}
+	name := c.PrimaryName()
+	if name != "" {
+		return name
+	}
+	displayName, _ := doc.M["displayName"].(string)
```
taratatach: Maybe we could add a `DisplayName` method to `Contact`. This way `primaryContactName(doc)` could be replaced by `contactDoc.PrimaryName() || contactDoc.DisplayName()` in the caller.

model/orgdirectory/groups.go
```diff
@@ -0,0 +1,613 @@
...
+	ID             string    `json:"id"`
+}
+
+// GroupMembersMessage is the payload for b2b.group.member.added and
+// b2b.group.member.removed.
+type GroupMembersMessage struct {
+	Timestamp      time.Time     `json:"timestamp"`
+	OrganizationID string        `json:"organizationId"`
+	ID             string        `json:"id"`
+	Members        []GroupMember `json:"members"`
```
taratatach, line 53: Does this hold every member of the group or only added members?

model/orgdirectory/groups.go
```diff
@@ -0,0 +1,613 @@
...
+}
+
+// GroupMember describes a B2B user present in a group membership event.
+type GroupMember struct {
```
taratatach, line 57: Why not use `ContactPatch` instead of creating this new type? Also, `GroupMember` does not have any phone field. Does that mean group members are less complete contacts?

model/orgdirectory/groups.go
```diff
@@ -0,0 +1,613 @@
...
+	Description    *string
+	Color          *string
+	CreatedAt      *time.Time
+}
+
+// GroupDocID returns the stable local document ID used for a replicated B2B
+// group. Raw external IDs stay in metadata for traceability.
+func GroupDocID(organizationID, externalID string) string {
+	sum := sha256.Sum256([]byte(strings.TrimSpace(organizationID) + "\x00" + strings.TrimSpace(externalID)))
+	return "b2b-group-" + hex.EncodeToString(sum[:16])
```
taratatach, line 80: Isn't there a risk of collision by reducing the ID to these 16 characters?

model/orgdirectory/groups.go
```diff
@@ -0,0 +1,613 @@
...
+
+	var existing contact.Group
+	err := couchdb.GetDoc(inst, consts.Groups, groupID, &existing)
+	if err == nil {
+		if !IsManagedDirectoryDoc(&existing.JSONDoc) {
+			return fmt.Errorf("group id %s already exists and is not managed", groupID)
+		}
+		doc.M = existing.M
+		doc.Type = consts.Groups
+		doc.SetID(groupID)
```
taratatach: Redundant with line 296.

taratatach: Unless you wanted to do `doc.SetID(existing.ID())`?

model/orgdirectory/groups.go
```diff
@@ -0,0 +1,613 @@
...
+		doc.SetRev(existing.Rev())
+	} else if !couchdb.IsNoDatabaseError(err) && !couchdb.IsNotFoundError(err) {
+		return err
+	}
+
+	if fields.Name != nil && strings.TrimSpace(*fields.Name) != "" {
+		doc.M["name"] = strings.TrimSpace(*fields.Name)
+	} else if doc.M["name"] == nil {
+		doc.M["name"] = fields.ExternalID
```
taratatach, line 313: No trimming here?

model/orgdirectory/groups.go
```diff
@@ -0,0 +1,613 @@
...
+	var errs []error
+	for _, member := range members {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+		input := contactPatchFromMember(organizationID, member)
+		if err := input.validate(); err != nil {
+			errs = append(errs, err)
+			continue
+		}
```
taratatach: Already done in `UpsertManagedContact`.

model/orgdirectory/store.go
```diff
@@ -0,0 +1,89 @@
...
+	var docs []*contact.Contact
+	req := &couchdb.FindRequest{
+		Selector: mango.Map{
+			"cozy": map[string]interface{}{
+				"$elemMatch": map[string]interface{}{
+					"url": cozyURL,
+				},
+			},
+		},
+		Limit: 10,
```
taratatach: Since we just want to know if we have one or more documents, shouldn't 2 be enough here?

model/orgdirectory/store.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	err := couchdb.FindDocsUnoptimized(db, consts.Contacts, req, &docs)
```
taratatach, line 40: Can't we create an index to make efficient queries?

model/orgdirectory/store.go
```diff
@@ -0,0 +1,112 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if len(docs) == 0 {
+		return nil, contact.ErrNotFound
+	}
+	return singleExternalContact(docs, "cozy URL "+cozyURL)
+}
+
+func singleExternalContact(matches []*contact.Contact, label string) (*contact.Contact, error) {
```
taratatach: Since the contact might not be external but managed:

```suggestion
func singleManagedOrExternalContact(matches []*contact.Contact, label string) (*contact.Contact, error) {
```

We should apply the same name pattern to the callers imho.

model/orgdirectory/groups.go
```diff
@@ -0,0 +1,613 @@
...
+			{"address": email, "primary": true},
+		}
+	}
+	name := strings.TrimSpace(input.Name)
+	if name == "" {
+		name = input.displayName()
+	}
+	if name != "" {
+		c.M["fullname"] = name
+		c.M["displayName"] = name
```
taratatach, line 419: We might want to always use `input.displayName()` here.

model/orgdirectory/groups.go
taratatach: Looks like it holds members to add. In this case, I think the RabbitMQ message routing keys should be `b2b.group.members.added` and `b2b.group.members.removed` to make it more explicit.

model/orgdirectory/groups.go
```diff
@@ -0,0 +1,613 @@
...
+}
+
+// GroupUpdatedMessage is the payload for b2b.group.updated.
+type GroupUpdatedMessage struct {
+	Timestamp      time.Time `json:"timestamp"`
+	OrganizationID string    `json:"organizationId"`
+	ID             string    `json:"id"`
+	Name           *string   `json:"name,omitempty"`
+	Description    *string   `json:"description,omitempty"`
+	Color          *string   `json:"color,omitempty"`
```
taratatach, line 37: Why do we have pointers here while we have simple strings in every other message?

model/orgdirectory/groups.go
```diff
@@ -0,0 +1,613 @@
...
+	}
+
+	if err := copyManagedGroups(ctx, orgInst, target, organizationID); err != nil {
+		return err
+	}
+	return copyManagedContactsWithGroups(ctx, orgInst, target, organizationID)
+}
+
+func copyManagedGroups(ctx context.Context, source, target *instance.Instance, organizationID string) error {
+	groups, err := listManagedJSONDocs(source, consts.Groups, organizationID)
```
taratatach: Why not use the `Group` type here rather than `couchdb.JSONDoc`?

model/orgdirectory/groups.go
```diff
@@ -0,0 +1,613 @@
...
+		copy.SetRev("")
+		if err := couchdb.Upsert(target, copy); err != nil {
+			return fmt.Errorf("copy group %s to %s: %w", copy.ID(), target.Domain, err)
+		}
+	}
+	return nil
+}
+
+func copyManagedContactsWithGroups(ctx context.Context, source, target *instance.Instance, organizationID string) error {
+	contacts, err := listManagedJSONDocs(source, consts.Contacts, organizationID)
```
taratatach: Why not use the `Contact` type here rather than `couchdb.JSONDoc`?

model/orgdirectory/groups.go
```diff
@@ -0,0 +1,613 @@
...
+	}
+	return changed, nil
+}
+
+func removeContactGroup(c *contact.Contact, groupID string) (bool, error) {
+	refs := contactGroupRefs(c)
+	if len(refs) == 0 {
+		return false, nil
+	}
+	next := make([]interface{}, 0, len(refs))
```
taratatach, line 481: nit: rename `next` into `remaining` or something similar.

model/orgdirectory/groups.go
```diff
@@ -0,0 +1,613 @@
...
+// CopyOrgDirectoryFromOrgInstance copies current managed group/contact replicas
+// from the organization instance to a newly created member instance. LDAP/B2B
+// remains authoritative; the organization instance is only a local copy source
+// for late onboarding.
+func CopyOrgDirectoryFromOrgInstance(ctx context.Context, target *instance.Instance, organizationID string) error {
+	organizationID = strings.TrimSpace(organizationID)
+	if organizationID == "" || target == nil {
+		return nil
+	}
+	orgInst, err := findOrganizationInstance(ctx, organizationID)
```
taratatach, line 212: nit: introduced after the first commit using it.

model/orgdirectory/groups.go
```diff
@@ -0,0 +1,613 @@
...
+func copyManagedGroups(ctx context.Context, source, target *instance.Instance, organizationID string) error {
+	groups, err := listManagedJSONDocs(source, consts.Groups, organizationID)
+	if err != nil {
+		return fmt.Errorf("copy org directory groups from %s: %w", source.Domain, err)
+	}
+	for _, g := range groups {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+		copy := cloneForCreateOrUpdate(g, consts.Groups)
```
taratatach: Why the copy here?

pkg/rabbitmq/handlers.go
```diff
@@ -223,11 +224,23 @@ func (h *UserCreatedHandler) Handle(ctx context.Context, d amqp.Delivery) error
...
-	if msg.OrganizationDomain != "" {
-		if err := SyncCreatedOrgContact(ctx, inst, msg); err != nil {
+	if strings.TrimSpace(msg.OrganizationID) != "" || strings.TrimSpace(msg.OrganizationDomain) != "" {
+		scope, err := orgdirectory.ResolveOrganizationInstances(msg.OrganizationID, msg.OrganizationDomain)
+		if err != nil {
+			err = fmt.Errorf("user.created: %w", err)
+			log.Errorf("user.created: failed to resolve organization instances for %s: %v", msg.WorkplaceFqdn, err)
+			return err
+		}
+		if err := syncCreatedOrgContact(ctx, inst, msg, scope); err != nil {
```
taratatach, line 234: nit: introduced after the first commit using it

pkg/rabbitmq/handlers.go
```diff
@@ -274,9 +287,6 @@ func (h *UserDeletedHandler) Handle(ctx context.Context, d amqp.Delivery) error
 	if msg.InternalEmail == "" {
 		return fmt.Errorf("user.deleted: missing internalEmail")
 	}
-	if msg.Domain == "" {
-		return fmt.Errorf("user.deleted: missing organization domain")
-	}
```
taratatach, line 279: What does this change have to do with contact synchronization?

model/orgdirectory/contact_patch.go
shepilov: #4849 shoud be addresses here

taratatach (approved):

PR #4847, opened 2026-07-07
feat: Introduce route to change group permission

zatteo:
We want to be able to change a group permission in a sharing. New routes are similar to routes to change member permission.

## Behavior

On existing member or group add, it is forbidden to add a member or a group that contains a member with a different permissions than an existing member. So if you have User 1 as readwrite and you try to add Group A that includes User 1 as readonly it is currently blocked by the stack.

I kept this behavior when changing permissions to avoid any secondary effects and complications. 

In the long term, we want to be able to set different permission per members and per groups and compute for members the highest permission. But it requires a new attribute to track the initial permission of a member for the following scenario:
- we add User 1 as readwrite
- we add Group A that includes User 1 as readwrite
- User becomes readwrite
- we downgrade Group A to readonly
=> should we set User 1 to readwrite or readonly?

Related: https://github.com/linagora/cozy-client/pull/1696

## Usecase tested manually

**One team**

- Alice creates “Dossier 1”
- Alice adds “Équipe A” as reader
    - Bob read access + can’t change permissions
    - Carol read access + can’t change permissions
- Alice updates “Équipe A” as writer
    - Bob write access + can change permissions
    - Carol write access + can change permissions
- Bob updates “Équipe A” as reader
    - Bob read access + can’t change permissions
    - Carol read access + can’t change permissions

**Two team**

- Alice creates “Dossier 2”
- Alice adds “Équipe B” as reader
    - Bob read access + can’t change permissions
- Alice adds “Équipe B” as writer
    - raises error
- Alice adds “Équipe A” as reader
    - Bob read access + can’t change permissions
    - Carol read access + can’t change permissions
- Alice updates “Équipe A” as writer
    - raises error
- Alice updates “Équipe B” as writer
    - raises error
- Alice removes “Équipe B”
    - Bob read access + can’t change permissions
    
**One team + one member**

- Alice creates “Dossier 3”
- Alice adds “Équipe A” as editor
    - Bob write access + can change permission
    - Alice write access + can change permission
- Alice adds Bob as write
    - raise error
- Alice adds Bob as editor
    - does nothing
    
At the end here, Bob’s "only_in_groups" attribute is true (existing behavior).

**One member + one team**

- Alice creates “Dossier 4”
- Alice adds Bob as editor
    - Bob write access + can change permission
- Alice adds “Équipe A” as editor
    - Bob write access + can change permission
    - Alice write access + can change permission
- Alice, Bob and Carol can not update Bob or “Équipe A” permission

At the end here, Bob’s "only_in_groups" attribute is false (existing behavior).

### DIFF
--- a/docs/sharing.md
+++ b/docs/sharing.md
@@ -1246,6 +1246,8 @@ This route is used to add the read-only flag on a recipient of a sharing.
 
 **Note**: 0 is not accepted for `index`, as it is the sharer him-self.
 
+It returns 400 if the member belongs to any non-revoked group.
+
 ##### Request
 
 ```http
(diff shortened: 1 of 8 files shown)


zatteo:
This feature will be part of broader change in permissions

PR #4853, opened 2026-07-16
Inline intent doc when fetching an intent index.html

zatteo:
Eliminate the `GET /intents/:id` fetch request that the intent service app (inside the iframe) makes to the stack, by inlining the full intent object into the HTML response when the stack serves the service app's `index.html`.

The stack already loads the intent from CouchDB in `handleIntent()` when serving a page with `?intent=<id>` but currently discards it after setting CSP `frame-ancestors` rules. This change keeps the loaded intent and serializes
it into a `<script>` tag in the `<head>`, so the service app can read it synchronously from `window.__COZY_INTENT__` instead of making a second network round-trip.

**Estimated save:** 1 blocking API call, 233ms (average of 5 call on stg.lin-saas.com)

**Front end:** [github.com/linagora/cozy-libs/pull/3082](https://github.com/linagora/cozy-libs/pull/3082)

### DIFF
--- a/web/apps/serve.go
+++ b/web/apps/serve.go
@@ -638,23 +638,14 @@ func (s serveParams) IntentData() (template.HTML, error) {
 	if err != nil {
 		return "", err
 	}
-	// <script> content is raw text — HTML entities are not decoded by the
-	// browser, so html/template's default JS escaping would corrupt the
-	// JSON. We bypass it with template.HTML and neutralize </script>
-	// breakout ourselves. </ only occurs inside JSON string values, and \/
-	// is a valid JSON escape, so the substitution is parse-equivalent.
-	sanitized := strings.ReplaceAll(string(raw), "</", "<\\/")
-	buf := new(bytes.Buffer)
-	if err := intentTemplate.Execute(buf, template.HTML(sanitized)); err != nil {
-		return "", err
-	}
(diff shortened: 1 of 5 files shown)


web/apps/serve.go
```diff
@@ -644,6 +669,10 @@ func BuildTemplates() {
 {{end}}
 {{end}}`,
 	))
+
+	intentTemplate = template.Must(template.New("intent-data").Parse(
+		`<script>window.__COZY_INTENT__ = JSON.parse({{.}})</script>`,
```
shepilov: Have you tested it locally? It's supposed to be blocked by CSP

zatteo: Oh interesting. I tested locally but by default I use --disable-csp. So in this case I can add it in a meta tag that will be fetched in the DOM and parsed like existing CozyData.

shepilov: `<script type="application/json">` shoudl help

zatteo: @shepilov Thank you for your suggestion. Please check the two new commits that uses `<script type="application/json">`. Also updated here for frontend: https://github.com/linagora/cozy-libs/pull/3082/changes/4982a31c296e4ca4958939ee5a9b5156d702833f

shepilov (approved):

PR #4854, opened 2026-07-20
feat(sharing): propagate group colors

shepilov:
## Summary

- Include contact group colors in sharing group attributes.
- Propagate group color changes to active sharings and recipients.
- Cover direct and delegated sharing flows and document the field.

### DIFF
--- a/model/job/trigger_share_group.go
+++ b/model/job/trigger_share_group.go
@@ -64,11 +64,11 @@ func (t *ShareGroupTrigger) matchGroup(e *realtime.Event) *ShareGroupMessage {
 	if e.Verb != realtime.EventUpdate {
 		return nil
 	}
-	newdoc, ok := e.Doc.(*couchdb.JSONDoc)
+	newdoc, ok := groupJSONDoc(e.Doc)
 	if !ok {
 		return nil
 	}
-	olddoc, ok := e.OldDoc.(*couchdb.JSONDoc)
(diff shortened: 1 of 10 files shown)


model/sharing/group.go
```diff
@@ -189,25 +196,25 @@ func UpdateGroups(inst *instance.Instance, msg job.ShareGroupMessage) error {
...
 			if group.ID == doc.ID() {
-				if name, ok := doc.M["name"].(string); ok {
-					group.Name = name
-					s.Groups[idx] = group
-					if err := couchdb.UpdateDoc(inst, s); err != nil {
-						errm = multierror.Append(errm, err)
-					}
-					cloned := s.Clone().(*Sharing)
-					go cloned.NotifyRecipients(inst, nil)
+				group.Name = updatedGroup.Name()
```
taratatach: There's a risk to remove the group's name now because we apply the change even if `doc.M["name"]` is an invalid string (which translates to `updatedGroup.Name()` returning `""`).

model/job/trigger_share_group.go
taratatach, line 1: I would expect this commit to be squashed with the previous one which introduces the documented function `groupJSONDoc`.

model/sharing/group.go
shepilov: I was not able imagine a scenario when it is missing, except we manually edited DB

taratatach (approved):

PR #4856, opened 2026-07-21
Feat/assistant knowledge base workspace

paultranvan:
An assistant (`io.cozy.ai.chat.assistants`) can now declare a knowledge base
pointing to a Drive folder:

```json
{ "knowledgeBase": [{ "doctype": "io.cozy.files", "dirId": "<dir-id>" }] }
```

Chat retrieval for conversations bound to such an assistant is then restricted
to that folder, using openRAG workspaces (a workspace mirrors the folder,
workspace_id = dirId).

**Chat path**

On each query, the stack ensures the workspace exists before forwarding the
completion with metadata.workspace:

- The workspace is created lazily on first use, and its membership is
backfilled from the folder subtree (chunked, with per-id retries when a
chunk contains non-indexed files, and a rollback on failure so the next
query restarts the backfill from scratch).
- Fail closed: if the assistant or the workspace cannot be resolved, the
query surfaces an error to the client rather than silently answering from
the whole instance.
- Concurrent queries on the same folder are serialized with a distributed
lock.

**Index path**

The rag-index worker keeps memberships in sync with the Drive tree:

- new/updated files carry their workspace_ids at upload;
- moves and renames (files, or whole directory subtrees via the changes feed)
are reconciled against the actual openRAG membership (GET + diff);
- the knowledge-base context is loaded once per batch, best-effort: an
indexing batch never fails because of workspace sync.

### DIFF
--- a/model/rag/chat.go
+++ b/model/rag/chat.go
@@ -148,6 +148,34 @@ func (a *chatAssistant) SetID(id string)    { a.DocID = id }
 func (a *chatAssistant) SetRev(rev string)  { a.DocRev = rev }
 func (a *chatAssistant) Clone() couchdb.Doc { c := *a; return &c }
 
+// knowledgeBaseDirID returns the Drive folder scoping the assistant's
+// retrieval, or "" when the assistant has no knowledge base. Only a single
+// folder per assistant is supported: extra io.cozy.files entries are ignored,
+// with a warning so the truncation is at least visible in the logs.
+func (a *chatAssistant) knowledgeBaseDirID(logger logger.Logger) string {
+	if a == nil {
+		return ""
+	}
+	dirID := ""
+	for _, entry := range a.KnowledgeBase {
+		if entry.Doctype != consts.Files || entry.DirID == "" {
+			continue
+		}
+		if dirID == "" {
+			dirID = entry.DirID
+		} else if entry.DirID != dirID {
+			logger.Warnf("assistant %s: multiple knowledge base folders are not supported, ignoring %s", a.DocID, entry.DirID)
+		}
+	}
+	return dirID
+}
+
+// ErrAssistantNotFound is returned when a conversation references an
+// assistant that is gone (deleted, or never created). The query must fail:
+// answering anyway would silently widen a possibly folder-scoped
+// conversation to whole-instance retrieval.
+var ErrAssistantNotFound = errors.New("assistant not found")
+
 func Chat(inst *instance.Instance, payload ChatPayload) (*ChatConversation, error) {
 	var chat ChatConversation
 	err := couchdb.GetDoc(inst, consts.ChatConversations, payload.ChatConversationID, &chat)
@@ -289,10 +317,11 @@ func getSources(event map[string]interface{}) ([]Source, error) {
 }
 
 // assistantForChat loads the assistant bound to the conversation. It returns
-// (nil, nil) when the conversation has no assistant or the assistant was
-// deleted. An error means the caller could not determine how the
-// conversation is configured (e.g. a transient CouchDB error) and MUST NOT
-// proceed as if it were unscoped.
+// (nil, nil) only when the conversation has no assistant at all. An error
+// means either the referenced assistant is gone (ErrAssistantNotFound) or
+// its configuration could not be read (e.g. a transient CouchDB error); in
+// both cases the caller MUST NOT proceed as if the conversation were
+// unscoped.
 func assistantForChat(inst *instance.Instance, chat *ChatConversation) (*chatAssistant, error) {
 	rel, ok := chat.Rels["assistant"]
 	if !ok {
@@ -305,15 +334,13 @@ func assistantForChat(inst *instance.Instance, chat *ChatConversation) (*chatAss
 	}
 	var assistant chatAssistant
 	if err := couchdb.GetDoc(inst, consts.ChatAssistants, assistantID, &assistant); err != nil {
-		// A missing document — or a missing assistants database, which is a
-		// reachable state since the database is only created when a first
-		// assistant doc is saved and Chat() does not check that the
-		// client-supplied assistant id exists — means no knowledge-base
-		// configuration can exist: treat the conversation as unscoped rather
-		// than failing it. Other errors (e.g. a transient CouchDB failure)
-		// keep failing closed: the assistant may exist and be folder-scoped.
+		// The conversation references an assistant that cannot be found
+		// (deleted, or never existed). It may have been scoped to a
+		// knowledge base: degrading to unscoped retrieval would silently
+		// widen it to the whole instance, so surface an explicit error and
+		// let the client deal with the dangling reference.
 		if couchdb.IsNotFoundError(err) {
-			return nil, nil
+			return nil, ErrAssistantNotFound
 		}
 		return nil, err
 	}
@@ -411,7 +438,7 @@ func Query(inst *instance.Instance, logger logger.Logger, query QueryMessage) er
 	if override := buildLLMOverride(inst, assistant); override != nil {
 		metadata["llm_override"] = override
 	}
-	if dirID := knowledgeBaseDirID(assistant, logger); dirID != "" {
+	if dirID := assistant.knowledgeBaseDirID(logger); dirID != "" {
 		if err := ensureWorkspace(inst, logger, dirID); err != nil {
 			logger.Warnf("cannot ensure RAG workspace %s: %s", dirID, err)
 			// A folder-scoped assistant must never answer from the whole
--- a/model/rag/chat.go
+++ b/model/rag/chat.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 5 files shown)


model/rag/chat.go
```diff
@@ -280,25 +288,46 @@ func getSources(event map[string]interface{}) ([]Source, error) {
...
 	if err := couchdb.GetDoc(inst, consts.ChatAssistants, assistantID, &assistant); err != nil {
+		// A missing document — or a missing assistants database, which is a
+		// reachable state since the database is only created when a first
+		// assistant doc is saved and Chat() does not check that the
+		// client-supplied assistant id exists — means no knowledge-base
+		// configuration can exist: treat the conversation as unscoped rather
+		// than failing it. Other errors (e.g. a transient CouchDB failure)
+		// keep failing closed: the assistant may exist and be folder-scoped.
+		if couchdb.IsNotFoundError(err) {
+			return nil, nil
```
shepilov: But it means that when we delete the assistant, all chat queries will search the whole instance

model/rag/workspace.go
```diff
@@ -0,0 +1,505 @@
...
+// workspaceBackfillChunkSize is the maximum number of file ids sent in a
+// single workspace membership backfill request.
+const workspaceBackfillChunkSize = 500
+
+// knowledgeBaseDirID returns the Drive folder scoping the assistant's
+// retrieval, or "" when the assistant has no knowledge base. Only a single
+// folder per assistant is supported: extra io.cozy.files entries are ignored,
+// with a warning so the truncation is at least visible in the logs.
+func knowledgeBaseDirID(assistant *chatAssistant, logger logger.Logger) string {
+	if assistant == nil {
```
shepilov: I think it makes sense to make it a part of the chatAssistant struct

shepilov:
How do we check permissions when we are creating a knowledge base route?

I don't see how the current routes are coupled with permission checks

In my understanding, it's supposed to be like this:
 - we should store this relationship 
 - validate it when starting a conversation 
 - use this auth scope in the async job
 - if we are going to support shared folders, this knowledge base probably should contain sharing IDs and another set of permission checks ...

model/rag/index.go
```diff
@@ -100,6 +105,16 @@ func callRAGIndexer(inst *instance.Instance, doctype string, change couchdb.Chan
...
 	if change.Doc.Get("type") == consts.DirType {
+		// A directory move or rename rewrites the path of the directory and
+		// of all its descendant directories, but never touches the file docs
+		// they contain: those files may have silently entered or left a
+		// knowledge-base folder. Every descendant directory shows up in this
+		// same changes feed, so reconciling the direct file children of each
+		// changed directory covers the whole moved subtree.
+		dirPath, _ := change.Doc.Get("path").(string)
+		if dirPath != "" && !strings.HasPrefix(dirPath, vfs.TrashDirName) {
+			reconcileDirChildren(inst, logger, kb(), change.DocID, dirPath)
```
shepilov, line 116: Why it's only nice to have to sync it with RAG? It's kind of a complicated sync process, and I hope it deserves a separate PR

model/rag/workspace.go
```diff
@@ -0,0 +1,505 @@
...
+// base folder exists, creating it and backfilling its membership from the
+// folder subtree when needed. An error means the query MUST NOT proceed
+// unscoped.
+func ensureWorkspace(inst *instance.Instance, logger logger.Logger, dirID string) error {
+	// Serialize the concurrent queries racing on the same folder (the
+	// rag-query worker runs several jobs in parallel): without this, two
+	// first-time chats could both create and backfill the workspace, and the
+	// rollback of a failed backfill could delete the workspace the other
+	// query just successfully ensured.
+	mu := config.Lock().ReadWrite(inst, "rag-workspace/"+dirID)
```
shepilov: it's only 20 sec by default; I think you need LongOperation here. 

Or maybe it's also not a good idea to have such a big critical section.

model/rag/workspace.go
```diff
@@ -0,0 +1,505 @@
...
+	if !createdOrExists(createRes.StatusCode) {
+		return fmt.Errorf("workspace creation status code: %d", createRes.StatusCode)
+	}
+
+	for chunk := range slices.Chunk(fileIDs, workspaceBackfillChunkSize) {
+		if err := backfillChunk(server, domain, dirID, chunk, logger); err != nil {
+			// The workspace was just created but is incomplete, and a later
+			// ensureWorkspace call would see it exists (GET 200) and never
+			// retry the backfill: delete it so the next call starts over.
+			deleteWorkspace(server, domain, dirID, logger)
```
shepilov, line 349: Are you sure about it? it's better than incomplete?

model/rag/workspace.go
```diff
@@ -0,0 +1,505 @@
...
+// non-indexed files (e.g. an image skipped by the indexer's class flags) and
+// are only logged, not treated as an ensure failure. A 5xx status (chunked
+// or per-id) means the indexer itself failed and is still an ensure failure
+// so the query does not proceed unscoped.
+func backfillChunk(server config.RAGServer, domain, dirID string, ids []string, logger logger.Logger) error {
+	status, err := postWorkspaceFiles(server, domain, dirID, ids)
+	if err != nil {
+		return err
+	}
+	if status != http.StatusNotFound {
```
shepilov, line 418: 403 we only log and return nil?

model/rag/workspace.go
```diff
@@ -0,0 +1,505 @@
+package rag
```
shepilov, line 1: sorry, I was struggling reading all in one, could we just structio it a little bit, and split on to componenets:

```

type WorkspaceClient interface {
      GetWorkspace(ctx context.Context, partition, id string) (*Workspace, error)
      CreateWorkspace(ctx context.Context, partition string, workspace Workspace) error
      DeleteWorkspace(ctx context.Context, partition, id string) error
      AddFiles(ctx context.Context, partition, workspaceID string, fileIDs []string) error
      GetFileWorkspaces(ctx context.Context, partition, fileID string) ([]string, error)
      RemoveFile(ctx context.Context, partition, workspaceID, fileID string) error
  }

 type WorkspaceManager struct {
      client WorkspaceClient
      files  FileTree
      locks  lock.Getter
  }

func (m *WorkspaceManager) EnsureReady(
      ctx context.Context,
      scope RetrievalScope,
  ) (WorkspaceID, error)
```

model/rag/workspace.go
```diff
@@ -0,0 +1,505 @@
...
+
```
shepilov, line 2: Maybe also something like to resolve scope and check permissions(not sure)

```golang

type RetrievalScopeResolver struct {
      assistants AssistantRepository
      permissions PermissionChecker
  }

```

shepilov: Also it would be greate to have type error returned from client:

 ErrAssistantNotFound
  ErrScopeDenied
  ErrFolderNotFound
  ErrWorkspaceUnavailable
  ErrMembershipOutOfSync

model/rag/index.go
```diff
@@ -71,7 +76,7 @@ func Index(inst *instance.Instance, logger logger.Logger, msg IndexMessage) erro
 	return errj
 }
 
-func callRAGIndexer(inst *instance.Instance, doctype string, change couchdb.Change) error {
+func callRAGIndexer(inst *instance.Instance, doctype string, change couchdb.Change, kb func() *kbContext, logger logger.Logger) error {
```
shepilov, line 79: Also, let's plan and split this function into components; it's just becoming bigger and bigger

shepilov: And in general, RAG index representation in the stack as a model

model/rag/workspace.go
paultranvan: I guess it's debatable, but I rather have an error sent to the client, than a silent incomplete creation. Especially for now that we are testing relevance of ai chat

model/rag/index.go
paultranvan: I guess we can make a separated PR about that, it is already a separated commit, but it seemed easier to me to group it so we have the full picture.
About the sync, the openRAG API is made in a way that we need a way to have this sync mechanism. This is something that I want to eventually change and make it dynamic so we don't require this sync anymore, that I agree is somewhat complicated. But in the meantime, we need to have it to deliver the workspace feature

model/rag/chat.go
paultranvan: Right, fixed in [73d5ecf](https://github.com/linagora/cozy-stack/pull/4856/commits/73d5ecf6f3455228f9035d39a2e9629063ea638a)

model/rag/workspace.go
paultranvan: Done in [73d5ecf](https://github.com/linagora/cozy-stack/pull/4856/commits/73d5ecf6f3455228f9035d39a2e9629063ea638a)

model/rag/workspace.go
paultranvan: Long operation done in [73d5ecf](https://github.com/linagora/cozy-stack/pull/4856/commits/73d5ecf6f3455228f9035d39a2e9629063ea638a)

shepilov (approved):

PR #4857, opened 2026-07-21
feat(sharings): Add effective access resolver

### LINKED ISSUE #4837: Add effective access resolver for additive nested shared folders
## Context

  Nested folder sharing is additive in v1:

  - parent folder access continues to apply to child folders
  - child folder shares can add users or grant higher rights
  - if a user has access through multiple sharing scopes, the highest permission wins
  - file shares remain additive and do not create boundaries
  - `limited_access` is reserved for a future feature

  Before endpoints can safely allow nested folder shares, Cozy Stack needs shared backend helpers to compute effective access.

### Option A: Copy inherited members into every child share

  Pros:

  - permission checks stay simple
  - each child sharing doc contains a complete member list
  - less resolver logic needed initially

  Cons:

  - parent membership changes require updating all nested child shares
  - removing a user from a parent share risks stale copied access in children
  - every inherited member must be replicated to every child share recipient
  - move/merge behavior becomes harder to reason about
  - child sharing docs become ambiguous: direct member, inherited member, copied member, or stale member

  ### Option B: Access on the fly

  Pros:

  - parent sharing remains the source of truth
  - no cascade updates when parent members change
  - revocation is immediate when parent access is removed
  - child sharing docs store only direct grants
  - works better with future `limited_access`

  Cons:

  - permission checks are more complex
  - requires efficient CouchDB indexes
  - needs request-level caching/batching to avoid repeated lookups
  - archive/realtime/sync flows need extra care later

## Decision

  Use **derived effective access**.

  Do not copy inherited parent members into child sharing docs during normal sharing updates.

  ## Domain Model

 Add/formalize access mode on io.cozy.sharings:

  io.cozy.sharings
  - access_mode: additive | limited_access

  Rules:

  - missing access_mode is treated as additive
  - additive is the only enabled mode for this task
  - limited_access is reserved for later
  - io.cozy.files.referenced_by remains the marker for shared roots
  - file-root shares are not restrictive boundaries

Internal service model:

```golang
 type SharingScope struct {
      SharingID  string
      RootID     string
      RootPath   string
      AccessMode string
      ReadOnly   bool
  }

  type EffectiveAccess struct {
      CanRead          bool
      CanWrite         bool
      SourceSharingIDs []string
  }

  type AccessResolver struct {
      inst *instance.Instance
      fs   vfs.VFS
  }
 
 func NewAccessResolver(inst *instance.Instance) *AccessResolver
 func (r *AccessResolver) Resolve(targetID string) (*EffectiveAccess, error)

access, err := resolver.Resolve(target vfs.Fetche)
  if err != nil {
      return err
  }
  if !access.Can(permission.GET) {
      return jsonapi.Forbidden(...)
  }

// Internal/private helpers can exist, but not as public API:

  func (r *AccessResolver) scopesFor(targetID string) ([]sharingScope, error)
  func (r *AccessResolver) ancestorPaths(targetID string) ([]string, error)
  func (r *AccessResolver) loadSharings(ids []string) ([]*Sharing, error)
```

### Index

No index

We can have a Mango query instead of this index when we query all the files with the parent path using the same path index and then getinng `sharing_id`s from the `reference_by` list

```
req := &couchdb.FindRequest{
        UseIndex: "dir-by-path",
        Selector: mango.And(
                mango.In("path", ancestorPathsAsInterfaces),
                mango.Equal("type", consts.DirType),
                mango.Exists(couchdb.SelectorReferencedBy),
        ),
        Fields: []string{"_id", "path", "referenced_by"},
        Limit:  len(ancestorPaths),
  }
```

 Flow:
   1. Read target path from the already loaded file/folder.
  2. Build ancestor folder paths.
  3. Query shared-folder-roots-by-path once with ViewRequest.Keys.
  4. Bulk-load returned sharing docs.
  5. Keep active additive sharing scopes.
  6. Merge current user permissions.

taratatach:
Add `AccessResolver` to compute the effective access to a file or
  folder from the additive sharing scopes on its path. `Resolve(targetID)`
  returns an `*EffectiveAccess` with `CanRead`/`CanWrite` aggregated
  highest-wins across all active additive scopes where the current
  instance is a member (via `Sharing.MemberFor`). The target's own
  share (for shared files) and every shared ancestor directory are
  considered. `EffectiveAccess.Can(verb)` maps `GET` to `CanRead` and
  the write verbs to `CanWrite`. `limited_access` scopes are filtered
  out in v1; `NearestRestrictiveBoundary` and `ChildSharedRootsUnder`
  are reserved as no-op stubs for that future mode.

  Fixes #4837

### DIFF
--- a/model/sharing/effective_access.go
+++ b/model/sharing/effective_access.go
@@ -1,0 +1,271 @@
+package sharing
+
+import (
+	"os"
+	"path"
+
+	"github.com/cozy/cozy-stack/model/instance"
+	"github.com/cozy/cozy-stack/model/permission"
+	"github.com/cozy/cozy-stack/pkg/consts"
+	"github.com/cozy/cozy-stack/pkg/couchdb"
+	"github.com/cozy/cozy-stack/pkg/couchdb/mango"
+)
+
+// SharingScope describes a single sharing scope that applies to a target file
+// or folder. It is produced by AccessResolver.scopesFor and consumed by
+// Resolve to aggregate an EffectiveAccess. ReadOnly reflects the current
+// instance's membership for that scope (via Sharing.MemberFor).
+type SharingScope struct {
+	SharingID  string
+	RootID     string
+	RootPath   string
+	AccessMode string
+	ReadOnly   bool
+}
+
+// EffectiveAccess is the merged result of all applicable additive sharing
+// scopes for a target, from the perspective of the current instance. CanRead
+// is true if the instance is a member of at least one applicable scope;
+// CanWrite is true if at least one such scope grants write (highest right
+// wins).
+type EffectiveAccess struct {
+	CanRead          bool
+	CanWrite         bool
+	SourceSharingIDs []string
+}
+
+// Can reports whether the effective access satisfies the given verb. GET
+// requires CanRead; POST, PUT, PATCH and DELETE require CanWrite.
+func (ea *EffectiveAccess) Can(verb permission.Verb) bool {
+	if ea == nil {
+		return false
+	}
+	switch verb {
+	case permission.GET:
+		return ea.CanRead
+	case permission.POST, permission.PUT, permission.PATCH, permission.DELETE:
+		return ea.CanWrite
+	}
+	return false
+}
+
+// AccessResolver computes the effective access for a target file or folder,
+// covering the target itself (if shared) plus all shared ancestors on its
+// path, filtered to scopes where the current instance is a member.
+type AccessResolver struct {
+	inst *instance.Instance
+}
+
+// NewAccessResolver builds an AccessResolver bound to an instance.
+func NewAccessResolver(inst *instance.Instance) *AccessResolver {
+	return &AccessResolver{inst: inst}
+}
+
+// Resolve returns the effective access for the given file or folder. The
+// target's own share (if it is a shared file) and the shares of every
+// ancestor directory are all considered. Scopes where the current instance
+// is not a member are ignored. The highest permission wins: CanRead if at
+// least one applicable scope, CanWrite if at least one non-read-only scope.
+func (r *AccessResolver) Resolve(targetID string) (*EffectiveAccess, error) {
+	scopes, err := r.scopesFor(targetID)
+	if err != nil {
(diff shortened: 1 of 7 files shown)


model/sharing/effective_access.go
```diff
@@ -0,0 +1,268 @@
...
+// ones. limited_access sharings are dropped in v1. Membership filtering is
+// done by the caller (scopesFor) so this helper stays reusable.
+func (r *AccessResolver) loadSharings(ids []string) ([]*Sharing, error) {
+	sharings, err := FindSharings(r.inst, ids)
+	if err != nil {
+		return nil, err
+	}
+	kept := make([]*Sharing, 0, len(sharings))
+	for _, s := range sharings {
+		if !s.Active {
```
shepilov, line 244: we should filter here also to only drive shares if we don't care about classical ones, and for classical ones we can have `read-only` rules

shepilov, line 244: and as far as we already have this if it would be great to add `!=nil` as far as FindSharings can return nil for missing ids

shepilov (approved):

model/sharing/sharing.go
```diff
@@ -67,6 +75,7 @@ type Sharing struct {
 	Active        bool      `json:"active,omitempty"`
 	Owner         bool      `json:"owner,omitempty"`
 	Open          bool      `json:"open_sharing,omitempty"`
+	AccessMode    string    `json:"access_mode,omitempty"` // additive (default) | limited_access (reserved)
```
shepilov, line 78: We need not forget to add this field to all clone/copy functions when we are sending sharing info to recipients, real-time events, etc ...

model/sharing/effective_access.go
taratatach: I don't think the current implementation of the resolver makes much sense for drive shares as we built scopes from local vfs docs and their references to sharings while we don't have these docs on drives recipients (and the owner always has full access).

taratatach: As discussed, we'll need more to check permission for drives recipient members but it's planned in the next issue.

PR #4858, opened 2026-07-21
fix(sharing): refresh delegated recipient removal tokens

shepilov:
## Summary

- Refresh an expired sharing OAuth token before retrying same-stack delegated recipient removal.
- Clear cached authorization state and stale authentication response headers between attempts.
- Cover local-owner removal with an expired token and persisted refreshed credentials.

### DIFF
--- a/web/sharings/revoke.go
+++ b/web/sharings/revoke.go
@@ -6,6 +6,8 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/cozy/cozy-stack/model/instance"
+	"github.com/cozy/cozy-stack/model/permission"
 	"github.com/cozy/cozy-stack/model/sharing"
 	"github.com/cozy/cozy-stack/pkg/jsonapi"
 	"github.com/cozy/cozy-stack/web/middlewares"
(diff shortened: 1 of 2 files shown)


web/sharings/revoke.go
```diff
@@ -55,23 +57,48 @@ func RevokeRecipient(c echo.Context) error {
...
+	}
+	if credentials.AccessToken == nil {
+		return wrapErrors(sharing.ErrNoOAuthClient)
+	}
+	return callRevokeRecipientOnOwner(c, owner, credentials.AccessToken.AccessToken)
+}
+
+func callRevokeRecipientOnOwner(c echo.Context, owner *instance.Instance, accessToken string) error {
+	// Re-enter the owner-side handler so delegated calls keep the same authorization path.
+	c.Response().Header().Del(echo.HeaderWWWAuthenticate)
```
taratatach, line 94: Why do we remove this header here?

taratatach (approved):

web/sharings/revoke.go
shepilov: The first attempt suppose to add WWW-Authenticate: Bearer error="invalid_token" and it will remain in a successfull response

PR #4871, opened 2026-07-23
fix(common): build matrix id from email localpart

### LINKED ISSUE #4870: [common settings] Matrix ID built from domain slug instead of email localpart
## Context

The stack pushes user profiles to the external common-settings API, and the bridge uses `matrix_id` as the canonical Matrix identity.

## Problem

The Matrix ID localpart is built from the instance domain slug, not from the user's email. The slug is dot-stripped, so a user whose email localpart is `al.ice.wonder.land` on `stg.lin-saas.com` is sent as `@alicewonderland:stg.lin-saas.com` instead of `@al.ice.wonder.land:stg.lin-saas.com`. The bridge then provisions a new, unrelated Matrix user.

The Matrix localpart can differ from the slug, and emails can contain dots, so the slug cannot reconstruct it.

## Expected behavior

Build the Matrix ID localpart from the email local part (segment before `@`), which preserves dots, keeping the homeserver part derived from the instance domain. For email `al.ice.wonder.land@stg.lin-saas.com` the payload should carry `@al.ice.wonder.land:stg.lin-saas.com`.

cc @pm-McFly

rezk2ll:
## Problem

The common-settings payload built the Matrix ID localpart from the instance domain slug. The slug is dot-stripped and can differ from the real localpart, so a user with email localpart `al.ice.wonder.land` was sent as `@alicewonderland:...` instead of `@al.ice.wonder.land:...`, and the bridge provisioned a new, unrelated Matrix user.

## Solution

Build the localpart from the email local part (the segment before `@`), lowercased, which preserves dots. The homeserver part stays domain-derived, and the slug is used only as a fallback when no email is set.

Closes #4870

### DIFF
--- a/model/settings/common/common.go
+++ b/model/settings/common/common.go
@@ -308,7 +308,15 @@ func buildRequest(inst *instance.Instance, settings *couchdb.JSONDoc) UserSettin
 		request.Payload.Phone = phone
 	}
 	if len(parts) > 1 {
-		id := fmt.Sprintf("@%s:%s", nickname, strings.Join(parts[1:], "."))
+		// The Matrix localpart comes from the email local part, which preserves
+		// dots and can differ from the domain slug. Fall back to the slug when
+		// no email is available. The homeserver part stays domain-derived.
+		localpart := nickname
+		email, _ := settings.M["email"].(string)
+		if local, _, found := strings.Cut(email, "@"); found && local != "" {
+			localpart = strings.ToLower(local)
+		}
+		id := fmt.Sprintf("@%s:%s", localpart, strings.Join(parts[1:], "."))
 		request.Payload.MatrixID = id
 	}
 
--- a/model/settings/common/common_test.go
+++ b/model/settings/common/common_test.go
@@ -10,6 +10,41 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestBuildRequest_MatrixIDFromEmail(t *testing.T) {
+	inst := &instance.Instance{Domain: "alicewonderland.stg.lin-saas.com"}
+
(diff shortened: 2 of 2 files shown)


model/settings/common/common.go
```diff
@@ -301,14 +301,22 @@ func buildRequest(inst *instance.Instance, settings *couchdb.JSONDoc) UserSettin
...
 	}
-	if email, ok := settings.M["email"].(string); ok {
+	email, _ := settings.M["email"].(string)
+	if email != "" {
 		request.Payload.Email = email
 	}
 	if phone, ok := settings.M["phone"].(string); ok {
 		request.Payload.Phone = phone
 	}
 	if len(parts) > 1 {
```
shepilov, line 310: why we are doing in in stack in general and not in registration? or whatever is near to user identity?

model/settings/common/common.go
```diff
@@ -301,14 +301,22 @@ func buildRequest(inst *instance.Instance, settings *couchdb.JSONDoc) UserSettin
 		request.Payload.FirstName = parts[0]
 		request.Payload.LastName = parts[len(parts)-1]
 	}
-	if email, ok := settings.M["email"].(string); ok {
+	email, _ := settings.M["email"].(string)
```
shepilov: mmm, how it's connected to matrixid, and why not empty emails?

shepilov (approved):

model/settings/common/common.go
rezk2ll: Yes, the stack has no notion of a Matrix identity, but it's currently the only one building and sending this information for common settings.

i agree it's shouldn't be here in the first place:
- either registration has to emit it
- or whatever consumer has to drop it ( since using the email you can tell which user to update )

PR #4872, opened 2026-07-23
fix(sharing): keep empty organization drives active

shepilov:
## Summary

- Keep owner-side organization-drive sharings active when their last recipient or group is revoked.
- Preserve the root sharing reference so empty organization drives remain in the active drive listing.
- Cover last-recipient, last-group, and recipient re-addition behavior with regression tests.

### DIFF
--- a/model/sharing/sharing.go
+++ b/model/sharing/sharing.go
@@ -754,9 +754,8 @@ func (s *Sharing) RemoveInteractPermissionsForAMember(inst *instance.Instance, m
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


web/sharings/drives_test.go
taratatach, line 1: Looks like we're actually testing adding recipients rather than reading them, no?

model/sharing/sharing.go
taratatach, line 1: This could be merged with the first commit.

taratatach (approved):

PR #4873, opened 2026-07-23
fix(bitwarden): find the org member by the old domain of the instance

Crash--:
## Summary

- Add `Organization.Member(inst)` which looks up the member entry by the instance domain, then falls back on `OldDomain` and on the domain aliases.
- Use it in `newOrganizationResponse`, `newCollectionResponse` and the two owner checks of `web/bitwarden/organization.go`, which read the map with `Members[inst.Domain]`.
- The members of an organization are keyed by the domains of their instances, so an instance migrated to a new domain no longer matches its own entry: the organization is returned with an empty key and an Invited status, and the shared ciphers cannot be decrypted.
- `PassphraseSalt` already uses the same `OldDomain` fallback for the master password salt.
- Add unit tests for the lookup.

### DIFF
--- a/model/bitwarden/organization.go
+++ b/model/bitwarden/organization.go
@@ -87,23 +87,34 @@ func (o *Organization) Clone() couchdb.Doc {
 // organization. The members are keyed by the instances domains, and an
 // instance that has been migrated to a new domain still has its entry stored
 // under its former domain: the old domain and the domain aliases are used as
-// fallbacks so that such an instance keeps its access to the organization. It
-// returns a zero value when the instance is not a member.
+// fallbacks so that such an instance keeps its access to the organization.
+// An entry with the organization key has the precedence, as a keyless entry
+// can have been created under the new domain while the key stayed on the
+// entry of the former domain. It returns a zero value when the instance is
+// not a member.
 func (o *Organization) Member(inst *instance.Instance) OrgMember {
-	if m, ok := o.Members[inst.Domain]; ok {
-		return m
-	}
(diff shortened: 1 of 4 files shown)


rezk2ll (approved):
👍 better than having to migrate immediately.

worth checking also model/sharing/member.go, the SaveBitwarden method
i know it's not directly related but , if the recipients of the share also migrate they will still continue to have the issue since they don't have the fallback

shepilov (approved):

Crash--:
Good catch, and the hole was wider than the read path: `SaveBitwarden` runs on the owner instance and keys the member on the host of `m.Instance`. For a migrated recipient that host is the new one, so it wrote a keyless entry under the new domain and left the original entry orphaned under the former one. That also defeated the fallback of this PR, since the lookup found an entry under the current domain and never reached `OldDomain`.

Two commits added:

- `SaveBitwarden` now looks the member up by user id, which is stable across a domain migration, reuses its organization key and removes the stale entry. `ConfirmUser` already matches members that way.
- `Organization.Member` gives the precedence to an entry that has the key, so instances already carrying the keyless duplicate are repaired instead of staying broken.

Crash--:
Correction to my previous comment: `SaveBitwarden` no longer removes the entry of the former domain. Recovering the organization key is what restores the access, the deletion was only cleanup, and it is the kind of thing that should not ride along in a fix on a document replicated to the other members. The duplicate entry is harmless since `Organization.Member` prefers the entry that has the key.

PR #4876, opened 2026-07-24
fix(ai): safe assistant system-prompt injection, scoped file permission

ewan102:
Assistant custom prompt was never applied to conversations : 
When a conversation was created with an assistantID, the assistant's prompt field was silently ignored — the LLM never saw it. The fix looks up the assistant doc in Chat() and, if it has a non-empty prompt, prepends it as a role: "system" message before the user's own message, consumed as-is by OpenRAG's chat completion format.

 Missing permission checks on files referenced by a conversation : 

Previously, the chat route only checked permission on io.cozy.ai.chat.conversations — any attachment ID or assistant-scoped knowledge-base folder passed in the payload was used without any permission check at all. The fix adds explicit checks: each attachment ID is resolved to its file doc and checked with middlewares.Allow(GET, fileDoc), and when an assistant is scoped to a knowledge-base folder, that folder is checked the same way — so a client can no longer read files it has no permission on just by referencing them in a chat request.

### DIFF
--- a/docs/ai.md
+++ b/docs/ai.md
@@ -173,7 +173,10 @@ Content-Type: application/json
 - `stream` enables streaming the response via SSE deltas (defaults to `true`).
 - `websearch` enables web search for the query (defaults to `false`).
 - `assistantID` (optional) associates the conversation with an `io.cozy.ai.chat.assistants`
-  document. When set, the response includes a `relationships` block.
+  document. When set, the response includes a `relationships` block. If the assistant has a
+  non-empty `prompt` field, a `role: "system"` message carrying that prompt is prepended to
+  `messages` when the conversation is created (before the user's own message). Clients should
+  not render `system` messages as regular conversation turns.
 - `attachmentIDs` (optional) array of ids, specifying which documents should be leveraged by the RAG.
   
 
--- a/web/ai/ai.go
+++ b/web/ai/ai.go
@@ -23,6 +23,30 @@ func Chat(c echo.Context) error {
 	}
 	payload.ChatConversationID = c.Param("id")
 	inst := middlewares.GetInstance(c)
+	for _, id := range payload.AttachmentIDs {
+		fileDoc, err := inst.VFS().FileByID(id)
+		if err != nil {
+			return middlewares.ErrForbidden
+		}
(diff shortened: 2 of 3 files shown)


web/ai/ai.go
```diff
@@ -23,6 +23,30 @@ func Chat(c echo.Context) error {
...
+		if dirID != "" {
+			dirDoc, err := inst.VFS().DirByID(dirID)
+			if err != nil {
+				return middlewares.ErrForbidden
+			}
+			if err := middlewares.Allow(c, permission.GET, dirDoc); err != nil {
+				return middlewares.ErrForbidden
+			}
+		}
+	}
```
paultranvan, line 49: You don't need to do all of this. Just this, at the beggining of the function, after the first permission check: 
```
if err := middlewares.AllowWholeType(c, permission.GET, consts.Files); err != nil {
  return middlewares.ErrForbidden
}
```

This will check the request has rights on files doctype. We don't need to check every single file.

docs/ai.md
```diff
@@ -173,7 +173,10 @@ Content-Type: application/json
 - `stream` enables streaming the response via SSE deltas (defaults to `true`).
 - `websearch` enables web search for the query (defaults to `false`).
 - `assistantID` (optional) associates the conversation with an `io.cozy.ai.chat.assistants`
-  document. When set, the response includes a `relationships` block.
+  document. When set, the response includes a `relationships` block. If the assistant has a
+  non-empty `prompt` field, a `role: "system"` message carrying that prompt is prepended to
+  `messages` when the conversation is created (before the user's own message). Clients should
+  not render `system` messages as regular conversation turns.
```
paultranvan, line 179: I don't think we should document it here, this should be restricted to the API endpoint. 
However this behaviour should indeed be documented, and I realized we never took the time to do the doc for io.cozy.ai.chat.assistants it on https://github.com/cozy/cozy-doctypes/tree/master/docs , so we should do it

shepilov:
And could you please split it into two pr, it's unrelated issues

ewan102:
This PR got divided into to PR as requested : #4881 and #4882, except the doc for io.cozy.ai.chat.assistants, which will be on cozy-doctype

PR #4878, opened 2026-07-27
feat: add forced session codes to intent hrefs

### LINKED ISSUE #4877: Add optional session code injection to intent service hrefs
## Context

PR #4850 lets the mobile intent/file-picker flow open a webapp URL with a `session_code` in the URL. The next step is to remove one UI round-trip when the intent is created.

Today the UI needs to create an intent, call `/auth/session_code`, then patch the selected service URL client-side. Instead, `/intents` should be able to return service hrefs that are already ready to open.

## Expected behavior

- Keep the default `/intents` behavior unchanged.
- When `POST /intents?force_session_id=true` is used, mint a session code and inject it into the returned `services[].href` values.
- Preserve the existing intent query parameter in each service href, for example `...?intent=<intent-id>&session_code=<code>`.
- Apply this only to the response; do not persist the generated `session_code` in the stored `io.cozy.intents` document.
- Reuse the same token-based authorization/session-code rules as `/auth/session_code` for eligible callers, especially the verified app-token-exchange flow. Do not trust claims unless token validation succeeded.
- If `force_session_id=true` is requested but the caller is not eligible to mint a session code, fail closed with an authorization error instead of returning hrefs without a session code.

## Implementation notes

A good implementation path is to extract a small auth helper for token-authorized session-code minting and reuse it from both `/auth/session_code` and `/intents`. The `/intents` path should not use the passphrase fallback from `/auth/session_code`; it should only work for already-authenticated token flows.

shepilov:
## Summary

- Add `force_session_id=true` handling on `/intents` to return service hrefs with a session code.
- Reuse token authorization and session-code minting helpers for eligible callers.
- Keep session codes out of stored intent documents.
- Closes #4877.

### DIFF
--- a/web/auth/flagship.go
+++ b/web/auth/flagship.go
@@ -30,6 +30,12 @@ const (
 	SessionCodeSourcePassword = "password"
 )
 
+// SessionCodeGrant authorizes a token caller to mint a session code.
+type SessionCodeGrant struct {
+	Permission *permission.Permission
+	Source     string
+}
+
 // CreateSessionCode is the handler for creating a session code by the flagship
 // app.
 func CreateSessionCode(c echo.Context) error {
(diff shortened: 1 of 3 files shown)


zatteo:
Tested on Android emulator and works well 👍

taratatach (approved):

PR #4881, opened 2026-07-27
fix(ai): inject assistant custom prompt as a system message

ewan102:
The assistant's prompt field was never applied to conversations. Now, when a conversation is created with an assistantID, its non-empty prompt is prepended as a role: system message before the user's message, consumed as-is by OpenRAG's chat completion format.

### DIFF
--- a/model/rag/chat.go
+++ b/model/rag/chat.go
@@ -55,6 +55,7 @@ type ChatMessage struct {
 const (
 	UserRole      = "user"
 	AssistantRole = "assistant"
+	SystemRole    = "system"
 	Temperature   = 0.3   // LLM parameter - Sampling temperature, lower is more deterministic, higher is more creative.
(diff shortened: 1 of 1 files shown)


model/rag/chat.go
```diff
@@ -197,6 +199,16 @@ func Chat(inst *instance.Instance, payload ChatPayload) (*ChatConversation, erro
 					},
 				},
 			}
+			var assistant chatAssistant
```
shepilov, line 202: Are we allowed to create chats without a system prompt, or is it optional thing?

shepilov (approved):

model/rag/chat.go
ewan102: Yes at the moment we can, the system prompt is just the current "description" paragraph whenever you create an assistant.

PR #4882, opened 2026-07-27
fix(ai): check files permission on chat requests referencing files

ewan102:
The chat route only checked permission on io.cozy.ai.chat.conversations; attachmentIDs and assistant-scoped knowledge base folders were used without any permission check. Now a whole-type GET check on io.cozy.files is performed whenever attachments or an assistant are referenced.

### DIFF
--- a/web/ai/ai.go
+++ b/web/ai/ai.go
@@ -23,10 +23,8 @@ func Chat(c echo.Context) error {
 	}
 	payload.ChatConversationID = c.Param("id")
 	inst := middlewares.GetInstance(c)
-	if len(payload.AttachmentIDs) > 0 || payload.AssistantID != "" {
-		if err := middlewares.AllowWholeType(c, permission.GET, consts.Files); err != nil {
-			return middlewares.ErrForbidden
-		}
+	if err := middlewares.AllowWholeType(c, permission.GET, consts.Files); err != nil {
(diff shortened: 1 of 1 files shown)


web/ai/ai.go
```diff
@@ -23,6 +23,11 @@ func Chat(c echo.Context) error {
 	}
 	payload.ChatConversationID = c.Param("id")
 	inst := middlewares.GetInstance(c)
+	if len(payload.AttachmentIDs) > 0 || payload.AssistantID != "" {
```
shepilov: Why do we have two fields in the payload: `AttachmentIDs` and `AssistantID`?

shepilov: Is this field mandatory? And shouldn't we always check these permissions?

shepilov:
And we are still missing a permission check during workspace creation; also, shouldn't we check that these files are inside the selected workspace ?

web/ai/ai.go
ewan102: Yes you are right... not mandatory and we should always check them

ewan102: Because AssistantID and AttachmentIDs serve purely different purposes : attachmentIDs is a "chat by chat" attribute, could change every chat, while AssistantID configures the whole conversation

shepilov (approved):

PR #4885, opened 2026-07-28
feat(sharings): Effective recipients endpoints

### LINKED ISSUE #4840: Effective recipients endpoints for nested shared folders
## Context

Nested shared folders use additive access: a child folder can add recipients or stronger rights, while parent access still applies to descendants.
`io.cozy.sharings.members` represents only direct members of one sharing document. The Drive UI needs another view: the combined list of everyone who can access a selected file or folder, including inherited recipients from parent shared folders.

  ## Endpoints

  - `GET /sharings/recipients/:file-id`
  - `GET /sharings/drives/:drive-id/recipients/:file-id`

## User journey

  1. User opens Drive.
  2. User selects one file or folder.
  3. User clicks Share.
  4. UI opens the share modal.
  5. UI calls the effective recipients endpoint for this single file/folder.
  6. UI displays:
     - direct recipients on this share,
     - inherited recipients from parent shared folders,
     - whether each recipient can be edited from this modal.

  ## Backend behavior

  - Resolve all additive sharing scopes that apply to the target.
  - Collect direct members from each applicable `io.cozy.sharings` document.
  - Deduplicate recipients by stable identity, preferably `instance`, then email fallback.
  - Merge permissions so read-write wins over read-only.
  - Return source metadata so UI can explain where access comes from.
  - Mark inherited-only recipients as `can_edit_here=false`.
  - Do not mutate sharing docs.
  - Do not copy inherited members into child sharing docs.

## Response

  ```json
     {
    "file_id": "target-file-id",
    "recipients": [
      {
        "name": "Alice",
        "email": "alice@example.com",
        "instance": "https://alice.example.com",
        "status": "ready",
        "read_only": false,
        "sources": [
          {
            "sharing_id": "parent-sharing-id",
            "root_id": "parent-folder-id",
            "root_name": "Project",
            "kind": "ancestor",
            "member_index": 1,
            "read_only": false,
            "manageable": false
          }
        ]
      }
    ]
  }
```

 ## Permission checks

  - Regular endpoint: caller must have at least read access to the requested file/folder.
  - Shared-drive endpoint: caller must be a member of :drive-id, and :file-id must belong to this shared drive.
  - The endpoint is read-only, so it must not require write permission just to display access.

## Client integration

  Add cozy-stack-client helpers:

  client.collection('io.cozy.sharings').fetchEffectiveRecipients({ id })

  client.collection('io.cozy.sharings').fetchSharedDriveEffectiveRecipients({
    driveId,
    id
  })

### PR DESCRIPTION
**Pull request by taratatach**
Closes #4840

## What

Two new read-only endpoints returning the combined list of people who can
access a file or folder, including access inherited from parent shared
folders:

- `GET /sharings/recipients/:file-id`
- `GET /sharings/drives/:id/recipients/:file-id`

## How

- `AccessResolver.applicableSharings` (extracted from `scopesFor`) resolves
  every active additive scope applying to the target — its own share plus
  shares inherited from ancestor folders — with no membership filtering.
- `EffectiveRecipients` collects the direct members of every applicable
  `io.cozy.sharings` doc (revoked members excluded), deduplicates by
  `instance` with an `email` fallback, lets read-write win over read-only,
  and keeps the most advanced status.
- Each recipient exposes its `sources` (`sharing_id`, `root_id`,
  `root_name`, `kind`, `member_index`, `read_only`, `manageable`) and a
  `can_edit_here` flag (true when at least one source is the target's own
  share).
- JSON-API response, with the target `file_id` in `meta`.

## Security / authorization

- Classic endpoint: the caller needs at least read access to the file.
- Drive endpoint: the caller must be a member of the drive (active drive)
  and the target must belong to that drive.
- Public share tokens (share-by-link, share preview) are rejected: a link
  holder must not be able to enumerate sharing members.

No mutation: these are read-only endpoints, no inherited member is copied
into child sharings.

## Tests

- `model/sharing/effective_recipients_test.go`: deduplication
  (instance/email/bridge between the two), read-only/status merging,
  revoked exclusion, `manageable`.
- `web/sharings/effective_recipients_test.go`: both endpoints, inheritance
  from ancestors, public token rejection, error cases (target outside the
  drive, inactive drive, non-member).

### DIFF
--- a/model/sharing/effective_recipients.go
+++ b/model/sharing/effective_recipients.go
@@ -1,0 +1,184 @@
+package sharing
+
+import (
+	"fmt"
+	"strings"
+)
+
+// RecipientSource describes one sharing scope through which a recipient has
+// access to the target: which sharing, which root, and how the access can be
+// managed from the target's share modal.
+type RecipientSource struct {
+	SharingID   string `json:"sharing_id"`
+	RootID      string `json:"root_id"`
+	RootName    string `json:"root_name"`
+	Kind        string `json:"kind"` // "self" | "ancestor"
+	MemberIndex int    `json:"member_index"`
+	ReadOnly    bool   `json:"read_only"`
+	Manageable  bool   `json:"manageable"`
+}
+
+// EffectiveRecipient is a deduplicated person who can access the target,
+// either through the target's own share ("self") or inherited from a shared
+// ancestor ("ancestor"). ReadOnly is merged across sources with read-write
+// winning over read-only. CanEditHere is true when at least one source is
+// the target's own share.
+type EffectiveRecipient struct {
+	Name        string            `json:"name"`
+	Email       string            `json:"email"`
+	Instance    string            `json:"instance"`
+	Status      string            `json:"status"`
+	ReadOnly    bool              `json:"read_only"`
+	CanEditHere bool              `json:"can_edit_here"`
+	Sources     []RecipientSource `json:"sources"`
+}
+
+// EffectiveRecipients returns the combined list of people who can access the
+// given file or folder: the direct members of every active additive sharing
+// scope applying to the target (its own share plus inherited ancestor
+// shares). Revoked members are excluded. Recipients are deduplicated by
+// instance, with an email fallback. It is a read-only view: no sharing
+// document is mutated and no inherited member is copied anywhere.
+func (r *AccessResolver) EffectiveRecipients(targetID string) ([]EffectiveRecipient, error) {
+	sharings, rootBySharing, err := r.applicableSharings(targetID)
+	if err != nil {
+		return nil, err
+	}
+
+	var recipients []EffectiveRecipient
+	index := make(map[string]int)
+	dropped := make(map[int]bool)
+	for _, s := range sharings {
+		info, ok := rootBySharing[s.SID]
+		if !ok {
+			continue
+		}
+		kind := "ancestor"
+		if info.RootID == targetID {
+			kind = "self"
+		}
+		canManage := false
+		if self := s.MemberFor(r.inst); self != nil && !self.ReadOnly {
+			canManage = true
+		}
+		for i := range s.Members {
+			m := &s.Members[i]
+			if m.Status == MemberStatusRevoked {
+				continue
+			}
+			candidate := EffectiveRecipient{
+				Name:     m.PrimaryName(),
+				Email:    m.Email,
+				Instance: m.Instance,
+				Status:   m.Status,
+				ReadOnly: m.ReadOnly,
+				Sources: []RecipientSource{{
+					SharingID:   s.SID,
+					RootID:      info.RootID,
+					RootName:    info.RootName,
+					Kind:        kind,
+					MemberIndex: i,
+					ReadOnly:    m.ReadOnly,
+					// i != 0: RevokeRecipient rejects the owner entry (index 0),
+					// so it must not be reported as manageable.
+					Manageable: kind == "self" && canManage && i != 0,
+				}},
+			}
+			candidate.CanEditHere = kind == "self"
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

**Comment by shepilov** [comment]
why do we need 2 apis? why not only `sharings/drives/:id/recipients/:file-id` ?

**Comment by shepilov** [comment]
> why do we need 2 apis? why not only `sharings/drives/:id/recipients/:file-id` ?

ok, I got it, but then we need to proxy call to an owner instance and check the permission on the owner side

and maybe it's also time to have only one API insted of two, especially if we don't need to support legacy ones and to have somehting like `GET /sharings/recipients/:file-id?drive_id=:drive-id `

**Comment by taratatach** [comment]
> > why do we need 2 apis? why not only `sharings/drives/:id/recipients/:file-id` ?
> 
> and maybe it's also time to have only one API insted of two, especially if we don't need to support legacy ones and to have somehting like `GET /sharings/recipients/:file-id?drive_id=:drive-id `

It is part of the spec and it seemed reasonable since we have separate routes for everything related to drives at the moment.
I'd suggest keeping it that way until we refactor that to use regular files routes.

File: web/sharings/effective_recipients.go [hunk 1]
```diff
@@ -0,0 +1,131 @@
...
+)
+
+// apiEffectiveRecipient serializes a sharing.EffectiveRecipient as JSON-API.
+type apiEffectiveRecipient struct {
+	*sharing.EffectiveRecipient
+	id string
+}
+
+func (a *apiEffectiveRecipient) ID() string                             { return a.id }
+func (a *apiEffectiveRecipient) Rev() string                            { return "" }
```
shepilov, line 25: why do we need all these fields in this model? links, relationships, rev? we don't have them, actually

taratatach: It's the JSON-API interface.

shepilov: Why do we need JSON:API response here? 

if you want to use it here I think you need to modify the contract more dramatically(define the resource type and stable ID, decide whether file_id belongs in meta or a relationship) than just wrap it with this interface and implement all the empty methods for apiEffectiveRecipient

model/sharing/effective_recipients.go
```diff
@@ -0,0 +1,191 @@
...
+		kind := "ancestor"
+		if info.RootID == targetID {
+			kind = "self"
+		}
+		// Same condition as authorizeRevokeRecipient: the owner can manage
+		// members, and a drive recipient can too (delegated to the owner)
+		// only with write access — read-only drive recipients get a 403
+		// from hasSharingWritePermissions when they try. A classic-sharing
+		// recipient cannot revoke.
+		canManage := s.Owner
```
shepilov: The request is proxied to the owner instance, so it's suppose to be always true

taratatach: That's true for federated sharings. Are we sure the front-end won't call this route for legacy replicated sharings?

web/sharings/effective_recipients.go
taratatach: This is for consistency with the other routes which use JSON-API responses.

The stable ID is defined by `effectiveRecipientID()` and the decision about the `file_id` has been made: it's stored in `meta`.
Do you see other things that should be done?

model/sharing/effective_recipients.go
shepilov: yes

shepilov (approved):

JF-Cozy:
> > why do we need 2 apis? why not only `sharings/drives/:id/recipients/:file-id` ?
> 
> ok, I got it, but then we need to proxy call to an owner instance and check the permission on the owner side

I asked myself the same thing. So what's your conclusion? Oh ok it's for sharing and shared drives 👍

PR #4888, opened 2026-08-24
feat(rabbitmq): store the matrix id received on user.created

### LINKED ISSUE #4887: store the Matrix ID received on user.created instead of deriving it
## Problem

The Matrix ID sent to the common settings service is derived from the email local part, falling back to the domain slug.

The homeserver mints the localpart from an OIDC claim that preserves dots, and the address does not always carry them, so the derived value can name an account that does not exist. The stack has no way to know the localpart, only the emitter does.

## Expected behavior

`user.created` carries the Matrix ID, the stack stores it on the instance and serves that value instead of recomputing one.

rezk2ll:
## Problem

The Matrix ID sent to the common settings service is derived from the email local part, which is not always the address of the account.

## Solution

Store the `matrixId` that `user.created` now carries, and send that. Instances without one keep the existing derivation.

Closes #4887

### DIFF
--- a/model/settings/common/common.go
+++ b/model/settings/common/common.go
@@ -127,6 +127,9 @@ func UpdateCommonSettings(inst *instance.Instance, settings *couchdb.JSONDoc) (b
 		if email, ok := settings.M["email"].(string); ok {
 			addDiff("email", remote.Payload.Email, email)
 		}
+		if matrixID, _ := settings.M["matrix_id"].(string); matrixID != "" {
+			addDiff("matrix_id", remote.Payload.MatrixID, matrixID)
+		}
 
 		log.Warnf("common settings out of sync: local=%d remote=%d", inst.CommonSettingsVersion, remote.Version)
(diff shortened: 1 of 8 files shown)


model/settings/common/common.go
```diff
@@ -307,7 +307,9 @@ func buildRequest(inst *instance.Instance, settings *couchdb.JSONDoc) UserSettin
 	if phone, ok := settings.M["phone"].(string); ok {
 		request.Payload.Phone = phone
 	}
-	if len(parts) > 1 {
+	if stored, _ := settings.M["matrix_id"].(string); stored != "" {
+		request.Payload.MatrixID = stored
```
taratatach:
```suggestion
	if matrixID, _ := settings.M["matrix_id"].(string); matrixID != "" {
		request.Payload.MatrixID = matrixID
```

taratatach (approved):

shepilov (approved):

PR #4889, opened 2026-08-25
Allow dataproxy to be framed inside an intent

zatteo:
To enable search and recent tabs in the File Picker, we need to add the dataproxy in the File Picker. So we need to set up the appropriate frame-ancestors CSP for the dataproxy.

However, the dataproxy does not fit with the current CSP when using an intent with external apps (like when tmail load the File Picker).

In https://github.com/linagora/cozy-stack/pull/4824, I implemented a first solution for external apps like tmail. When an intent is created by an app associated to an external app, it adds both URL in CSP (cozy container url + external app url).

But as we can see in the schema below the dataproxy intent is not created by the tmail container or the tmail app. It is created by the File Picker so by tdrive. So this first solution does not work (it only adds tdrive in dataproxy frame-ancestors). Moreover the dataproxy intent can not know if it has been created by a usual cozy apps (eg tdrive) or by an app associated to an external app (eg tmail).

Screenshot 2026-08-25 at 10 07 22

That's why here I propose to add in CSP _only for the dataproxy_ **all apps that have an external app attached**. So for example, the dataproxy will allows to be framed by calendar container & calendar app & mail container & mail app when it is loaded in an intent. We open "more what we need" but I do not see other solution and the blast radius stay controlled because we only allow app that can open the dataproxy anyway.

Tell me what you think. I am open to other solution.

### DIFF
--- a/web/apps/serve.go
+++ b/web/apps/serve.go
@@ -149,6 +149,33 @@ func handleIntent(c echo.Context, i *instance.Instance, slug, intentID string) *
 	return intent
 }
 
+// allowContainerAppsFrame adds to frame-ancestors the cozy subdomain and the
+// client URL (via its client_url_flag feature flag) of every installed app
+// that declares one. Dataproxy can be nested several iframes deep in an
+// intent flow (e.g. a container app framing its external client, which frames
+// the intent client, which frames dataproxy), and frame-ancestors must list
+// every ancestor in the chain.
+func allowContainerAppsFrame(c echo.Context, i *instance.Instance) {
+	if config.GetConfig().CSPDisabled {
+		return
+	}
+	manifests, _, err := app.ListWebappsWithPagination(i, 0, "")
+	if err != nil {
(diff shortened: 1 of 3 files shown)


web/apps/serve.go
```diff
@@ -271,6 +298,10 @@ func ServeAppFile(c echo.Context, i *instance.Instance, fs appfs.FileServer, web
 	var intentDoc *intent.Intent
 	if intentID := c.QueryParam("intent"); intentID != "" {
 		intentDoc = handleIntent(c, i, slug, intentID)
+
+		if slug == consts.DataProxySlug {
```
shepilov: we should check here that intentDoc is not empty, or the first check doesn't make sense as well

shepilov (approved):

web/apps/serve.go
zatteo: Good catch, done

PR #4890, opened 2026-08-26
feat: Use effective access in shared drive files routes

### LINKED ISSUE #4842: Use Effective Access In Shared-Drive File Routes
## Context

 Shared-drive file routes currently authorize users mostly through the route shared drive ID, using direct membership checks on the corresponding io.cozy.sharings document. This is not enough for nested shared folders.

  The backend should authorize file operations against the effective access of the target file or folder, not only against the direct members of /sharings/drives/:id.

  ## Goal

  Update shared-drive file routes so they use the effective access resolver for permission checks.

  The check should answer:

  - can the current user read this target file/folder?
  - can the current user write/update/delete this target file/folder?
  - can the current user create content inside this destination folder?

  ## Scope

  Update the shared-drive route authorization layer in web/sharings/drives.go, especially:

  - proxy
  - sharedDrivePermissionCheck
  - checkSharedDrivePermission
  - validateSharedDrivePermissionTargetID
  - any middlewares.AllowVFS integration used by these routes

  The route guard should become target-aware when a :file-id or destination folder is involved.

  ## Related Endpoints

  Read-like routes:

  HEAD   /sharings/drives/:id/download/:file-id
  GET    /sharings/drives/:id/download/:file-id
  HEAD   /sharings/drives/:id/download/:file-id/:version-id
  GET    /sharings/drives/:id/download/:file-id/:version-id
  GET    /sharings/drives/:id/_changes
  HEAD   /sharings/drives/:id/:file-id
  GET    /sharings/drives/:id/metadata
  GET    /sharings/drives/:id/:file-id
  GET    /sharings/drives/:id/:file-id/size
  GET    /sharings/drives/:id/:file-id/thumbnails/:secret/:format
  POST   /sharings/drives/:id/downloads
  GET    /sharings/drives/:id/downloads/:secret/:fake-name
  POST   /sharings/drives/:id/archive
  GET    /sharings/drives/:id/archive/:secret/:fake-name
  GET    /sharings/drives/:id/notes/:file-id/open
  GET    /sharings/drives/:id/office/:file-id/open
  GET    /sharings/drives/:id/shortcuts/:file-id
  GET    /sharings/drives/:id/realtime

  Write-like routes:

  POST   /sharings/drives/:id/:file-id/versions
  PATCH  /sharings/drives/:id/:file-id
  POST   /sharings/drives/:id/
  POST   /sharings/drives/:id/:file-id
  PUT    /sharings/drives/:id/:file-id
  POST   /sharings/drives/:id/upload/metadata
  POST   /sharings/drives/:id/:file-id/copy
  POST   /sharings/drives/:id/trash/:file-id
  DELETE /sharings/drives/:id/trash/:file-id
  DELETE /sharings/drives/:id/:file-id
  POST   /sharings/drives/:id/notes

  ## Permission Rules

  For read routes: require effective read on the target file/folder
  For write/update/delete routes: require effective write on the target file/folder
  For create/upload routes:  require effective write/create on the destination folder
  For metadata update / rename / backend folder move through PATCH: require effective write on the source item
  
  ## Out Of Scope

  This task should not include:

  - POST /sharings/drives/move, handled by a dedicated backend move task
  - recipient management endpoints
  - share-by-link permissions endpoints
  - effective recipients API for the Share modal
  - limited access / restrictive folder support

taratatach:
Closes #4842 

  Shared-drive file routes will authorize requests against the effective
  access of the calling member, resolved on the owner's instance where
  the drive content lives (recipients hold no local copies of drive
  files). `AccessResolver` gains `ResolveForMember`, which matches the
  member across sharings by email or instance host via the new
  `Sharing.MemberMatching`, so nested shared folder scopes combine as
  they already do for the instance-scoped `Resolve`.

### DIFF
--- a/web/sharings/drives.go
+++ b/web/sharings/drives.go
@@ -255,8 +255,56 @@ func HeadDirOrFile(c echo.Context, inst *instance.Instance, s *sharing.Sharing) 
 	return nil
 }
 
+// authorizeDriveTarget resolves the effective access of the given drive
+// member on the target; 404 hides targets outside every sharing scope, to
+// avoid revealing which paths exist on the instance.
+func authorizeDriveTarget(inst *instance.Instance, member *sharing.Member, targetID string) (*sharing.EffectiveAccess, error) {
+	ea, err := sharing.NewAccessResolver(inst).ResolveForMember(targetID, member)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return nil, jsonapi.NotFound(errors.New("shared drive target not found"))
+		}
+		return nil, wrapErrors(err)
+	}
+	if !ea.CanRead {
+		return nil, jsonapi.NotFound(errors.New("shared drive target not found"))
+	}
+	return ea, nil
+}
+
+// checkDriveMemberRead asserts the calling drive member (if any) has
+// effective read access on the target. No-op for non-member tokens (owner,
+// share-by-link).
+func checkDriveMemberRead(c echo.Context, inst *instance.Instance, targetID string) error {
+	member := GetSharedDriveMember(c)
+	if member == nil {
+		return nil
+	}
+	_, err := authorizeDriveTarget(inst, member, targetID)
+	return err
+}
+
 // ReadMetadataFromPath allows to get file/dir information for a path.
 func ReadMetadataFromPath(c echo.Context, inst *instance.Instance, s *sharing.Sharing) error {
+	// Drive tokens are checked against the member's effective read access on
+	// the resolved target; other tokens keep the VFS permission check done
+	// by files.ReadMetadataFromPath. A target outside every sharing scope is
+	// answered 404 to avoid revealing which paths exist on the instance.
+	if GetSharedDriveMember(c) != nil && c.QueryParam("Path") != "" {
+		dir, file, err := inst.VFS().DirOrFileByPath(c.QueryParam("Path"))
+		if err != nil {
+			return jsonapi.NotFound(errors.New("shared drive target not found"))
+		}
+		var targetID string
+		if dir != nil {
+			targetID = dir.DocID
+		} else {
+			targetID = file.DocID
+		}
+		if err := checkDriveMemberRead(c, inst, targetID); err != nil {
+			return err
+		}
+	}
 	return files.ReadMetadataFromPath(c, s)
 }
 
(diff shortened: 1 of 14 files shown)


web/sharings/drives.go
```diff
@@ -255,8 +258,80 @@ func HeadDirOrFile(c echo.Context, inst *instance.Instance, s *sharing.Sharing)
...
+		return wrapErrors(err)
+	}
+	if !ea.Can(permission.GET) {
+		return jsonapi.NotFound(errors.New("shared drive target not found"))
+	}
+	// Echo caches query params on first c.QueryParam call (the permission
+	// middlewares run before us) and never invalidates the cache: mutate the
+	// cached values so the open handlers see the forced read-only, and keep
+	// RawQuery in sync for the reverse-proxy path, which forwards RawQuery.
+	if c.QueryParams().Get("ReadOnly") != "true" && !ea.Can(permission.PUT) {
```
shepilov: function with the name checkDriveOpenAccess modifies the internal state of the passed params, which looks like a bad design.

Can we have a helper in EffectiveAccess like: 
```
  access, err = func authorizeDriveTarget(
      inst *instance.Instance,
      drive *sharing.Sharing,
      member *sharing.Member,
      targetID string,
      verb permission.Verb,
  ) (*sharing.EffectiveAccess, error)
```
and use it everywhere; in this case, it will be 
```
readOnly := c.QueryParam("ReadOnly") == "true" || !access.CanWrite
  return notes.OpenNoteURLWithReadOnly(c, readOnly)
```

we don't need this helper specifically for this place, but it can be helpful in multiple places

and here we should at least return readonly boolean and don't modify request context

web/sharings/drives.go
```diff
@@ -255,8 +258,80 @@ func HeadDirOrFile(c echo.Context, inst *instance.Instance, s *sharing.Sharing)
...
+	}
+	ea, err := sharing.NewAccessResolver(inst).ResolveForMember(fileID, member)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return jsonapi.NotFound(errors.New("shared drive target not found"))
+		}
+		return wrapErrors(err)
+	}
+	if !ea.Can(permission.GET) {
+		return jsonapi.NotFound(errors.New("shared drive target not found"))
```
shepilov: 403?

taratatach: The idea is not to advertise existing documents to users without read access. This is coherent with the behavior of other routes.

web/sharings/drives.go
taratatach: Ah yes, I agree. I wanted to change that name and forgot.

web/sharings/drives.go
```diff
@@ -1437,6 +1533,20 @@ func proxy(fn func(c echo.Context, inst *instance.Instance, s *sharing.Sharing)
...
+			// sharing: the recipient-side member would authorize nothing here.
+			ownerMember, err := sharedDriveInteractMember(c, owner, ownerSharing)
+			if err != nil {
+				return err
+			}
+			SetSharedDriveMember(c, ownerMember)
+			if err := guardSharedDriveRouteForMember(c, owner, ownerSharing); err != nil {
+				return err
+			}
 			return fn(c, owner, s)
```
shepilov, line 1559: should we have owner sharing here?

taratatach: Hmm, I would say no. The request is made by a member so I think we want to pass the member's sharing to the handlers. Otherwise the requester will be considered as the owner, no?

shepilov (approved):

PR #4891, opened 2026-08-27
feat(banners): evaluate and materialize io.cozy.banners

rezk2ll:
The `io.cozy.banners` doctype was declared but nothing wrote to it, so no client could ever show a platform message. This is the producing half.

**What it does**

- Evaluates banner rules in the stack when an input changes, and stores the result as a document in the user's instance. Clients only read, sort and render.
- Quota: a banner at 90% and another at the limit. Removed when space is freed or the limit is raised.
- Billing: consumes payment events from the bus and shows the restricted dialog once Stripe has given up on a subscription.
- Text is stored already localized in the instance language (en, fr, ru, vi).
- Nothing is written unless the instance context enables it.

**What it does not do yet**

- No warning while a card can still be fixed. Both Figma payment designs describe an already restricted workspace, so there is no approved wording for a user whose access is intact.
- No admin broadcasts, no time-based trigger, no rollout bootstrap.

**For reviewers**

- Each category has one document with a fixed id (`banner-quota`, `banner-billing`), so CouchDB refuses a second concurrent create, and every evaluation runs under a per-instance lock so a losing write is re-evaluated rather than dropped. No new index and no `IndexViewsVersion` change.
- The disk quota alert now accepts several subscribers, and the banner takes only its rising edge. On a deletion that alert fires before the removal is indexed, so the banner re-evaluates from the usage the destroy will leave.
- The fr, ru and vi translations need a native reviewer.

Evaluation and materialization: linagora/twake-workplace-private#1625. The quota and payment triggers from linagora/twake-workplace-private#1626. Both under linagora/twake-workplace-private#1583.


model/banner/trigger.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/cozy/cozy-stack/model/vfs"
+	"github.com/cozy/cozy-stack/pkg/logger"
+)
+
+func init() {
+	// A write that crosses the threshold, where reading the usage back is safe
+	// because the file is already indexed. The rising edge only: on the downward
+	// crossing this runs after the callback below and reads a usage the destroy
+	// has not reached yet, resurrecting the banner it just removed.
+	vfs.RegisterDiskQuotaAlertCallback(func(domain string, exceeded bool) {
```
MontaGhanmy, line 18: quota.exceeded is basically unreachable from this path. capsize comes from CheckAvailableDiskSpace, and it stays 0 as soon as diskUsage is already past 90%, so PushDiskQuotaAlert(fs, true) only ever fires on the rising edge of 90%. someone filling 91% to 100% never re-evaluates, so the banner sits on almost-full and the exceeded one only appears after a destroy or a quota patch. should we also refresh when a write is rejected with ErrFileTooBig?

shepilov (approved):

PR #4895, opened 2026-08-28
feat(rag): record the index status on a callback route

ewan102:
The RAG indexer reports the indexation status of a file by posting on /ai/index/status, and the status is saved on an io.cozy.ai.chat.rag document.

Callbacks are ordered on the revision of the file they carry, which is mandatory and kept on the status document: one about a revision that is not newer than the one already described is dropped, so that it cannot claim the file is indexed when a more recent status says otherwise.

### DIFF
--- a/model/rag/status.go
+++ b/model/rag/status.go
@@ -1,0 +1,81 @@
+package rag
+
+import (
+	"errors"
+	"os"
+	"time"
+
+	"github.com/cozy/cozy-stack/model/instance"
+	"github.com/cozy/cozy-stack/pkg/consts"
+	"github.com/cozy/cozy-stack/pkg/couchdb"
+	"github.com/cozy/cozy-stack/pkg/couchdb/revision"
+)
+
+const (
+	StatusSuccess      = "success"
+	StatusError        = "error"
+	StatusNotSupported = "notsupported"
+)
+
+// IndexStatusPath is the path of the route on which the RAG indexer posts the
+// indexation status of a file. It is what inst.PageURL takes to build the
+// callback_url given to the indexer.
+const IndexStatusPath = "/ai/index/status"
+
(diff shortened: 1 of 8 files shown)


model/rag/status.go
```diff
@@ -0,0 +1,81 @@
...
+		return err
+	}
+
+	doc := NewIndexStatus(fileID)
+	err := couchdb.GetDoc(inst, consts.ChatRAG, fileID, doc)
+	switch {
+	case err == nil:
+		// Dropped whole: such a callback must not claim the file is indexed, as
+		// a more recent status (e.g. a delete) may say otherwise.
+		if isOutdated(rev, doc.FileRev) {
```
shepilov: you shoudl do read and write under the lock, probably by fileId, otherwise you can have the same proble

web/ai/index_status.go
```diff
@@ -0,0 +1,91 @@
...
+
+// statusMessage is the payload posted by the RAG indexer on the callback URL.
+type statusMessage struct {
+	Partition string `json:"partition"`
+	FileID    string `json:"file_id"`
+	Status    string `json:"status"`    // "success" | "error" | "notsupported"
+	Timestamp string `json:"timestamp"` // RFC3339Nano
+	// The "version" field holds the revision of the file the status is about.
+	// Callbacks are ordered on it.
+	Metadata struct {
```
shepilov, line 23: are you sure about nested object for one field?

ewan102: It's because version is composed of the metadata the stack is sending to OpenRAG so version is composed of md5sum, datetime, doctype and version. I agree that now that we have a filerev field the name Version might be an error...

model/rag/status.go
ewan102: yes you are right. Fixing it

shepilov (approved):

PR #4896, opened 2026-08-28
refactor(rag): send the callback URL and the file revision to the ind…

ewan102:
callRAGIndexer was a 253-line function; it is now a 35-line dispatch that delegates. 

The indexer is given the callback URL to report the indexation status on, and the revision of the file it is about, which it echoes back. A file is indexed again when its status says it is not, so that a status lost while the file was being indexed is recovered. Its status document is deleted along with the file, and a class the flags refuse is recorded as notsupported.

The md5sum sent to the indexer was the hexadecimal encoding of a base64 string, and is now the digest itself.

### DIFF
--- a/model/rag/index.go
+++ b/model/rag/index.go
@@ -13,7 +13,6 @@ import (
 	"net/url"
 	"strings"
 	"sync"
-	"time"
 
 	"github.com/cozy/cozy-stack/model/feature"
 	"github.com/cozy/cozy-stack/model/instance"
@@ -134,7 +133,7 @@ func isClassAllowed(flags *feature.Flags, class string) bool {
 // that decides a file will not be indexed.
 func markNotSupported(inst *instance.Instance, change couchdb.Change) error {
 	rev, _ := change.Doc.Get("_rev").(string)
-	return SetIndexStatus(inst, change.DocID, StatusNotSupported, rev, time.Now())
+	return SetIndexStatus(inst, change.DocID, StatusNotSupported, rev)
 }
 
 func deleteFromRAG(inst *instance.Instance, fileID string) error {
--- a/model/rag/index.go
+++ b/model/rag/index.go
@@ -84,12 +84,10 @@ func callRAGIndexer(inst *instance.Instance, doctype string, change couchdb.Chan
 	}
 
 	if change.Doc.Get("type") == consts.DirType {
-		// A directory move or rename rewrites the path of the directory and
-		// of all its descendant directories, but never touches the file docs
-		// they contain: those files may have silently entered or left a
-		// knowledge-base folder. Every descendant directory shows up in this
-		// same changes feed, so reconciling the direct file children of each
-		// changed directory covers the whole moved subtree.
+		// Moving a directory rewrites the path of every descendant directory
+		// but never touches the file docs, which may have silently entered or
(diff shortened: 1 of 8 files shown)


model/rag/status.go
```diff
@@ -70,6 +70,20 @@ func SetIndexStatus(inst *instance.Instance, docID, newStatus, rev string, times
 	return couchdb.Upsert(inst, doc)
 }
 
+// DeleteIndexStatus removes the indexation status of a document. One that never
+// had a status is not an error.
+func DeleteIndexStatus(inst *instance.Instance, docID string) error {
```
shepilov, line 81: This function call should be done also under the same lock as the status update

model/rag/index.go
```diff
@@ -118,215 +98,347 @@ func callRAGIndexer(inst *instance.Instance, doctype string, change couchdb.Chan
...
+		allowed, _ := flags.M["rag.index.audio.enabled"].(bool)
+		return allowed
+	}
+	return true
+}
+
+// markNotSupported records a status no emitter ever reports: it is the stack
+// that decides a file will not be indexed.
+func markNotSupported(inst *instance.Instance, change couchdb.Change) error {
+	rev, _ := change.Doc.Get("_rev").(string)
```
shepilov: one-line function, called only once, you should handle errors here ot make any sense in its existence

model/rag/index.go
```diff
@@ -118,215 +98,347 @@ func callRAGIndexer(inst *instance.Instance, doctype string, change couchdb.Chan
...
+
+// decodeMD5Sum turns the md5sum carried by the changes feed into the
+// hexadecimal digest the RAG server is given. CouchDB serializes the bytes of
+// the digest in base64.
+func decodeMD5Sum(v interface{}) string {
+	s, _ := v.(string)
+	if s == "" {
+		return ""
+	}
+	if raw, err := hex.DecodeString(s); err == nil && len(raw) == md5Length {
```
shepilov: do we need hex support here, in tests it's only base64?

model/rag/index_http_test.go
```diff
@@ -0,0 +1,160 @@
...
+	"net/http"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// formFields reads back the multipart body the RAG server received.
+func formFields(t *testing.T, contentType string, body []byte) map[string]string {
```
shepilov: we can use multipart.Reader.ReadForm as far as it's refactoring and return Form.Value

model/rag/index.go
ewan102: No, you're right, it's a vestige from when we were using md5sum to check isOutdated

model/rag/index.go
ewan102: I just wanted to put all methods by used indexation out of Index and CallRagIndexer

ewan102: Changing it tho, you are right, it might be useless in that case

model/rag/index_http_test.go
ewan102: Ok thank you, I corrected it

shepilov (approved):

PR #4897, opened 2026-08-31
feat(load): add concurrent upload capacity test suite

shepilov:
## Summary

- Add a Dockerized k6 concurrent upload scenario with a local mock smoke flow.
- Generate reusable upload binaries for six file sizes and make each upload unique.
- Add adaptive capacity search with configurable baselines, absolute or relative p95 limits, sustained confirmation, safety guards, and JSON reports.
- Support the personal-drive upload lifecycle and a work-in-progress single-instance shared-drive API mode.
- Verify every successful upload with an immediate chunked download and incremental MD5 comparison, aborting the campaign on any mismatch.
- Keep upload-binary generation and mock responses streamed or chunked to limit memory consumption.
- Keep the load harness as strict TypeScript with local type checks and regression tests.
- Organize source, service configuration, and generated data separately, with unit tests next to their modules.
- Add an opt-in local cozy-stack environment with CouchDB, Redis, and RabbitMQ.
- Provision Prometheus and Grafana with campaign filters and stack telemetry panels.
- Document the architecture, k6 and Node runtime boundaries, scenario strategy, and local and dedicated-environment workflows.

image


tests/load/src/fixtures/upload-data.ts
```diff
@@ -0,0 +1,45 @@
...
+    state ^= state >>> 17
+    state ^= state << 5
+    uploadDataBytes[index] =
+      index < uniqueHeader.length
+        ? uniqueHeader.charCodeAt(index) & 0xff
+        : state & 0xff
+  }
+
+  return uploadData
+}
```
taratatach: I have no idea what this function does.

PR #4903, opened 2026-09-07
feat: Check destination on move-like files PATCH

taratatach:
A PATCH on /files can move a file or folder by changing its parent,
  but only the target was authorized. Now checkMoveDestination requires
  the destination to exist and POST permission on it, and rejects with
  422 any move whose source or destination belongs to a shared drive:
  those must go through POST /sharings/drives/move until the
  shared-aware move task implements delegation.

### DIFF
--- a/web/files/files.go
+++ b/web/files/files.go
@@ -722,6 +722,10 @@ func applyPatch(c echo.Context, fs vfs.VFS, patch *docPatch) (err error) {
 		return err
 	}
 
+	if err = checkMoveDestination(c, fs, patch, dir, file); err != nil {
+		return err
+	}
+
 	if patch.Delete {
 		if dir != nil {
 			inst := middlewares.GetInstance(c)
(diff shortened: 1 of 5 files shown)


web/files/files.go
```diff
@@ -2419,6 +2432,54 @@ func checkPerm(c echo.Context, v permission.Verb, d *vfs.DirDoc, f *vfs.FileDoc)
...
+		return err
+	}
+	if newParent == nil {
+		return jsonapi.BadRequest(errors.New("destination is not a directory"))
+	}
+	if err := middlewares.AllowVFS(c, permission.POST, newParent); err != nil {
+		return err
+	}
+	resolver := sharing.NewAccessResolver(middlewares.GetInstance(c))
+	docs := [2][2]any{{dir, file}, {newParent, nil}}
```
shepilov: can we replace untyped matrix by anything else?

shepilov: probably we can replace it with two explicit calls

shepilov (approved):

PR #4904, opened 2026-09-08
feat: Add trial ending banner

MontaGhanmy:
Materializes a dismissible banner from billing trial events: "your trial ends on X", with a link to the manager.

The banner carries its own end date so clients expire it on their own, since the stack has no scheduled re-evaluation. Any status other than `trialing`, or an end date already past, deletes it. Post-trial states (conversion, dunning, grace) stay with the existing billing banners.

### DIFF
--- a/model/banner/banner_test.go
+++ b/model/banner/banner_test.go
@@ -14,7 +14,7 @@ import (
 // TestMain loads the real catalogs, so a message id that no longer exists
 // fails here rather than rendering its own name to a user.
 func TestMain(m *testing.M) {
-	for _, locale := range []string{"en", "fr"} {
+	for _, locale := range []string{"en", "fr", "ru", "vi"} {
 		po, err := os.ReadFile("../../assets/locales/" + locale + ".po")
 		if err != nil {
 			panic(err)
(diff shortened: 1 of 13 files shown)


model/banner/trial.go
```diff
@@ -0,0 +1,95 @@
...
+		return nil
+	}
+	// A window ending at the zero time is dropped by every client while still
+	// occupying the slot, and one that has already closed says nothing true.
+	// The second case is what a redelivered trialing event looks like after the
+	// trial is over, which supersededBy cannot catch once the document it
+	// ordered against has been deleted.
+	if state.EndsAt.IsZero() || !state.EndsAt.After(now) {
+		return nil
+	}
```
rezk2ll, line 58: This claims that the guard catches a redelivered `trialing` event after the trial is over.

Here, ' now ' is the event's own Stripe timestamp (`RefreshTrial` passes `eventAt` twice, L86-93), and a redelivery carries the original timestamp, which is always before `EndsAt`.

Compare against `time.Now()`, or fix the comment.

Separately, `state.EndsAt.IsZero() ||` is unreachable: the zero time is year 1, `!EndsAt.After(now)` already covers it.

model/banner/trial.go
```diff
@@ -0,0 +1,95 @@
...
+// TriggerTrialChanged is recorded on documents produced by a trial event.
+const TriggerTrialChanged = "trial.changed"
+
+// The wording, as a message id of the stack locales. It takes the end date.
+const textTrialEnding = "Banners Trial Ending Text"
+
+// trialDateLayout is the end date as the wording reads it. Day before month,
+// which is the order en, fr, ru and vi all write.
+//
+// ponytail: monday v1.0.2 has no vi locale, so Vietnamese wording gets an
```
rezk2ll, line 24: i'm not sure we need to mention this.

Let's make a pass to remove the useless comments

pkg/rabbitmq/handlers.go
```diff
@@ -892,7 +892,16 @@ func (h *BillingLifecycleHandler) Handle(ctx context.Context, d amqp.Delivery) e
...
-		if err := banner.RefreshBilling(domain, status, eventAt); err != nil {
+		var err error
+		// A trial carries its own statuses, and the rewrite above is about
+		// payments: the routing keys are distinct, so it cannot have reached
+		// one and the status arrives here as the Cloudery sent it.
+		if d.RoutingKey == RoutingKeyTrialChanged {
+			err = banner.RefreshTrial(domain, status, msg.TrialEndsAt, eventAt)
+		} else {
+			err = banner.RefreshBilling(domain, status, eventAt)
+		}
```
rezk2ll, line 903: A `trial.changed` event only ever touches the trial category.

I checked locally: setting the status to unpaid on this routing key removes the trial banner and writes no billing banner.

pkg/rabbitmq/rabbitmq.go
```diff
@@ -277,6 +278,14 @@ func BuildExchangeSpecs(exchangesCfg []config.RabbitExchange) []ExchangeSpec {
...
 			case QueueBillingLifecycle:
 				handler = NewBillingLifecycleHandler()
+				// Bindings come from the deployment config, not from the code,
+				// so a routing key missing there is silence: the Cloudery
+				// publishes, the exchange drops it, and no banner ever appears.
+				for _, key := range []string{RoutingKeyPaymentFailed, RoutingKeyPaymentRecovered, RoutingKeyTrialChanged} {
+					if !slices.Contains(configQueue.Bindings, key) {
+						log.Warnf("Queue %s is not bound to %s: those events will never be consumed", configQueue.Name, key)
+					}
+				}
```
rezk2ll, line 288: The billing exchange is kind: `topic` where payment.* is a legally binding contract that covers everything this checks for and still logs three warnings.

and i think we shouldn't do anything about it here. it's a gozy-salt configuration per context ( if you configure the queues you get the messages otherwise no )

let's get rid of this and focus on banners and trials

pkg/rabbitmq/handlers_test.go
```diff
@@ -95,3 +98,106 @@ func TestUserCreatedHandlerStoresMatrixID(t *testing.T) {
...
+// TestBillingLifecycleMessageParsesTheTrialDeadline pins the wire shape: the
+// Cloudery renders the date with a numeric offset rather than Z. It has to
+// land as a usable time, or the whole message dead-letters and no banner ever
+// appears.
+func TestBillingLifecycleMessageParsesTheTrialDeadline(t *testing.T) {
+	var msg rabbitmq.BillingLifecycleMessage
+	require.NoError(t, json.Unmarshal([]byte(
+		`{"workplaceFqdn":"a.example","status":"trialing","timestamp":1784721600,"trialEndsAt":"2026-08-05T12:00:00+00:00"}`), &msg))
+	assert.True(t, time.Date(2026, 8, 5, 12, 0, 0, 0, time.UTC).Equal(msg.TrialEndsAt))
+}
```
rezk2ll, line 203: This is testing that encoding/json parses RFC3339 with a numeric offset. has nothing to do with the logic / PR.

useless test imo remove it

model/banner/banner_test.go
```diff
@@ -353,3 +353,91 @@ func TestEvaluateBilling(t *testing.T) {
...
+		for _, locale := range []string{"en", "fr", "ru", "vi"} {
+			s := state("trialing")
+			s.Locale = locale
+			b := EvaluateTrial(s, now)
+			require.NotNil(t, b, locale)
+			assert.Contains(t, b.Text, "2026", "%s dropped the placeholder", locale)
+			assert.NotContains(t, b.Text, "%!", locale)
+		}
+	})
+}
```
rezk2ll, line 443: The second test is the first one but in a loop. Keep the loop and assert the exact en/fr strings inside it.

pkg/rabbitmq/contracts.go
```diff
@@ -58,6 +61,12 @@ type BillingLifecycleMessage struct {
 	// Delivery is at-least-once and unordered, so this, not the arrival time,
 	// decides which event wins.
 	Timestamp int64 `json:"timestamp"`
+	// TrialEndsAt is when the free trial ends, known from the day it starts,
+	// so the banner can state an absolute date and no deadline trigger is
+	// needed. It rides on every lifecycle message rather than on a trial one
+	// because the Cloudery declares it on the shared base class the payment
+	// and trial messages both extend, so it is zero on a payment message.
+	TrialEndsAt time.Time `json:"trialEndsAt,omitempty"`
```
rezk2ll, line 69: - The last clause is wrong: trialEndsAt is not zero on a payment message.
- LifecycleMessage.for populates it unconditionally for all three subclasses from instance.trial_ends_at ( cloudery )
- so the payload ( after we merge your PR on cloudery ):
```
payment.failed {...,"status":"past_due","attemptCount":2,"trialEndsAt":"2026-09-21T14:13:20.000Z"}

```

A real past date, on a payment message, for anyone who ever trialed. We're safe today because the handler only reads the field on `trial.changed`, but the comment would justify the opposite.

Fix the comment, and keep gating on routing key + status rather than on the field's presence.

Also ,omitempty on a time.Time never omits, a struct is never empty. A zero value marshals to {"trialEndsAt":"0001-01-01T00:00:00Z"}.

rezk2ll (commented):
and let's get rid of useless comments. no need to have them if the code is self explanatory

PR #4905, opened 2026-09-08
feat(banners): warn during the payment grace period

rezk2ll:
## Context

The billing banner only had something to say once the workspace was already restricted. While Stripe is still retrying, the status is `past_due` and the plan is intact, so there is still time for the user to fix the card.

## Solution

`past_due` materializes a dismissible inline banner that escalates with the invoice attempt count. The attempt is part of the banner id rather than just the severity, because a dismissal is carried forward only while the id is unchanged: an escalation reappears, a repeat of the same attempt writes nothing.

The restricted dialog now takes the event origin. Only an organization keeps its plan into `unpaid`, so a single subscriber gets nothing there instead of a dialog about a restriction that did not happen.

## Worth knowing

The three grace sentences are provisional, so swapping the `Banners Billing Grace {Info,Warning,Error} Text` entries is the whole change. They are wired in `ru` and `vi` as well, but `consts.SupportedLocales` is `{en, fr}`, so those two render English until that list grows.


model/banner/billing.go
```diff
@@ -78,7 +139,7 @@ func EvaluateBilling(state BillingState, now time.Time) *Banner {
 // RefreshBilling re-evaluates the billing banner of an instance from a payment
 // event. eventAt is the moment Stripe recorded the event, not the moment this
 // runs, so it both stamps the document and orders it against what is stored.
-func RefreshBilling(domain, status string, eventAt time.Time) error {
+func RefreshBilling(domain, status string, attemptCount int, b2b bool, eventAt time.Time) error {
```
MontaGhanmy: we already load the instance here and it has OrgDomain. why do we need to pass b2b down from the handler? if an instance level event ever lands on a member of an org we'll silently show nothing at all.

can't we do `b2b := inst.OrgDomain != ""` and drop the param?

model/banner/billing.go
```diff
@@ -41,14 +67,49 @@ type BillingState struct {
...
+
+// The attempt is part of the identifier because Merge carries a dismissal
+// forward only while the identifier is unchanged: an escalation has to read as
+// a new occurrence, and a re-evaluation of the same attempt has to not.
+func graceBanner(state BillingState, now time.Time) *Banner {
+	attempt := min(max(state.AttemptCount, 1), maxGraceAttempt)
+	wording := graceWording[attempt-1]
+
+	startsAt := now
+	banner := &Banner{
```
MontaGhanmy, line 100: the quota ids are exported consts, with a comment saying clients key their dismissal on them. why is this one an inline string? at least the billing.grace.attempt-N shape should be written down somewhere, clients will depend on it.

model/banner/billing.go
```diff
@@ -41,14 +67,49 @@ type BillingState struct {
...
+	default:
 		return nil
 	}
+}
+
+// The attempt is part of the identifier because Merge carries a dismissal
+// forward only while the identifier is unchanged: an escalation has to read as
+// a new occurrence, and a re-evaluation of the same attempt has to not.
+func graceBanner(state BillingState, now time.Time) *Banner {
+	attempt := min(max(state.AttemptCount, 1), maxGraceAttempt)
```
MontaGhanmy, line 96: the attempt goes back to 1 when a new invoice opens, so a new cycle mints billing.grace.attempt-1 again. if the user dismissed the previous cycle's attempt-1 and there was no recovery in between, Merge carries the dismissal over and he never sees the new failure. ( i know it's a narrow path )

should the invoice id be part of the identifier?

PR #4908, opened 2026-09-08
feat(rag): index the assistants' knowledge base folders on openRAG

paultranvan:
## RAG indexing scoped by the assistants' knowledge bases

The `rag-index` worker now indexes the folders listed in the `knowledgeBase` of the `io.cozy.ai.chat.assistants` documents, and only them. The root folder (`io.cozy.files.root-dir`) means the whole Drive.

### How it works

- One job per instance, one lock, one checkpoint on the `io.cozy.files` changes feed. Two `@event` triggers wake it up, created by the app in charge of the assistants (the worker is no longer reserved): `io.cozy.files` with a 30s debounce, and `io.cozy.ai.chat.assistants` without debounce.
- Every knowledge base folder has its openRAG workspace, named after the folder id (well-known dotted ids are mapped, e.g. `io-cozy-files-root-dir`). At every run the worker reconciles the workspaces with the folders: a new folder gets its workspace then a job that walks its subtree; a folder no assistant uses any more loses its workspace and its files are detached.
- A file is indexed in the workspaces of every folder containing it, its memberships follow it when it moves, and it is deleted from openRAG when no folder contains it. The chat scopes its retrieval to the workspace of the assistant's folder.
- Uploads are decided from the md5 openRAG holds. Its two 409 are told apart: an id it already holds is re-uploaded with a PUT, a content it already indexed under another id is skipped for good (openRAG deduplicates by content per partition). Files the indexer refuses or whose content is missing from the storage are skipped; only transient errors are retried.

### Operator tools

Admin routes `POST /instances/:domain/rag/{reset,reconcile?dir_id=,prune,purge}`, documented in `docs/admin.md` and `docs/ai.md`. The CLI wrappers (`cozy-stack rag …`) come in a separate PR (`feat/rag-cli`).

### Tests

`model/rag` integration tests run against an in-memory fake of the openRAG API (workspaces, memberships, both 409s, pending uploads, content deduplication). Validated end to end on a local instance with the cozy-search and cozy-drive counterparts.

### DIFF
--- a/model/rag/prune.go
+++ b/model/rag/prune.go
@@ -1,0 +1,219 @@
+package rag
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+	"os"
+	"path"
+	"slices"
+
+	"github.com/cozy/cozy-stack/model/instance"
+	"github.com/cozy/cozy-stack/model/job"
+	"github.com/cozy/cozy-stack/model/vfs"
+	"github.com/cozy/cozy-stack/pkg/config/config"
+	"github.com/cozy/cozy-stack/pkg/consts"
+	"github.com/cozy/cozy-stack/pkg/logger"
+	"github.com/labstack/echo/v4"
+)
+
+// ErrUnknownFolder is returned by Reconcile when the given folder is in no
+// assistant's knowledge base: there is nothing to index in it.
+var ErrUnknownFolder = errors.New("the folder is in no assistant's knowledge base")
+
+// PruneResult summarizes a Prune run.
+type PruneResult struct {
+	FilesScanned      int `json:"files_scanned"`
+	FilesDeleted      int `json:"files_deleted"`
+	WorkspacesDeleted int `json:"workspaces_deleted"`
+}
+
+// claimsFile tells whether a knowledge base folder claims the file: it must
+// be live in the VFS, with a parent directory inside such a folder.
+func (s *scopes) claimsFile(fs vfs.VFS, fileID string) (bool, error) {
+	file, err := fs.FileByID(fileID)
+	if errors.Is(err, os.ErrNotExist) {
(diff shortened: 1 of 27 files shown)


model/rag/prune.go
```diff
@@ -0,0 +1,219 @@
...
+		}
+	}
+	return ids, nil
+}
+
+// Prune deletes from openRAG what no knowledge base folder claims any more:
+// the files gone or trashed in the VFS, the files outside every knowledge
+// base folder, and the workspaces of the folders that are in no knowledge
+// base. The index status of a deleted file is dropped too.
+func Prune(inst *instance.Instance, logger logger.Logger) (PruneResult, error) {
```
shepilov: why do we need a logger as argument if we are getting it from the instance?

web/instances/instances.go
```diff
@@ -775,6 +775,10 @@ func Routes(router *echo.Group) {
 	router.GET("/:domain/swift-prefix", getSwiftBucketName)
 	router.GET("/:domain/sharings/:sharing-id/unxor/:doc-id", unxorID)
 	router.POST("/:domain/notifications", sendNotification)
+	router.POST("/:domain/rag/reset", ragReset)
+	router.POST("/:domain/rag/prune", ragPrune)
```
shepilov, line 779: when do we call this endpoint?

shepilov (approved):

model/rag/prune.go
paultranvan: No particular reason, removed in aaa0168730eade6bc0e7591f37ee456750be80b6

web/instances/instances.go
paultranvan: For now, never, it will be useful for future CLI tooling
it's more a safeguard if things go wrong

PR #4909, opened 2026-09-10
refactor(banners): Replace the payment-failure evaluator with a generic backend command intake

MontaGhanmy:
## Context

The stack was deciding when a payment problem deserves a banner. It read the Stripe subscription status and the invoice attempt count off the bus, ran the escalation rules itself, and picked the wording from its own translation catalog. None of that is ours: the stack should be writing banners, not deciding them. Every wording change or rule change meant a stack release, and the billing rules already live on the backend side that owns the subscription.

## Solution

`billing.go`, the `BillingLifecycleMessage` contract, its handler, its queue and the four billing translations are removed. In their place a backend publishes a `banner.materialize` or a `banner.clear` command on the `platform` exchange, carrying the wording it already decided, and the stack validates it and stores it. `docs/banners.md` describes that contract for the producing side.

The intake is not a pass-through. A command is refused if its category is not listed in `banner_command_categories` for the instance context, and the `quota` category is refused always, because the stack measures disk usage itself and no backend should be able to speak for it. The routing key decides materialize versus clear, never the payload, so a producer cannot clear a category by naming a field. The CTA has to be an absolute `https` URL, the payload has to fit `MaxCommandBytes`, and the `bannerId` has to match the format the doctype documents.

Ordering is the part worth reading. Delivery is at-least-once and unordered, so the intake needs to know what it already applied. That record is a new stack-private doctype, `io.cozy.banners.commands`, blocklisted so no application can reach it, holding the last accepted revision per category. It is separate from `io.cozy.banners` on purpose: an app has write access there to carry a dismissal, and if the ordering rode on the visible document that app could rewind it. A stale revision is refused and the same revision applied twice converges, so a redelivery is a no-op. A command that retrying cannot fix is nacked and the broker dead letters it after `delivery_limit` attempts.

The second commit is unrelated and reads on its own. `Extend()` returned nothing, so the renewal goroutine of a long operation could not tell a renewed lease from a lost one: it kept renewing a lease Redis no longer held, and it outlived `Unlock`. It now returns an error and the goroutine stops on the first failure. The two callers, the RAG workspace and the VFS migration, do not observe the lease themselves, so for them this is a goroutine that stops instead of one that lies.

## Worth knowing

Tested locally against a real broker: materialize inline and modal, clear, a stale revision refused, and malformed and reserved-category commands dead lettered. `TestHandlers/InstallApp` fails, but it fails identically on master.

One gap is open and I do not think it blocks: a commanded banner does not re-localize when the user changes their language, because we keep only the locale picked when the command arrived. The backend sends the full set, we drop the rest. Follow-up.

The old `stack.billing.lifecycle` queue and its dead letter queue stay declared on the broker. That is a deployment cleanup, not a code change.


assets/locales/en.po
```diff
@@ -1345,18 +1345,3 @@ msgstr "AI Assistant"
 
 msgid "Banners Quota Exceeded Text"
 msgstr "You have reached your storage limit."
-
-
-msgid "Banners Billing Restricted Text"
```
rezk2ll, line 1350: Why did you remove the translations here? 

The idea is that the stack knows the instance language and it will materialize the banner using that instance language.

we should just keep that system, but the events and calculations are done in cloudery

docs/banners.md
rezk2ll, line 1: too much text here, and it feels we are just repeating what the ADR said

model/banner/command.go
```diff
@@ -0,0 +1,404 @@
...
+	at := time.Unix(cmd.Timestamp, 0).UTC()
+	locale := cmd.locale(instanceLocale)
+	b := &Banner{
+		BannerID:     cmd.BannerID,
+		Category:     cmd.Category,
+		Severity:     cmd.Severity,
+		Surface:      cmd.Surface,
+		Title:        cmd.Title[locale],
+		Text:         cmd.Text[locale],
+		Lang:         locale,
```
rezk2ll, line 258: A banner made from a command is stored in one language only. The Cloudery sends
every language it has, but here we pick one and throw the others away.

This breaks something ADR 054 decided. The trigger table has a row for it:

> | Instance locale change | Every document for that instance, re-localized |

and the Translation section spells out the mechanism: "a language change
re-materializes the documents, which clients pick up on their next refetch".

The stack already does exactly that. Instance patching flags a banner refresh
when the locale changes, and the refresh hook runs. That hook is only wired to
the quota rule, so quota banners get reworded and command banners do not.

I tested this locally. An instance in French with both kinds of banner, then one
language change to English: the storage warning became English, the payment
warning stayed French and still declared `lang: "fr"`. The user sees two banners
in two languages, side by side.

The fix does not need the deleted translation files, and does not need the
Cloudery to resend anything. The command already carried the English text, we
just dropped it. Keep the full set of languages on the private command document
and let the existing refresh hook pick again.

Worth separating two things here. Moving the wording to the Cloudery is right ( in a sense ),
and it removes a cost the ADR accepted: that new wording and new locales need a
stack release ( and it's fine ). Dropping re-localization is a different change, and those two
were separable. If we do want to change the decision rather than fix the code,
that needs an ADR amendment, not a paragraph in `docs/banners.md`.

model/banner/command.go
```diff
@@ -0,0 +1,404 @@
...
+// commandState records the last command accepted for a category, so ordering
+// survives clears (which leave no public document) and unchanged decisions.
+type commandState struct {
+	DocID  string `json:"_id,omitempty"`
+	DocRev string `json:"_rev,omitempty"`
+
+	Category string `json:"category"`
+	Revision int64  `json:"revision"`
+	Clear    bool   `json:"clear"`
+	EventID  string `json:"eventId,omitempty"`
```
rezk2ll: This is the right place to keep the languages we currently discard. It is already private to the stack and already unreachable from applications, so nothing else needs to change to store them here.

model/banner/command.go
```diff
@@ -0,0 +1,404 @@
...
+	"github.com/cozy/cozy-stack/pkg/prefixer"
+)
+
+// TriggerCommand is recorded on documents a backend asked for rather than a
+// rule the stack runs itself.
+const TriggerCommand = "banner.command"
+
+// ErrInvalidCommand marks a command no retry can fix. A transport rejects such
+// a delivery instead of requeuing it; anything else is a storage failure worth
+// retrying.
```
rezk2ll: This comment says a transport rejects an invalid command instead of putting it
back on the queue. Nothing does that. The queue runner puts every failed message
back, no matter the error, and no code outside the tests ever checks for this
error type.

we need to fix this comment

model/banner/command.go
```diff
@@ -0,0 +1,404 @@
...
+		Severity:     cmd.Severity,
+		Surface:      cmd.Surface,
+		Title:        cmd.Title[locale],
+		Text:         cmd.Text[locale],
+		Lang:         locale,
+		CTA:          cmd.CTA.pick(locale),
+		SecondaryCTA: cmd.SecondaryCTA.pick(locale),
+		Dismissible:  cmd.Dismissible,
+		Priority:     cmd.Priority,
+		StartsAt:     cmd.StartsAt,
```
rezk2ll, line 263: `startsAt` is read from the command here, but it gets thrown away later if the
banner already exists with the same `bannerId`. `endsAt` and the text are
applied, `startsAt` is not.

I tested this locally. I sent a first command with a window in August 2026, then
a second command moving both ends into March 2027. The result was a banner
running from August 2026 to March 2027, which is neither of the two windows I
asked for, and no error was raised.

So the Cloudery extending a window gets half of the change applied. Either honour
an explicit `startsAt`, or document that the start is locked once an occurrence
exists and a new window needs a new `bannerId`.

model/banner/command.go
```diff
@@ -0,0 +1,404 @@
...
+	if err := mu.Lock(); err != nil {
+		return err
+	}
+	defer mu.Unlock()
+
+	stored, err := storedCommand(inst, cmd.Category)
+	if err != nil {
+		return err
+	}
+	if stored != nil && cmd.Revision <= stored.Revision {
```
rezk2ll, line 150: The stored revision is keyed only by category, but the docs tell the Cloudery to
count revisions per target and category. Those two do not match.

I tested this locally. I sent an organization-wide command at revision 50, then a
command aimed at one member instance at revision 7. The second one was silently
ignored, because 7 is not greater than 50, even though it was counting in its own
sequence.

Either say the counter is shared per instance and category however, it was
addressed, so a category must only ever be addressed one way, or key the record
on the target too.

model/banner/command.go
```diff
@@ -0,0 +1,404 @@
...
+	instances, err := cmd.targets()
+	if err != nil {
+		return err
+	}
+	for _, inst := range instances {
+		if inst.HasBannersEnabled() && !inst.AllowsBannerCategory(cmd.Category) {
+			return fmt.Errorf("%w: the context of %s does not accept commands for the %s category",
+				ErrInvalidCommand, inst.Domain, cmd.Category)
+		}
+	}
```
rezk2ll: This behaves exactly as documented, but the consequence is worth a warning in the
docs. One member instance in a badly configured context blocks the command for
the entire organization.

I confirmed it locally: adding one instance whose context does not list the
category stopped billing banners for every other member, and the command went to
the dead letter queue. Nothing recovers on its own, someone has to fix the config
and replay.

model/banner/command.go
```diff
@@ -0,0 +1,404 @@
...
+		return err
+	}
+	if !cmd.complete(consts.DefaultLocale) {
+		return fmt.Errorf("%w: every text and label is required in the %s fallback locale",
+			ErrInvalidCommand, consts.DefaultLocale)
+	}
+	raw, err := json.Marshal(cmd)
+	if err != nil || len(raw) > MaxCommandBytes {
+		return fmt.Errorf("%w: command must encode to at most %d JSON bytes", ErrInvalidCommand, MaxCommandBytes)
+	}
```
rezk2ll, line 394: This re-encodes the command to check its size, but the handler already checked
the size of the raw message body before decoding, and that check is stricter
because it counts whitespace and unknown fields.

This is the only caller, so this second check never catches anything the first one missed. Removing it also removes the `encoding/json` import.

model/banner/command.go
```diff
@@ -0,0 +1,404 @@
...
+var _ couchdb.Doc = &commandState{}
+
+func (cmd Command) state() *commandState {
+	return &commandState{
+		Category: cmd.Category,
+		Revision: cmd.Revision,
+		Clear:    cmd.Clear,
+		EventID:  cmd.EventID,
+	}
+}
```
rezk2ll, line 228: This helper has a single caller, just above at line 166. Building the struct inline there would be shorter and save the reader a jump.

model/banner/command.go
```diff
@@ -0,0 +1,404 @@
...
+
+// commandState records the last command accepted for a category, so ordering
+// survives clears (which leave no public document) and unchanged decisions.
+type commandState struct {
+	DocID  string `json:"_id,omitempty"`
+	DocRev string `json:"_rev,omitempty"`
+
+	Category string `json:"category"`
+	Revision int64  `json:"revision"`
+	Clear    bool   `json:"clear"`
```
rezk2ll: Clear is written to the document but never read back anywhere. A cleared category has no banner document at all, which already tells you the same thing.

model/banner/command.go
```diff
@@ -0,0 +1,404 @@
...
+	return nil
+}
+
+// validate checks every locale, not just the one that will be picked.
+func (l Localized) validate(field string, max int) error {
+	if len(l) > maxLocales {
+		return fmt.Errorf("%w: %s carries more than %d locales", ErrInvalidCommand, field, maxLocales)
+	}
+	for locale, text := range l {
+		if len(locale) == 0 || len(locale) > maxLocaleLen {
```
rezk2ll, line 404: This checks the length of the language key. The overall command size is already capped, and a language key that nobody looks up is never used for anything, so this does not prevent any real problem.

model/banner/command.go
rezk2ll, line 184: This is the right place to keep the languages we currently discard. It is already
private to the stack and already unreachable from applications, so nothing else
needs to change to store them here.

model/banner/command.go
```diff
@@ -0,0 +1,404 @@
...
+}
+
+func (c *CommandCTA) validate(field string) error {
+	if c == nil {
+		return nil
+	}
+	if len(c.URL) > maxURLLen {
+		return fmt.Errorf("%w: %s.url is longer than %d bytes", ErrInvalidCommand, field, maxURLLen)
+	}
+	if ctaTarget(c.URL) == "" {
```
rezk2ll, line 422: This only checks that the URL is HTTPS and has a host. ADR 054 asks for an
allowlist by name:

> a `cta.url` allowlist is the stack's job at materialization time, so a locally
> authored document cannot turn into a platform-styled link anywhere it likes.

That mattered less while the stack built the URL itself from Cloudery's
manager URL. This PR is the first time the URL arrives from outside, so it is
the case the ADR was written for. As it stands, anything that can publish on the
exchange can point a platform-styled call to action at any https address.

A per-context host allowlist alongside `banner_command_categories` would fit the
shape this PR already uses for authorization.

model/banner/command.go
rezk2ll, line 183: `Clear` is written to the document but never read back anywhere. A cleared
category has no banner document at all, which already tells you the same thing.

model/banner/command.go
```diff
@@ -0,0 +1,404 @@
...
+	maxLabelLen   = 128
+	maxTitleLen   = 256
+	maxTextLen    = 1024
+	maxURLLen     = 2048
+	maxPriority   = 1000
+	maxEventIDLen = 256
+	maxLocaleLen  = 35
+	// MaxCommandBytes bounds the JSON body at the transport boundary.
+	MaxCommandBytes = 256 * 1024
+	// maxLocales bounds the locale map, not each value.
```
rezk2ll: This comment just repeats what the constant name already says.

model/banner/command.go
```diff
@@ -0,0 +1,404 @@
...
+	if cmd.Timestamp <= 0 {
+		return fmt.Errorf("%w: timestamp is required", ErrInvalidCommand)
+	}
+	if _, err := time.Unix(cmd.Timestamp, 0).UTC().MarshalJSON(); err != nil {
+		return fmt.Errorf("%w: timestamp must be epoch seconds within the RFC3339 range", ErrInvalidCommand)
+	}
+	if len(cmd.EventID) > maxEventIDLen {
+		return fmt.Errorf("%w: eventId is longer than %d bytes", ErrInvalidCommand, maxEventIDLen)
+	}
+	// A clear carries no presentation fields.
```
rezk2ll: This comment restates the condition written directly below it.

docs/banners.md
```diff
@@ -0,0 +1,238 @@
...
+
+Two kinds of producer write the same documents through the same code:
+
+- **In-process rules**, for what the stack owns. Disk usage is the only one
+  today (`model/banner/quota.go`): the stack measures it, so the stack decides,
+  and the wording comes from its own locale catalogs.
+- **A backend on the bus**, for what the stack cannot verify. A payment status,
+  a dunning step, a trial conversion are decisions another service already
+  made, so they arrive as commands carrying their own wording. The stack
+  validates, targets, orders, localizes and stores them; it decides nothing
```
rezk2ll: "localizes" is right, and it should stay right. Today the code only does half of it: it picks a language when the command arrives and never revisits it. Keep this line as written and make the code match, rather than weakening the sentence to describe what the code currently does.

The stack owns the instance language. It should keep owning what language the user reads, whoever wrote the words.

docs/banners.md
```diff
@@ -0,0 +1,238 @@
...
+
+`text`, `title` and every label are rendered by the backend, not by the stack.
+The stack picks **one** locale for the whole banner: the instance's, if every
+string the document needs exists in it, and `en` otherwise. `lang` names the
+language the user actually reads. Falling back field by field would put a
+French sentence above an English button.
+
+The stack keeps no copy of the locales it did not use, so changing an
+instance's language leaves a commanded banner in the language it was
+materialized in until the backend publishes its next command. Only the banners
```
rezk2ll: This is the sentence to change, and the code with it. Because the unused languages are discarded, a user who switches languages ends up with commanded banners in the old language sitting next to stack-written banners in the new one.

ADR 054 decided the opposite, and I want us to keep that decision: a locale change re-localizes every document for that instance. Keep the languages the Cloudery sends on the private command document and let the existing refresh hook re-pick; this paragraph becomes a description of the stack keeping every locale rather than an admission that it drops them.

docs/banners.md
```diff
@@ -0,0 +1,238 @@
...
+  "category": "billing"
+}
+```
+
+| Field | Required | Notes |
+| --- | --- | --- |
+| `category` | always | The slot to write. One document per category per instance. `quota` is refused. |
+| `workplaceFqdn` | one of the two | A single instance. |
+| `domain` | one of the two | A B2B organization: every instance under it gets the banner. |
+| `revision` | always | A positive counter the backend increments per target and category. It is what orders commands. |
```
rezk2ll: Telling the Cloudery to count revisions "per target" is what creates the collision. The stack stores one counter per category, shared by organization-wide and instance-specific commands, so two independent sequences overwrite each other.

docs/banners.md
```diff
@@ -0,0 +1,238 @@
...
+| `eventId` | no | The backend's correlation id, at most 256 bytes. Logged and retained, never a second ordering mechanism. |
+| `bannerId` | materialize | Identifies the occurrence: a new one clears a dismissal, the same one keeps it. |
+| `severity` | materialize | `info`, `warning` or `error`. |
+| `surface` | materialize | `banner` or `modal`. |
+| `text` | materialize | A map keyed by locale, complete in `en`. |
+| `title` | no | Same shape as `text`. A client with no title names the dialog from the text. |
+| `cta`, `secondaryCta` | no | `url` must be an absolute `https` URL. A secondary action needs a primary one. |
+| `dismissible` | no | Defaults to false. A modal with neither a call to action nor a dismissal is made dismissible. |
+| `priority` | no | 0 to 1000. The stack's own quota banners sit at 50 and 100. |
+| `startsAt`, `endsAt` | no | RFC 3339. `startsAt` defaults to the decision time. |
```
rezk2ll: `startsAt` is listed as a normal settable field, but it is silently ignored once a banner exists with the same `bannerId`. Needs a caveat here, or a fix in the code.

docs/banners.md
```diff
@@ -0,0 +1,238 @@
...
+  including whitespace and unknown fields at the transport boundary.
+- clear commands reject nonempty presentation fields, including wording and
+  windows; their addressing, timestamp and correlation fields are still validated.
+
+An instance whose context has no `enable_banners` is a no-op rather than a
+rejection: the backend knows its customers, not which of them display banners.
+A workplace that is not here is retried rather than rejected, because the stack
+cannot tell a deleted instance from one still being provisioned; the queue's
+delivery limit is what bounds those retries.
+
```
rezk2ll: The reasoning here is that retrying gives a still-being-provisioned instance time to appear. It does not. Putting a message back on a quorum queue redelivers it straight away, with no delay anywhere in the path.

I measured this locally: every rejected command was delivered six times within 7 to 15 milliseconds, then dead-lettered. No instance is going to finish provisioning in 14 milliseconds, so this is a rejection with five extra log lines.

Either add a real delay (dead letter with a time-to-live, routed back to the queue, and the dead letter exchange is already declared) or drop the provisioning explanation and say a missing instance dead-letters and needs a replay.

docs/banners.md
```diff
@@ -0,0 +1,238 @@
...
+the stack does write itself, the quota banners, and shipping a `.po` file is
+not what enables one.
+
+### Validation
+
+The command is rejected, never repaired. An authorized backend can put
+arbitrary text in front of a user, so anything unexpected in a payload is a
+backend bug worth surfacing rather than something to guess at. A rejected
+command fails the delivery, so the broker redelivers it up to the queue's
+`delivery_limit` and then dead letters it.
```
rezk2ll: Off by one. A delivery limit of 5 gave six handler runs, not five, because RabbitMQ counts deliveries rather than retries.

docs/banners.md
```diff
@@ -0,0 +1,238 @@
...
+```yaml
+contexts:
+  b2b_twake_default:
+    enable_banners: true
+    banner_command_categories:
+      - billing
+      - trial
+```
+
+**One category, one owner.** Two producers writing the same category means last
```
rezk2ll: "One category, one owner" covers two different producers. It does not cover the Cloudery using both addressing modes, which the contract explicitly allows and which results in the same collision.

docs/banners.md
```diff
@@ -0,0 +1,238 @@
...
+
+Re-materializing the same `bannerId` keeps a dismissal the user recorded, and
+keeps the moment the occurrence began rather than the last evaluation. A new
+`bannerId` is a message the user has not seen, so it clears the dismissal.
+Escalating a dunning cycle, or starting a new one, is a new occurrence; changing
+the wording of the current one is not.
+
+### What the stack does not do
+
+Nothing replays on its own: enabling `enable_banners` on a context materializes
```
rezk2ll: This contradicts the ADR, which has a trigger row for exactly this case:

> | Rollout and instance creation | A bootstrap evaluation, so a condition already true on day one produces a banner without waiting to be re-crossed |

Quota still bootstraps, because the stack owns the input and can re-read it. Billing and trial now depend on the Cloudery replaying, and nothing in this PR asks it to. So turning `enable_banners` on for a context leaves every customer already in dunning with no banner until their next Stripe event, which for a healthy subscription could be a month away.

Either ask the Cloudery for a replay-on-demand endpoint as part of this contract, or say plainly here that rollout needs a manual bulk replay and who owns running it.

pkg/rabbitmq/handlers.go
```diff
@@ -837,66 +836,47 @@ func maskSensitiveData(data string) string {
...
-		}
-		if len(domains) == 0 {
-			log.Infof("billing.lifecycle: no instances found for organization %s", msg.Domain)
-			return nil
-		}
+	// A command the queue cannot fix is nacked like any other failure. The
+	// queue is declared with a delivery limit, so the broker dead letters it
+	// after those attempts rather than redelivering it forever.
+	if err := banner.ApplyCommand(cmd); err != nil {
+		return fmt.Errorf("banner.commands: %s revision %d: %w", d.RoutingKey, cmd.Revision, err)
```
rezk2ll, line 872: This logs "applied" even when the command was ignored. `ApplyCommand` returns no error both when it wrote something and when it skipped a stale revision, so the handler cannot tell them apart.

Locally, one command produced both of these lines:

```
billing: ignoring revision 7, not newer than the stored 50
banner.commands: banner.materialize applied to bob.cozy.example
```

Since ignoring a stale revision is the normal outcome of at-least-once delivery, this is the line someone will be reading when they are trying to work out why a banner never appeared, and it tells them the opposite. Either return an applied/ignored result, or say "accepted" instead of "applied".

pkg/rabbitmq/handlers.go
```diff
@@ -837,66 +836,47 @@ func maskSensitiveData(data string) string {
...
-		for _, inst := range list {
-			domains = append(domains, inst.Domain)
-		}
-		if len(domains) == 0 {
-			log.Infof("billing.lifecycle: no instances found for organization %s", msg.Domain)
-			return nil
-		}
+	// A command the queue cannot fix is nacked like any other failure. The
+	// queue is declared with a delivery limit, so the broker dead letters it
+	// after those attempts rather than redelivering it forever.
```
rezk2ll: This states a deployment setting as a fact from inside the code. It is true of the example config, but not of any deployment that leaves out `delivery_limit`, and the code has no way to know. The rabbitmq doc is the right place for it and already says it.

pkg/rabbitmq/handlers.go
```diff
@@ -837,66 +836,47 @@ func maskSensitiveData(data string) string {
...
-	for _, domain := range domains {
-		if err := banner.RefreshBilling(domain, status, eventAt); err != nil {
-			return fmt.Errorf("billing.lifecycle: materialize for %s: %w", domain, err)
-		}
-		log.Infof("billing.lifecycle: %s applied to %s (status %s, attempt %d, event %s, at %s)",
-			d.RoutingKey, domain, status, msg.AttemptCount, msg.EventID, eventAt.Format(time.RFC3339))
+	target := cmd.WorkplaceFqdn
+	if target == "" {
+		target = cmd.Domain
 	}
```
rezk2ll: Four lines to build one log value. Exactly one of the two fields is set, so concatenating them works, and that is the idiom the command file already uses at line 295.

pkg/lock/lock.go
rezk2ll, line 81: The fix is right and the two new tests cover both halves. Worth saying out loud that the renewer stopping does not stop the operation. The VFS migration never checks its lease, so it keeps running while the lock expires and another process can take it. That was already true before this change, and the new warning is the first signal it has ever had.

pkg/lock/lock.go
```diff
@@ -43,37 +45,44 @@ type ErrorRWLocker interface {
 
 type longOperationLocker interface {
 	ErrorLocker
-	Extend()
+	Extend() error
 }
 
+// ErrLockLost means that an operation no longer owns its distributed lock.
+var ErrLockLost = errors.New("lock ownership lost")
```
rezk2ll: `ErrLockLost` is exported but nothing outside this package ever checks for it. Could stay unexported until something needs it.

model/banner/command.go
```diff
@@ -0,0 +1,404 @@
...
+	URL   string    `json:"url"`
+}
+
+// Command is what a backend asks for, independent of how it arrived.
+type Command struct {
+	Category string `json:"category"`
+
+	// Exactly one of Domain and WorkplaceFqdn is set. Domain addresses a
+	// B2B organization, and every instance under it gets the banner.
+	Domain        string `json:"domain,omitempty"`
```
shepilov: can we use here tenant(org id) instead of domain?

model/banner/command.go
```diff
@@ -0,0 +1,404 @@
...
+		return err
+	}
+
+	instances, err := cmd.targets()
+	if err != nil {
+		return err
+	}
+	for _, inst := range instances {
+		if inst.HasBannersEnabled() && !inst.AllowsBannerCategory(cmd.Category) {
+			return fmt.Errorf("%w: the context of %s does not accept commands for the %s category",
```
shepilov: should we just skip this instance and not return error for others?

pkg/lock/lock.go
```diff
@@ -1,9 +1,11 @@
 package lock
 
 import (
+	"errors"
```
shepilov: This definitely has to be a separate PR, the change is too sensitive, with also strong argumentation why do we need it

model/banner/command_test.go
```diff
@@ -0,0 +1,667 @@
...
+
+const decidedAt = 1788944400
+
+// fixture decodes one of the shared wire fixtures, which are what the backend
+// publisher is developed against. A field this package stops reading, or a
+// field it starts requiring, breaks here rather than in production.
+func fixture(name string) Command {
+	raw, err := os.ReadFile("testdata/" + name + ".json")
+	if err != nil {
+		panic(err)
```
shepilov: use t.Helper() and reqire.NoError() instead

model/banner/banner_test.go
```diff
@@ -21,6 +25,15 @@ func TestMain(m *testing.M) {
 		}
 		i18n.LoadLocale(locale, "", po)
 	}
+	if err := config.LoadTestFile(); err != nil {
+		panic(err)
+	}
+	ctx := context.Background()
+	if _, err := couchdb.CheckStatus(ctx); err == nil {
+		if err := couchdb.InitGlobalDB(ctx); err != nil {
```
shepilov, line 33: mmmm, why do we need this init here?

model/banner/command.go
```diff
@@ -0,0 +1,404 @@
...
+		next.DocID, next.DocRev = stored.DocID, stored.DocRev
+		return couchdb.UpdateDoc(inst, next)
+	}
+	next.DocID = docID(cmd.Category)
+	return couchdb.CreateNamedDocWithDB(inst, next)
+}
+
+// commandState records the last command accepted for a category, so ordering
+// survives clears (which leave no public document) and unchanged decisions.
+type commandState struct {
```
shepilov, line 177: Why do we need a new doc type? Why banners isn't enough?

assets/locales/en.po
MontaGhanmy: The stack **doesn't choose billing wording anymore**, so those entries would never be read again. What we did lose is re-localization, and that's a separate thing: **fixed in `36d95f94b`**, where the private command document keeps every locale the Cloudery sends and the refresh hook picks one again.

model/banner/command.go
MontaGhanmy: **Fixed in `36d95f94b`.** The private command document now keeps every locale, and the existing refresh hook picks the language again when the instance changes it. I kept the whole command rather than only the wording, so we never have to rebuild a banner from the public document that apps can write.

model/banner/command.go
MontaGhanmy: **You're right, nothing checks it.** The comment now says the runner nacks every error the same way. Fixed in `16cb17427`.

model/banner/command.go
MontaGhanmy: **Fixed in `de627cb63`.** A command that sends no `startsAt` keeps the start the occurrence already had, and one that sends it replaces the stored value. **There was a worse case than the one you hit:** validation checked the window the command sent, not the one we actually store, so moving a window backwards could store a start after its end.

model/banner/command.go
MontaGhanmy: **I don't think that holds.** `json.Marshal` escapes `<`, `>` and `&`, and a publisher can legally send those raw, so a body can pass the handler's check and still re-encode above the cap. The two checks bound different things, so I kept it.

model/banner/command.go
MontaGhanmy: **Kept.** Since `36d95f94b` it also builds the command we retain, so it isn't just a struct literal behind a function anymore.

model/banner/command.go
MontaGhanmy: **Kept for now.** It's three lines, and since `36d95f94b` those locale keys are stored rather than read once and dropped. Happy to drop it if you'd still rather.

model/banner/command.go
MontaGhanmy: **Agreed**, and that's exactly where they went in `36d95f94b`.

model/banner/command.go
MontaGhanmy: **It's read now**, in `36d95f94b`. The refresh has to tell a category we deliberately cleared from one whose document an app simply deleted.

model/banner/command.go
MontaGhanmy: **Removed in `16cb17427`.**

model/banner/command.go
MontaGhanmy: **Removed in `16cb17427`.**

docs/banners.md
MontaGhanmy: **Kept as written**, and the code does it now, `36d95f94b`.

docs/banners.md
MontaGhanmy: **Both changed in `36d95f94b`.** One thing I added: the refresh rewrites a banner, it doesn't restore one, so a category whose document was deleted stays empty until the next command.

docs/banners.md
MontaGhanmy: **Took the fix** rather than the caveat, `de627cb63`.

docs/banners.md
MontaGhanmy: **You're right, nothing in the path delays a redelivery.** I removed the provisioning explanation in `4ab9f1d42`. **I didn't add a TTL loop:** our dead letter queue routes with `banner.commands.dead`, and the handler needs `banner.materialize` or `banner.clear` to know which operation it's replaying, so sending the message straight back wouldn't work.

docs/banners.md
MontaGhanmy: **Fixed in `4ab9f1d42`.** Our own rabbitmq test expects `DeliveryLimit + 1` handler calls, so your six matches. I wrote it as true of the version our test fixture pins rather than as a promise, because how RabbitMQ counts a redelivery has changed between releases and we requeue with `basic.nack`.

docs/banners.md
MontaGhanmy: **It already exists on the Cloudery:** `rake banners:billing:reconcile`, scoped by owner, context or partner, with a dry run, plus a `sweep` task for work a crash left pending. It re-decides from live facts rather than replaying old failure events, which is the part that matters here.

Publishing is gated there by `BANNER_COMMAND_CONTEXTS`, empty by default, so a rollout is three steps: enable the context on our side, set that variable, run reconcile.

pkg/rabbitmq/handlers.go
MontaGhanmy: **Fixed in `16cb17427`.** It says accepted now, and the line names the other outcomes that reach it, a stale revision and an instance with banners off.

pkg/rabbitmq/handlers.go
MontaGhanmy: **Fixed in `16cb17427`.** The comment now leaves the delivery limit to the rabbitmq doc.

pkg/rabbitmq/handlers.go
MontaGhanmy: **Fixed in `16cb17427`.**

pkg/lock/lock.go
MontaGhanmy: **Moved to #4911** and written into its description. Neither caller watches its lease, so the operation keeps running while the lock expires. That was already true before the change, the new warning is just the first time anything says so.

pkg/lock/lock.go
MontaGhanmy: **Unexported in #4911.**

pkg/lock/lock.go
MontaGhanmy: **Done, #4911.** Nothing in this PR depends on it.

model/banner/command_test.go
MontaGhanmy: **Done in `3d4810c75`**, for both `fixture` and `valid`.

model/banner/banner_test.go
MontaGhanmy: It brings up the global database, which the command tests need because they create instances. It's spelled out here instead of calling our shared test helper because importing `tests/testutils` from inside `package banner` is an **import cycle**: testutils pulls in `model/stack`, which pulls in `pkg/rabbitmq`, which imports `model/banner`. That was already true before this PR, `pkg/rabbitmq` imported `model/banner` on master too.

model/banner/command.go
MontaGhanmy: Because **`io.cozy.banners` isn't in the permission blocklist and `io.cozy.banners.commands` is.** An app needs write access to the banners doctype to record a dismissal, so if the revision lived there the same app could rewind it and we'd accept a stale command. A clear also deletes the public document, which would leave nothing to compare the next revision against.

docs/banners.md
MontaGhanmy: Trimmed in 49db1bff4

model/banner/command.go
MontaGhanmy: Fixed in 861eb4f05: organization commands now use `tenant` (org ID) with `ListOrgInstancesByID`, with updated validation, fixtures, documentation, and passing banner, RabbitMQ, and API tests.

model/banner/command.go
MontaGhanmy: Updated in 2328c5079, now we skip instances whose context disallows the category and continue processing eligible instances, while still returning storage errors for retries.

model/banner/command.go
MontaGhanmy: Do we need both addressing modes for the same instance and category? I'd keep one sequence in Cloudery; separate counters on the stack would still leave us deciding which command wins.

docs/banners.md
MontaGhanmy: You're right, ‘per target’ allows two sequences that the stack cannot compare; we need to make the same ordering rule explicit on both sides.

docs/banners.md
MontaGhanmy: Agreed, one producer can still create the collision; we need to prevent overlapping tenant and workplace sequences for the same category in Cloudery.

model/banner/command.go
MontaGhanmy: Agreed, HTTPS alone doesn't restrict where the button leads; let's configure allowed hosts per context so this also works for partner and on-prem deployments.

model/banner/command.go
MontaGhanmy: Updated in 2328c5079, now we skip instances that disallow the category and continue processing the others.

model/banner/command.go
shepilov: Then we shouldn't delete the public document, or whatever. To have a permssion blacklist as a separate type is kind of everkill. There is multiple variants how you can implement it, to have multiple banners of the same catergoy and the filter/group on get query or just let the app do anything it wants as we have actually for files. Could you please describe the use case: when do we need this "blacklist",  I just don't understand what problem we can have with this "dismissal" record. If it's ABA problem, then just, ok, let's record timestamp or optimistiv locking on a revision

model/banner/banner_test.go
shepilov: that's why it's better then to put test in test package, actually to banner_test package

model/banner/command.go
shepilov: but for errors as well, if we had an error for one instance, why should we stop processing others?

model/banner/command.go
```diff
@@ -0,0 +1,426 @@
...
+	}
+	inst, err := lifecycle.GetInstance(cmd.WorkplaceFqdn)
+	if err != nil {
+		return nil, err
+	}
+	return []*instance.Instance{inst}, nil
+}
+
+func (cmd Command) applyTo(inst *instance.Instance) error {
+	// Instances that disable banners or disallow this category are skipped.
```
rezk2ll, line 133: But the single-instance path now has no signal at all. Nothing written, nothing dead-lettered, and the only line an operator sees is the handler's, which says the command was accepted:

banner.commands: banner.materialize accepted for dave.cozy.example (category billing, revision 9, event skip1)

rezk2ll (commented):
still two things missing:
- a disallowed category is now silently swallowed
- no CTA allowlist yet
- the Revision collision issue

model/banner/command.go
MontaGhanmy: We picked one rule in a27af446e: one revision per instance and category, shared by tenant and workplace commands, so a category is addressed only one way. If the Cloudery ever needs both modes on the same category we key on the addressing mode, but I don't see that case today.

model/banner/command.go
MontaGhanmy: Added in a27af446e as `banner_cta_hosts` per context, checked when a command is applied and again on refresh. A host missing from the list skips the instance with a warning, and there is a test for it.

model/banner/command.go
MontaGhanmy: Agreed, done in a0811b11a. We now try every member and return all the failures together so the delivery is retried, and the test checks that a member after the failing ones still gets its banner.

model/banner/banner_test.go
MontaGhanmy: Done in 3a9328040. The command tests moved to `banner_test` and use the shared `testutils` setup, so the inlined TestMain is gone.

model/banner/command.go
MontaGhanmy: You're right, the separate doctype was overkill, so we dropped it in 2816d5605. The revision and the accepted command now live on the banner document itself, a clear expires that document instead of deleting it, and dismissals go through the normal `_rev`.

model/banner/command.go
MontaGhanmy: Right, fixed in a27af446e. A skipped instance now logs a warning with the reason, either the category or the CTA host, and its revision doesn't move.

model/banner/command.go
```diff
@@ -0,0 +1,381 @@
...
+	URL   string    `json:"url"`
+}
+
+// Command is what a backend asks for, independent of how it arrived.
+type Command struct {
+	Category string `json:"category"`
+
+	// Exactly one of Tenant and WorkplaceFqdn is set. Tenant is the B2B
+	// organization ID, and every instance under it gets the banner.
+	Tenant        string `json:"tenant,omitempty"`
```
shepilov: OrgID

cozy.example.yaml
```diff
@@ -535,6 +535,22 @@ contexts:
     # Feature flags
     features:
       - hide_konnector_errors
+    # Materialize the platform banners (io.cozy.banners) for the instances of
+    # this context. Off by default, so the rules can ship before the clients
+    # that render them. Turning it back off stops the writes and leaves the
+    # documents already materialized in place.
+    enable_banners: true
```
shepilov: group to the banner: and one struct

cozy.example.yaml
```diff
@@ -535,6 +535,22 @@ contexts:
...
+    # The banner categories the stack.banner.commands queue is allowed to
+    # write. Its publisher is authenticated by its broker credentials and
+    # bindings, so this says what it may say, not who it is. The quota
+    # category is always refused: the stack measures disk usage itself.
+    # See docs/banners.md.
+    banner_command_categories:
+      - billing
+      - trial
+    # The hosts a banner command's call to action may link to.
+    banner_cta_hosts:
```
shepilov: how I can enable everything?

shepilov (approved):

PR #4913, opened 2026-09-11
feat(s3): support explicit bucket layouts

### LINKED ISSUE #4912: Allow S3 deployments to use pre-created buckets
## Problem

The S3 backend always calls `ListBuckets` during connection setup and `MakeBucket` for the fixed secondary-storage buckets and each organization VFS bucket. This requires runtime credentials with `s3:ListAllMyBuckets` and `s3:CreateBucket`.

Some deployments provision buckets through Terraform or another infrastructure tool and want cozy-stack to run with credentials limited to the existing buckets and their objects.

Some S3 deployments also issue separate credentials for each bucket. Cozy-stack currently uses one global S3 client, so every bucket must share the same credentials.

Follow-up to #2183.

## Proposal

Allow automatic bucket creation to be disabled and allow an optional connection URL per concrete bucket. The existing `fs.url` remains the default connection and fallback, preserving current configurations.

One possible configuration shape is:

```yaml
fs:
  url: s3://s3.example.net?access_key=DEFAULT_KEY&secret_key=DEFAULT_SECRET&region=gra&bucket_prefix=cozy
  s3:
    auto_create_buckets: false
    bucket_urls:
      cozy-default: s3://s3.example.net?access_key=VFS_KEY&secret_key=VFS_SECRET&region=gra
      cozy-apps-web: s3://s3.example.net?access_key=APPS_KEY&secret_key=APPS_SECRET&region=gra
      cozy-previews: s3://s3.example.net?access_key=PREVIEWS_KEY&secret_key=PREVIEWS_SECRET&region=gra
```

The exact configuration keys can be adjusted during implementation, but it should support exact pre-created bucket names with distinct endpoints or credentials.

When automatic creation is disabled:

- Do not call `ListBuckets` or `MakeBucket`.
- Verify each fixed bucket is accessible during startup using its configured connection.
- Verify an organization VFS bucket is accessible when its filesystem is initialized.
- Return an explicit error containing the expected bucket name when a bucket is missing, inaccessible, or has no usable credentials.
- Use the global S3 connection when a bucket does not have an explicit override.
- Keep the existing bucket naming and per-instance object-key prefixes.
- Never include access keys or secret keys in logs or errors.

Automatic creation should remain enabled by default to preserve current behavior.

## Acceptance criteria

- Existing single-connection configurations keep working and automatically creating buckets.
- Pre-created-bucket mode works without `s3:CreateBucket` and `s3:ListAllMyBuckets`.
- Each pre-created bucket can use its own configured S3 endpoint and credentials.
- Buckets without an override use the global `fs.url` connection.
- Missing or inaccessible buckets fail with a clear, non-secret error.
- Tests cover at least two buckets using different credentials.
- S3 configuration documentation covers both modes.

shepilov:
## Summary

- Configure default and per-type S3 destinations with bucket names and optional connection URLs together in one map.
- Separate objects by type and instance so shared-bucket reads, listings, and deletion stay scoped.
- Support pre-created buckets with bounded startup checks, optional automatic creation, and environment overrides.
- Document the configuration and object layout.

Closes #4912.


docs/s3.md
```diff
@@ -9,24 +9,94 @@ Scaleway, etc.) and does not depend on the AWS SDK.
...
+
+Invalid names, unknown storage types, missing destinations and invalid
+connection URLs cause startup errors. Configuration errors do not print
+connection credentials.
+
+### Bucket provisioning
+
+`auto_create_buckets` defaults to `true`: startup creates each distinct bucket
+once, using its connection's region. Existing buckets are accepted. Bucket
+creation requires `s3:CreateBucket`; bucket enumeration is not used.
```
Crash--: It means that if we deploy the stack when the S3 is down, then we'll not be able to deploy, right?

docs/s3.md
```diff
@@ -9,24 +9,94 @@ Scaleway, etc.) and does not depend on the AWS SDK.
...
+
+With `auto_create_buckets: false`, create the configured buckets before starting
+the stack. Startup checks each distinct bucket with `HeadBucket`, without listing
+or creating buckets. Missing or inaccessible buckets fail startup with a
+bucket-specific error. All startup bucket operations have a 30-second timeout.
+Loading or initializing an instance does not probe S3.
+
+`HeadBucket` requires bucket-scoped `s3:ListBucket` permission on AWS S3; see the
```
Crash--: do not use AWS S3 as exemple :p

docs/s3.md
```diff
@@ -116,50 +190,25 @@ Comment out the S3 URL in your config and restart cozy-stack:
...
-| `<prefix>-exports`           | Instance export archives                |
-
-Buckets are created automatically on first use.
+| Storage type | Object prefix | Content |
+|--------------|---------------|---------|
+| `files` | `files/<DBPrefix>/` | Files, versions, avatars and thumbnails |
+| `apps_web` | `apps-web/` | Installed web applications, markers and tarballs |
+| `apps_konnectors` | `apps-konnectors/` | Installed konnectors, markers and tarballs |
+| `assets` | `assets/` | Dynamic assets, grouped by context |
+| `previews` | `previews/` | PDF previews and icons |
```
Crash--, line 209: why do we have a dedicated bucket for this one? It's the same as "files"? Is it not easier to handle that in the same bucket / place? 

Specially since you say that files is also for thumbnails 

(shit this is something I've done  ^^)

pkg/config/config/config.go
```diff
@@ -849,6 +864,28 @@ func UseViper(v *viper.Viper) error {
...
+		return errors.New("s3: invalid fs.s3.auto_create_buckets configuration")
+	}
+	if s3.Buckets == nil {
+		s3.Buckets = make(map[string]FsS3Bucket)
+	}
+	for _, kind := range append([]string{"default"}, s3StorageKinds...) {
+		entry := s3.Buckets[kind]
+		key := "fs.s3.buckets." + kind
+		if v.IsSet(key+".name") || v.IsSet(key+".url") {
+			entry.Name = v.GetString(key + ".name")
```
Crash--, line 883: how do you ensure it's a valid key name?

pkg/config/config/s3.go
```diff
@@ -2,105 +2,129 @@ package config
...
-	// Sanitize bucket prefix: lowercase, only alphanumeric and hyphens
-	s3BucketPrefix = strings.ToLower(s3BucketPrefix)
-	s3BucketPrefix = strings.Map(func(r rune) rune {
-		if (r >= 'a' && r <= 'z') || (r >= '0' && r <= '9') || r == '-' {
-			return r
+	for kind, entry := range fs.S3.Buckets {
+		if kind != "default" && !slices.Contains(s3StorageKinds, kind) {
+			return fmt.Errorf("s3: unknown storage type %q in fs.s3.buckets", kind)
+		}
+		if s3utils.CheckValidBucketNameStrict(entry.Name) != nil {
```
Crash--, line 57: https://github.com/linagora/cozy-stack/pull/4913/changes#r4004781577

ok it's done here

docs/s3.md
shepilov: Yes, and it was like this before this PR, the stack is kind of useless without storage

docs/s3.md
shepilov: and it's some legacy stuff, there was somelogic with checksum cache that I don't understand why do we need, so we can drop it, and it's just backwardcompatibility just in case

pkg/config/config/s3.go
shepilov: yep

model/instance/instance.go
```diff
@@ -294,10 +294,9 @@ func (i *Instance) AvatarFS() vfs.Avatarer {
 			panic(ErrInvalidSwiftLayout)
 		}
 	case config.SchemeS3:
-		client := config.GetS3Client()
-		bucket := vfss3.BucketName(i.GetOrgID(), config.GetS3BucketPrefix())
-		keyPrefix := i.DBPrefix() + "/"
-		return vfss3.NewAvatarFs(client, bucket, keyPrefix)
+		storage := config.GetS3Storage("files")
```
taratatach: nit: Shouldn't we use some constant here rather than the `"files"` string? Like we do for konnectors for example (`appfs.S3AppsStorageType(consts.KonnectorType)`).

taratatach (approved):

model/instance/instance.go
shepilov: yep, I'll add one

PR #4921, opened 2026-09-14
Enable upload in a magic folder in 1 network call instead of 5

zatteo:
How to create or fetch a magic folder is summarized here: https://github.com/linagora/tmail-flutter/issues/4827#issuecomment-5664161656

But it is a 5 network call action. We want to reduce this to simplify the usage of magic folders for non Cozy apps.

More context here: https://github.com/linagora/tmail-flutter/issues/4827#issuecomment-5664343100

**To test locally**

```
curl -s -X POST "http://alice.cozy.localhost:8080/files/?Type=file&Name=attachment.txt&MagicFolder=io.cozy.apps/mail" -H "Authorization: Bearer $(cozy-stack instances token-cli alice.cozy.localhost:8080 io.cozy.files)" -H "Content-Type: text/plain" --data 'hello'
```


shepilov:
@zatteo I've changed the API a little bit, I think it will be more straightforward, not to use an additional param but to use it as just DirID

zatteo:
> @zatteo I've changed the API a little bit, I think it will be more straightforward, not to use an additional param but to use it as just DirID

Great!

shepilov (approved):

docs/files.md
```diff
@@ -22,6 +22,35 @@ The root of the virtual file system is a special directory with id
...
+
+These endpoints accept a magic folder ID in the directory path parameter:
+
+- `POST /files/:dir-id` with `Type=file` or `Type=directory`
+- `GET /files/:dir-id`
+- `GET /files/:dir-id/relationships/contents`
+- `GET /files/:dir-id/size`
+
+The supported IDs are `io.cozy.apps/mail`, `io.cozy.apps/notes`, and
+`io.cozy.files.shared-drives-dir` (the container for shared drives).
```
Crash--, line 35: tell me that this is just an exemple? And we support any kind of magic folder?

zatteo: No 🙈

We could:
- allow all magic folders allowed on cozy-client (see https://github.com/linagora/cozy-client/blob/a50c1ee50cc0a9fc9910473309add9629bba4d6c/packages/cozy-client/src/models/folder.js#L11)
- allow all magic folders but for this we need to find the name of the folder (either add a param in API or we invent a name and the user rename it later)

WDYT? I don't know why they have been restricted on cozy-client

Crash--: we're on stack side, so we should not rely on cozy-client check or whatever. So can we create the IDs we want?

zatteo: Not right now but I can make a PR. I will just need to add a new param for the name of the folder.

PR #4923, opened 2026-09-15
feat(banners): default CTA hosts to the csp_allowlist hosts

MontaGhanmy:
Hosts listed in `csp_allowlist`, global and for the instance's context, are now allowed as banner CTA hosts by default, so `cta_hosts` only needs to list the others.

CSP wildcards, keywords (`'self'`), scheme sources (`data:`) and non-https URLs are ignored, so there is still no wildcard CTA host.


shepilov (approved):

shepilov:
dropped

PR #4926, opened 2026-09-16
feat: Enforce effective member access on shared drive moves

### LINKED ISSUE #4844: Use Effective Access In Backend Move Operations
## Context

  The backend already supports move/copy operations even if the UI does not expose move yet.

  With nested shared folders, moving content is not only a VFS metadata update. A move can change which sharing scopes apply to the moved file or folder, especially when content is moved:

  - into a shared folder,
  - out of a shared folder,
  - between two shared folders,
  - with nested shared folders inside the moved tree.

  So move operations must use effective access for authorization and must go through a shared-aware move path instead of being applied as a simple metadata patch.

  ## Goal

 Update the backend move endpoint so it uses effective access checks for source and destination.

  ## Related Endpoints

  Explicit backend move endpoint:

  POST /sharings/drives/move

  PATCH endpoints that can become move operations:

  PATCH /files/:file-id
  PATCH /files/metadata
  PATCH /files/
  PATCH /sharings/drives/:id/:file-id

  ## Scope
 
 web/sharings/move.go
    MoveHandler
    checkSharedDrivePermission calls inside MoveHandler

  Current logic checks permissions roughly like this:

  source sharing:  require write when moving, require read when copying
  destination sharing: require write

This should be replaced by target-based effective access checks.

  This includes:

  - detecting parent changes in PATCH requests,
  - routing move-like PATCH requests away from simple metadata update logic,
  - applying effective access checks to source and destination,
  - preventing move operations from bypassing nested shared-folder rules.

  ## Permission Rules

  For copy:
  require effective read on the source file/folder
  require effective write/create on the destination parent folder

  For move:
  require effective write on the source file/folder
  require effective write/create on the destination parent folder

taratatach:
Closes #4844 

  Moves and copies on shared drives were authorized by a flat
  membership plus `ReadOnly` check, wrong with nested shared folders,
  and cross-stack moves bypassed member authorization entirely via the
  share-interact token.

  - Route cross-stack moves through the drive routes: `client.InDrive`
    scopes a client to a shared drive, so the hosting stack authorizes
    each request via `guardSharedDriveRouteForMember`
  - Resolve effective access on source and destination in
    `MoveHandler` via `AccessResolver.ResolveForMember` — 404 out of
    scope, 403 on insufficient access
  - Support `permanent_delete` on drive PATCH, previously a silent
    no-op
  - Reserve drive root trash/destroy to the owner: members get a 403


client/files.go
```diff
@@ -97,11 +97,31 @@ type FilePatch struct {
 // listing directory children. Tests can override this to force pagination.
 var ListChildrenPageSize = "100"
 
+// InDrive scopes the client to the given shared drive: file operations go
+// through the /sharings/drives/<id> endpoints instead of /files, so the
+// stack hosting the drive enforces member authorization. All the operations
+// routed through filesPath have a drive-route equivalent (reads, creation,
+// metadata patch, upload, overwrite, trash, restore, destroy). It mutates
+// and returns the receiver for chaining.
+func (c *Client) InDrive(driveID string) *Client {
```
shepilov, line 106: why new method instead of constructor? It's always called NewRemoteClient.InDrive

taratatach: `NewRemoteClient` is part of the `sharings` package while the `Client` struct (which keeps hold of the drive ID) is part of the `client` package.

The new `InDrive` method allows other clients to set the `driveID` attribute.

shepilov (approved):

client/files.go
shepilov: add one more param to NewRemoteClient or override?

taratatach: Then only clients created via `NewRemoteClient` could set the `driveID` attribute while it's an attribute of `Client`.

shepilov: it can be set also with Client struct init by any consumer, as in NewRemoteClient

taratatach: We'd need to export the attribute then. Which could be a good idea actually.

PR #4929, opened 2026-09-17
fix(lock): stop the renewal goroutine on unlock

shepilov:
## Problem
A job acquires a 20-second Redis lease at 0 seconds and finishes at 1 second. 
The renewal goroutine is already waiting for the first tick at about 6.7 seconds. 
Previously, `Unlock()` stopped the ticker without closing its channel, leaving that goroutine blocked forever.
The explicit shutdown signal now wakes it so it can exit.

## Summary

- End the long-operation renewal goroutine when `Unlock()` closes its shutdown channel. The goroutine stops its own ticker when it exits.
- Serialize renewal with unlock and check the current acquisition so a ready tick cannot renew a released or reused lock. Add regression checks for goroutine exit, lock reuse, and a tick arriving during unlock.
- Return the existing mutex directly for in-memory long operations, since those locks never expire and need no renewal goroutine.


pkg/lock/lock.go
```diff
@@ -49,31 +49,34 @@ type longOperationLocker interface {
...
-			l.mu.Unlock()
-			<-ch
-			l.mu.Lock()
-			if l.tick == nil {
+			select {
+			case <-done:
 				return
+			case <-tick.C:
+				l.mu.Lock()
+				if l.done != done {
```
taratatach, line 73: In which situations do we replace `l.done`? 🤔

pkg/lock/lock.go
```diff
@@ -49,31 +49,34 @@ type longOperationLocker interface {
...
 		return err
 	}
-	l.tick = time.NewTicker(l.timeout / 3)
+	l.mu.Lock()
+	done := make(chan struct{})
+	l.done = done
+	l.mu.Unlock()
 	go func() {
-		defer l.mu.Unlock()
+		tick := time.NewTicker(l.timeout / 3)
```
taratatach, line 65: What does it change to move the tick instantiation to the goroutine?

shepilov: Actually, nothing, just have it in the same place where it is used, there no need to have in the struct

pkg/lock/lock.go
shepilov: for example, when you call lock after unlock

PR #4931, opened 2026-09-21
fix(oauth): say workplace on the authorize screen

Crash--:
## Summary

- Reword the `Authorize Client presentation` string in all locales to read "would like to access the following data from your workplace", replacing the leftover "cozy" and "Twake" wording.


Crash--:
In order to fix that. 

Capture d’écran 2026-09-21 à 11 33 05

shepilov (approved):

shepilov:
why not twake?

PR #4932, opened 2026-09-21
fix(sharing): retire the magic shared drives container

shepilov:
## Summary

- Treat the legacy shared-drives container as an ordinary folder without creating it on reads.
- Create new named shared folders and recipient shortcuts at root while preserving existing shortcut locations.
- Keep the legacy creation endpoint returning an ordinary folder with a generated ID for new folders.


docs/shared-drives.md
```diff
@@ -157,7 +157,7 @@ When an existing root is used through `folder_id` or `file_id`, that root must:
 Additional rules for directory roots:
 
 - Must not be a system folder (`root`, `trash`, `shared-with-me`,
-  `shared-drives`, `no-longer-shared`)
+  `no-longer-shared`)
```
taratatach, line 160: `shared-with-me` and `no-longer-shared` are not listed in the magic folders section of the doc.
Are they missing or are these different kinds of special directories?

taratatach (approved):

docs/shared-drives.md
shepilov: I didn't look do we use them anywhere right now, it's legacy for the old type sharing, so I don't know

